Issue #381, opened 2016-05-20
Detach failure during VM shutdown brings ESX service down


pdhamdhere:
Setup: Photon RC, TP2 release

Steps to repro;
1. Create volume: `docker volume create -d vmdk --name tv1`
2. Run container: `docker run -d -it --restart="always" -v tv1:/data busybox`
3. Shudown VM from ESX UI
4. Noticed Server crash and disk tv1.vmdk remains attached to VM

```
05/20/16 18:47:04 2896177 [INFO   ] *** detachVMDK: /vmfs/volumes/4bad174b-b67f5cfc-2fc6-0026b966a303/dockvols/tv1.vmdk from Photon1 VM uuid = 564db5e5-a5a9-5d76-b349-70cd1326d9fd
05/20/16 18:47:04 2896177 [ERROR  ] (vim.fault.InvalidState) {
   dynamicType = <unset>,
   dynamicProperty = (vmodl.DynamicProperty) [],
   msg = 'The operation is not allowed in the current state.',
   faultCause = <unset>,
   faultMessage = (vmodl.LocalizableMessage) []
}
Traceback (most recent call last):
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 691, in main
    handleVmciRequests()
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 672, in handleVmciRequests
    details["Name"], opts)
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 370, in executeRequest
    response = detachVMDK(vmdk_path, vm_name)
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 320, in detachVMDK
    return disk_detach(vmdk_path, vm)
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 613, in disk_detach
    wait_for_tasks(si, [vm.ReconfigVM_Task(spec=spec)])
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 764, in wait_for_tasks
    raise task.info.error
vim.fault.InvalidState: (vim.fault.InvalidState) {
   dynamicType = <unset>,
   dynamicProperty = (vmodl.DynamicProperty) [],
   msg = 'The operation is not allowed in the current state.',
   faultCause = <unset>,
   faultMessage = (vmodl.LocalizableMessage) []
}
[...]

Issue #374, opened 2016-05-19
Plugin treats delta disks as docker volumes


pdhamdhere:
Steps to Repro:
See #373 Failed snapshot left behind delta files in dockvols;

- ESX
```
[root@wdc-vmgft-203-dhcp239:/vmfs/volumes/4bad174b-b67f5cfc-2fc6-0026b966a303/dockvols] ls -al *v6*
-rw-------    1 root     root          4096 May 19 05:50 v6-000001-delta.vmdk
-rw-------    1 root     root           273 May 19 05:50 v6-000001.vmdk
-rw-------    1 root     root           161 May 19 05:49 v6-7bc3bf935f99d29e.vmfd
-rw-------    1 root     root     104857600 May 19 05:50 v6-flat.vmdk
```
- VM:
```
root@burningman:~# docker volume ls
DRIVER              VOLUME NAME
...
vmdk                v6
vmdk                v6-000001
```

Admin CLI too is not happy with delta VMDKs since there is no KV file for v6-000001 
@andrewjstone Should we gracefully handle this case rather than crashing CLI? Not for TP.
```
vmdkops_admin ls -l
Traceback (most recent call last):
  File "/bin/vmdkops_admin", line 540, in <module>
    main()
  File "/bin/vmdkops_admin", line 37, in main
    args.func(args)
  File "/bin/vmdkops_admin", line 319, in ls
    (header, data) = ls_dash_l()
  File "/bin/vmdkops_admin", line 341, in ls_dash_l
    rows = generate_ls_dash_l_rows()
  File "/bin/vmdkops_admin", line 383, in generate_ls_dash_l_rows
    attached_to = get_attached_to(metadata)
  File "/bin/vmdkops_admin", line 406, in get_attached_to
    if metadata[u'status'] == u'attached':
TypeError: 'NoneType' object has no attribute '__getitem__'
```

PR #433, opened 2016-06-01
Align kv files to 4k boundary and misc fixes.

govint:
Aligns the kv side car, padding to the next 4K boundary. This isn't an issue at present with  independent disks, SVM and clone works fine as is. But dependent disks will fail and hence this related issue is being fixed for beta.

Testing Done:
Tested snapshot of a VM with docker volumes attached (as persistent disks - dependent) and able to take the snap successfully.

### DIFF
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -35,10 +35,12 @@ lib = None
 useSideCarCreate = False
 DVOL_KEY = "docker-volume-vsphere"
 
-# Maps to OPEN_BUFFERED | OPEN_LOCK | OPEN_NOFILTERS
+# Results in a buffered, locked, filter-less open,
 # all vmdks are opened with these flags
 VMDK_OPEN_FLAGS = 524312
 
+# Default kv side car alignment
+KV_ALIGN = 4096
 
 # Load the disk lib API library
 def loadDiskLib():
@@ -120,7 +122,7 @@ def volOpenPath(volpath):
 
 
 # Create the side car for the volume identified by volpath.
-def create(volpath, kvDict):
+def create(volpath, kv_dict):
     disk = c_uint32(0)
     objHandle = c_uint32(0)
 
@@ -144,7 +146,7 @@ def create(volpath, kvDict):
     lib.DiskLib_SidecarClose(disk, DVOL_KEY, byref(objHandle))
     lib.DiskLib_Close(disk)
 
-    return save(volpath, kvDict)
+    return save(volpath, kv_dict)
 
 
 # Delete the the side car for the given volume
@@ -166,35 +168,44 @@ def delete(volpath):
     return True
 
 
+# Align a given string to the specified block boundary.
+def align_str(kv_str, block):
+   # Align string to the next block boundary. The -1 is to accommodate
+   # a newline at the end of the string.
+   aligned_len = int((len(kv_str) + block - 1) / block) * block - 1
+   return '{:<{width}}\n'.format(kv_str, width=aligned_len)
+
+
 # Load and return dictionary from the sidecar
 def load(volpath):
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
 
     try:
-        fh = open(metaFile, "r+")
-    except IOError:
-        logging.exception("Failed to open %s", metaFile);
+       with open(metaFile, "r") as fh:
+          kv_str = fh.read()
+    except:
+        logging.exception("Failed to access %s", metaFile);
         return None
 
-    kvDict = json.load(fh)
-
-    fh.close()
-
-    return kvDict
+    try:
+       return json.loads(kv_str)
+    except ValueError:
+       logging.exception("Failed to decode meta-data for %s", volpath);
+       return None
 
 
 # Save the dictionary to side car.
-def save(volpath, kvDict):
+def save(volpath, kv_dict):
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
 
+    kv_str = json.dumps(kv_dict)
+
     try:
-        fh = open(metaFile, "w+")
-    except IOError:
-        logging.exception("Failed to open %s", metaFile);
+       with open(metaFile, "w") as fh:
+          fh.write(align_str(kv_str, KV_ALIGN))
+    except:
+        logging.exception("Failed to save meta-data for %s", volpath);
         return False
 
-    json.dump(kvDict, fh)
-    fh.write("\n")  # newline makes it easier for human to read
-
-    fh.close()
     return True
+


esx_service/vmdk_ops.py
```diff
@@ -314,15 +314,25 @@ def find_child(vm_name):
 
 #returns error, or None for OK
 def attachVMDK(vmdk_path, vm_name):
-    vm = findVmByName(vm_name)
+    try:
+       vm = findVmByName(vm_name)
```
pdhamdhere: Should we rather change findVmByName to return error than throwing exception? @andrewjstone

esx_service/vmdk_ops.py
```diff
@@ -590,6 +600,7 @@ def disk_attach(vmdk_path, vm):
     spec = vim.vm.ConfigSpec()
     spec.deviceChange = disk_changes
 
+    # Handle the most generic exception for a VM reconfig.
```
pdhamdhere: I doubt if this comment is giving reader any addition information. I suggest dropping it from here and other places.

esx_service/vmdk_ops.py
```diff
@@ -634,9 +645,10 @@ def disk_detach(vmdk_path, vm):
     dev_changes.append(disk_spec)
     spec.deviceChange = dev_changes
 
+    # Handle the most generic exception for a VM reconfig.
     try:
         wait_for_tasks(si, [vm.ReconfigVM_Task(spec=spec)])
-    except vim.fault.GenericVmConfigFault as ex:
+    except vim.fault.VimFault as ex:
```
pdhamdhere: It seems consistent to change GenericVmConfigFault =>VimFault. However, this doesn't really fix #381.

Unrelated to your change but does `logging.warning(f.message)` logs exception code too?

govint: There is nothing that can be done about a VM changing state. The plugin can
at the most handle any errors that can cause it to crash. Other than that,
the plugin has no way to handle VM state changes.  If a VM is powering off
or reset or suspended then the VM reconfig will fail. The design for
handling configuration changes to the VM will have be a separate issue.
I'll raise that.

This issue is only going to prevent plugin-server exit because of unhandled
exception.

esx_service/vmdk_ops.py
govint: Its got to be handled somewhere and in fact I see no point raising an
exception in findChild() thats only going to be ignored. The plugin-server
has no idea how to handle any of the exceptions thrown by the ESX except
create an error message and fail the requested op. It does cause issues in
at least the detach case.

esx_service/vmdk_ops.py
govint: Lets take a step back, once we catch the exception and prevent plugin crash whats going to happen is that the plugin has rightfully failed the request. Following can happen,

1. The VM (docker) may or may not have lost the context about the volume. If docker still remembers that the volume is attached then its a docker issue to detach the volume (doubt it though).

2. If docker doesn't have any context of the volume detach having failed, which would be a docker bug btw, then we can't expect the plugin to keep state of the failure. The volume remains attached (meta-data also) and will need something like VC/vim-cmd as the user chooses, to remove the volume from the Vm config.

3. The plugin is by design stateless w.r.t errors that happen beyond its control. Its like a VC user initiates a reconfig and the ESX crashes, we have no idea if the op went thru or not.

For this issue I see no more change except to handle the correct exception and avoid crashing.

esx_service/utils/kvESX.py
```diff
@@ -175,24 +185,28 @@ def load(volpath):
     except IOError:
         return None
 
-    kvDict = json.load(fh)
-
+    kv_str = fh.read()
     fh.close()
+    try:
+       kv_dict = json.loads(kv_str)
```
msterin: why this try/except is needed ? Does not upstairs need to check for return value anyways, so it could haqve handles exceptions instead?  I think we need to catch exception only when a decision is needed, and up to this moment we can simply pass it up (which is default behavior)

esx_service/vmdk_ops.py
```diff
@@ -314,15 +314,25 @@ def find_child(vm_name):
 
 #returns error, or None for OK
 def attachVMDK(vmdk_path, vm_name):
-    vm = findVmByName(vm_name)
+    try:
+       vm = findVmByName(vm_name)
+    except Exception as e:
+       msg = "VM {0} not found".format(vm_name)
```
msterin: catching generic exception is frown upon in python. We should either stick with C-style error code unrolling, or use specific exceptions

esx_service/vmdk_ops.py
```diff
@@ -314,15 +314,25 @@ def find_child(vm_name):
...
                  vmdk_path, vm_name, vm.config.uuid)
     return disk_attach(vmdk_path, vm)
 
 
 #returns error, or None for OK
 def detachVMDK(vmdk_path, vm_name):
-    vm = findVmByName(vm_name)
+    try:
+       vm = findVmByName(vm_name)
+    except Exception as e:
```
msterin: I agree with Prashant - it looks strange to mix-n-match exceptions and return code without any visible system
We can agree on "catch exceptoins as soon as they happen and pass error code around" or "catch exceptions only when decision needs to be made". findVmByName handling mixes both - I think it needs to be changed to a single style

msterin:
This PR  seems to contains the following: 
*  some good cosmetic changes and comments (which I agree with)
* KV alignment , which seems to provide  incorrect alignment, and IMO is not needed  until we have at least one real use case which tests it. As far as I know there are **no** use cases we currently support which need KV alignment (since we don't take volume snapshots, and clone/SVM is blocked anyways). However I would be fine with this sectoin if there was a set of unit tests. 
* exception handling change - I do not understand the reason or the impact - what specifically is fixed or changed by this ?

esx_service/vmdk_ops.py
govint: In which case is there a point to findChild() re-raising an exception at
all?

govint:
- some good cosmetic changes and comments (which I agree with)
   - KV alignment , which seems to provide incorrect alignment, and IMO is
   not needed until we have at least one real use case which tests it. As far
   as I know there are *no* use cases we currently support which need KV
   alignment (since we don't take volume snapshots, and clone/SVM is blocked
   anyways). However I would be fine with this sectoin if there was a set of
   unit tests.

Since you had raised this issue, do we close as won't fix, defer till vvol
is supported (vvol needs it)? I've not added tests because this will be
tested by default on any create/attach/detach req to the plugin. There
can't be unit tests to do snapshot unless the disks are dependent.

   - exception handling change - I do not understand the reason or the
   impact - what specifically is fixed or changed by this ?

Thats confusing, on the one hand we have code that raising exceptions which
if they were to occur won't be handled and cause the plugin to crash. Are
you saying there is no need to fix code that's not handling exceptions.

esx_service/vmdk_ops.py
govint: Sorry, is there a point in findVmByName() raising an exception, I'd
mentioned in response to Prashant as well, unless we are going to be
handling exceptions differently there is no point in raising exceptions in
the plugin server.

esx_service/utils/kvESX.py
govint: The try-except is in case the kv_str is a null dict from json.loads. Like
if the meta-data is corrupt.

esx_service/vmdk_ops.py
govint: We have no idea what exceptions ESX is going to return save VimFault which
is a catch all and how does it help  catching a VimFault and ignoring it.

esx_service/vmdk_ops.py
govint: I'd prefer that we don't raise exceptions anywhere in the plugin code and
instead handle the exception at the point that we could get one and
translate to an appropriate error return. In which case findVmByName()
mustn't be raising exceptions.

andrewjstone:
I don't have a particular problem changing code to immediately catch exceptions and return errors instead of catching them higher up, but this does make tend to make error handling a little more verbose and un-pythonic. I do wonder what we are supposed to do when we catch an exception that we don't know how to handle though. Exceptions are nice in that they report detailed tracebacks about where the error occurred. I'm not sure we want to give that up. 

I also doubt we will successfully catch all runtime exceptions in code systematically without running into them first when our code crashes during testing or in production. While we can certainly handle them at that point, we should really have proper supervision of the process, so when it crashes it is restarted automatically and we can fix the code later.

esx_service/vmdk_ops.py
andrewjstone: The reason to ignore it is so that we know what it is and get a traceback on crash to help us debug later. What are we catching and converting to an error code otherwise? I didn't see a good way to do this systematically.

msterin:
@andrewstone - I agree, and you said what I thought, only eloquently :-). We dhould strive to use exceptions and catch it up the call stack, when a decision needs to be made IMO , for reasons you have mentioned 

What concerns me is that we do heavy mix-n-match and this PR brings more exception handlers down and up the stack. 

Still, my main objection is that alignment calc seems to be broken , plus I do not think we need it today, and even if we let it I it should have unit tests

msterin:
@govint - yes, we can close "4k aligned" as won't fix, or (my preference) push it out and link to "snapshots" discussion. I raised the issue when snapshot was failing. With "independent" disk it's no longer a problem .
But the rest of your PR (cleanup/cosmetic, and exceptions) are independent, the cleanup is LGMT, and the exceptions I 'd like to take a look as a separate PR - could be some minor comments

On the alignment - unit tests DO NOT require attach/detach. Just checking that your alignment function produces strings , and safes them in json files. of expected size should not be too hard.

govint:
Mark, can you say why the alignment calculation is broken.

govint:
Ok, like after a test - do ls -l of the side car to show its 4K aligned.
Didn't strike me as a unit test.

govint:
This is what the code is generating with 4K aligned side cars.

Than keep commenting that the alignment calculation is broken I'd like to
know exactly how its broken. The dict thats been saved is padded with
spaces to the next 4K boundary (what ever the size of the meta-data).

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols]
ls -l clonevol-7*

-rw-------    1 root     root          4096 Jun  1 05:38
clonevol-7-f4f403117dcdc36d.vmfd

-rw-------    1 root     root     104857600 Jun  1 05:20
clonevol-7-flat.vmdk
-rw-------    1 root     root           566 Jun  1 05:20 clonevol-7.vmdk

govint:
@andrewjstone @msterin, on the issue of handling exceptions or not, do we decide that errors must happen and bugs created to go figure out that an exception must be handled. That way most of try-except chunks in the server code have never been seen before they were added up nor do we have "unit tests" for any of those.

Thanks for the sermon though.

esx_service/utils/kvESX.py
msterin: let me rephrase. Why catch here (where we just return None to be taken care upstairs) instead of catching upstairs and do whatever upstairs is doing on "none" result ?

msterin:
about the "broken" part: 

The code was hard to read and looked suspicious, I tried to follow it manually and did not get aligned result - so I simply  I cut-n-pasted the code and got the following result:

```python
>>> def align_str(kv_str):
...   aligned_len = ((len(kv_str) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN - 1
...   padded_str = '{:<{width}}'.format(kv_str, width=aligned_len)
...   return '{0}{1}'.format(padded_str, '\n')
...
>>> KV_ALIGN=4096
>>> s = align_str("hello")
>>> len(s)
4100
>>>
```

4100 is not aligned on 4096.
I have no idea why json.save generates 4K aligned pieces, but align_str does not behave the way I expected it to behave

---
About the exceptions: as @andrewjstone has mentioned , handling exception at the earliers point tends to generate C-style routing of error code though all call chain, and is non-python-ish (and not exception-ish). So I'd like to move toward handling them higher the stack when a decision needs to be made. It's not a hard stop , just a desired direction

msterin:
oops... accidentally pushed the wrong button, I did not mean to close. Reopening

esx_service/utils/kvESX.py
```diff
@@ -166,6 +168,14 @@ def delete(volpath):
     return True
 
 
+def align_str(kv_str):
+   # Align a given KV meta-data string to the next 4K boundary.
+   aligned_len = ((len(kv_str) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN - 1
+   padded_str = '{:<{width}}'.format(kv_str, width=aligned_len)
+
+   # Add a trailing newline to bring up string length to a 4K boundary.
+   return '{0}{1}'.format(padded_str, '\n')
```
msterin: why not use default (space) for padding ? Something like 
`('{:<%d}` % round_up(len(kv_str, 4096))).format(kv_str)`
(assuming `def round_up(n, block): return (( n + block - 1 ) / block ) * block`

msterin:
An update: now we do have a use case (veeam backup). I guess as long as the snapshot it collapsed (deleted) after backup, we can use dependent disk - it the snapshot is not collapsed, the volume will stick to the VM forever (well, until snapshot exists). 
We still need test coverage to merge this. I suggest checking if creating a child disk would impact the sidecars - could be a good test.

govint:
Mark, agree. I'll run this test and update. Looks like I still need to
rebase and generate a fresh set of changes for the PR.

govint:
Mark, tried the same test as you did. How is ths working different,

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/service] python
Python 2.7.10 (default, Dec 30 2015, 10:42:45)
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> KV_ALIGN=4096
>>> def align_str(kv_str):
...    aligned_len = ((len(kv_str) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN - 1
...    padded_str = '{:<{width}}'.format(kv_str, width=aligned_len)
...    return '{0}{1}'.format(padded_str, '\n')
... ^D

>>> s='hello'
>>> len (align_str(s))
4096

>>> s = align_str("hello")
>>> len(s)
4096
>>>


Tried by cut pasting your earlier update,

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/service] python
Python 2.7.10 (default, Dec 30 2015, 10:42:45)
[GCC 4.6.3] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> def align_str(kv_str):
...    aligned_len = ((len(kv_str) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN - 1
...    padded_str = '{:<{width}}'.format(kv_str, width=aligned_len)
...    return '{0}{1}'.format(padded_str, '\n')
... ^D
>>> KV_ALIGN=4096
>>> s = align_str("hello")
>>> len(s)
4096
>>>

govint:
Tested a snapshot (opened up some more issues) with a "persistent" mode disk attached via the plugin (modified vmdk_ops.py to attach disk as "persistent" which makes it a dependent disk"

1. VM configured with a docker volume and as "dependent".

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/2] grep vmdk *vmx
scsi0:0.fileName = "2-000001.vmdk"
scsi3:0.fileName = "2_1.vmdk"
scsi3:1.fileName = "/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/clonevol-20.vmdk"
scsi3:2.fileName = "/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/clonevol-21.vmdk"
scsi3:3.fileName = "/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/clonevol-23.vmdk"    <------------------   dependent disk

2. Docker volume has a nice 4K aligned side car.

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/2] ls -l ../dockvols/clonevol-23*
-rw-------    1 root     root          4096 Jun  8 11:04 ../dockvols/clonevol-23-adcca9bd68a9990c.vmfd
-rw-------    1 root     root     104857600 Jun  8 10:52 ../dockvols/clonevol-23-flat.vmdk
-rw-------    1 root     root           568 Jun  8 10:52 ../dockvols/clonevol-23.vmdk

3. After the snapshot we have a delta-disk created for the docker volume:

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/2] grep scsi3:3 *vmx
scsi3:3.deviceType = "scsi-hardDisk"
scsi3:3.fileName = "/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols/clonevol-23-000001.vmdk"   <----------   delta disk
scsi3:3.present = "true"
scsi3:3.redo = ""

4. Issue: Admin CLI shows the delta-disk (should it?)

clonevol-23         bigone     2           Wed Jun  8 10:52:35 2016  2            N/A     100.00MB  14.00MB
clonevol-23-000001  bigone     2           Wed Jun  8 10:52:35 2016  2            N/A     100.00MB  8.00KB

5. Issue: docker volume ls also shows the delta disks (it shouldn't - these are internal disks and the user didn't create these)

vmdk                clonevol-23
vmdk                clonevol-23-000001

6. Side car is properly snapped too.
[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols] ls -l clonevol-23*
-rw-------    1 root     root          4096 Jun  8 11:09 clonevol-23-000001-7b39b103c85cf656.vmfd
-rw-------    1 root     root          4096 Jun  8 11:09 clonevol-23-000001-delta.vmdk
-rw-------    1 root     root           462 Jun  8 11:09 clonevol-23-000001.vmdk
-rw-------    1 root     root          4096 Jun  8 11:04 clonevol-23-adcca9bd68a9990c.vmfd
-rw-------    1 root     root     104857600 Jun  8 10:52 clonevol-23-flat.vmdk
-rw-------    1 root     root           568 Jun  8 10:52 clonevol-23.vmdk
[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/dockvols]

pdhamdhere:
Thanks, for running test.

> Issue: Admin CLI shows the delta-disk (should it?)
> Issue: docker volume ls also shows the delta disks

Yes. Captured in #374

msterin:
> Mark, tried the same test as you did. How is ths working different

Good question. It looks like I was testing on Python3.... and the 
code relies on Python 2 int arithmetic - so it does not work on Python 3. Since ESX is moving to python3 it needs to run on both.

The suggested code works on both Python 2 and Python 3, and (IMO :-)) is easier to read too.

```python
def round_up(n, block_size):
  '''aligns <n> on <block_size> by rounding up'''
  return int(( n+ block_size- 1 ) / block_size) * block_size

kv_str="test"
padded_str = ('{:<%d}' % round_up(len(kv_str), 4096)).format(kv_str)

len(padded_str)
```

esx_service/utils/kvESX.py
```diff
@@ -175,24 +185,26 @@ def load(volpath):
...
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
 
     try:
         fh = open(metaFile, "w+")
     except IOError:
         return False
 
-    json.dump(kvDict, fh)
-    fh.write("\n")  # newline makes it easier for human to read
+    kv_str = json.dumps(kv_dict)
```
msterin: if json or write throws, this will leave file open.

esx_service/utils/kvESX.py
```diff
@@ -175,24 +185,26 @@ def load(volpath):
     except IOError:
         return None
 
-    kvDict = json.load(fh)
-
+    kv_str = fh.read()
```
msterin: if read throws. this will leave file open

esx_service/utils/kvESX.py
```diff
@@ -175,24 +185,26 @@ def load(volpath):
...
     fh.close()
-
-    return kvDict
+    if kv_str: 
+       return json.loads(kv_str)
```
msterin: can json.loads throw ? if it does, is it expected upstairs ?

esx_service/utils/kvESX.py
govint: Thanks I'll post the changes

esx_service/utils/kvESX.py
```diff
@@ -175,24 +185,26 @@ def load(volpath):
...
+    else:
+       return None
 
 
 # Save the dictionary to side car.
-def save(volpath, kvDict):
+def save(volpath, kv_dict):
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
```
govint: That's why I'm my original change I was handling the exception here. No, calling code doesn't handle exceptions at all. These modules don't define exceptions as yet for any of these APIs, which may be a todo via a separate issue.

pdhamdhere:
@govint Looks like you have updated PR to only deal with KV alignment issue. Can you please update PR summary & description to reflect same. Also include manual test that you did under "Testing Done" in description. Thanks for making this change.

govint:
Yes since we decided on one change only for a PR which is easier. I'll make
the updates

govint:
Debugged this and such a weird way to do integer division. Python 3 changed
division to floating point which is why this code was broken there..

>>> ((len(s) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN - 1
4097.0
>>> ((len(s) + (KV_ALIGN - 1)) / KV_ALIGN) * KV_ALIGN
4098.0
>>> ((len(s) + (KV_ALIGN - 1)) / KV_ALIGN) <---------------------   gives a
floating point result

1.00048828125

>>> ((len(s) + (KV_ALIGN - 1)) // KV_ALIGN) <-----------   The double //
results ini integer division
1

>>> def align_str(kv_str):
...       aligned_len = ((len(kv_str) + (KV_ALIGN - 1)) // KV_ALIGN) *
KV_ALIGN - 1 <--------- use a double /
...       print (aligned_len)
...       padded_str = '{:<{width}}'.format(kv_str, width=aligned_len)
...       return '{0}{1}'.format(padded_str, '\n')
...
>>> KV_ALIGN = 4096
>>> s='zzz'
>>> t =align_str(s)
4095
>>> len(t)
4096

govint:
Tested snapshot of a VM with docker volumes attached (as persistent disks - dependent) and able to take the snap successfully.

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,45 @@ def delete(volpath):
     return True
 
 
+# Align a given string to the specified block boundary.
+def align_str(kv_str, block):
+   # Align string to the next block boundary.
+   new_len = int((len(kv_str) + block - 1) / block) * block - 1
+   padded_str = '{:<{width}}'.format(kv_str, width=new_len)
```
msterin: why not simply
```python 
return  '{:<{width}}\n'.format(kv_str, width=new_len)
```
instead of lines 175-178 ?

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,45 @@ def delete(volpath):
     return True
 
 
+# Align a given string to the specified block boundary.
+def align_str(kv_str, block):
+   # Align string to the next block boundary.
```
msterin: nit: please add "-1" is to accommodate extra \n" to the comments - I was really confused by the -1 until I read the follow up

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,45 @@ def delete(volpath):
...
+   new_len = int((len(kv_str) + block - 1) / block) * block - 1
```
msterin: nit: would be better to name it aligned_len (just a nit, it's fine as is too)

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,45 @@ def delete(volpath):
...
+
 # Load and return dictionary from the sidecar
 def load(volpath):
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
 
     try:
-        fh = open(metaFile, "r+")
-    except IOError:
-        logging.exception("Failed to open %s", metaFile);
+       with open(metaFile, "r+") as fh:
```
msterin: why 'r+', not 'r' ? we don't see to be writing anything, so 'r' should work, should not it ?

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,45 @@ def delete(volpath):
...
+def save(volpath, kv_dict):
     metaFile = lib.DiskLib_SidecarMakeFileName(volpath, DVOL_KEY)
 
+    kv_str = json.dumps(kv_dict)
+
     try:
-        fh = open(metaFile, "w+")
-    except IOError:
-        logging.exception("Failed to open %s", metaFile);
+       with open(metaFile, "w+") as fh:
```
msterin: why 'w+' not 'w' ?

pdhamdhere:
LGTM.  I took liberty to update PR description.

esx_service/utils/kvESX.py
govint: The function is returning a string thats ready to be written. With backend like VVOL every write has to be aligned to 4k boundary and also sized to a 4K multiple. Else caller still needs to add a newline and then issue a single IO thats aligned to a 4K IO size and also to a 4K boundary (in this case its always offset 0).

esx_service/utils/kvESX.py
govint: For both r+ and w+ this is left over from earlier when the code was being developed and was reading and updating the meta-data in the same functions. Its not needed.

esx_service/utils/kvESX.py
msterin: the single line I suggested is returning **exactly** the same string (including newline), so I am not sure what are you talking about

esx_service/utils/kvESX.py
msterin: I am not sure I understand the answer/ Are you saying it's a leftover and you'll make it "w" and "r", or are you saying the "+" is needed? if it's the latter, I still don't get why?

esx_service/utils/kvESX.py
govint: Missed the \n in your suggestion, in which case thats better.

esx_service/utils/kvESX.py
govint: No, "r" and 'w" it is.

esx_service/utils/kvESX.py
```diff
@@ -166,35 +168,44 @@ def delete(volpath):
...
-    kvDict = json.load(fh)
-
-    fh.close()
-
-    return kvDict
+    try:
+       return json.loads(kv_str)
+    except ValueError:
+       logging.exception("Failed to decode meta-data for %s", volpath);
+       return None
```
msterin: nit: in both 'returns' in this function, just 'return' would be better (to indicate no return value is expected)

msterin:
LGTM (with a small nit, and assuming CI passes - it current failure is unrelated to this PR, but we still need to get a pas)

Issue #206, opened 2016-03-25
VMCI failures are all reported as "code: -1. can't connect to ESX". 


msterin:
Both vmci_client.c and vmci_server.c  return "-1" on any error, and we report it to user (and log) as "Can't connect to ESX" or "Peer connection dropped".

Error code is now passed back in VMCI packet, and if empty string is passed it's considered Success.
At a minimum, this API and error handling has to be documented and the error string needs to be bubbled up to logs

/CC @govint

PR #435, opened 2016-06-02
Report errno and proper strings on vSocket communcation failures

msterin:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/206

* Client code passes errno up and GO code handles it.
  CGO auto-returns (return_code, error) tuple and error info was there all along, we just use it now.
* Server code makes sure  errno is properly set (and not cleaned up by close() and the likes) ,
  and sets errno when needed. Python code then generates proper messages
* **removed uneeded custom error codes - it turned out to be throwaway code**
* a few cosmetic changes - mainly dropped single letter variables since they are hard to  maintain

Tested with CI and 'make all'. **some** error codes tested manually, to make sure the path works: 

* **Run python service while it is already up**. Log says:
```
05/14/16 15:19:51 562104 [ERROR  ] Failed to initialize vSocket listener: Address already in use (errno=98)
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 725, in main
    handleVmciRequests()
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 667, in handleVmciRequests
    %  (os.strerror(errno), errno))
OSError: Failed to initialize vSocket listener: Address already in use (errno=98)
````

* **Run `docker volume ls` when the ESX service is down.** Docker says 
```
root@ubuntu-srv14:~# docker volume ls
list vmdk: VolumeDriver.List: 'list' failed: connection reset by peer (errno=104). Check that ESX service is running.
root@ubuntu-srv14:~# 
```

### CODE REVIEW

File: esx_service/vmdk_ops.py [hunk 1]
```diff
@@ -695,9 +702,11 @@ def handleVmciRequests():
             opts = details["Opts"] if "Opts" in details else {}
             ret = executeRequest(vm_name, cfg_path, req["cmd"],
                                  details["Name"], opts)
-            logging.debug("executeRequest ret = %s", ret)
+            logging.info("executeRequest '%s' completed with ret=%s",
+                         req["cmd"], ret)
 
         response = lib.vmci_reply(c, c_char_p(json.dumps(ret)))
+        errno = get_errno()
```
andrewjstone: What are we doing with this errno? Should we log it?

andrewjstone:
LGTM

esx_service/vmdk_ops.py
msterin: yep, I missed this. Will add. Thanks !

esx_service/vmci/vmci_client.c
```diff
@@ -175,6 +172,7 @@ vsock_get_family(void)
 }
 
 // Create and connect VMCI socket.
+// return success (0) or failure (-1)
```
pdhamdhere: nit: return CONN_SUCCESS (0) or CONN_FAILURE(-1)

esx_service/vmci/vmci_client.c
```diff
@@ -183,13 +181,12 @@ vsock_init(be_sock_id *id, int cid, int port)
    int sock;     // socket id
 
    if ((af = vsock_get_family()) == -1) {
-      perror("Failed to get address family.");
-      return CONN_FAILED_VMCI_ADDRESS_FAMILY_GET;
+      errno = EAFNOSUPPORT;
```
pdhamdhere: vsock_get_family sets errno. Why do we have to set it again here?

esx_service/vmci/vmci_server.c
```diff
@@ -68,18 +70,17 @@ vmci_init(void)
...
 
    /*
     * Open a STREAM socket using our address family.
     */
 
-   s = socket(af, SOCK_STREAM, 0);
-   if (s == -1) {
+   socket_fd = socket(af, SOCK_STREAM, 0);
+   if (socket_fd == -1) {
       perror("Failed to open socket");
```
pdhamdhere: Do you need this perror? What if perror runs into issues in opening STDERR and resets errno? Also errno valid after perror returns successfully?

esx_service/vmci/vmci_server.c
```diff
@@ -33,32 +32,35 @@
...
    }
+
    return af;
 }
 
-// returns socket to listen to, or -1
-// also fills in af and vmciFd
+// Returns vSocket to listen on, or -1.
+// errno indicates the reason for a failure, if any.
 int
```
pdhamdhere: returns be_sock_status and change comment as well in this and other functions.

esx_service/vmdk_ops.py
```diff
@@ -654,26 +654,33 @@ def signal_handler_stop(signalnum, frame):
...
     # Load and use DLL with vsocket shim to listen for docker requests
-    lib = cdll.LoadLibrary(os.path.join(LIB_LOC, "libvmci_srv.so"))
+    lib = CDLL(os.path.join(LIB_LOC, "libvmci_srv.so"), use_errno=True)
 
     bsize = MAX_JSON_SIZE
     txt = create_string_buffer(bsize)
 
     cartel = c_int32()
     sock = lib.vmci_init()
+    if sock == -1:
```
kerneltime: Use a define instead of -1?

esx_service/vmci/vmci_client.c
msterin: done

esx_service/vmci/vmci_client.c
msterin: Actually i twas not set in client .It is a leftover from totally separate client/server files. 
I moved the vsock_get_family to the header fie (so it's shared) and dropped extra errno set in both client and server

esx_service/vmci/vmci_server.c
msterin: in these cases it's ignored. I want to keep it here as it allows to run the service in the foreground and see the messages without opening/traking the log. If you feel strongly about it, I'll remove

esx_service/vmci/vmci_server.c
msterin: No, this one **does** return socket to listen on

esx_service/vmdk_ops.py
msterin: done

kerneltime:
LGTM but this code needs another round of clean up. Will see if I can get it done as part of the capabilities change.

PR #438, opened 2016-06-06
Make the vsocket port configurable.

kerneltime:
The port is passed in as a command line param.
VM: Customer can use standard techniques for systemd and upstart to
change the startup scripts for port.
ESX: Since the init scripts are readonly, an optional file is read
which exports the port number override.

Testing: Running basic commands on dev box using the new port.
CI will use new ports as well.

### DIFF
--- a/esx_service/vmdk-opsd
+++ b/esx_service/vmdk-opsd
@@ -18,8 +18,6 @@ OPSD_GROUP_NAME="vmdkops"
 OPSD_GROUP="$OPSD_GROUP_PATH/$OPSD_GROUP_NAME"
 OPSD_SCHED_PARAM="++memreliable,group=$OPSD_GROUP"
 
-# Pass these params to service.
-OPSD_PARAMS="--log-level=info"
 
 LOCAL_CLI_SCHED="localcli --plugin-dir=/usr/lib/vmware/esxcli/int sched group"
 
@@ -29,6 +27,14 @@ MAX_RETRY=10
 MAX_QUICK_FAILURES=5
 SIGTERM=15
 SIGKILL=9
+
+# Config
+VMDK_OPSD_PORT=1019 # Override using CONFIG_FILE
+
+# Create the following file if defaults need to be overridden
+# Example:
+# export VMDK_OPSD_PORT=1020
+CONFIG_FILE=/etc/vmware/vmdkops/service_config.sh
 
 # The numbers below are to setup the framework for
 # resource limits but not setting them up. -1 is unlimited
@@ -47,6 +53,12 @@ getpid() {
 }
 
 start() {
+   if [ -f $CONFIG_FILE ]
+   then
+      echo "Loading $CONFIG_FILE"
+      . $CONFIG_FILE
+   fi
+
    local PID=$(getpid)
    if [ -n "${PID}" ]; then
       echo "Failed to start. $OPSD_TAG already running."
@@ -64,6 +76,9 @@ start() {
       fi
    fi
 
+
+   # Pass these params to service.
+   OPSD_PARAMS="-p $VMDK_OPSD_PORT"
 
    ${LOCAL_CLI_SCHED} setmemconfig -g ${OPSD_GROUP} --min=${MINMEM} --max=${MAXMEM} --minlimit=${MINLIMIT} -u mb
    ${LOCAL_CLI_SCHED} setcpuconfig -g ${OPSD_GROUP} --min=${MINCPU} --max=${MAXCPU} -u pct
@@ -111,13 +126,13 @@ stop() {
          check_running_status
          exit 4
       fi
-   fi 
+   fi
 
    check_running_status
    if [ $? -eq 0 ]; then
       echo "Failed to stop $OPSD_TAG"
       exit 5
-   fi 
+   fi
    return 0
 }
 
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -30,18 +30,18 @@ Commands ("cmd" in request):
 
 '''
 
-from ctypes import *
+import atexit
+import getopt
 import json
+import logging
 import os
 import os.path
-import subprocess
-import atexit
-import time
-import logging
-import signal
-import sys
 import re
+import signal
+import subprocess
+import sys
 import time
+from ctypes import *
 
 from vmware import vsi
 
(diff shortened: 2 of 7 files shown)


esx_service/vmdk-opsd
```diff
@@ -30,6 +28,14 @@ MAX_QUICK_FAILURES=5
 SIGTERM=15
 SIGKILL=9
 
+# Config
+VMDK_OPSD_PORT=1019 # Override using CONFIG_FILE
```
msterin: why 1019 ? Just curious :-) ?

kerneltime: popped up in my head..

esx_service/vmdk-opsd
```diff
@@ -30,6 +28,14 @@ MAX_QUICK_FAILURES=5
...
+
+# Create the following file if defaults need to be overridden
+# Example:
+# export VMDK_OPSD_PORT=1020
+CONFIG_FILE=/etc/vmdk_opsd/config.sh
```
msterin: our log config is in `/etc/vmware/vmdkops/log_config.json`
I think  the config.sh should be in the same place, and desirably named using the same convention (e.g. service_config.sh to differentiate from log config) 
Or log_config could move

esx_service/vmdk-opsd
```diff
@@ -47,6 +53,13 @@ getpid() {
 }
 
 start() {
+   echo $CONFIG_FILE
```
msterin: why do we need the echo ?

esx_service/vmdk_ops.py
```diff
@@ -715,18 +715,33 @@ def handleVmciRequests():
...
     lib.close(sock)  # close listening socket when the loop is over
 
+def usage():
+    print "Usage: %s -p <vSocket Port to listen on>" % sys.argv[0]
 
 def main():
     log_config.configure()
-    logging.info("=== Starting vmdkops service ===")
-
+    logging.info("=== Starting vmdkops service PID====")
```
msterin: Why adding 'PID' string?

misc/scripts/install/docker-volume-vsphere.conf
```diff
@@ -7,4 +7,4 @@ respawn
 
 kill timeout 20
 
-exec /usr/local/bin/$UPSTART_JOB
+exec /usr/local/bin/$UPSTART_JOB -port 1019
```
msterin: (not an issue, just a question) - do we need to pass port number ? Can the  default be hardcoded in client server

vmdk_plugin/main.go
```diff
@@ -96,11 +97,13 @@ func main() {
...
 	flag.Parse()
 
+	vmdkops.EsxPort = *port
+
 	logInit(logLevel, nil, configFile)
 
 	log.WithFields(log.Fields{
 		"version":   version,
-		"port":      *port,
+		"port":      vmdkops.EsxPort,
```
msterin: where would I configure it? I did not see a new flag for it...

msterin:
Looks good with the exception of config file location, and a better info on test (or even adding auto-test). Plus a few nits as usual :-)

esx_service/vmdk-opsd
kerneltime: makes sense. will change it.

esx_service/vmdk-opsd
kerneltime: will remove it.

esx_service/vmdk_ops.py
kerneltime: Yup. will clean it up, I had a PID print there that I did not clean up.

misc/scripts/install/docker-volume-vsphere.conf
kerneltime: This helps in debugging cause the PS lists the port.. considering the difficulty we have been having with tracking sockets and that the code is shipped as binary.

vmdk_plugin/main.go
kerneltime: Sure.

misc/scripts/install/docker-volume-vsphere.conf
msterin: I see - and agreed

esx_service/vmdk-opsd
pdhamdhere: Why not have one config which has both log and service specific config options?

esx_service/vmdk_ops.py
```diff
@@ -715,18 +715,33 @@ def handleVmciRequests():
...
     signal.signal(signal.SIGINT, signal_handler_stop)
     signal.signal(signal.SIGTERM, signal_handler_stop)
+    try:
+        port=1019
+        opts, args = getopt.getopt(sys.argv[1:], 'hp:')
```
pdhamdhere: Why not get this from "/etc/docker-volume-vsphere.conf" vs command line?

esx_service/vmdk-opsd
msterin: log config file is a standard python logging config, we just reuse it. 

I guess we can add something to json there as long as it is ignored by logging subsystem, or filter it out on read ourselves - but this adds more moving parts for questionable value IMO. I am not against this, just pointing out it's extra work and I am not sure it's worth it

esx_service/vmdk_ops.py
kerneltime: We do not ship that file by default. Also, ps will tell you the exact value vs. not being sure which version of config was loaded in..

kerneltime: Had a chat with Prashant. A config file might be nice for all things plugin.. have filed https://github.com/vmware/docker-volume-vsphere/issues/439 for the vmci clean up where I will look at the config file needs as well. Currently there is a config file that is only used for logging and also, there is no VMCI module which can be configured on power up that leaves either a global variable for port or change the function call signatures.. For now I am not changing this PR which is needed for take 2 of TP.

esx_service/vmdk-opsd
andrewjstone: Tis prime

misc/scripts/install/docker-volume-vsphere.conf
andrewjstone: What if the config uses a different port though? Won't this hardcoding be incorrect?

vmdk_plugin/main.go
andrewjstone: Should we change the default port on line 93?

misc/scripts/install/docker-volume-vsphere.conf
kerneltime: The config file is on ESX side and it overrides (but is a command line
param)
On VMware side there is only command line param.
If we do go to a config file model then that will be the only thing on the
vm side.

andrewjstone: Ok.

andrewjstone:
LGTM

misc/scripts/install/docker-volume-vsphere.conf
kerneltime: The current model to change the port on the VM side is to change the init scripts..

vmdk_plugin/main.go
kerneltime: Will do. Had left it there to catch bugs if the value is not being set.

PR #443, opened 2016-06-07
Change default log level to INFO in plugin

### LINKED ISSUE #396: Default log level should be INFO in plugin
currently it's DEBUG on Linux (main.go) , but already INFO on ESX (log_config.py)
We need to set it to INFO on both  ASAP after TP, per prior offline discussions

pdhamdhere:
Change default log level for plugin from "debug" to "info".

Testing done: Manually confirmed in logs that only info log statements are logged.

Fixes #396


msterin:
The fix is good- unfortunately it does break screen-scraping in misc/scripts/refcnt_test.sh, so it has to come together with test fix.

msterin:
LGTM

PR #447, opened 2016-06-08
Log exceptions in ESX service

pdhamdhere:
Log exceptions to make debugging easier. 

**Testing Done:**
Manually verified exceptions in log file for e.g.
- Before
```
[WARNING] Failed to open /vmfs/volumes/4bad174b-b67f5cfc-2fc6-0026b966a303/dockvols/myvol1.vmdk for descriptor check
```
- After
```
[ERROR  ] Failed to open /vmfs/volumes/4bad174b-b67f5cfc-2fc6-0026b966a303/dockvols/myvol1.vmdk for descriptor check
  File "/usr/lib/vmware/vmdkops/Python/vmdk_utils.py", line 96, in vmdk_is_a_descriptor
    with open(filepath) as f:
IOError: [Errno 16] Device or resource busy: '/vmfs/volumes/4bad174b-b67f5cfc-2fc6-0026b966a303/dockvols/myvol1.vmdk'
```

### DIFF
--- a/esx_service/utils/vmdk_utils.py
+++ b/esx_service/utils/vmdk_utils.py
@@ -94,7 +94,7 @@ def vmdk_is_a_descriptor(filepath):
                 line = f.readline()
                 return line.startswith('# Disk DescriptorFile')
         except:
-            logging.warning("Failed to open %s for descriptor check", filepath)
+            logging.exception("Failed to open %s for descriptor check", filepath)
 
     return False
 
--- a/esx_service/vsan_policy.py
+++ b/esx_service/vsan_policy.py
@@ -16,6 +16,7 @@
 # Module for VSAN storage policy creation and configuration
 
 import os
+import logging
 import vmdk_utils
 import volume_kv as kv
 
@@ -67,6 +68,7 @@ def create_policy_file(filename, content):
             f.write(content)
             f.write('\n')
     except:
+        logging.exception("Failed to open %s for writing", filename)
         return 'Error: Failed to open {0} for writing'.format(filename)
 
     return None
@@ -87,6 +89,7 @@ def delete(name):
     try:
         os.remove(policy_path(name))
     except:
+        logging.exception("Failed to remove %s policy file", name)
         return 'Error: {0} does not exist'.format(name)
 
     return None
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -173,6 +173,7 @@ def load(volpath):
     try:
         fh = open(metaFile, "r+")
     except IOError:
+        logging.exception("Failed to open %s", metaFile);
         return None
 
     kvDict = json.load(fh)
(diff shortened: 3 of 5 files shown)


esx_service/utils/vmdk_utils.py
```diff
@@ -94,7 +94,7 @@ def vmdk_is_a_descriptor(filepath):
                 line = f.readline()
                 return line.startswith('# Disk DescriptorFile')
         except:
-            logging.warning("Failed to open %s for descriptor check", filepath)
+            logging.exception("Failed to open %s for descriptor check", filepath)
```
govint: Since we aren't catching a specific exception any idea what message we get in this case? Can we catch a generic exception that will get logged?

esx_service/vsan_policy.py
```diff
@@ -67,6 +68,7 @@ def create_policy_file(filename, content):
             f.write(content)
             f.write('\n')
     except:
+        logging.exception("Failed to open %s for writing", filename)
```
govint: Same as before - can we catch a generic exception to get the log for that.

govint: There is one place in vmdk_ops.c: setStatusDetached() where we do a try: - except: and in the except part we do a "pass" if there is an exception trying to delete the VM name and UUID. This is wrong because meta-data is missing and should be logged.

govint: There are a lot of places in vmdk_ops.py where exceptions are handled with no logs. Will be great to log all exceptions (findVMByName(), findChild())

esx_service/utils/vmdk_utils.py
pdhamdhere: I've added this specific one in PR description. What do you mean by catch generic exception? "except:" will cause any exception in this case rt?

esx_service/vsan_policy.py
pdhamdhere: Added logging to find_child. I left setStatusDetached as-is since exception handler there is used to auto-correct an issue.

kerneltime:
LGTM

esx_service/vsan_policy.py
```diff
@@ -87,6 +89,7 @@ def delete(name):
     try:
         os.remove(policy_path(name))
     except:
+        logging.exception("Failed to remove %s policy file", name)
```
msterin: nit: not sure this one is needed , if the file does not exist silent fail is better. But it's good as is as well

msterin:
Thanks for taking on this !

LGTM

andrewjstone:
LGTM

esx_service/vsan_policy.py
pdhamdhere: Yes, thought about it and decided to leave it there just in case for other OSErrors.

pdhamdhere:
Thanks everyone for quick review!

PR #450, opened 2016-06-09
Add VSAN policy update code to admin api

### LINKED ISSUE #329: Add vmdkops_admin.py policy update
Add the update call as well as updated the docs to reflect it being added.

I'm not putting this in the initial PR as it relies on some objtool stuff Mark is working on and that PR is getting pretty big already. Should be easy to add though.

andrewjstone:
Add unit test for vsan_policy.update in admin cli

Add docs for vsan policy update in admin cli

Tested via make all and CI, as well as manual tests.

Fixes #329

### DIFF
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -151,6 +151,21 @@ def commands():
                     'func': policy_ls,
                     'help':
                     'List storage policies and volumes using those policies'
+                },
+                'update': {
+                    'func': policy_update,
+                    'help': ('Update the definition of a storage policy and all'
+                              'VSAN objects using that policy'),
+                    'args': {
+                        '--name': {
+                            'help': 'The name of the policy',
+                            'required': True
+                        },
+                        '--content': {
+                            'help': 'The VSAN policy string',
+                            'required': True
+                        }
+                    }
                 }
             }
         },
@@ -484,6 +499,14 @@ def policy_ls(args):
         rows.append([name, content.strip(), active])
 
     print cli_table.create(header, rows)
+
+
+def policy_update(args):
+    output = vsan_policy.update(args.name, args.content)
+    if output:
+        print output
+    else:
+        print 'Successfully updated policy: {0}'.format(args.name)
 
 
 NOT_IMPLEMENTED = "Not implemented"
--- a/esx_service/vsan_policy.py
+++ b/esx_service/vsan_policy.py
@@ -17,8 +17,10 @@
 
 import os
 import logging
+import shutil
 import vmdk_utils
 import volume_kv as kv
+import vsan_info
 
 ERROR_NO_VSAN_DATASTORE = 'Error: VSAN datastore does not exist'
 
@@ -42,6 +44,119 @@ def create(name, content):
         return 'Error: Invalid policy string'
 
     return create_policy_file(filename, content)
+
+
+def update(name, content):
+    """
+    Update the content of an existing VSAN policy in the VSAN datastore.
+    Update the policy content in each VSAN object currently using the policy. If
+    a VSAN policy of the given name does not exist return an error string.
+    Return None on success.
+    """
+    path = policy_path(name)
+    if not path:
+        return ERROR_NO_VSAN_DATASTORE
+
+    err = update_policy_file_content(path, content)
+    if err:
+        return err
+
+    return update_vsan_objects_with_policy(name, content)
+
+
+def update_policy_file_content(path, content):
+    """
+    Update the VSAN policy file content. Return an error msg or None on success.
+    """
+    try:
+        with open(path) as f:
+            existing_content = f.read()
+    except OSError as e:
+        if not os.path.isfile(path):
+            return 'Error: Policy {0} does not exist'.format(
+                    os.path.basename(path))
+        else:
+            return 'Error opening existing policy file {0}: {1}'.format(path, e)
+
+    if existing_content.strip() == content.strip():
+            return 'Error: New policy is identical to old policy. Ignoring.'
+
+    # Create a temporary file so we don't corrupt an existing policy file
+    tmpfile = '{0}.tmp'.format(path)
+    err = create_policy_file(tmpfile, content)
+    if err:
+        return err
+
+    # Copy the original policy file to a backup file (.old)
+    # The backup will be maintained in case the policy content is invalid, when
+    # attempting to apply it to existing volumes.
+    # Do an atomic rename of the tmpfile to the real policy file name
+    try:
+        shutil.copy(path, backup_policy_filename(path))
+        os.rename(tmpfile, path)
+    except OSError:
+        print ('Internal Error: Failed to update policy file contents: '
+                '{0}').format(path)
+        raise
+
+    return None
+
+
+def update_vsan_objects_with_policy(name, content):
+    """
+    Find all VSAN objects using the policy given by `name` and update the policy
+    contents in their objects. Returns an error string containing the list of
+    volumes that failed to update, or a msg if there were no volumes to update.
+    Returns None if all volumes were updated successfully.
+
+    Note: This function assumes datastore_path exists.
+    """
+    update_count = 0
+    failed_updates = []
+    dockvols_path = vmdk_utils.get_vsan_dockvols_path()
+    print "This operation may take a while. Please be patient."
+    for v in list_volumes_and_policies():
+        if v['policy'] == name:
+            volume_name = v['volume']
+            vmdk_path = os.path.join(dockvols_path, volume_name)
+            if vsan_info.set_policy(vmdk_path, content):
+                update_count = 1
+            else:
+                failed_updates.append(volume_name)
+
+    if len(failed_updates) != 0:
+        if update_count == 0:
+            # All volumes failed to update, so reset the original policy
+            os.rename(policy_path(backup_policy_filename(name)),
+                      policy_path(name))
+        else:
+            log_failed_updates(failed_updates, name)
+
+        return ('Successfully updated {0} volumes.\n'
+                'Failed to update volumes: {0}').format(update_count,
+                                                        failed_updates)
+
+    # Remove old policy file on success
+    os.remove(policy_path(backup_policy_filename(name)))
+    return None
+
+def backup_policy_filename(name):
+    """ Generate a .old file from a policy name or path """
+    return '{0}.old'.format(name)
+
+def log_failed_updates(volumes, policy_name):
+    """
+    During policy update, some volumes may fail to have their VSAN policies
+    updated. We create a file containing these volumes for debugging purposes.
+    """
+    filename = policy_path('{0}.failed_volume_updates'.format(policy_name))
+    try:
+        with open(filename, 'w') as f:
+            f.write(volumes)
+            f.write('\n')
+    except:
+        print ("Failed to save volume names that failed to update to file."
+                "Please record them for future use.")
 
 
 def make_policies_dir(datastore_path):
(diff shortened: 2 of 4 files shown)


esx_service/cli/vmdkops_admin.py
```diff
@@ -151,6 +151,21 @@ def commands():
                     'func': policy_ls,
                     'help':
                     'List storage policies and volumes using those policies'
+                },
+                'update': {
```
msterin: Nice - and easy to review / read :-)

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+def update_policy_file_content(path, content):
+    """
+    Update the VSAN policy file content. Return an error msg or None on success.
+    """
+    try:
+        with open(path) as f:
+            existing_content = f.read()
+    except OSError as e:
+        if not os.path.isfile(path):
+            return 'Error: Policy does not exist'
```
msterin: I think policy name should be in the error

esx_service/cli/vmdkops_admin.py
```diff
@@ -151,6 +151,21 @@ def commands():
...
+                    'func': policy_update,
+                    'help': ('Update the content of a storage policy and all'
```
pdhamdhere: nit: "update the content" => "update definition of a storage policy"?

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+        else:
+            return 'Error opening existing policy file: {0}'.format(e)
+
+    if existing_content.strip() == content.strip():
+            return 'Error: New policy is identical to old policy. Ignoring.'
+
+    # Create a temporary file so we don't corrupt an existing policy file
+    tmpfile = '{0}.tmp'.format(path)
```
msterin: nit: will it be created in the same folder ? Maybe it's bettet to do mktemp() or whatever it's called in python, to avoid leftrovers if something fails ?

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+    err = create_policy_file(tmpfile, content)
+    if err:
+        return err
+
+    # Do an atomic rename of the tmpfile to the real policy file
+    try:
+        os.rename(tmpfile, path)
+    except OSError:
+        return 'Failed to rename {0} to {1}. Do both files still exist?'
```
msterin: I would be very confused with this message... it looks like an internal error though. Maybe just trace exception and add "internal error" to the "failed" message ?

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+    Note: This function assumes datastore_path exists.
+    """
+    update_count = 0
+    failed_updates = []
+    dockvols_path = vmdk_utils.get_vsan_dockvols_path()
+    for v in list_volumes_and_policies():
+        if v['policy'] == name:
+            volume_name = v['volume']
+            vmdk_path = os.path.join(dockvols_path, volume_name)
+            if vsan_info.set_policy(vmdk_path, content):
```
msterin: it may be a heavy op - I strongly recommend adding an info log message

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+                update_count = 1
+            else:
+                failed_updates.append(volume_name)
+
+    if update_count == 0:
+        return 'No volumes using policy {0}'.format(name)
```
msterin: is it treated as an error ?  I'd expect "no volumes" to be a success ?

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+
+    # Create a temporary file so we don't corrupt an existing policy file
+    tmpfile = '{0}.tmp'.format(path)
+    err = create_policy_file(tmpfile, content)
+    if err:
+        return err
+
+    # Do an atomic rename of the tmpfile to the real policy file
+    try:
+        os.rename(tmpfile, path)
```
pdhamdhere: We do not validate vSAN policy string. We won't know if new policy string is valid unless we apply new policy. In case of errors, you have lost original definition of policy. Should we rename or update policy definition after it is successfully set for at least one volume?

esx_service/vsan_policy.py
```diff
@@ -43,6 +44,87 @@ def create(name, content):
...
+            if vsan_info.set_policy(vmdk_path, content):
+                update_count = 1
+            else:
+                failed_updates.append(volume_name)
+
+    if update_count == 0:
+        return 'No volumes using policy {0}'.format(name)
+
+    if len(failed_updates) != 0:
+        return ('Successfully updated {0} volumes.\n'
```
pdhamdhere: In case of failed update, don't you want to preserve both original and new policy definition since some (failed) volumes are still using original policy?

msterin:
All in all the code  looks good.
*  A few nits and suggestions (mainly about logging).
*  **Main concern** - the CI actually fails (out of space - see below), but the status does show OK - so
 * (1) something is still broken in CI reporting, most likely 'make test-esx' (I thought I've fixed it though) , and it may hide other issues 
 * (2)  you mentioned manual tests - can you also provide what exactly did you test manually, what was the result and why did you decide against piutting it in automatesd testing 

CI log says:

```
Running unit tests in /tmp/vmdk_ops_unittest11080/vmdk_ops_test.py...
....FF
FAIL: testPolicy (__main__.VmdkCreateRemoveTestCase)
  File "/tmp/vmdk_ops_unittest11080/vmdk_ops_test.py", line 125, in testPolicy
AssertionError: {u'Error': 'Failed to create /vmfs/volumes/vsan:5204bc8bc03e1c20-5a4696067b772017/dockvols/vol_UnitTest_Create0.vmdk. Failed to create virtual disk: There is not enough space on the file system for the selected operation (13).\n'}
```

esx_service/cli/vmdkops_admin.py
andrewjstone: haha :D

esx_service/cli/vmdkops_admin.py
andrewjstone: Sure. I'll make that change.

esx_service/vsan_policy.py
andrewjstone: Will do.

esx_service/vsan_policy.py
andrewjstone: Yep, you are right. I will do that.

esx_service/vsan_policy.py
andrewjstone: Hmmm, That's a great point.  There needs to be some form of 2 phase commit here :D. Maintaining the original file as a backup should be done, and then replacing the new one with the old one if we can't update the volume. 

We can't just not do the rename first, because we actually use the policy file via objtool to update the VMs. Although we could just use a copy. Either way I'll fix it.

This brings up another problem though. The code currently allows updating policies with no existing volumes. We won't be able to do the check in that case. Ideally we'd have a way to check outside of updating/creating a volume.

esx_service/vsan_policy.py
andrewjstone: Will do.

esx_service/vsan_policy.py
andrewjstone: So, the way it's used is actually a success, since the cli itself just prints the return value. But I agree that this is hacky, and it also makes it hard to test. I will create a better return type for this.

esx_service/vsan_policy.py
andrewjstone: Yes. That's a good point. We should probably also log the failed updates with the name of the original policy. Something like `policy-name.0` for the policy and `policy-name-failed-volume-updates.0' for the list of volumes that were not updated.

andrewjstone:
The automated tests are actually more comprehensive than the manual tests I performed. Manually, I only created and updated a policy file to check the output, and then tried to update wiht the same content to watch it fail. I didn't actually create any volumes to use a policy manually.

The only reason I did the manual tests was to sanity check my vsan/nimbus install and see the output before I ran `make all`.

esx_service/vsan_policy.py
andrewjstone: Hmm. How many old files do we really want to keep around? Are we going to allow multiple iterations of failed policy updates, or just keep one backup file and a list of the volumes that failed to update?

esx_service/vsan_policy.py
andrewjstone: I actually just removed that check altogether. No reason to output it.

esx_service/vsan_policy.py
msterin: one backup IMO

msterin:
I still see 
```
AssertionError: {u'Error': 'Failed to create /vmfs/volumes/vsan:5204bc8bc03e1c20-5a4696067b772017/dockvols/vol_UnitTest_Create.vmdk. Failed to create virtual disk: There is not enough space on the file system for the selected operation (13).\n'}

```
in the CI log (which still shows as PASSED). Please take a look (the rest looks good, after merge conflicts cleanup)

esx_service/vsan_policy.py
pdhamdhere: Should we be printing volume name being worked on or something to show progress like "...." or "x/y volume" etc?

esx_service/vsan_policy.py
```diff
@@ -43,6 +45,119 @@ def create(name, content):
...
+        if v['policy'] == name:
+            volume_name = v['volume']
+            vmdk_path = os.path.join(dockvols_path, volume_name)
+            if vsan_info.set_policy(vmdk_path, content):
+                update_count = 1
+            else:
+                failed_updates.append(volume_name)
+
+    if len(failed_updates) != 0:
+        # All volumes failed to update, so reset the original policy
```
pdhamdhere: nit: All => Some

esx_service/vsan_policy.py
```diff
@@ -43,6 +45,119 @@ def create(name, content):
...
+            existing_content = f.read()
+    except OSError as e:
+        if not os.path.isfile(path):
+            return 'Error: Policy {0} does not exist'.format(
+                    os.path.basename(path))
+        else:
+            return 'Error opening existing policy file {0}: {1}'.format(path, e)
+
+    if existing_content.strip() == content.strip():
+            return 'Error: New policy is identical to old policy. Ignoring.'
```
pdhamdhere: Why is this error? User may want to have different name for same policy string? Update policy name and succeed?

esx_service/vsan_policy.py
```diff
@@ -43,6 +45,119 @@ def create(name, content):
...
+def backup_policy_filename(name):
+    """ Generate a .old file from a policy name or path """
+    return '{0}.old'.format(name)
+
+def log_failed_updates(volumes, policy_name):
+    """
+    During policy update, some volumes may fail to have their VSAN policies
+    updated. We create a file containing these volumes for debugging purposes.
+    """
+    filename = policy_path('{0}.failed_volume_updates'.format(policy_name))
```
pdhamdhere: That's a long name but I dont have better suggestion so it's fine as-is

pdhamdhere:
Overall LGTM. Some minor suggestions above.

esx_service/vsan_policy.py
andrewjstone: I meant All. I just moved it inside the next conditional to be more clear.

esx_service/vsan_policy.py
andrewjstone: This is an identical policy for the same name. It's a no-op, but we don't want to waste time updating volumes.

pdhamdhere: Contents are same but intention may be to update policy name?

pdhamdhere: I take it back. If we allow user to rename policy then we will have to go & update KV for each disk.

andrewjstone:
This branch was rebased on main and comments were addressed. `make all` passes against my nimbus setup with VSAN. Since this merge requires VSAN testing, it should probably wait until CI covers that, although I believe it's safe to merge now.

msterin:
LGTM  - but please wait with merge until VSAN in CI is back on-line and CI passes (/cc @kerneltime )

msterin:
LGTM

PR #452, opened 2016-06-10
Add additional logging for vmci failures. Fixes #207

### LINKED ISSUE #207: Check that vm-open-tools are installed
Summary:
need to check that either vm-open-tools or vmware-tools  (or just vmci kernel module) are installed

Details:
The code works only if the Linux VM (running Docker) is running under ESX and vm-open-tools or vmware-tools are installed. If they are not, some mystery bugs will be reported.
Plugin GO code needs to check that the vm-open-tools or vmware-tools are installed, and that it is indeed ESX VM (we can see it in BIOS). This will simplify troubleshooting and accelerated solving customer install issues. 

As an alternative, we can write a "check-prerequisites" script thst does all that and (eventually) more.

kerneltime:
Since the cause of communication failures need more info than what can be packed into a log message. Forward the customer to the FAQ.

```
2016-06-11 00:15:06.869406637 +0530 IST [WARNING] 'list' failed: connection reset by peer (errno=104). Hit communication issue with ESX (vmci or ESX service)
     Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq
```
Testing:
* Installed plugin with no vmware tools
* Insured no /dev/vsock (rmmod)
* Ran docker volume ls

### DIFF
--- a/vmdk_plugin/vmdkops/esx_vmdkcmd.go
+++ b/vmdk_plugin/vmdkops/esx_vmdkcmd.go
@@ -92,7 +92,8 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 		errno = err.(syscall.Errno)
 		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 		if errno == syscall.ECONNRESET {
-			msg += " Check that ESX service is running."
+			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
+			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
 		}
 		log.Warnf(msg)
 		return nil, errors.New(msg)


vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -93,6 +93,9 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 		if errno == syscall.ECONNRESET {
 			msg += " Check that ESX service is running."
+			msg += " Check vmware tools or open vm tools is installed."
```
pdhamdhere: We need to document somewhere how to do all of the above! and possibly include link here.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -93,6 +93,9 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 		if errno == syscall.ECONNRESET {
 			msg += " Check that ESX service is running."
```
pdhamdhere: Rather than embedding these error messages here, how about introducing separate init-handshake op. Tools/VMCI/Port error messages doesnt make sense after first command is completely successfully.

kerneltime: The vmci module can be unloaded between docker commands (in theory)

msterin:
I think some noise in error log in this case is OK

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -92,7 +92,8 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 		errno = err.(syscall.Errno)
 		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 		if errno == syscall.ECONNRESET {
-			msg += " Check that ESX service is running."
+			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
+			msg += " Please refer FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
```
andrewjstone: 'Please refer to the FAQ'

kerneltime: Thank you :-)

pdhamdhere:
LGTM

PR #456, opened 2016-06-10
Print the VM name for which request is being processed.

kerneltime:
This change prints the name for the VM that issued the request being served.

Testing:
make sure /etc/vmware/vmdkops/log_config.json is deleted
make deploy-esx
run docker volume ls
```
[root@promc-2n-dhcp24:~] 06/10/16 21:56:26 2795426 [Ubuntu 64-bit] [INFO   ] executeRequest 'list' completed with ret=[{u'Attributes': {}, u'Name': 'MyVolume'}, {u'Attributes': {}, u'Name': 'rg'}, {u'Attributes': {}, u'Name': 'radio2016'}, {u'Attributes': {}, u'Name': 'MyVolume2'}]
```

### DIFF
--- a/esx_service/utils/log_config.py
+++ b/esx_service/utils/log_config.py
@@ -51,7 +51,7 @@ LOG_CONFIG_DEFAULT = {
     "formatters": {
         "standard": {
             "format":
-            "%(asctime)-12s %(process)d [%(levelname)-7s] %(message)s",
+            "%(asctime)-12s %(process)d [%(threadName)s] [%(levelname)-7s] %(message)s",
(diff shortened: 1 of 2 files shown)


kerneltime:
When we do use multiple threads this change should still work.
From: https://docs.python.org/2/library/threading.html
```
name
A string used for identification purposes only. It has no semantics. Multiple threads may be given the same name. The initial name is set by the constructor.
```

andrewjstone:
LGTM

pdhamdhere:
Nice. LGTM. Thanks!

PR #457, opened 2016-06-10
Change some headers for admin cli ls

### LINKED ISSUE #434: Volume to VM to Container Mapping
Taking the below output:

```
vmdkops_admin.py ls -c volume,datastore,created-by,policy,attached-to,capacity,used

Volume  Datastore   Created By  Attached To                           Policy  Capacity  Used
------  ----------  ----------  ------------------------------------  ------  --------  --------
Ryan    datastore1  PhotonOS_1    564dae2e-cc4c-36a4-1371-a2422132add3  N/A     5.00GB    130.00MB
```

It's my understanding that if a vmdk created by one VM (read: the "Created By" field) can't be leveraged in another VM through the docker volume commands (a limitation of Docker Volume command's scope). Assuming that's indeed the case, I'd prefer to see "VM Owner" instead of "Created By" and "Attached to Container" as oppose to the "Attached To" field... which from what I've gathered is the unique BIOS ID?

andrewjstone:
Attached To -> Attached To VM
Created By -> Created By VM

Tested with make all and CI

Fixes #434

### DIFF
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -345,8 +345,8 @@ def format_header_as_arg(header):
 
 def all_ls_headers():
     """ Return a list of all header for ls -l """
-    return ['Volume', 'Datastore', 'Created By', 'Created',
-            'Attached To', 'Policy', 'Capacity', 'Used']
+    return ['Volume', 'Datastore', 'Created By VM', 'Created',
+            'Attached To VM', 'Policy', 'Capacity', 'Used']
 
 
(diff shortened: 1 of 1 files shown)


kerneltime:
LGTM

pdhamdhere:
LGTM.

msterin:
LGTM. This one was easy :-)

andrewjstone:
As discussed with Mark the other day, this is just a log message during a negative test, which is why CI still passes. Will work to somehow disable or annotate it.

PR #458, opened 2016-06-13
Switch to VMFS only on the new setup while vsan is being configured.

kerneltime:
Only VMFS for now.

Testing: CI should be green.


andrewjstone:
LGTM

Issue #399, opened 2016-05-24
ESX services throws on detach task failure


msterin:
A disk was a part of snapshot (before we changed disks to independent) , and now `docker volume rm ` fails.  ESX log shows \ 
 `Cannot remove virtual disk from the virtual machine because it or one of its parent disks is part of a snapshot of the virtual machine.'` The failure leads to ESX service throwing exception and dying, (and then restarted by watchdog) and vmdk plugin showing "cannot connect to ESX" 

this exposes a general issue - task (vim.fault.InvalidDeviceSpec) exceptions are not handled, so there may be more glitches

pdhamdhere:
PR #460 should fix this rt? @msterin

msterin:
yes

msterin:
Fixed by #460

PR #460, opened 2016-06-14
Catch all vim exceptions in disk_detach

msterin:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/381

* Catch all vim exceptions in disk_detach
* Return error message from exception, and log a warning.
* Also change logging in findVMByUuid to be warning (since we retry after the failure)

test: manually shutdown Photon OS with OS->Shutdown menu in ESXi embedded UI. 
Before fix: exception logged and ESX service exit
After fix: exception logged and ESX Service keeps goings

Error report in the log:
```
6/14/16 02:53:20 1010878803 [INFO   ] *** detachVMDK: /vmfs/volumes/vsan:572904f8c031435f-06/14/16 18:02:52 1011049660 [WARNING] Failed to detach /vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82/dockvols/vol.vmdk (The operation is not allowed in the current state.)
(vim.fault.InvalidState) {
   dynamicType = <unset>,
   msg = 'The operation is not allowed in the current state.',
   faultCause = <unset>,
   faultMessage = (vmodl.LocalizableMessage) []
}
06/14/16 18:02:52 1011049660 [INFO   ] executeRequest 'detach' completed with ret={u'Error': u'Failed to detach /vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82/dockvols/vol.vmdk (The operation is not allowed in the current state.)'}

```


esx_service/vmdk_ops.py
```diff
@@ -648,10 +648,13 @@ def disk_detach(vmdk_path, vm):
 
     try:
         wait_for_tasks(si, [vm.ReconfigVM_Task(spec=spec)])
-    except vim.fault.GenericVmConfigFault as ex:
-        for f in ex.faultMessage:
-            logging.warning(f.message)
-        return err("Failed to detach " + vmdk_path)
+    except Exception as ex:
```
pdhamdhere: Why not just catch & handle "vim.Fault.VimFault" exception and pass others to caller? This would eliminate `if isinstance(...)`

msterin: Ah (forehead hit), yes ! A brain hiccup I guess. Will fix. Thanks.

pdhamdhere:
Fixes #399

esx_service/vmdk_ops.py
```diff
@@ -307,13 +307,14 @@ def listVMDK(path):
...
 def findVmByUuid(vm_uuid):
+    vm = None
     try:
         vm = si.content.searchIndex.FindByUuid(None, vm_uuid, True, False)
-        if vm:
-            return vm
-    except Exception as e:
-        logging.exception("Failed to find VM uuid=%s (traceback below), " \
-                          "retrying...", vm_uuid)
+    except Exception as ex:
```
kerneltime: is there a guarantee that the second search will always complete without exception?

kerneltime:
LGTM but the error handling seems built with assumptions.

esx_service/vmdk_ops.py
msterin: there is no guarantee, but we know that on connection timeout retry fixes it, and on other issues we want to rethrow if we fail in retry

esx_service/vmdk_ops.py
```diff
@@ -648,10 +649,10 @@ def disk_detach(vmdk_path, vm):
 
     try:
         wait_for_tasks(si, [vm.ReconfigVM_Task(spec=spec)])
-    except vim.fault.GenericVmConfigFault as ex:
-        for f in ex.faultMessage:
-            logging.warning(f.message)
-        return err("Failed to detach " + vmdk_path)
+    except vim.Fault.VimFault as ex:
+        msg = "Failed to detach %s (%s)" % (vmdk_path, ex.msg)
+        logging.warning(msg + "\n" + str(ex))
```
pdhamdhere: `str(ex)` logs traceback and exception info?

pdhamdhere:
LGTM.

esx_service/vmdk_ops.py
```diff
@@ -648,10 +649,10 @@ def disk_detach(vmdk_path, vm):
...
+        return err(msg)
```
govint: This will leave the disk attached to the VM. Like I'd mentioned in PR #433 (original set of changes) there is nothing the plugin can do to help this situation except log the error. Does the error notify docker that the unmount has failed and docker must keep track that the volume is still in use? The volume remains in attached state if docker doesn't care to unmount again.

esx_service/vmdk_ops.py
msterin: see the bug description - there is an example there

esx_service/vmdk_ops.py
msterin: What is your question or request to change ?

andrewjstone:
LGTM

msterin:
Updated code slightly to print out backtrace (so we know who calls us) rather that exception details (which have nothing beyond "failure reason), which we print anyways). The log looks like this now: 
```
06/14/16 20:14:51 1011078223 [INFO   ] *** detachVMDK: /vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82/dockvols/vol.vmdk from PhotonRC_1 VM uuid = 564d0062-db25-a480-72e2-0596eaf7fcb4
06/14/16 20:14:51 1011078223 [WARNING] Failed to detach /vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82/dockvols/vol.vmdk: The operation is not allowed in the current state.
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 652, in disk_detach
    wait_for_tasks(si, [vm.ReconfigVM_Task(spec=spec)])
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 838, in wait_for_tasks
    raise task.info.error

06/14/16 20:14:51 1011078223 [INFO   ] executeRequest 'detach' completed with ret={u'Error': u'Failed to detach /vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82/dockvols/vol.vmdk: The operation is not allowed in the current state.'}
```

PR #467, opened 2016-06-15
Moved get_vsan_dockvols_path to vsan_info from vmdk_utils

msterin:
Just the move - no functional changes, 
The function was in the wrong file and generated almost-circular (and certainly violating layering) dependency on vmdk_ops.
This is in preparation to multiple datastore support

test: `make all` and CI (in progress)


pdhamdhere:
LGTM

PR #468, opened 2016-06-15
Add vmdkops_admin.py status

### LINKED ISSUE #359: Implement vmdkops_admin.py status

andrewjstone:
Add a status command to the admin CLI.

Tested via manual run with output:

```
[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
Version: 1.0.0-0.0.1
Status: Running
Pid: 82287
Port: 1019
LogLevel: INFO
```

Also tested with an addition to vmdkops_admin_sanity_test.

Fixes #359

### DIFF
--- a/doc/vmdkops-admin-cli-spec.md
+++ b/doc/vmdkops-admin-cli-spec.md
@@ -116,18 +116,18 @@ simplest implementation and isolation of admin information inside a single scrip
 
 #### `Status`
 Show any interesting information about the service. This includes file paths of config files, version
-information, and location of running service. A simple example is shown here, although it's likely
+information, and PID of running service. A simple example is shown here, although it's possible
 that the exact format may be somewhat different.
 
 ```
->vmdkops-admin service status
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
+Version: 1.0.0-0.0.1
 Status: Running
-Health: Healthy
-Version: 1.0.0
-LogFile: /var/log/vmdkops.log
-LogLevel: Debug
-Datastore: DS1, Size: 100GB, Used: 10GB
-Datastore: DS2, Size: 50GB, Used: 0GB
+Pid: 73114
+Port: 1019
+LogConfigFile: /etc/vmware/vmdkops/log_config.json
+LogFile: /var/log/vmware/vmdk_ops.log
+LogLevel: INFO
 ```
 
 #### `help`
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -29,6 +29,7 @@ import cli_table
 import vsan_policy
 import vmdk_utils
 import vsan_info
+import log_config
 
 NOT_AVAILABLE = 'N/A'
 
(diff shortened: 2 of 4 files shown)


doc/vmdkops-admin-cli-spec.md
```diff
@@ -116,18 +116,18 @@ simplest implementation and isolation of admin information inside a single scrip
...
-LogFile: /var/log/vmdkops.log
-LogLevel: Debug
-Datastore: DS1, Size: 100GB, Used: 10GB
-Datastore: DS2, Size: 50GB, Used: 0GB
+Pid: 73114
+Port: 1019
+LogConfigFile: /etc/vmware/vmdkops/log_config.json
+LogFile: /var/log/vmware/vmdk_ops.log
+LogLevel: INFO
 ```
```
govint: Can include a build number of the service. May be a separate issue perhaps, but thats a good data to have.

esx_service/cli/vmdkops_admin.py
```diff
@@ -509,6 +510,59 @@ def policy_update(args):
...
+
+def get_listening_port(pid):
+    """ Return the configured port that the service is listening on """
+    try:
+        cmd = "{0}{1}{2}".format(PS, pid, GREP_V_GREP)
+        output = subprocess.check_output(cmd, shell=True).split()[6]
+        return output
+    except:
+        return NOT_AVAILABLE
+
```
govint: Maybe instead have the plugin server print its entire meta-data into the log file whenever its started/restarted and use that to get this data. This approach is also fine, just suggesting. Logging the meta-data (with build number) is nice when we get a log from a user and it has all the data in it.

govint: In fact the admin CLI should be able to send a request to the plugin server and get its "status" directly. High time we have a way to "talk" to the plugin server on the host itself to run what ever commands we want to. Could avoid these functions here to collect this information. Much like localcli.

doc/vmdkops-admin-cli-spec.md
govint: Can also include a start time - since when is the server  running. For example want to check if plugin crashed, status shows since when the plugin has been running.

pdhamdhere:
LGTM. Additional details if you decide to add can be separate PR.

doc/vmdkops-admin-cli-spec.md
andrewjstone: How is the build number different from the version?

andrewjstone: @govint I'm not sure how to retrieve the start time of the process on ESX. ps doesn't seem to provide it.

esx_service/cli/vmdkops_admin.py
andrewjstone: I agree that it would be great if there was an admin port to allow the cli to ask questions of the plugin server. I think that's an enhancement we should make in the future. @pdhamdhere @msterin @kerneltime Thoughts?

msterin: http://stackoverflow.com/questions/21650370/setting-up-an-http-server-that-listens-over-a-file-socket

msterin: Adding extra info (version,. log config location, whatever) to log on start is goodness.
Having a listen port and http server there (or something like that) is a heavy overkill IMO
We can dump it to /tmp/pid.stats on start (and clean on interupt/exit) and then read - would be simple

andrewjstone: That makes sense. I created an issue to track this: https://github.com/vmware/docker-volume-vsphere/issues/473

msterin:
LGTM

Issue #436, opened 2016-06-02
Validate client side socket port bind capability


kerneltime:
We only want processes with root privileges to be able to communicate with the ESX backend.
This requires
1. Ability to bind a port dynamically below 1024 on the client side
2. Validate the client port on the server side.

[Relevant doc](https://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.vmci.pg.doc/vsockSecurity.7.4.html?path=7_6_0_4_2_2#1006237)
and from [vSockets Programming Guide, pg 17](https://pubs.vmware.com/vsphere-60/topic/com.vmware.ICbase/PDF/ws9_esx60_vmci_sockets.pdf)
```
The port number is arbitrary, although server (listener) and client (connector) must use the same number,
which must designate a port not already in use. Only privileged processes can use ports < 1024.
```

kerneltime:
Fixed in #469

lipingxue:
I saw a issue which may related to this client side socket port bind issue.
I use the following simple bash script to create 60 docker volumes, 
```
#!/bin/bash

for i in {1..60}
do
        vol_name_prefix="t"
        vol_name="${vol_name_prefix}${i}"
        echo $vol_name
        docker volume create --driver=vmdk --name=$vol_name -o size=100mb
        #docker run -it -v $vol_name:/data busybox
done
```


I  do see some intermittent errors like this 
 ```
           root@photon-machine [ ~ ]# ./create_vol.sh 
......
t6
t7
Error response from daemon: create t7: VolumeDriver.Create: 'create' failed: address already in use (errno=98).
......
t30
Error response from daemon: create t30: VolumeDriver.Create: 'create' failed: address already in use (errno=98).
......
Error response from daemon: create t53: VolumeDriver.Create: 'create' failed: address already in use (errno=98).

```

I talked with Mark, and he said that this may relate to the client socket bind issue.

We check the "docker-volume-vsphere.log", and we did see other commands (such as "get") 
also failed with the same error.



```
vi /var/log/docker-volume-vsphere.log

2016-08-13 22:00:12.094932863 +0000 UTC [INFO] Volume created name=t38
2016-08-13 22:00:12.251016903 +0000 UTC [INFO] Volume created name=t39
2016-08-13 22:00:12.395618983 +0000 UTC [INFO] Volume created name=t40
2016-08-13 22:00:12.506006561 +0000 UTC [INFO] Volume created name=t41
[...]

PR #469, opened 2016-06-15
Only process requests from clients with privilege

kerneltime:
To be able to bind a port below 1024 on the client side, the plugin
needs to be run as root or given neccessary privileges. This change
enforces the policy of only allowing privileged clients.

To keep things simple the client port to bind is the same as the server.

Testing:
1. Install client without code change to use priviledge port
2. Validate that docker requests fail and log messages are generated.
3. Upgrade to client with code change for priviledge port
4. Validate that docker requests pass

Will be filling a separate issue to resolve the std out logging
on ESX. We need to invest a bit more to override the std out buffer
and redirect it to a log file.

### DIFF
--- a/esx_service/vmci/connection_types.h
+++ b/esx_service/vmci/connection_types.h
@@ -25,6 +25,9 @@
 // 0 is usually success. Note: sometimes we return socket FD on success
 #define CONN_SUCCESS  (0)
 
+// First non privileged port
+#define START_NON_PRIVILEGED_PORT 1024
+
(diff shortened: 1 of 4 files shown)


kerneltime:
fixes https://github.com/vmware/docker-volume-vsphere/issues/436

msterin:
Thanks for the fix. LGTM (contingent on CI passing)

andrewjstone:
LGTM

PR #476, opened 2016-06-16
Fix checks of ctype return values.

govint:
Fix KV open flags for volume open to use a no-io mode. And fix checks of ctype value returned from volOpenPath().


esx_service/utils/kvESX.py
```diff
@@ -35,9 +35,9 @@
 useSideCarCreate = False
 DVOL_KEY = "docker-volume-vsphere"
 
-# Results in a buffered, locked, filter-less open,
+# Results in a no-io, filter-less open,
 # all vmdks are opened with these flags
-VMDK_OPEN_FLAGS = 524312
```
kerneltime: Is there any documentation that you can refer to for the value picked?

govint: These are disklib flags - you can check opengrok.eng.vmware.com and search
for OPEN_LOCK (for example) in diskLibTypes.h. A few of those flags are put
together for the opens we do.

PR #477, opened 2016-06-16
Implemented Volume.Get()

### LINKED ISSUE #464: volume.Get() should not send list() to server
Currently Get() request from Docker is implemented via List() request to the server with further filtering on the client. THis is super slow  when we have more than a handful of volumes, as Docker uses Get() a  lot - and each time the server has to send a big chunk of mainly useless info to Client

We need a simple get() implementation which does not go via list(). Easy to test - create 100 volumes and delete 10 , and observe the logs

msterin:
Fixes #464 ("volume.Get() should not send list() to server")

As volume count grows, the impact of this one gets higher
E.g. with ~40 volumes, it accelerates "docker volume inspect" x20  (0.5 sec -> 0.02 sec).

Volume.Get() was implemented via Volume.List() - all data were delivered to the client and then
filtered there. This change implements native Volume.Get() (used in many cases, including `docker volume rm` and `docker volume inspect`)

The end result is much faster response time when we have a few dozen (or more) volumes

**Test**: since we do many 'docker volume ls' and 'docker volume rm' already , no new tests, only
`make all`, CI  and quick perf validation: 

Time tested on dev environment, physical ESX and PhotonRC VM. 42 volumes. 
Pre  fix: `real	0m0.553s`
Post fix: `real	0m0.018s`


pdhamdhere:
LGTM

andrewjstone:
LGTM

PR #478, opened 2016-06-17
Added log format to README.md

### LINKED ISSUE #409: Does the plugins support multi-ESX host?
For example, there are two ESX host, a VM (docker host) on one ESX host, create a volume on the other ESX host and VM can use the volume on the other ESX host. I think it is very useful and convenient for docker immigration. I do not know whether the plugin support and whether it can be implement?

Another small question: what is the configuration file (/etc/docker-volume-vsphere.conf) format?
Can you give an example with some description?

msterin:
Fixes #409 

Note:  #409 contains (1) a question "does plugin works with multiple ESX hosts" - and the answer is YES (details #409).  and (2) a request to provide log format info. This PR addressed the second part  and closes the issues.

Only README.md is impacted. See formatted version here https://github.com/vmware/docker-volume-vsphere/tree/readme-log.msterin

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,25 +85,37 @@ systemctl start docker
 ```
 
 ## Logging
-The relevant logging for debugging consists of
+The relevant logging for debugging consists of the following:
+* Docker Logs
+* Plugin logs - VM (docker-side)
+* Plugin logs - ESX (server-side)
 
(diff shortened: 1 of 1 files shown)


andrewjstone:
LGTM

pdhamdhere:
LGTM

PR #479, opened 2016-06-17
Fixes to support python 3.5.1 and 64-bit libs

govint:
These are fixes for syntax errors with Python 3.5.1

### DIFF
--- a/esx_service/Makefile
+++ b/esx_service/Makefile
@@ -33,7 +33,10 @@ OFFLINE_DEPOT := vmware-esx-vmdkops-$(PKG_VERSION).zip
 VIB_BIN  = $(BIN)/$(VIB)
 
 # Shared lib. Note that it is referred by name from vmci_srv.py
-VMCI_SRV_LIB := $(BIN)/libvmci_srv.so
+VMCI_SRV_LIB_NAME := libvmci_srv.so
+VMCI_SRV_LIB   := $(BIN)/$(VMCI_SRV_LIB_NAME)
+VMCI_SRV_LIB64 := $(BIN)/$(VMCI_SRV_LIB_NAME).64
+VMCI_SRV_LIBS  := $(VMCI_SRV_LIB) $(VMCI_SRV_LIB64)
 
 # Source code - all .py stuff which is not tests
 PY_SRC   := $(filter-out %_test.py, $(wildcard *.py))
@@ -41,11 +44,11 @@ PY_CLI   := $(filter-out %_test.py, $(wildcard cli/*.py))
 PY_UTILS := $(filter-out %_test.py, $(wildcard utils/*.py))
 
 C_SRC  := vmci/vmci_server.c
-# We need to build C code as 32 bits.
-# Reason: vSockets return cartelId to ID connecting VM. We use VSI to convert
-#       it to VM name/ID. ESX is shipped with 32bit VSI python module only, so
-#       we use 32bit python and 32 bit shared libs for VMCI/vSockets
+
+# Build both a 32-bit and a 64-bit library for vmci.
 CFLAGS  := -fPIC -m32 -shared
+CFLAGS64  := -fPIC -shared
+
 cc := gcc
 
 INIT_SCRIPT := vmdk-opsd
@@ -79,6 +82,7 @@ VMDKOPS_PAYLOAD := $(PAYLOAD)/vmdkops
 VMDKOPS_BIN     := $(VMDKOPS_PAYLOAD)$(ESX_LOC)/bin
 VMDKOPS_PY      := $(VMDKOPS_PAYLOAD)$(ESX_LOC)/Python
 VMDKOPS_LIB     := $(VMDKOPS_PAYLOAD)$(ESX_LOC)/lib
+VMDKOPS_LIB64   := $(VMDKOPS_PAYLOAD)$(ESX_LOC)/lib64
 VMDKOPS_INITD   := $(VMDKOPS_PAYLOAD)/etc/init.d
 
 # this is used from a Makefile upstairs, and could be used manually
@@ -89,24 +93,28 @@ build: $(VIB_BIN)
 TO_WDIR := $(DESCRIPTOR)
 TO_ESX_BIN := $(PY_SRC) $(PY_CLI) $(TOOLS)
 TO_ESX_PY  := $(PY_UTILS)
-TO_ESX_LIB := $(VMCI_SRV_LIB)
+TO_ESX_LIB := $(VMCI_SRV_LIBS)
 TO_ESX_INITD := $(INIT_SCRIPT)
 
 # copy files to staging area and then build the VIB
 $(VIB_BIN): $(TO_WDIR) $(TO_ESX_BIN) $(TO_ESX_PY) $(TO_ESX_LIB) $(TO_ESX_INITD)
 	@echo === Staging in $(PAYLOAD) ===
-	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_INITD)
+	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_LIB64) $(VMDKOPS_INITD)
 	@chmod -R a+w $(PAYLOAD)
 	@cp $(TO_WDIR) $(WDIR)
 	@cp $(TO_ESX_BIN) $(VMDKOPS_BIN)
 	@cp $(TO_ESX_PY)  $(VMDKOPS_PY)
-	@cp $(TO_ESX_LIB) $(VMDKOPS_LIB)
+	@cp $(VMCI_SRV_LIB) $(VMDKOPS_LIB)/$(VMCI_SRV_LIB_NAME)
+	@cp $(VMCI_SRV_LIB64) $(VMDKOPS_LIB64)/$(VMCI_SRV_LIB_NAME)
 	@cp $(TO_ESX_INITD) $(VMDKOPS_INITD)
 	$(VIBAUTHOR) $(VIBAUTHOR_PARAMS)
 
 
 $(VMCI_SRV_LIB): $(C_SRC)
 	$(CC) $(CFLAGS) -o $@ $(C_SRC)
+
+$(VMCI_SRV_LIB64): $(C_SRC)
+	$(CC) $(CFLAGS64) -o $@ $(C_SRC)
 
 # Run lint on python source. Currently this is run manually and
 # needs pylint and dependent packages to be installed.
@@ -120,7 +128,7 @@ lint:
 
 .PHONY:	clean
 clean:
-	rm -rf $(VIB) $(VMCI_SRV_LIB) $(WDIR)/$(DESCRIPTOR) $(PAYLOAD)
+	rm -rf $(VIB) $(VMCI_SRV_LIBS) $(WDIR)/$(DESCRIPTOR) $(PAYLOAD)
 
 .PHONY: test
 test: test-esx build
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -331,7 +331,7 @@ def ls(args):
         header = all_ls_headers()
         rows = generate_ls_rows()
 
-    print cli_table.create(header, rows)
+    print(cli_table.create(header, rows))
 
 
 def ls_dash_c(columns):
@@ -466,17 +466,17 @@ def human_readable(size_in_bytes):
 def policy_create(args):
     output = vsan_policy.create(args.name, args.content)
     if output:
-        print output
+        print(output)
     else:
-        print 'Successfully created policy: {0}'.format(args.name)
+        print('Successfully created policy: {0}'.format(args.name))
 
 
 def policy_rm(args):
     output = vsan_policy.delete(args.name)
     if output:
-        print output
+        print(output)
     else:
-        print 'Successfully removed policy: {0}'.format(args.name)
+        print('Successfully removed policy: {0}'.format(args.name))
 
 
 def policy_ls(args):
@@ -499,27 +499,27 @@ def policy_ls(args):
             active = 'Unused'
         rows.append([name, content.strip(), active])
 
-    print cli_table.create(header, rows)
+    print(cli_table.create(header, rows))
 
 
 def policy_update(args):
     output = vsan_policy.update(args.name, args.content)
     if output:
-        print output
+        print(output)
     else:
-        print 'Successfully updated policy: {0}'.format(args.name)
+        print('Successfully updated policy: {0}'.format(args.name))
 
 
 def status(args):
-    print "Version: {0}".format(get_version())
+    print("Version: {0}".format(get_version()))
     (status, pid) = get_service_status()
-    print "Status: {0}".format(status)
+    print("Status: {0}".format(status))
     if pid:
-        print "Pid: {0}".format(pid)
-        print "Port: {0}".format(get_listening_port(pid))
-    print "LogConfigFile: {0}".format(log_config.LOG_CONFIG_FILE)
-    print "LogFile: {0}".format(log_config.LOG_FILE)
-    print "LogLevel: {0}".format(log_config.get_log_level())
+        print("Pid: {0}".format(pid))
+        print("Port: {0}".format(get_listening_port(pid)))
+    print("LogConfigFile: {0}".format(log_config.LOG_CONFIG_FILE))
+    print("LogFile: {0}".format(log_config.LOG_FILE))
+    print("LogLevel: {0}".format(log_config.get_log_level()))
 
 
 VMDK_OPSD = '/etc/init.d/vmdk-opsd'
@@ -567,23 +567,23 @@ NOT_IMPLEMENTED = "Not implemented"
 
 
 def role_create(args):
-    print NOT_IMPLEMENTED
+    print(NOT_IMPLEMENTED)
 
 
 def role_rm(args):
-    print NOT_IMPLEMENTED
+    print(NOT_IMPLEMENTED)
 
 
 def role_ls(args):
-    print NOT_IMPLEMENTED
+    print(NOT_IMPLEMENTED)
 
 
 def role_set(args):
-    print NOT_IMPLEMENTED
+    print(NOT_IMPLEMENTED)
 
 
 def role_get(args):
-    print NOT_IMPLEMENTED
+    print(NOT_IMPLEMENTED)
 
 
 if __name__ == "__main__":
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -22,6 +22,10 @@ from ctypes import \
         c_void_p, c_char_p, c_int32, c_bool, c_uint32, c_uint64
 import json
 import logging
+import sys
+
+# Python version 3.5.1
+PYTHON64_VERSION = 50659824
 
 # Side car create/open options
 KV_SIDECAR_CREATE = 0
@@ -30,9 +34,10 @@ KV_SIDECAR_CREATE = 0
 KV_CREATE_SIZE = 0
 
 # VSphere lib to access ESX proprietary APIs.
+DISK_LIB64 = "/lib64/libvmsnapshot.so"
 DISK_LIB = "/lib/libvmsnapshot.so"
 lib = None
-useSideCarCreate = False
+use_sidecar_create = False
 DVOL_KEY = "docker-volume-vsphere"
 
 # Results in a buffered, locked, filter-less open,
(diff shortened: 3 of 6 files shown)


govint:
Fixes #480

andrewjstone:
I prefer removing the space between `print` and `(` so it looks more like a standard function call. Other than that, LGTM.

pdhamdhere:
+1 for Andrew's suggestion otherwise LGTM.

msterin:
+1

govint:
Shall I submit this for beta or wait till after that.

msterin:
please submit after Beta tag

govint:
Fixed print statements in vmdkops_admin.py.

govint:
Fixed loading 64-bit libvmsnapshot.so in case loading the 32-bit (default) fails.

esx_service/utils/kvESX.py
```diff
@@ -31,6 +31,7 @@
 
 # VSphere lib to access ESX proprietary APIs.
 DISK_LIB = "/lib/libvmsnapshot.so"
+DISK_LIB64 = "/lib64/libvmsnapshot.so"
```
msterin: Can you elaborate on how the loading of correct versions of esx  and our Libs supposed to work end to end, ? Are  you saying in esx 6.5 the 32 bit lib is missing and you use it for determine the bitness ?

govint: Both 6.0 and 6.5 have 32 and 64 bit libraries installled on them. But up to
python 3.4 there isn't an issue with using the 32-bit libraries. Python 3.5
doesn't load a 32-bit library.. It depends on what python version is
installed on the host. So, attempt to load the 32-bit lib first and if that
fails then load the 64-bit version.

msterin: I am still not sure how would it help for esx_service to run. Don't we need to build our vmci lib in 32 and 64 and load the correct version? And do you plan to figure out what is the correct version based on try/except above ? And how it is tested? 
All in all, I think it is better to merge the syntax repair code (print ->print() ) and the likes - with related pycompile test. 
```python
import py_compile
py_compile.compile('my_script.py')
```

And then have a separate PR to build our vmci libs in 64, and load based on version of python or version of ESX - - with related tests on python 3.5

msterin:
All in all, I think it is better to merge the syntax repair code (the original change you have all LGTMs for)  , with a small pycompile test on 3.5, something like 
```python
import py_compile
py_compile.compile('my_script.py')
```

And then have a separate PR to build our vmci libs in 64, and load based on version of python or version of ESX - - with related tests on python 3.5

govint:
Mark, no this is not done yet. As posted in telegram json.loads() has a
problem with byte strings that we get from vmcii_get_one_op(). I'll update
with the complete changes.

For now I'm building a 64-bit vmci library to test locally.. I'll send over
all changes for review. And also update the PR.

govint:
Pls. don't review these changes as more changes are coming in. I'll update once those are complete.

govint:
Fixes as below:

1. c_char_p() no longer accepts an str, needs bytes - fix is to use <str>.encode() in all places where c_char_p() was used.
2. json.loads() doesn't accept a string, needs bytes - use "<str>.decode() in places where json.loads() was complaining (only in vmdk_ops.py). KV load and store works fine with data read from the KV file.
3. Building a 64-bit vmci_srv library and installing on ESX.
4. Using a 64-bit libvmsnapshot.so on ESX and updated disklib API arg/result list in kvESX.py to use 64-bit types depending on python version used. Choice of library (32 or 64 bits) depends on >=python3.5.1
5. Fixed print, exception statements as reported by Python.

govint:
CI deploy is still failing.

msterin:
Since there have been MANY functional changes since my last review, I withdraw my LGTM  - I need to do another review now.

pdhamdhere:
Hey @govint Is this PR ready for review? Otherwise, do you want to close this PR and publish when it is ready for review?

My understanding so far is we need to support following versions;
- Python 2 32-bit bundle ( < ESX 6.x)
- Python 3 64-bit bundle ( >= ESX 2016)

Build should build both of the above and deploy/run should pick appropriate binaries (32-bit vs 64-bit).

esx_service/Makefile
```diff
@@ -33,19 +33,21 @@ OFFLINE_DEPOT := vmware-esx-vmdkops-$(PKG_VERSION).zip
 VIB_BIN  = $(BIN)/$(VIB)
 
 # Shared lib. Note that it is referred by name from vmci_srv.py
-VMCI_SRV_LIB := $(BIN)/libvmci_srv.so
+VMCI_SRV_LIB   := $(BIN)/libvmci_srv.so
+VMCI_SRV_LIB64 := $(BIN)/libvmci_srv64.so
```
msterin: should it be  something like ? 
```
VMCI_SRV_LIB_NAME   := libvmci_srv.so
VMCI_SRV_LIB32   := $(BIN)/$(VMCI_SRV_LIB_NAME)
VMCI_SRV_LIB64 := $(BIN)/$(VMCI_SRV_LIB_NAME).64
```

since you will be copying it to staging area explicitly to $(VMCI_SRV_LIB_ NAME) anyways ?

esx_service/Makefile
```diff
@@ -89,25 +92,29 @@ build: $(VIB_BIN)
...
 	@echo === Staging in $(PAYLOAD) ===
-	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_INITD)
+	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_LIB64) $(VMDKOPS_INITD)
 	@chmod -R a+w $(PAYLOAD)
 	@cp $(TO_WDIR) $(WDIR)
 	@cp $(TO_ESX_BIN) $(VMDKOPS_BIN)
 	@cp $(TO_ESX_PY)  $(VMDKOPS_PY)
-	@cp $(TO_ESX_LIB) $(VMDKOPS_LIB)
+	@cp $(VMCI_SRV_LIB) $(VMDKOPS_LIB)
+	@cp $(VMCI_SRV_LIB64) $(VMDKOPS_LIB64)/libvmci_srv.so
```
msterin: please do not use the explicit name here. Define VMCI_SRV_LIB_NAME  as I suggested, or simply use $(basename)

esx_service/utils/kvESX.py
```diff
@@ -22,6 +22,10 @@
         c_void_p, c_char_p, c_int32, c_bool, c_uint32, c_uint64
 import json
 import logging
+import sys
+
+# Python version 3.5.1 
+PYTHON64_VERSION = 50659824
```
msterin: if python clicks to 3.5.2, will it still work ?

esx_service/utils/kvESX.py
```diff
@@ -42,24 +47,78 @@
...
+    lib.DiskLib_Close.restype = int
+
+    # Check if this library supports create API
+    try:
+        lib.DiskLib_SidecarCreate.argtypes = [c_uint64, c_char_p, c_uint64,
+                                              c_int32, POINTER(c_uint64)]
+        lib.DiskLib_SidecarCreate.restype = int
+        useSideCarCreate = True
+    except:
+        # do nothing
```
msterin: "do nothing" comment is incorrect as you do print a  message. 
You can always use 
```python
except:
  logging.debug(...)
  pass
```
if you want to emphasize the lack of specific action

esx_service/utils/kvESX.py
```diff
@@ -42,24 +47,78 @@
...
+                                        POINTER(c_uint64)]
+    lib.DiskLib_SidecarOpen.restype = int
+
+    lib.DiskLib_SidecarClose.argtypes = [c_uint64, c_char_p, POINTER(c_uint64)]
+    lib.DiskLib_SidecarClose.restype = int
+
+    lib.DiskLib_SidecarMakeFileName.argtypes = [c_char_p, c_char_p]
+    lib.DiskLib_SidecarMakeFileName.restype = c_char_p
+
+    lib.ObjLib_Pread.argtypes = [c_uint32, c_void_p, c_uint64, c_uint64]
```
msterin: where do we use pread/pwrite ?

esx_service/utils/kvESX.py
```diff
@@ -42,24 +47,78 @@
...
+       lib.DiskLib_Init.restype = c_bool
+       lib.DiskLib_Init()
 
 
 # Loads the ESX library used to access proprietary ESX disk lib API.
 # Create arg/result definitions for those that we use.
-def kvESXInit():
+
+# Initialize 32-bit disk lib interfaces
+def kv_ESX_lib64_init():
```
msterin: why 64 bit init is different from 32 bit init? Looking into function defintion, they seem to using 32 or 64 bit explicitly, so why different functions? 

[nit] if it **is** needed, I'd prefer to hide this in kv_esx_lib_init() rather that having it done upstairs

esx_service/utils/kvESX.py
```diff
@@ -102,17 +161,33 @@ def kvESXInit():
...
+
+   return
 
 # Open a VMDK given its path, the VMDK is opened locked just to
 # ensure we have exclusive access and its not already in use.
 def volOpenPath(volpath):
-    dhandle = c_uint32(0)
-    ihandle = c_uint32(0)
+    if using_64bits:
+       dhandle = c_uint64(0)
```
msterin: Can't you just have a wrapper for c_uint (get_c_uint) and call it when needed, instead of doing ' if using_64bits:' all over the place ?

esx_service/utils/kvESX.py
```diff
@@ -123,42 +198,57 @@ def volOpenPath(volpath):
 
 # Create the side car for the volume identified by volpath.
 def create(volpath, kv_dict):
-    disk = c_uint32(0)
-    objHandle = c_uint32(0)
+    if using_64bits:
```
msterin: same comment - please wrap c_uint conversion and use it instead of multiple  'if using_64bits:'

esx_service/vmdk_ops.py
```diff
@@ -648,7 +653,9 @@ def disk_attach(vmdk_path, vm):
                      if type(dev) == vim.VirtualDisk and dev.controllerKey ==
                      controller_key])
         # search in 15 slots, with unit_number 7 reserved for scsi controller
-        availSlots = set(range(0, 6) + range(8, 16)) - taken
+        range1 = set(range(0, 7))
```
msterin: why do we need 3 lines instad of one now ? I understand  you fixed a bug (Missing slot #6), but why not simply leave it as 
```python
availSlots = set(range(0, 7) + range(8, 16)) - taken
``` 
?

esx_service/vmdk_ops.py
```diff
@@ -738,13 +745,19 @@ def signal_handler_stop(signalnum, frame):
     logging.warn("Received signal num: ' %d '", signalnum)
     sys.exit(0)
 
+def load_vmci():
+    global lib
+
+    try:
```
msterin: this should be based on 32/64 bit switch, not simply failing to load

govint:
Yes please all do the review again. There was no point fixing a few things
and then the server still doesn't even run.

govint:
Prashant,
    I've posted the changes for review. The appropriate libs are built and
used depending on the python version used.

esx_service/utils/kvESX.py
govint: Ok, I'm checking if the Python version is >= 3.5.1. This is lowest version
thats reporting the issues and what 6.5 is going to be shipping with
(main-stage has this).

msterin: Ah, makes sense - thanks for the comment. I missed the >= the check.

esx_service/utils/kvESX.py
govint: These aren't used presently nor are the DDB get/set but in place in case we
need to support <6.0 ESX or want to write to side cars with these. Although
pread/write() is not usable (I'll remove that). Have raised a PR in
bugzilla to create disklib functions that will do generic read/write IO -
handle all of the the direct IO and buffer alignment.

esx_service/utils/kvESX.py
govint: Args are 64-bit handles passed to disklib in the 64-bit init and 32-bit
handles otherwise. I didn't want to have one large function with most
functions declared in a if - else statement. I can separate out the ones
that are not impacted into a kv_esx_lib_common() so that just the 32-64-bit
declarations are in one function (separated of course).

msterin: You can wrap 2 sets of definitions it
2 big blocks in one function - essentially 
the same way you did one KV init function

esx_service/vmdk_ops.py
govint: Python 3.5 doesn't allow '+' in a set definition:

Python 3.5.1+ (default, Apr  4 2016, 12:10:58)
[GCC 4.6.3] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> y=set(range(1,5) + range(10,15))
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'range' and 'range'


Works perfect in 2.7,
Python 2.7.6 (default, Mar 22 2014, 22:59:56)
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> y=set(range(1,5) + range(10,15))
>>> print y
set([1, 2, 3, 4, 10, 11, 12, 13, 14])
>>>

msterin: OK, still no need for 3 lines
availSlots = (set(range(0, 7) | range(8, 16)) - taken

govint: Nope, fails as well,
Python 3.4.0 (default, Apr 11 2014, 13:05:11)
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> y=set(range(0, 7) | range(8, 16))
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'range' and 'range'

msterin: @govint - reading manual helps.
```python
availSlots = ( set(range(0,7))  | set(range(7, 15)) )  - taken
```
works fine on both pyth2 and pyth3

govint: I'll do this - set(range(0, 7)) | set(range(8, 16)) - taken

govint: Missed the "saving lines" section :-)), anyway I got the same approach
above.

esx_service/utils/kvESX.py
msterin: let's add only what we need at the moment. We are not providing generic wrapper, we just use what is needed

esx_service/Makefile
```diff
@@ -89,25 +93,29 @@ build: $(VIB_BIN)
...
 $(VIB_BIN): $(TO_WDIR) $(TO_ESX_BIN) $(TO_ESX_PY) $(TO_ESX_LIB) $(TO_ESX_INITD)
 	@echo === Staging in $(PAYLOAD) ===
-	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_INITD)
+	@mkdir -p $(VMDKOPS_BIN) $(VMDKOPS_PY) $(VMDKOPS_LIB) $(VMDKOPS_LIB64) $(VMDKOPS_INITD)
 	@chmod -R a+w $(PAYLOAD)
 	@cp $(TO_WDIR) $(WDIR)
 	@cp $(TO_ESX_BIN) $(VMDKOPS_BIN)
 	@cp $(TO_ESX_PY)  $(VMDKOPS_PY)
-	@cp $(TO_ESX_LIB) $(VMDKOPS_LIB)
+	@cp $(VMCI_SRV_LIB) $(VMDKOPS_LIB)
```
msterin: nit: just for consistency, you may want to change this line to 
`@cp $(VMCI_SRV_LIB) $(VMDKOPS_LIB)//$(VMCI_SRV_LIB_NAME)`

esx_service/utils/kvESX.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 VMware, Inc. All Rights Reserved.
+	# Copyright 2016 VMware, Inc. All Rights Reserved.
```
msterin: nit: no need for the extra spaces here

esx_service/utils/kvESX.py
```diff
@@ -42,130 +47,156 @@
...
-    lib.DiskLib_SidecarOpen.restype = int
+# Initialize 32-bit disk lib interfaces
+def kv_esx_init():
+    global is_64bits
+    global useSideCarCreate
 
-    lib.DiskLib_SidecarClose.argtypes = [c_uint32, c_char_p, POINTER(c_uint32)]
-    lib.DiskLib_SidecarClose.restype = int
+    # Define arg types for disk lib apis.
+    if sys.hexversion >= PYTHON64_VERSION:
```
msterin: I am not sure why did you take all disklib init stuff from loadDiskLib() and moved to kv_esx_init() ? 
If it was to address my comment about spread of 32 vs 64 bit code, I think it's fine in init functions (which is also what I said). 

I think disklib-related init correct place in in loadDiskLib (or disklib_init(), using the proper convention). The current arrangement feels wrong. E..g. if I want to use DiskLib_OpenWithInfo, I for some reason need to call kv_esx_init() first, instead of just disklib_init().

esx_service/vmdk_ops.py
```diff
@@ -738,13 +745,19 @@ def signal_handler_stop(signalnum, frame):
     logging.warn("Received signal num: ' %d '", signalnum)
     sys.exit(0)
 
+def load_vmci():
+   global lib
+
+   if sys.hexversion >= PYTHON64_VERSION:
```
msterin: if possible, please add an info log message

msterin:
Overall LGTM, pending one issues resolution  - the location of disklib init code seems wrong. Please review and either fix or explain.
Also a couple of small nits ( feel free to ignore).

esx_service/utils/kvESX.py
```diff
@@ -42,130 +47,156 @@
...
+          res = lib.DiskLib_SidecarOpen(dhandle, DVOL_KEY.encode(), KV_SIDECAR_CREATE,
+                                        byref(objHandle))
+       if res != 0:
+          logging.warning("Side car create for %s failed - %x", volpath, res)
+          lib.DiskLib_Close(dhandle)
+          return False
 
-    lib.DiskLib_SidecarClose(disk, DVOL_KEY, byref(objHandle))
-    lib.DiskLib_Close(disk)
+       lib.DiskLib_SidecarClose(dhandle, DVOL_KEY.encode(), byref(objHandle))
```
andrewjstone: Are these indentations correct? I'm not sure they aren't, but it looks like they changed to move under
 `if disk_is_valid(dhandle):`

In this case there is no other return value when the disk is invalid and None will be returned. I'm not sure if this is correct or not since there is no specification of the return value or any doc comments explaining what is happening.

esx_service/utils/kvESX.py
```diff
@@ -42,130 +47,156 @@
...
 
-    res = lib.DiskLib_SidecarDelete(disk, DVOL_KEY)
-    if res != 0:
-        logging.warning("Side car delete for %s failed - %x", volpath, res)
-        lib.DiskLib_Close(disk)
-        return False
-
-    lib.DiskLib_Close(disk)
-    return True
+       lib.DiskLib_Close(dhandle)
```
andrewjstone: Same issue as my comment on create. In this function however, it appears that returning None is absolutely incorrect and False should be returned if `disk_is_valid` returns false.

msterin: good catch !

esx_service/vmdk_ops.py
```diff
@@ -808,7 +821,9 @@ def handleVmciRequests(port):
             logging.info("executeRequest '%s' completed with ret=%s",
                          req["cmd"], ret)
 
-        response = lib.vmci_reply(c, c_char_p(json.dumps(ret)))
+            ret_string = json.dumps(ret)
```
andrewjstone: Should these 2 lines really be moved inside the else clause? If so, response will be None if there is an exception and we will never match `if response == VMCI_ERROR:`

andrewjstone:
Local `make all` is showing an error. Note that the same error is not seen on master.
```
Running unit tests in /tmp/vmdk_ops_unittest/vmdkops_admin_sanity_test.py...
..E
ERROR: test_status (__main__.TestVmdkopsAdminSanity)
  File "/tmp/vmdk_ops_unittest/vmdkops_admin_sanity_test.py", line 58, in test_status
    output = subprocess.check_output(cmd, shell=True, stderr=self.devnull)
  File "/build/mts/release/bora-3620759/bora/build/esx/release/vmvisor/sys-boot/lib/python2.7/subprocess.py", line 573, in check_output
CalledProcessError: Command '/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status' returned non-zero exit status 1

Ran 3 tests in 1.876s

FAILED (errors=1)
Running unit tests in /tmp/vmdk_ops_unittest/vmdkops_admin_test.py...
.....................
Ran 21 tests in 2.606s

OK
Running unit tests in /tmp/vmdk_ops_unittest/vol_test.py...
VOL_TEST kv_store skippped
Running unit tests in /tmp/vmdk_ops_unittest/vsan_info_test.py...
ss
Ran 2 tests in 0.000s

OK (skipped=2)
Running unit tests in /tmp/vmdk_ops_unittest/vsan_policy_test.py...
sssss
Ran 5 tests in 0.000s

OK (skipped=5)
Makefile:145: recipe for target 'test-esx' failed
make[2]: *** [test-esx] Error 1
make[2]: Leaving directory '/home/ajs/go/src/github.com/vmware/docker-volume-vsphere/esx_service'
Makefile:298: recipe for target 'test-esx' failed
make[1]: *** [test-esx] Error 2
make[1]: Leaving directory '/home/ajs/go/src/github.com/vmware/docker-volume-vsphere/vmdk_plugin'
Makefile:33: recipe for target 'all' failed
make: *** [all] Error 2
```

esx_service/cli/vmdkops_admin.py
```diff
@@ -499,27 +499,27 @@ def policy_ls(args):
...
     if pid:
-        print "Pid: {0}".format(pid)
-        print "Port: {0}".format(get_listening_port(pid))
-    print "LogConfigFile: {0}".format(log_config.LOG_CONFIG_FILE)
-    print "LogFile: {0}".format(log_config.LOG_FILE)
-    print "LogLevel: {0}".format(log_config.get_log_level())
+        print("Pid: {0}".format(pid))
+        print("Port: {0}".format(get_listening_port(pid)))
+    print("LogConfigFile: {0}".format(log_config.LOG_CONFIG_FILE))
+    printr("LogFile: {0}".format(log_config.LOG_FILE))
```
andrewjstone: I just debugged the make all failure and tracked it to this line. Should be `print` instead of `printr`.

esx_service/utils/kvESX.py
govint: Correct, will revert.

esx_service/utils/kvESX.py
govint: Thanks, my change was to put the code into the if disk valid clause while earlier if the disk was invalid the function returned there itself. Reverted.

esx_service/utils/kvESX.py
govint: Yes, reverted the code to its earlier form and avoid moving all of the code into the "if disk valid clause". Thanks

esx_service/vmdk_ops.py
govint: Correct, thanks

govint:
1. Rebased on current code in master

2. Ran pylint and fixed those lint-warnings resulting from the new changes.

3. Tested on Python 3.5.1 and 64-bit libs.

4. CI is green at https://ci.vmware.run/vmware/docker-volume-vsphere/839

esx_service/utils/kvESX.py
```diff
@@ -30,9 +34,10 @@
 KV_CREATE_SIZE = 0
 
 # VSphere lib to access ESX proprietary APIs.
+DISK_LIB64 = "/lib64/libvmsnapshot.so"
 DISK_LIB = "/lib/libvmsnapshot.so"
 lib = None
-useSideCarCreate = False
+use_sidecar_create = False
```
andrewjstone: Nit: This should probably be in CAPS as well

esx_service/utils/kvESX.py
```diff
@@ -42,132 +47,161 @@
...
+def kv_esx_init():
+   disk_lib_init()
 
+def get_uint(val):
+   if is_64bits:
+      return c_uint64(val)
+   else:
+      return c_uint32(val)
+
+def disk_is_valid(dhandle):
```
andrewjstone: Nit: Could shorten this function using `get_uint(0)`

esx_service/vmdk_ops.py
```diff
@@ -841,23 +853,26 @@ def main():
...
         usage()
-        return 1
+        return(1)
     for a, v in opts:
         if a == '-p':
             port = int(v)
         if a == '-h':
             usage()
-            return 0
+            return(0)
```
andrewjstone: Nit: I find it somewhat weird to wrap a return statement in parens, but it's no big deal

andrewjstone:
LGTM

esx_service/vmdk_ops.py
govint: This is enforced by Python 3.5

esx_service/utils/kvESX.py
```diff
@@ -42,132 +47,161 @@
...
+      return dhandle.value != c_uint32(0).value
 
 # Open a VMDK given its path, the VMDK is opened locked just to
 # ensure we have exclusive access and its not already in use.
-def volOpenPath(volpath):
-    dhandle = c_uint32(0)
-    ihandle = c_uint32(0)
+def vol_open_path(volpath):
+    dhandle = get_uint(0)
+    ihandle = get_uint(0)
```
msterin: nit: could be `dhandle = ihandle = get_uint(0)`

esx_service/vmdk_ops.py
```diff
@@ -751,13 +756,19 @@ def signal_handler_stop(signalnum, frame):
     logging.warn("Received signal num: ' %d '", signalnum)
     sys.exit(0)
 
+def load_vmci():
+   global lib
+
+   logging.info("Loading VMCI server lib.")
```
msterin: nit: please add which lib is loaded (64 or 32) to thew log as info

esx_service/vmdk_ops.py
```diff
@@ -841,23 +853,26 @@ def main():
...
         port = 1019
         opts, args = getopt.getopt(sys.argv[1:], 'hp:')
-    except getopt.error, msg:
+    except getopt.error as msg:
         if msg:
-            print >> sys.stderr, msg
+           logging.exception(msg)
         usage()
-        return 1
+        return(1)
```
msterin: actually, I do not think we should use return (val) instead of `return val`. See https://www.python.org/dev/peps/pep-0008/

msterin:
LGTM, contingent on addressing a small issue - I think we should use `return 1`, not 'return (1)` style - see PEP8

Issue #484, opened 2016-06-24
Wait for attach to complete before mount.


kerneltime:
We need to wait for linux kernel to create the /dev entry for the attached volume before the go code mounts.

Copy paste internal chat message
```
Looks like we have a race.. I pushed a branch of master and the build failed.. then I added a 3 second gap between attach command to esx completing to us mounting the FS and ci passed
failed: https://ci.vmware.run/vmware/docker-volume-vsphere/655
success: https://ci.vmware.run/vmware/docker-volume-vsphere/656
diff: https://github.com/vmware/docker-volume-vsphere/commit/d5df95064553af899566ffeaf97ff10ab3b65b25
The out of resource thingy is some thing else and needs to be looked at but it is not the reason for CI failing...
```

kerneltime:
Checking in https://github.com/vmware/docker-volume-vsphere/pull/485 as a temporary fix.

govint:
But esx attach completes and only then does ESX respond to the client. But
it's possible that the Linux kernel in the guest hasn't refreshed the devfs
entry yet in which case mount will fail.

kerneltime:
Yup.

kerneltime:
Fixed in https://github.com/vmware/docker-volume-vsphere/pull/520

PR #485, opened 2016-06-24
Temporary fix, till #484 is not fixed

kerneltime:


msterin:
LGTM - but Please add "added sleep" to merge message.

PR #486, opened 2016-06-28
On create, reject volume or datastore names > 100 chars.

### LINKED ISSUE #284: On create, reject file names > 255 chars on ‘create’
This is from a TODO in refcnt.go

msterin:
Fixes #284 

Test: adjusted name tests, run 'make all' and CI and manual (generated long names and made sure the error is issued)

Note that CI is now failing badly for other reasons ("out of resources") , but the naming tests pass

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -165,7 +165,7 @@ def make_create_cmd(opts, vmdk_path):
         # datastore is not VSAN
         policy_file = vsan_policy.policy_path(opts[kv.VSAN_POLICY_NAME])
         return "{0} -d {1} -c {2} --policyFile {3} {4}".format(VMDK_CREATE_CMD, format, size,
-                                                    policy_file, vmdk_path)
+                                                               policy_file, vmdk_path)
     else:
         return "{0} -d {1} -c {2} {3}".format(VMDK_CREATE_CMD, format, size, vmdk_path)
 
@@ -390,7 +390,7 @@ def parse_vol_name(full_vol_name):
     On parse errors raises ValidationError with syntax explanation
     """
     # note: \w in regexp is [a-zA-Z0-9_]
-    groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", full_vol_name)
+    groups = re.match(r"\A([a-zA-Z_][\w\_.]*)(@([a-zA-Z_][\w\_\-.]*))?$", full_vol_name)
     if not groups:
         raise ValidationError("Invalid syntax: '{0}'.\n"
                               "Valid syntax is volume@datastore, where 'volume' or 'datastore' "
@@ -837,19 +837,19 @@ def main():
     signal.signal(signal.SIGINT, signal_handler_stop)
     signal.signal(signal.SIGTERM, signal_handler_stop)
     try:
-        port=1019
+        port = 1019
         opts, args = getopt.getopt(sys.argv[1:], 'hp:')
     except getopt.error, msg:
         if msg:
             print >> sys.stderr, msg
         usage()
-        return(1)
-    for a,v in opts:
+        return 1
+    for a, v in opts:
         if a == '-p':
-            port=int(v)
+            port = int(v)
         if a == '-h':
             usage()
-            return(0)
+            return 0
 
     try:
         kv.init()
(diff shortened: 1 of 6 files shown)


esx_service/vmdk_ops.py
```diff
@@ -386,12 +390,19 @@ def parse_vol_name(full_vol_name):
...
     """
     # note: \w in regexp is [a-zA-Z0-9_]
-    groups = re.match("\A([\w\-.]+)@?([\w\-.]+)?$", full_vol_name)
+    groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", full_vol_name)
     if not groups:
-        raise ValidationError("Invalid syntax: '{0}'. " \
-                           "Valid syntax is volume@datastore, where volume or datastore" \
-                           "can contain [a-zA-Z0-9_-.]".format(full_vol_name))
-    return groups.groups()[0], groups.groups()[1]
+        raise ValidationError("Invalid syntax: '{0}'.\n"
```
govint: This allows the user to specify the volume name as "a.b.c" but "a.b.c@xyz" is failed. Do we want to allow multiple '.'  in a file name? Nothing wrong with that. But the above example can behave the same way whether datastore is provided or not.
>>> s='a.b.c@x'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
None
>>> s='a.b.c'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
<_sre.SRE_Match object; span=(0, 5), match='a.b.c'>

govint: Related to above example, the behavior isn't consistent, 'a.b.c@xxx' works fine, while 'a.b.c@x' doesn't.

>>> s='a.b.c@xxx'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
<_sre.SRE_Match object; span=(0, 9), match='a.b.c@xxx'>
>>> s='a.b.c@x'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
None

govint: Numbers should be allowed in the volume name as mentioned in the help message.

>>> s='1@x.y.z'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
None
>>> s='1@1'
>>> groups = re.match(r"\A([a-zA-Z_][\w\_.]+)(@([a-zA-Z_][\w\_\-.]+))?$", s)
>>> print(groups)
None

esx_service/vmdk_ops_test.py
```diff
@@ -41,10 +41,11 @@ def test_name_parse(self):
...
         testInfo = [
             #    full_name                       vol_name   datastore  success ?
-            ["MyVolume123-a_.vol@vsanDatastore", "MyVolume123-a_.vol", "vsanDatastore", True],
+            ["MyVolume123_a_.vol@vsanDatastore_11", "MyVolume123_a_.vol", "vsanDatastore_11", True],
+            ["no-dashes-please@datastore",       None,                 None,            False],
             ["Spaces NotGood@vsan",              None,                 None,            False],
             ["SGoodVold@bad ds with spaces",     None,                 None,            False],
-            ["Volume-123@dots.dot",              "Volume-123",        "dots.dot",       True],
+            ["Volume.123@dots.dot",              "Volume.123",        "dots.dot",       True],
             ["simple_volume",                    "simple_volume",      None,            True],
```
govint: Could include a test for ".volume" and ".volume@xxx" - these should fail and tests for volume and DS names starting with numbers

esx_service/vmdk_ops.py
msterin: yes, the current regexp does not accept single letter names (bug introduced when I forced 1st symbol to be a letter). Thanks for the catch, will fix. 

And numbers **are** allowed, just not in the first position

govint: Can we allow the same naming convenction that a user can use today to create vmdks - '-' is allowed and numbers are allowed anywhere in the volume name. Names like 'dvol-1' or '1-dvol' are allowed. Do we need to restrict that.

msterin: I do not want '-' to make sure names like "disk-001" can be created by snapshots/moves and never by `docker volume create`. We can revisit it later if needed.
On the second point (allow to start from any legal character) - I agree, will fix

andrewjstone:
LGTM

govint:
Ship it.

PR #487, opened 2016-06-28
Force build and clean before making deploy-esx and deploy-vm

### LINKED ISSUE #445: make deploy doesn't remove the VIB thats present and install fails
make deploy doesn't first remove the existing vib and hence the deploy of the VIB itself fails. "make all" seems to do this as part of a cleanup that happens before the build starts and hence installs the vib on the ESX host.

But if we run "make;make deploy ...." then VIB install fails. Whichever way its run the make must install components as expected - either as "make all" or as "make deploy". "make deploy" must deploy.

Most of the time user wants to build and deploy and run tests and hence "make;make deploy ..." is the command used.

Right now this is what happens, vib is skipped.

make --directory=vmdk_plugin deploy
make[1]: Entering directory `/home/administrator/vpl-kv/docker-volume-vsphere/vmdk_plugin'
../misc/scripts/deploy-tools.sh deployesx "10.110.48.29" "../build/vmware-esx-vmdkops-0.1.0.tp.2.172872e.vib"

=> Deploying to ESX root@10.110.48.29

Installation Result:
   Message: Host is not changed.
   Reboot Required: false
   VIBs Installed:
   VIBs Removed:
   VIBs Skipped: VMWare_bootbank_esx-vmdkops-service_1.0.0-0.0.1 <-------
vmdkops-opsd is running pid=1013003366
../misc/scripts/deploy-tools.sh deployvm "10.110.48.56 10.110.48.38" ../build

msterin:
Fixes #445

Forces makefile to build the code and run clean-esx or clean-vm before building deploy-esx and deploy-vm respectively.

It does make a few extra cycles re-building packages when it's not really needed (since "package" goal does not check dependencies, and is being called from build-all) - but IMO it's well worth the improved reliability of manual builds

Tested: make all, and CI (may be failing until CI is fixed)


msterin:
oops... looks like CI won't work with it too well. Withdrawing for now

PR #488, opened 2016-06-28
Force clean before making deploy-esx and deploy-mv

### LINKED ISSUE #445: make deploy doesn't remove the VIB thats present and install fails

msterin:
Fixes #445

Forces makefile  run clean-esx or clean-vm before building deploy-esx and deploy-vm
respectively.

It could take a few extra cyclesto try to clean when it's not really needed- but IMO it's well worth the improved reliability of manual builds

Tested: make all, and CI (Ci if failing until CI is fixed)


andrewjstone:
LGTM

kerneltime:
LGTM. I will clean up the ci scripts to not call clean explicitly as this will call clean.

PR #492, opened 2016-06-29
Debug OOR

govint:
PR only for debug - pls. don't review.


govint:
Served to identify why the volume create errors happened. Closing.

PR #498, opened 2016-06-30
Re-raise any exception from running unittests in vmdk_ops_test.py

### LINKED ISSUE #491: Failing Tests do not cause CI to fail
The test failure shown in #490 can still result in CI passes. My guess is that whatever command is running the tests is not consuming the error code somehow.

andrewjstone:
Fixes #491

### DIFF
--- a/esx_service/vmdk_ops_test.py
+++ b/esx_service/vmdk_ops_test.py
@@ -37,7 +37,7 @@ class VolumeNamingTestCase(unittest.TestCase):
     """Unit test for operations with volume names (volume@datastore)"""
 
     def test_name_parse(self):
-        """checks name parsing and error checks 
+        """checks name parsing and error checks
         'volume[@datastore]' -> volume and datastore"""
         testInfo = [
             #    full_name                       vol_name   datastore  success ?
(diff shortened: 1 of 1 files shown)


esx_service/vmdk_ops_test.py
```diff
@@ -238,5 +238,6 @@ def test_failure(self):
         if not paths:
             logging.debug("Directory clean up - removing  %s", path)
             os.removedirs(path)
-    #suite = unittest.TestLoader().loadTestsFromTestCase(TestStringMethods)
-    #unittest.TextTestRunner(verbosity=2).run(suite)
+
+        # If the unittest failed, re-raise the error
+        raise
```
govint: Once raised which code handles the exception. Its best if all the tests or test scripts run thru as they a single test run tells us what all problems are there than if the test itself stops on the first exception.

andrewjstone: All tests in this module and all other python tests in test-esx will continue to run. Any tests not under the `test-esx` target in the Makefile will not however. This is just fixing a bug that I assume was recently introduced here, in which a failing test didn't return a failing (non-zero) exit code from the process. All other python test files end with `unittest.main()` which will run through all tests even if one fails, but return a non-zero exit code if there is actually a failure after they all run.

msterin:
How the fix was tested ?

andrewjstone:
This fix was tested by commenting out the @unittest.SkipIf on lines 113 and 114 of vmdk_ops_test.py and running the test on a local vm setup, then checking the return code with $?. The same thing was done in vsan_policy.py. vsan_policy.py returned 1 as the error code, but vmdk_ops_test returned 0. Making this change caused vmdk_ops_test to return 2. I then uncommented the @unittest.SkipIf and re-ran to ensure that vmdk_ops_test passed and returned 0 if no error occurred.

msterin:
lgtm

PR #499, opened 2016-06-30
Fixes for issues found on CI.

kerneltime:
Remove no-op cleanup
Add retry for requests on the client side
Restart docker before test run

### DIFF
--- a/misc/scripts/deploy-tools.sh
+++ b/misc/scripts/deploy-tools.sh
@@ -184,21 +184,11 @@ function cleanvm {
     for IP in $IP_LIST
     do
         TARGET=root@$IP
-        cleanupVolumes
         setupVMType
         log "cleanvm: Cleaning up on $TARGET : $FILE_EXT"
         cleanupVMPre
         cleanupVM
         cleanupVMPost
-    done
-}
-
-function cleanupVolumes {
-    log "cleanupVolumes: Asking docker $TARGET to remove volumes ($VOLUMES)"
-    for vol in $VOLUMES
-    do
-        $SSH $TARGET "if docker volume ls | $GREP -q $vol; then \
-        docker volume rm $vol; fi "
     done
 }
 
@@ -208,8 +198,10 @@ function cleanupVMPre {
     then
         $SSH $TARGET systemctl stop $PLUGIN_NAME
         $SSH $TARGET $PIDOF $PLUGIN_NAME
+        $SSH $TARGET systemctl restart docker
     else
         $SSH $TARGET service $PLUGIN_NAME stop
+        $SSH $TARGET service docker restart
     fi
 }
 
--- a/vmdk_plugin/vmdkops/esx_vmdkcmd.go
+++ b/vmdk_plugin/vmdkops/esx_vmdkcmd.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	log "github.com/Sirupsen/logrus"
 	"syscall"
+	"time"
 	"unsafe"
 )
 
@@ -39,6 +40,7 @@ type EsxVmdkCmd struct{}
 
 const (
 	commBackendName string = "vsocket"
+	maxRetryCount          = 5
 )
 
 // A request to be passed to ESX service
@@ -85,19 +87,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 	defer C.free(unsafe.Pointer(ans))
 
 	// connect, send command, get reply, disconnect - all in one shot
-	_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
-
-	if err != nil {
-		var errno syscall.Errno
-		errno = err.(syscall.Errno)
-		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
-		if errno == syscall.ECONNRESET {
-			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
-			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+	retryCount := 0
+	for {
+		retryCount++
+		_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+		if err == nil {
+			break
+		} else {
+			var errno syscall.Errno
+			errno = err.(syscall.Errno)
+			msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
+			if retryCount >= maxRetryCount {
+				if errno == syscall.ECONNRESET {
+					msg += " Hit communication issue with ESX (vmci or ESX service)\n"
+					msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+				}
+				log.Warnf(msg)
+				return nil, errors.New(msg)
+			}
+			msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
+			log.Warnf(msg)
+			time.Sleep(3 * time.Millisecond)
 		}
-		log.Warnf(msg)
-		return nil, errors.New(msg)
 	}
+
 	response := []byte(C.GoString(ans.buf))
 	C.free(unsafe.Pointer(ans.buf))
 	err = unmarshalError(response)
--- a/vmdk_plugin/vmdkops/esx_vmdkcmd.go
+++ b/vmdk_plugin/vmdkops/esx_vmdkcmd.go
@@ -88,7 +88,8 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 
 	// connect, send command, get reply, disconnect - all in one shot
 	retryCount := 0
-	for {
+
+   for {
 		retryCount++
 		_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
 		if err == nil {
@@ -105,9 +106,15 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 				log.Warnf(msg)
 				return nil, errors.New(msg)
 			}
-			msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
-			log.Warnf(msg)
-			time.Sleep(3 * time.Millisecond)
+
+         if errno == syscall.ECONNRESET {
+            msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
+            log.Warnf(msg)
+            time.Sleep(3 * time.Millisecond)
+         } else {
+            log.Errorf(msg)
+            return nil, errors.New(msg)
+         }
 		}
 	}
 
--- a/misc/drone-scripts/deploy-and-test-wrapper.sh
+++ b/misc/drone-scripts/deploy-and-test-wrapper.sh
@@ -83,9 +83,6 @@ truncate_esx_logs
 
 log "starting deploy and test"
 
-make clean-vm
-make clean-esx
-
 if make deploy-esx deploy-vm testasroot testremote TEST_VOL_NAME=vol-build$BUILD_NUMBER;
 then
   dump_esx_info $ESX
--- a/misc/scripts/refcnt_test.sh
+++ b/misc/scripts/refcnt_test.sh
@@ -27,13 +27,6 @@
 # For now (TP) we still need basic validation
 #
 
-# make sure all vars are set before usage
-set -o nounset
-
-# make sure scripts terminate (and test fails) on first failure
-set -e
-
-
 DIR=$(dirname ${BASH_SOURCE[0]})
 . $DIR/wait_for.sh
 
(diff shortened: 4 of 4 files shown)


misc/scripts/deploy-tools.sh
```diff
@@ -193,23 +192,16 @@ function cleanvm {
...
-    done
-}
-
 function cleanupVMPre {
     $SSH $TARGET "$IS_SYSTEMD"
     if [ $? -eq 0 ]
     then
         $SSH $TARGET systemctl stop $PLUGIN_NAME
         $SSH $TARGET $PIDOF $PLUGIN_NAME
+	$SSH $TARGET systemctl restart docker
```
andrewjstone: Nit: Spacing is 3 instead of 4

misc/scripts/deploy-tools.sh
```diff
@@ -193,23 +192,16 @@ function cleanvm {
...
     else
         $SSH $TARGET service $PLUGIN_NAME stop
+	$SSH $TARGET service docker restart
```
andrewjstone: Nit: spacing

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +86,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-	_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
-
-	if err != nil {
-		var errno syscall.Errno
-		errno = err.(syscall.Errno)
-		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
-		if errno == syscall.ECONNRESET {
-			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
-			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+	retryCount := 10
```
pdhamdhere: I am not comfortable with retry logic unless you have reasoned through all ESX OPs and absolutely sure they are idempotent.

pdhamdhere: For e.g. in unmount=>detach, assume ESX received 'detach' and completed successfully. Retry in this case is going to return error.

govint: Yes, these ops aren't idempotent and a retry will fail while a disk may have been created or worse attached and the client doesn't know about it.

Instead perk these failures upwards and return a "transient" error to the caller and the caller can decide whether to retry. The retry must be done by first fetching current status of the volume (say on attach)  or figure if the volume is there (create/delete) and only then decide whether a retry is needed.

For this issue can may be generate the error and a separate issue to do the error handling appropriately.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +86,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
+				}
+				log.Warnf(msg)
+				return nil, errors.New(msg)
+			}
+			msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
+			log.Warnf(msg)
+			time.Sleep(3 * time.Millisecond)
+		} else {
+			break
 		}
```
govint: As mentioned this can be dangerous and shouldn't be handled here.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
govint: For example, plugin server "always" tries to attach the volume provided. And errors out if the volume is attached. So, if the first attach was OK and a network error happened the client is going to redo the attach and this time the server sends back an error which the client receives and interprets wrongly (having already attached the disk).

Pls. identify errors that are transient and push upwards and let the upper layer (plugin.go) decide on getting current status of the volume and decide to retry.

misc/scripts/deploy-tools.sh
kerneltime: Will do.. the Mac vim was not setup.. and I have not yet gotten to see why my dev box is down..

misc/scripts/deploy-tools.sh
kerneltime: will do

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: Is there something inherent about the API as to why they cannot be idempotent.. ? If the volume is attached second attach should be no-op and the same is true for detach.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: I agree that our current implementation might not be idempotent and this PR should wait but I do not see why our ESX APIs cannot be idempotent.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
msterin: Actually, that's exactly how the current code works (If the volume is attached second attach is a no-op). 
For detach it's different (for historical reasons) but could be easily changed with no side effects - just open a bug and assign to me, or fix detach_disk:713 (return err(msg)) to return None , and adjust the log msg to "Info"

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +86,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-		errno = err.(syscall.Errno)
-		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
-		if errno == syscall.ECONNRESET {
-			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
-			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+	retryCount := 10
+	for {
+		retryCount--
+		_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+		if err != nil {
```
msterin: Nit: I think doing 
```Go
if err == nil {
   break
} 

.... the rest of the code , for err != nil
```
is easier to read , and it makes less offset/nesting too.
Please consider.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +86,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
+			var errno syscall.Errno
+			errno = err.(syscall.Errno)
+			msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
+			if retryCount == 0 {
+				if errno == syscall.ECONNRESET {
+					msg += " Hit communication issue with ESX (vmci or ESX service)\n"
```
msterin: I think it's better to say "Communication issue...." instead of "hit communication issue"

vmdk_plugin/vmdkops/esx_vmdkcmd.go
msterin: on retry count - please make it a module const

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +86,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
+					msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+				}
+				log.Warnf(msg)
+				return nil, errors.New(msg)
+			}
+			msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
+			log.Warnf(msg)
+			time.Sleep(3 * time.Millisecond)
```
msterin: Please add a comment "// TBD - remove when #xxx is fixed"

msterin:
It is not clear to me if this fix cures the CI (I understand it's a patch for now). If it does - LGTM pending fixes or replies to my nits. If it does not then I see no point in applying the "Add retry for requests on the client side" commit. the other 2 (Remove no-op cleanup, Restart docker before test run) are LGTM is you want to separate them

vmdk_plugin/vmdkops/esx_vmdkcmd.go
govint: Not really, the plugin server returns an error that the disk is already
attached on a second attach (without detaching the volume). Now, do we
really want the esx_vmdkcmd.go handling server errors? It knows nothing of
volumes.

a. A retry can cause the server to return an error.

b. If the error is passed up then plugin.go would assume that the disk
attach failed and will return to the caller (in plugin.go:mountVolume()).

c. It will not try to mount the volume on the client.

To be idempotent, the server must return OK status vs. error. Or the client
must fetch status and figure where the volume is mounted.

Docker 1.12 allows returning plugin specific status from server and that
can be leveraged to figure if the volume is attached to the client.

msterin: @govint - please do check the code before posting comments. The disk_attach() code says: 
```python
  # Check if this disk is already attached, and if it is - skip the attach
    device = findDeviceByPath(vmdk_path, vm)
    if device:
        # Disk is already attached.
        logging.warning("Disk %s already attached. VM=%s",
                        vmdk_path, vm.config.uuid)
        setStatusAttached(vmdk_path, vm)
        return busInfo(device.unitNumber,
                       device.controllerKey - offset_from_bus_number)
```

govint: Oops, absolutely right. Sorry, its fine as is.

kerneltime:
Based on the feedback for the vSocket bug in 4.3 4.4 and 4.5,  I am trying a few more things on 4.2 kernel to make sure that the only pending issue on VMCI front is the bug with vSocket. Then will get back to a workaround for Photon GA.

govint:
LGTM.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +87,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-		if errno == syscall.ECONNRESET {
-			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
-			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+	retryCount := 0
+	for {
+		retryCount++
+		_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+		if err == nil {
+			break
+		} else {
```
pdhamdhere: Why are we retrying all errors? If there is genuine error from backend, retry may return irrelevant error.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -85,19 +87,30 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
-		if errno == syscall.ECONNRESET {
-			msg += " Hit communication issue with ESX (vmci or ESX service)\n"
-			msg += " Please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+	retryCount := 0
+	for {
+		retryCount++
+		_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+		if err == nil {
+			break
```
pdhamdhere: nit: spacing seems off either here or at L102.

pdhamdhere:
Please update PR description with analysis of issue and how this change resolves it. Also add "Testing Done".

Regarding PR itself, I want to register my concerns about retry logic. Backend operations are not idempotent. Took a quick look and here is what I see in code;
**Create**: Retry returns error if first attempt is successful
**Attach**: Retry returns success if already attached
**detach**: Retry returns error if already detached
**remove**: returns error if already successful

It seems like we are introducing "retry" logic without reasoning through possible repercussions. 

If you want to merge PR with retry code then I suggest following;
1. Make ESX backend operations idempotent
2. Test change by injecting failure on first attempt assuming operation is successful
3. Retry iff error are related to socket communication?

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: I don't see it.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: will make it specific to conn reset

kerneltime:
Plan of action. Will address minor review comments here and file an issue for API idempotency. Merge this PR to unblock CI.

kerneltime:
https://github.com/vmware/docker-volume-vsphere/issues/501

kerneltime:
The retry logic is primarily to work around VMCI errors where requests do not reach ESX.. with this most cases of failures will be handled. In the absence of idempotent APIs, there are few cases where the original request was processed by ESX and a retry was issued but without the retry logic the operation would fail either ways. There are challenges in making API idempotent which is part of #501 and this change for now reduces the number of failures we see but does not eliminate it.

msterin:
LGTM

PR #504, opened 2016-06-30
Retry on ETIMEDOUT as well

kerneltime:


msterin:
I still do not understand why and how this timeout can happen. Let's nor rush with the patches, and instead check if this is indeed VMCI failure - with latest kernels all the tests should pass since the bug is not there, right ?

kerneltime:
Yup or an older kernel.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -107,14 +107,14 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-         if errno == syscall.ECONNRESET {
-            msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
-            log.Warnf(msg)
-            time.Sleep(3 * time.Millisecond)
-         } else {
-            log.Errorf(msg)
-            return nil, errors.New(msg)
-         }
+			if errno == syscall.ECONNRESET || errno == syscall.ETIMEDOUT {
+				msg += fmt.Sprintf(" retryCount=%d Retrying Request\n", retryCount)
```
govint: nit - define a function to validate if given error is among errors identified as retryable.

govint: Pls. print what error was returned in the log below.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -107,14 +107,14 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
+				log.Warnf(msg)
+				time.Sleep(3 * time.Millisecond)
+			} else {
+				log.Errorf(msg)
+				return nil, errors.New(msg)
+			}
 		}
```
govint: Add a warning log that retries are "done and giving up"?

PR #505, opened 2016-07-01
Fix vsan policy test failures on CI

andrewjstone:
Tests were failing with "out of resources" messages due to the fact that we were
using a `(("hostFailuresToTolerate" i1))` for the original config. This works
when there are actually multiple ESXs in the VSAN cluster, but fails when a
single ESX is used because you can't tolerate one host failuer with one host.

This change uses `(("hostFailuresToTolerate" i0))` for both the old and new
configs and adds `(("proportionalCapacity" i0))` to the original config which is
harmless, but allows for 2 different configs to allow for testing policy update
code paths.

This was tested via `make all` against both our internal testbed with 3 ESX VSAN
cluster and CI with one ESX.

### DIFF
--- a/esx_service/vmdk_ops_test.py
+++ b/esx_service/vmdk_ops_test.py
@@ -37,7 +37,7 @@ class VolumeNamingTestCase(unittest.TestCase):
     """Unit test for operations with volume names (volume@datastore)"""
 
     def test_name_parse(self):
-        """checks name parsing and error checks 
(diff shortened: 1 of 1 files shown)


msterin:
LGTM

govint:
Ship it. I'm thinking we must run lint FWIW on our code on each checkin (easily we get typos that won't ever be found till the code runs)

PR #506, opened 2016-07-01
Debug oor

govint:
Not for review


govint:
Done debugging issue of file handle limit error during CI tests.

PR #512, opened 2016-07-07
Remove DevFS entry for VSAN objects after format.

kerneltime:
Testing:
Ran make all on local setup over VSAN and VMFS. Manually checked ESX for devfs being in a clean state.

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -199,7 +199,7 @@ def validate_opts(opts, vmdk_path):
                + 'Valid options and defaults: ' \
                + '{0}'.format(zip(list(valid_opts), defaults))
         raise ValidationError(msg)
-    
+
     if kv.SIZE in opts:
         validate_size(opts[kv.SIZE])
     if kv.VSAN_POLICY_NAME in opts:
@@ -239,7 +239,8 @@ def validate_disk_allocation_format(format):
         raise ValidationError('Disk Allocation Format \'{0}\' does not exist. Valid options are: {1}'.format(format, kv.VALID_ALLOCATION_FORMATS))
 
 
-def getVMDKUuid(vmdk_path):
+# Returns the UUID if the vmdk_path is for a VSAN backed.
+def get_vsan_uuid(vmdk_path):
     f = open(vmdk_path)
     data = f.read()
     f.close()
@@ -252,31 +253,64 @@ def getVMDKUuid(vmdk_path):
     except:
         return None
 
+# Create and return the devFS path for a VSAN object UUID.
+# Also returns true if clean up is needed.
+# cleanup_vsan_devfs should be called to clean up before
+# removing the VSAN object.
+def get_vsan_devfs_path(uuid):
+
+    logging.debug("Got volume UUID %s", uuid)
+
+    # Objtool creates a link thats usable to 
+    # read write to vsan object.
+    cmd = "{0} {1}".format(OBJ_TOOL_CMD, uuid)
+    rc, out = RunCommand(cmd)
+    fpath="/vmfs/devices/vsan/{0}".format(uuid)
+    if rc == 0 and os.path.isfile(fpath):
+        return fpath, True
+    logging.error("Failed to create devFS node for %s", uuid)
+    return None, False
+
+# Clean up vsan devfs path
+def cleanup_vsan_devfs_path(devfs_path):
+    try:
+        os.remove(devfs_path)
+        logging.debug("Unlinked %s", devfs_path)
+	return True
+    except OSError as ex:
+        logging.error("Failed to remove backing device %s",
+                      devfs_path, str(ex))
+    return False
+
+# Returns the flat file for a VMDK.
+def get_backing_flat_file(vmdk_path):
+    return vmdk_path.replace(".vmdk", "-flat.vmdk")
 
 # Return a backing file path for given vmdk path or none
-# if a backing can't be found.
-def getVMDKBacking(vmdk_path):
-    flatBacking = vmdk_path.replace(".vmdk", "-flat.vmdk")
+# if a backing can't be found. Returns True if clean up 
+# is needed. Do not forget to call cleanup_backing_device when done.
+def get_backing_device(vmdk_path):
+    flatBacking = get_backing_flat_file(vmdk_path)
     if os.path.isfile(flatBacking):
-        return flatBacking
+        return flatBacking, False
 
-    uuid = getVMDKUuid(vmdk_path)
+    uuid = get_vsan_uuid(vmdk_path)
 
     if uuid:
-        logging.debug("Got volume UUID %s", uuid)
-        # Objtool creates a link thats usable to format the
-        # vsan object.
-        cmd = "{0} {1}".format(OBJ_TOOL_CMD, uuid)
-        rc, out = RunCommand(cmd)
-        fpath = "/vmfs/devices/vsan/{0}".format(uuid)
-        if rc == 0 and os.path.isfile(fpath):
-            return fpath
-    return None
 
+        return get_vsan_devfs_path(uuid)
+
+    return None, False
+
+def cleanup_backing_device(backing, cleanup_device):
+    if cleanup_device:
+    	return cleanup_vsan_devfs_path(backing)
+    return True
 
 def formatVmdk(vmdk_path, vol_name):
-    # Get backing for given vmdk path.
-    backing = getVMDKBacking(vmdk_path)
+    # Get backing for given vmdk path. This is the backing
+    # device that will be formatted.
+    backing, needs_cleanup = get_backing_device(vmdk_path)
 
     if backing is None:
         logging.warning("Failed to format %s.", vmdk_path)
@@ -285,6 +319,10 @@ def formatVmdk(vmdk_path, vol_name):
     # Format it as ext4.
     cmd = "{0} {1} {2}".format(MKFS_CMD, vol_name, backing)
     rc, out = RunCommand(cmd)
+
+    # clean up any resources backing file might have created
+    # during get_backing_device function call.
+    cleanup_backing_device(backing, needs_cleanup)
 
     if rc != 0:
         logging.warning("Failed to format %s - %s", vmdk_path, out)
@@ -296,7 +334,6 @@ def formatVmdk(vmdk_path, vol_name):
                 % vmdk_path)
     return None
 
-
 # Return error, or None for OK
 def removeVMDK(vmdk_path):
     logging.info("*** removeVMDK: %s", vmdk_path)
@@ -304,6 +341,7 @@ def removeVMDK(vmdk_path):
     rc, out = RunCommand(cmd)
     if rc != 0:
         return err("Failed to remove %s. %s" % (vmdk_path, out))
+
     return None
 
 
@@ -317,12 +355,12 @@ def getVMDK(vmdk_path, vol_name):
 
 def listVMDK(vm_datastore):
     """
-    Returns a list of volume names (note: may be an empty list). 
+    Returns a list of volume names (note: may be an empty list).
     Each volume name is returned as either `volume@datastore`, or just `volume`
     for volumes on vm_datastore
     """
     vmdks = vmdk_utils.get_volumes()
-    # build  fully qualified vol name for each volume found 
+    # build  fully qualified vol name for each volume found
     return [{u'Name': get_full_vol_name(x['filename'], x['datastore'], vm_datastore),
              u'Attributes': {}} \
             for x in vmdks]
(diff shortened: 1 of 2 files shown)


esx_service/vmdk_ops.py
```diff
@@ -294,16 +318,17 @@ def formatVmdk(vmdk_path, vol_name):
             return err(
                 "Unable to format %s and unable to delete volume. Please delete it manually."
                 % vmdk_path)
+    cleanupVSANDevFSNode(vmdk_path)
```
pdhamdhere: Is there any harm in doing this independent of return code from MKFS command? Otherwise there is still possibility of leak if mkfs failed internally with genuine error.

kerneltime: Good catch. Will fix this.

esx_service/vmdk_ops.py
```diff
@@ -268,11 +282,21 @@ def getVMDKBacking(vmdk_path):
         # vsan object.
         cmd = "{0} {1}".format(OBJ_TOOL_CMD, uuid)
         rc, out = RunCommand(cmd)
-        fpath = "/vmfs/devices/vsan/{0}".format(uuid)
+        fpath = getVSANDevFSPath(uuid)
```
msterin: I think the fix is doing some redundant steps and it complicates it.
We already have all data in getVMDKBacking()
I'd suggest just return (fpath, devfs_path) from getVMKDBacking() - i.e `return flatBacking, None`, `return fpath, fpath` and `return None, None)` in 3 `return` in the functions. And then in formatVmdk() simply do 
```python
 backing, devfs_path = getVMDKBacking(vmdk_path)
 ...
  # Cleanup temporary devFS node:
  if devfs_path:
       try: 
            os.remove(dev_fs_path)
            logging.debug(...)
       except: 
            logging.exception(....)
```

esx_service/vmdk_ops.py
```diff
@@ -268,11 +282,21 @@ def getVMDKBacking(vmdk_path):
...
         if rc == 0 and os.path.isfile(fpath):
             return fpath
     return None
 
+def cleanupVSANDevFSNode(vmdk_path):
+    devFSNode = getVSANDevFSNode(vmdk_path)
```
msterin: we should simply pass the devfs_path instead of reconstructing it (not relevant if you take my suggestion)

msterin:
Thanks for the fix. It looks correct but I think it should be simpler - no need to recomstruct the path and no need for new functions. Please see my suggestion.

kerneltime:
After discussing this further with @msterin, the answer is to make the code a bit more symmetric and explicit in terms of creation and cleanup. Will update soon.

esx_service/vmdk_ops.py
```diff
@@ -239,7 +239,8 @@ def validate_disk_allocation_format(format):
         raise ValidationError('Disk Allocation Format \'{0}\' does not exist. Valid options are: {1}'.format(format, kv.VALID_ALLOCATION_FORMATS))
 
 
-def getVMDKUuid(vmdk_path):
+# Returns the UUID if the vmdk_path is for a VSAN backed.
+def get_vsan_uuid(vmdk_path):
```
msterin: yep, this is more accurate and style compliant too :-)

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,57 @@ def getVMDKUuid(vmdk_path):
     except:
         return None
 
+# Create and return the devFS path for a VSAN object UUID.
+def get_vsan_devfs_path(uuid):
+
+    logging.debug("Got volume UUID %s", uuid)
+
+    # Objtool creates a link thats usable to
```
msterin: Pls add to the comment something like "please note that simply deleting the VMDK and backing object does not remove the link, so it needs to be taken care of explicitly after use"

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,57 @@ def getVMDKUuid(vmdk_path):
     except:
         return None
 
+# Create and return the devFS path for a VSAN object UUID.
```
msterin: comment is missing the second return value ("needs cleanup" ?)

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,57 @@ def getVMDKUuid(vmdk_path):
...
+# Returns the flat file for a VMDK.
+def get_backing_flat_file(vmdk_path):
+    return vmdk_path.replace(".vmdk", "-flat.vmdk")
 
 # Return a backing file path for given vmdk path or none
-# if a backing can't be found.
-def getVMDKBacking(vmdk_path):
-    flatBacking = vmdk_path.replace(".vmdk", "-flat.vmdk")
+# if a backing can't be found. Returns True if clean up 
+# is needed. Always call cleanup_backing_device when done.
```
msterin: nit: "do not forget to call cleanup_backing_device" is probably a better comment.

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,57 @@ def getVMDKUuid(vmdk_path):
...
-            return fpath
-    return None
 
+        return get_vsan_devfs_path(uuid)
+
+    return None, False
+
+def cleanup_backing_device(backing, cleanup_device):
+    if cleanup_device:
+    return cleanup_vsan_devfs_path(backing)
```
msterin: is this offset right ? (it does look wrong on the githuv review board)

msterin:
One offset looks wrong and  few nits, mainly comments-related. 
LGTM contingent on the offset fix (or explanation)

esx_service/vmdk_ops.py
kerneltime: Will fix this.

esx_service/vmdk_ops.py
kerneltime: done.

esx_service/vmdk_ops.py
kerneltime: done

govint:
Does os.remove() in Python end up doing the call to the vmkernel to remove the device node in the vmkernel? Otherwise we may be removing the FS nodes but the vmkernel nodes won't get removed.

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,62 @@ def getVMDKUuid(vmdk_path):
...
+# if a backing can't be found. Returns True if clean up 
+# is needed. Do not forget to call cleanup_backing_device when done.
+def get_backing_device(vmdk_path):
+    flatBacking = get_backing_flat_file(vmdk_path)
     if os.path.isfile(flatBacking):
-        return flatBacking
+        return flatBacking, False
 
-    uuid = getVMDKUuid(vmdk_path)
+    uuid = get_vsan_uuid(vmdk_path)
```
govint: This can be vvol (too) later, we could not have to special case function names as vsan and vvol(later) or other supported types. Just mentioning.

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,62 @@ def getVMDKUuid(vmdk_path):
...
+        logging.debug("Unlinked %s for flag %s", backing, cleanup_device)
+	return True
+    except OSError as ex:
+        logging.error("Failed to remove backing device %s",
+                      backing, str(ex))
+    return False
+
+# Returns the flat file for a VMDK.
+def get_backing_flat_file(vmdk_path):
+    return vmdk_path.replace(".vmdk", "-flat.vmdk")
```
govint: This won't work once snaps are supported (which is pretty soon). Instead please use the vmdk descriptor file to get the backing file. For a delta disk there is no "-flat" in the file name. A delta disk looks like this:
Ubuntu-14.04-64-LTS-IMG-000001-delta.vmdk

And the corresponding descriptor file will clearly say whats the backing, lets use that vs. doing a string replace (was ok till now, not further).

[root@localhost:/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/Ubuntu-14.04-64-LTS-IMG] cat Ubuntu-14.04-64-LTS-IMG-000001.vmdk
# Disk DescriptorFile
version=1
encoding="UTF-8"
CID=764b5bfd
parentCID=91a8e92f
isNativeSnapshot="no"
createType="vmfsSparse"
parentFileNameHint="/vmfs/volumes/523c14b7-e5561042-5e2f-002481aaea6c/Ubuntu-14.04-64-LTS-IMG/Ubuntu-14.04-64-LTS-IMG.vmdk"
# Extent description
RW 67108864 VMFSSPARSE "Ubuntu-14.04-64-LTS-IMG-000001-delta.vmdk"

# The Disk Data Base
#DDB

ddb.longContentID = "caa378fba2e6a2f300876aa2764b5bfd"


Use the extent file mentioned in the descriptor.

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,62 @@ def getVMDKUuid(vmdk_path):
...
-    return None
 
+        return get_vsan_devfs_path(uuid)
+
+    return None, False
+
+def cleanup_backing_device(backing, cleanup_device):
+    if cleanup_device:
+    	return cleanup_vsan_devfs_path(backing)
+    return True
```
govint: Nit - just as a check could verify if the backing is a VSAN backing, something like "vsan:......".  What would be an assertion in C code :-)

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,62 @@ def getVMDKUuid(vmdk_path):
...
+
+    logging.debug("Got volume UUID %s", uuid)
+
+    # Objtool creates a link thats usable to 
+    # read write to vsan object.
+    cmd = "{0} {1}".format(OBJ_TOOL_CMD, uuid)
+    rc, out = RunCommand(cmd)
+    if rc == 0 and os.path.isfile(fpath):
+        return "/vmfs/devices/vsan/{0}".format(uuid), True
+    return None, False
```
govint: A debug log  to say that we could not get a device path for the UUID.

esx_service/vmdk_ops.py
msterin: the current code *explicitly* works for vsan only (see node names, i.e. /dev/**vsan**) , if we add vvol or oher support we will add functions or rename

esx_service/vmdk_ops.py
kerneltime: We need to move formatting into the VM and this function for now is only called for create.

msterin: there are many other things which will not work with snaps. They are all will be fixed with snaps work. 
Let's not pollute this PR with "future-maybes" , this specific code is not impacted by the PR - just the function is renamed to comply with code style

esx_service/vmdk_ops.py
kerneltime: Cleanup is used as a symmetric function for get, so this would be a nice check but ideally done only for debug. Also, once this code moves into VM these functions will no longer be present.

esx_service/vmdk_ops.py
msterin: good point - but this code is really the old one, just an extra return value is added

kerneltime: Will do.

msterin:
lgtm

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,64 @@ def getVMDKUuid(vmdk_path):
...
+
+    logging.debug("Got volume UUID %s", uuid)
+
+    # Objtool creates a link thats usable to 
+    # read write to vsan object.
+    cmd = "{0} {1}".format(OBJ_TOOL_CMD, uuid)
+    rc, out = RunCommand(cmd)
+    fpath="/vmfs/devices/vsan/{0}".format(uuid)
+    if rc == 0 and os.path.isfile(fpath):
+        return fpath, True
```
andrewjstone: Nit: Isn't True always returned if fpath isn't None? Seems like this variable isn't needed at all.

esx_service/vmdk_ops.py
```diff
@@ -252,31 +253,64 @@ def getVMDKUuid(vmdk_path):
...
-# if a backing can't be found.
-def getVMDKBacking(vmdk_path):
-    flatBacking = vmdk_path.replace(".vmdk", "-flat.vmdk")
+# if a backing can't be found. Returns True if clean up 
+# is needed. Do not forget to call cleanup_backing_device when done.
+def get_backing_device(vmdk_path):
+    flatBacking = get_backing_flat_file(vmdk_path)
     if os.path.isfile(flatBacking):
-        return flatBacking
+        return flatBacking, False
```
andrewjstone: Same comment as above. We can be conditional on the None and always do cleaunp if flatBacking is valid.

kerneltime: Makes sense, will make the change.

andrewjstone:
LGTM - Feel free to ignore my nits.

One thing for the future. I think we should be using docstrings for function comments instead of comments above the function.

esx_service/vmdk_ops.py
kerneltime: I will have to add further checks on top for clean up. This way the decision is bubbled up for clean up.

esx_service/vmdk_ops.py
kerneltime: On second thoughts, the presence of string does not imply clean up.

kerneltime:
ignored :-)

PR #514, opened 2016-07-07
WIP: Add a draft of the permissions spec

andrewjstone:
WIP: DO NOT MERGE

This is useful for discussion/commenting and is fairly minimal right now.

### DIFF
--- a/doc/volume-permissions-spec.md
+++ b/doc/volume-permissions-spec.md
@@ -1,0 +1,67 @@
+### Introduction
+
+Currently the Docker volume driver for VSphere allows any container on any VM with access to a given datastore
+to create, mount, and delete volumes at will. There are no limits to the size of the volumes that
+can be created or quotas for datastore usage for container volumes. This lack of access control
+takes the management of storage out of the hands of the IT administrator and puts it in the hands of
+the developer. This is a worrying proposition to many IT admins, and since they are the ones that
+must allow deployment of the driver and it's associated host agent, we must allow them to regain
+control with proper enforcement mechanisms that limit damage due to malfeasance or mistake.
+
+### Permissions Model
+
+Since docker containers run inside VMs, and can have root permissions, we cannot rely on the VM user
+for access control. Permissions must be granted to a VM, since VMs are created and managed
+by the IT administrator. This brings up a few issues:
+
+  1. How are VMs identified in order to grant permissions?
+  2. What is the granularity of a permission?
+  3. Where are permissions stored?
+  4. What are the actual permissions?
+
+In order to answer these questions, we must consider 2 scenarios. The first scenarios is permissions
+for a single VM with a local datastore, and the second is multiple VMs with a shared datastore. The
+model must work for both scenarios in order to be useful.
+
+1. VMs can share names so must be identified by unique IDs. These IDs should be unique across a
+   cluster of hosts. If desired, we can associate many IDs with a Role that can be used to grant
+   permissions. In this case permissions granted to a role apply to all VMs that have that role
+   applied.
+2. Permissions are managed per datastore. This means that VMs can have different privileges on
+   different datastores. For instance, on one datastore a VM may be allowed to create volumes of a
+   given size, while on another datastore the VM cannot create volumes at all. Note that in order to
+   provide out of the box usability and experimentation, all datastores allow all operations with no
+   restrictions. Permissions must be added per VM per Datastore by an administrator as necessary.
+3. Since there is no useful centralized storage outside of VC for storing permissions, and we don't
+   want to rely on VC, VM permissions for a given datastore shall reside inside the datastore. There
+   must be a way to manage concurrency control and eliminate data races, so using a plain old file
+   is probably not the solution. A preliminary idea is to store this data in a database such as
+   Postgres or SQlite which already provides these capabilities. The downside of this is that now,
+   we need to manage (possibly many) DB instances as part of our solution.
+4. The list below shows available privileges per VM and any further restrictions allowed for each one.
+
+  * Create a volume on a given datastore
+    * Maximum size of volume
+    * Maximum numbers of volumes
+  * Remove a volume on a given datastore
+    * Remove any volume
+    * Remove only volumes created by this VM
+  * Mount a volume on a given datastore
+    * Mount any volume
+    * Mount only volumes created by this VM
+
+Additionally we can constrain each datastore with:
+  * A total size for all volumes created
+  * A total number of volumes created
+
+### UI
+
+In order to quickly implement a usable permissions system, we will re-use existing tools. The
+current way that administrators manage volumes is via the admin CLI on each ESX host. This CLI will
+be enhanced with permissions management and will be able to grant or revoke permissions for any VM
+and datastore visible to it.
+
+In a later stage, in order to simplify centralized management, an API providing the same
+functionality as the CLI will be implemented for each host agent. This api can then be used to
+create a VC plugin that allows one click provisioning of Swarm and Kubernetes clusters with proper
+permissions configured.


doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+  1. How are VMs identified in order to grant permissions?
+  2. What is the granularity of a permission?
+  3. Where are permissions stored?
+  4. What are the actual permissions?
+
+In order to answer these questions, we must consider 2 scenarios. The first scenarios is permissions
+for a single VM with a local datastore, and the second is multiple VMs with a shared datastore. The
+model must work for both scenarios in order to be useful.
+
+1. VMs can share names so must be identified by unique IDs. These IDs should be unique across a
```
pdhamdhere: Are you proposing to introduce new UUID or use VM-UUID to apply permission. I am not sure if there is easy way for Admin to find UUID of a selected VM.

Admin CLI examples explaining how role & vm ids going to work would be nice.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+   cluster of hosts. If desired, we can associate many IDs with a Role that can be used to grant
+   permissions. In this case permissions granted to a role apply to all VMs that have that role
+   applied.
+2. Permissions are managed per datastore. This means that VMs can have different privileges on
+   different datastores. For instance, on one datastore a VM may be allowed to create volumes of a
+   given size, while on another datastore the VM cannot create volumes at all. Note that in order to
+   provide out of the box usability and experimentation, all datastores allow all operations with no
+   restrictions. Permissions must be added per VM per Datastore by an administrator as necessary.
+3. Since there is no useful centralized storage outside of VC for storing permissions, and we don't
```
pdhamdhere: I agree depending on VC doesn't seem wise. docker volume operations shouldn't depend on VC always being up & available. Secondly, permissions are best enforced where action happens i.e. on ESX host.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+   want to rely on VC, VM permissions for a given datastore shall reside inside the datastore. There
+   must be a way to manage concurrency control and eliminate data races, so using a plain old file
+   is probably not the solution. A preliminary idea is to store this data in a database such as
+   Postgres or SQlite which already provides these capabilities. The downside of this is that now,
+   we need to manage (possibly many) DB instances as part of our solution.
```
pdhamdhere: Hmm DB seems overkill. We do have other use cases coming up which can leverage DB so may be fine. I need to think little hard. If you decide to go DB route, then I would appreciate more details like is this DB per DS or ESX, schema etc

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+    * Maximum numbers of volumes
+  * Remove a volume on a given datastore
+    * Remove any volume
+    * Remove only volumes created by this VM
+  * Mount a volume on a given datastore
+    * Mount any volume
+    * Mount only volumes created by this VM
+
+Additionally we can constrain each datastore with:
+  * A total size for all volumes created
```
pdhamdhere: +1 on total size for all volumes created.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+  * A total number of volumes created
+
+### UI
+
+In order to quickly implement a usable permissions system, we will re-use existing tools. The
+current way that administrators manage volumes is via the admin CLI on each ESX host. This CLI will
+be enhanced with permissions management and will be able to grant or revoke permissions for any VM
+and datastore visible to it.
```
pdhamdhere: Is this acceptable limitation that admin-cli can only operate on VM and datastore accessible to ESX. Also does VM have to be registered on ESX host where I run admin CLI? 

Should Admin have a mechanism to blacklist datastores entirely?

doc/volume-permissions-spec.md
msterin: As another note, VMs already have both local-uuid (shown in BIOS to the guest) and VC UUID, so infrastructure for using uuid is in place. But I do not think Admins need to deal with UUID if we can avoid it.

doc/volume-permissions-spec.md
msterin: python sqlite API is trivial to use and is not much harder that regular files. Given the amount of data there is low , maintenance would be a no-issue.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+   Postgres or SQlite which already provides these capabilities. The downside of this is that now,
+   we need to manage (possibly many) DB instances as part of our solution.
+4. The list below shows available privileges per VM and any further restrictions allowed for each one.
+
+  * Create a volume on a given datastore
+    * Maximum size of volume
+    * Maximum numbers of volumes
+  * Remove a volume on a given datastore
+    * Remove any volume
+    * Remove only volumes created by this VM
```
msterin: How do you think Admins would use control over "volumes created by this VM" ?

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+
+In order to quickly implement a usable permissions system, we will re-use existing tools. The
+current way that administrators manage volumes is via the admin CLI on each ESX host. This CLI will
+be enhanced with permissions management and will be able to grant or revoke permissions for any VM
+and datastore visible to it.
+
+In a later stage, in order to simplify centralized management, an API providing the same
+functionality as the CLI will be implemented for each host agent. This api can then be used to
+create a VC plugin that allows one click provisioning of Swarm and Kubernetes clusters with proper
+permissions configured.
```
msterin: I second @pdhamdhere here  - a set of CLI commands is needed to complete the write up. 
Also: we need to discuss defaults - is it "ALLOW ALL" by default or is it "DENY ALL"  for create/remove/mount

msterin:
Andrew - good write up ! My comments are inline. 2 major points: (1) need CLI examples (2) need to define and agree on defaults

As for persistence, I think sqlite will work great for us

doc/volume-permissions-spec.md
andrewjstone: I was thinking of using the the existing UUIDs. I knew there were 2 of them, but didn't know what the differences were exactly. Thanks @msterin for the tips!  Hopefully the admin won't have to touch these at all. They just need to be maintained in the per-datastore DBs. The only reason I can think of for the admin to use them is if VMs are named the same and the UUIDs are used for disambiguation.

doc/volume-permissions-spec.md
andrewjstone: The sole rationale for using a DB is concurrency control. Note that that is a very good reason. By implementing our own concurrency control we'd likely end up with complex bit of code that is hard to verify and debug. We'd also need support for schema upgrades etc... All this is already done in a widely used / production ready SQlite. I was hesitant against using a DB myself, but SQlite is very lightweight and battle hardened. I'm firmly in the sqlite camp at this point. Hell it comes bundled with python even.

Also note that sqlite databases are a single portable file. So it's basically just using their concurrency code and format for the file instead of our own ad-hoc format or worse (JSON)

andrewjstone: One thing to take note of is using sqlite on NFS stores. Apparently posix advisory file locking is broken on NFS. It appears that there is a workaround using dotfiles if needed. The only other weird thing was that posix advisory locking behaves stupidly with threads. Our python code is single threaded though, so no issue there.

This is the only major issue I could find to be aware of looking here: https://www.sqlite.org/howtocorrupt.html

doc/volume-permissions-spec.md
andrewjstone: > Is this acceptable limitation that admin-cli can only operate on VM and datastore accessible to ESX.

Whether it's acceptable or not, I don't see how we can get around it without implementing our own host clustering or using VC. 

> Also does VM have to be registered on ESX host where I run admin CLI?

I'm not sure what `registered` means here.

> Should Admin have a mechanism to blacklist datastores entirely?

Yes! Absolutely.

doc/volume-permissions-spec.md
andrewjstone: In multi-tenancy scenarios, admins may want to isolate volumes completely from one VM to another. They may actually not even want these volumes visible. Of course that would require the vmdk_ops server to be able to respond appropriately to `docker volume ls` and `docker volume inspect` commands from different VMs.

andrewjstone:
I agree with the needs for commands. Let's discuss a bit in our meeting tomorrow and I will add them in for review after that.

doc/volume-permissions-spec.md
govint: Can we consider classifying storage into categories or classes (based on capabilities defined commonly) and not expose storage as datastores. And allow users a quota or entitlement on that storage. The idea of needing users to "know" what datastores are available at the backend (that too without knowing their capability, we aren't advertising anything about the storage itself) and then deciding where to place their volumes makes it "non-SDDC-like" (:-) ).  Users may crowd on the same datastore giving the system no way to optimize placement.

Instead, the user should be able to query what classes/categories of storage are available in terms of storage capabilities (TBD) and then acquire an entitlement on those (as they need or say choose to pay for). Then the user gets to create volumes against that entitlement. And the system has the flexibility to "decide" where to locate volumes for that user based on their entitlements.

Its possible that multiple datastores available on a host are all the same category, and the system would consolidate like storage into pools or containers and then allocations are made on that pool - across datastores in the pool).

So we have abstractions like,

a. Storage pools
b. Pool elements - datastores (disks)
c. Pool properties  - essentially the capabilities of the pool's member datastores.
d. Pool stats.
e. Pools ops - add/remove datastores (disks).

Thats a more user friendly approach and abstracts the backend completely than exposing backend datastores.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+the developer. This is a worrying proposition to many IT admins, and since they are the ones that
+must allow deployment of the driver and it's associated host agent, we must allow them to regain
+control with proper enforcement mechanisms that limit damage due to malfeasance or mistake.
+
+### Permissions Model
+
+Since docker containers run inside VMs, and can have root permissions, we cannot rely on the VM user
+for access control. Permissions must be granted to a VM, since VMs are created and managed
+by the IT administrator. This brings up a few issues:
+
```
govint: Suggest not linking up with VMs, instead allow users to be created and managed for storage entitlement. Each user is given a unique ID for the cluster (for example) and they can acquire storage capacity entitlements. Users provide their ID when allocating storage and the system manages the allocations per user.

The VM is purely a vehicle for running the user's containers and a way of plumbing backend VSphere storage to third party user containers.

The user can be running their containers on any number of VMs on any hosts in a cluster but they have a centrally managed entitlement to storage accessible via the cluster (see below on an alternate way of exposing storage to users abstracing out the storage architecture and instead positioning storage as classes based on capabilities).

doc/volume-permissions-spec.md
govint: But if the datastore is lost or corrupt then the database too will be lost. We must have a separate storage thats used to keep storage entitlements.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+   given size, while on another datastore the VM cannot create volumes at all. Note that in order to
+   provide out of the box usability and experimentation, all datastores allow all operations with no
+   restrictions. Permissions must be added per VM per Datastore by an administrator as necessary.
+3. Since there is no useful centralized storage outside of VC for storing permissions, and we don't
+   want to rely on VC, VM permissions for a given datastore shall reside inside the datastore. There
+   must be a way to manage concurrency control and eliminate data races, so using a plain old file
+   is probably not the solution. A preliminary idea is to store this data in a database such as
+   Postgres or SQlite which already provides these capabilities. The downside of this is that now,
+   we need to manage (possibly many) DB instances as part of our solution.
+4. The list below shows available privileges per VM and any further restrictions allowed for each one.
```
govint: This should be per user vs. per VM.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+
+  * Create a volume on a given datastore
+    * Maximum size of volume
+    * Maximum numbers of volumes
+  * Remove a volume on a given datastore
+    * Remove any volume
```
govint: Is there a use case where one user wants to remove another's volumes.

doc/volume-permissions-spec.md
```diff
@@ -0,0 +1,67 @@
...
+    * Remove only volumes created by this VM
+  * Mount a volume on a given datastore
+    * Mount any volume
+    * Mount only volumes created by this VM
+
+Additionally we can constrain each datastore with:
+  * A total size for all volumes created
+  * A total number of volumes created
+
```
govint: This may not have a use case, but the total size for all volumes is exactly the storage entitlement per user.

doc/volume-permissions-spec.md
govint: Does it help to use a central DB where all hosts in a cluster can update user entitlements - from any host.

andrewjstone:
Closing in light of using VC and a new proposal under discussion.

PR #517, opened 2016-07-08
Gvt changes in preperation for changes to Mount device only after OS has discovered it.

kerneltime:
The following changes are part of this PR.

1. Add new make gvt target that drops user into a container in which the user can add vendor sw. Documentation for gvt updated.
2. Many trailing white spaces removed.

Testing:

Each run command part of CI will run this code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing 
+# Contributing
 
 ## Topics
 * [Code contribution guidelines](#code-contribution-guidelines)
@@ -14,19 +14,19 @@ Read the [FAQ on the Wiki](https://github.com/vmware/docker-volume-vsphere/wiki#
 * Each PR must be accompanied with unit/integration tests
 * Add detailed description to pull request including reference to issues.
 * Add details of tests in "Testing Done".
-* Locally run integration tests. 
+* Locally run integration tests.
 * Squash your commits before you publish pull request.
 * If there are any documentation changes then they must be in the same PR.
 * We don't have formal coding conventions at this point.
 Make sure your code follows same style and convention as existing code.
 
-See  [Typical Developer Workflow](#typical-developer-workflow) to get started. 
+See  [Typical Developer Workflow](#typical-developer-workflow) to get started.
 
 
 ### Merge Approvals:
 * Pull request requires a minimum of 2 approvals, given via "Ship it",
 "LGTM" or "+1" comments.
-* Author is responsible for resolving conflicts, if any, and merging pull request. 
+* Author is responsible for resolving conflicts, if any, and merging pull request.
 * After merge, you must ensure integration tests pass successfully.
 Failure to pass test would result in reverting a change.
 
@@ -51,7 +51,7 @@ Expected Result:
 Actual Result:
 
 Triage:
-Copy-paste relevant snippets from log file, code. Use github markdown language for basic formatting. 
+Copy-paste relevant snippets from log file, code. Use github markdown language for basic formatting.
 
 ```
 ## Typical Developer Workflow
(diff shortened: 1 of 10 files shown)


vmdk_plugin/plugin.go
```diff
@@ -117,19 +116,53 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	}
+
+	watcher, err := inotify.NewWatcher()
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint},
+		).Error("Failed to create watcher ")
+		return mountpoint, err
+	}
+	err = watcher.Watch("/dev/disk/by-path")
```
govint: Is it possible to watch the specific device (do the watch on the device thats to be mounted. Do it on the device returned by get_device_path(). Else will the watcher return if some device gets created, not necessarily the one for the attach below.

govint:
Could add a test where multiple (say 10 - 20) containers are started in parallel and each of those are able to mount their volumes successfully.

vmdk_plugin/plugin.go
kerneltime: The select loop filters out devices we do not care about.

vmdk_plugin/plugin.go
```diff
@@ -117,19 +116,53 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	select {
+	case ev := <-watcher.Event:
+		log.Debug("event: ", ev)
+		if ev.Name == device {
+			// Log when the device is discovered
+			log.WithFields(
+				log.Fields{"name": name, "event": ev},
+			).Info("Attach complete ")
+			break
+		}
```
msterin: timeout is needed here

vmdk_plugin/plugin.go
```diff
@@ -117,19 +116,53 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	if d.useMockEsx {
+		return mountpoint, fmt.Errorf("No device to mount.")
+	}
+
+	watcher, err := inotify.NewWatcher()
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint},
+		).Error("Failed to create watcher ")
+		return mountpoint, err
```
pdhamdhere: Watcher errors should not result in mountVolume error. We should log & proceed to check if disk is mounted anyway.

vmdk_plugin/plugin.go
```diff
@@ -117,19 +116,53 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	}
+	err = watcher.Watch("/dev/disk/by-path")
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint},
+		).Error("Failed to watch /dev ")
+		return mountpoint, err
```
pdhamdhere: Same here. Let's log & proceed.

vmdk_plugin/plugin.go
pdhamdhere: +1 on timeout

vmdk_plugin/plugin.go
```diff
@@ -117,19 +116,53 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+			// Log when the device is discovered
+			log.WithFields(
+				log.Fields{"name": name, "event": ev},
+			).Info("Attach complete ")
+			break
+		}
+	case err := <-watcher.Error:
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint, "error": err},
+		).Error("Hit error during watch")
```
pdhamdhere: Same here. We should ignore any errors reported by watcher and proceed optimistically.

kerneltime:
Reducing this PR to be just gvt.

msterin:
LGTM , just please update the PR description

PR #520, opened 2016-07-09
Mount only after the OS is done discovering new dev.

kerneltime:
This is a continuation of https://github.com/vmware/docker-volume-vsphere/pull/517
There is a race between the plugin completing the attach command and
it requesting the OS to mount before the OS discovers the new device.
Add support to be notified when the device is attached.

Testing:
On my local box, the mount was triggered only after the device was
detected.
Tested timeout by watching on an incorrect path.
Tested looping by watching /dev
Each run command part of CI will run this code.

### DIFF
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -26,18 +26,21 @@ package main
 
 import (
 	"fmt"
-	"path/filepath"
-	"sync"
-	"time"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/go-plugins-helpers/volume"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/fs"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops"
+	"golang.org/x/exp/inotify"
+	"path/filepath"
+	"sync"
+	"time"
 )
 
 const (
-	mountRoot = "/mnt/vmdk" // VMDK block devices are mounted here
+	devWaitTimeout   = 1 * time.Second
+	mountRoot        = "/mnt/vmdk" // VMDK block devices are mounted here
+	sleepBeforeMount = 1 * time.Second
+	watchPath        = "/dev/disk/by-path"
 )
 
 type vmdkDriver struct {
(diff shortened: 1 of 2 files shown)


vmdk_plugin/plugin.go
```diff
@@ -117,19 +120,70 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	}
+
+	skipInotify := false
+
+	watcher, err := inotify.NewWatcher()
+
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint},
+		).Error("Failed to create watcher ")
```
pdhamdhere: nit: Failed to create watcher, skipping INotify

vmdk_plugin/plugin.go
```diff
@@ -117,19 +120,70 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+		skipInotify = true
+	} else {
+		err = watcher.Watch(watchPath)
+		if err != nil {
+			log.WithFields(
+				log.Fields{"name": name, "dir": mountpoint},
+			).Error("Failed to watch /dev ")
+			skipInotify = true
```
pdhamdhere: Same here. Add skipping INotify

vmdk_plugin/plugin.go
kerneltime: will do.

vmdk_plugin/plugin.go
```diff
@@ -117,19 +120,70 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
 	// Have ESX attach the disk
 	dev, err := d.ops.Attach(name, nil)
 	if err != nil {
 		return mountpoint, err
 	}
 
-	time.Sleep(3 * time.Second)
+	device, err := fs.GetDevicePath(dev)
+	if err != nil {
+		return mountpoint, err
```
pdhamdhere: Log error?

vmdk_plugin/plugin.go
```diff
@@ -117,19 +120,70 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+	}
 
-	if d.useMockEsx {
-		return mountpoint, fs.Mount(mountpoint, nil, "ext4")
+	if skipInotify {
+		time.Sleep(sleepBeforeMount * time.Second)
+		return mountpoint, fs.Mount(mountpoint, "ext2", device)
```
pdhamdhere: Hopefully we won't reach here often but code forces sleep though device may be already mounted

kerneltime: yes.. cause we cannot depend on inotify, it is better than failing a valid mount.

vmdk_plugin/plugin.go
kerneltime: No, these are forwarding errors generated some place else and the caller will log it.

pdhamdhere:
LGTM some minor nits above. Do get another review from Go expert.
Minor suggestion to test trigger ignorable events to exercise ignore event path.

kerneltime:
Currently all events in the path are triggered into the select, I tested the timeout by watching /dev that sent vsocket events but no volume got mounted and the code timed out.. I was looking into a good cheap way to mock this but looks like the code will need more elaborate interfaces that we then mock.

vmdk_plugin/plugin.go
```diff
@@ -26,18 +26,21 @@ package main
...
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops"
+	"golang.org/x/exp/inotify"
+	"path/filepath"
+	"sync"
+	"time"
 )
 
 const (
-	mountRoot = "/mnt/vmdk" // VMDK block devices are mounted here
+	devWaitTimeout   = 1
```
msterin: One of what :-) ? Pls add units to either bar name devNotifyTimeoutSec or just comments

vmdk_plugin/plugin.go
```diff
@@ -26,18 +26,21 @@ package main
...
+	mountRoot        = "/mnt/vmdk" // VMDK block devices are mounted here
+	sleepBeforeMount = 1
```
msterin: Same comment about units

msterin:
LGTM with two minor items about time  units

vmdk_plugin/plugin.go
kerneltime: :-)

kerneltime:
Will merge this in once CI passes.

vmdk_plugin/plugin.go
```diff
@@ -117,19 +120,70 @@ func (d *vmdkDriver) mountVolume(name string) (string, error) {
...
+			}
+		case err := <-watcher.Error:
+			log.WithFields(
+				log.Fields{"name": name, "device": device, "error": err},
+			).Error("Hit error during watch ")
+			break loop
+		case <-time.After(devWaitTimeout):
+			log.WithFields(
+				log.Fields{"name": name, "timeout": devWaitTimeout, "device": device},
+			).Error("Reached timeout while waiting for device ")
```
govint: What if the device is attached now by the OS, the device is attached and does docker detach if there is an error?

Should have a test to force both the error and this timeout path (return error from here even though the disk is attached) and figure that the disk isn't stuck to the VM and unusable.

kerneltime:
Docker is supposed to send unmount for each mount including the failed ones. It is not related to this code change.

PR #522, opened 2016-07-09
[Testing do not merge] Runci/kerneltime

kerneltime:


govint:
Finally, finally some one other than me is bitten by the same bug - commits outside of my change get included in the PR ;-))

kerneltime:
The reason here is that I am doing a pull request across random branches.. It is working as expected..

PR #525, opened 2016-07-11
[Do not merge] Trigger ci tests only on certain conditions.

kerneltime:
**Ran into a bug with how github is linking to the CI runs triggered for PRs and Pushes.. have pinged CNA and Drone folks**

Developer branches should not trigger tests.
If developer wants to trigger a test run, prefix branch name with runci/ Example: runci/kerneltime 
Pull Requests should always trigger tests

This will result in quicker runs when an open PR is updated with review comments. Currently the push into the branch as well as the open PR results in the tests to be run twice serially.

Testing:

triggermehappy.kerneltime branch is not prefixed and the CI run did not run tests for push https://ci.vmware.run/vmware/docker-volume-vsphere/935

runci/kerneltime is prefixed and the same commit when pushed into runci/kerneltime caused all tests to run https://ci.vmware.run/vmware/docker-volume-vsphere/937

This PR without a prefix branch lead to all tests being run.https://ci.vmware.run/vmware/docker-volume-vsphere/936


kerneltime:
saw a bug with PR linking to CI runs vmware/docker-volume-vsphere#525 triggered https://ci.vmware.run/vmware/docker-volume-vsphere/936 but github reports on https://ci.vmware.run/vmware/docker-volume-vsphere/937 which is the same commit on a different branch pushed to github
Sequence: 1. push a branch, 2 start a PR, 3 checkout different branch hard reset to the branch used for PR. 3 push new branch to github.. the PR for the original branch now links to the CI run for the new branch..
From drone channel
without knowing the full details, the GitHub status API accepts the following input POST /repos/:owner/:repo/statuses/:sha. If the same status is being shared it is due to multiple pull requests or branches having the same sha

PR #526, opened 2016-07-11
Trigger ci tests only on certain conditions.

kerneltime:
* Developer branches should not trigger tests.
* If developer wants to trigger a test run, prefix branch name with runci/* Example: runci/reducewaste.kerneltime
* Pull Requests should always trigger builds

Testing:

reducewaste.kerneltime branch is not prefixed and the CI run did not run tests for push https://ci.vmware.run/vmware/docker-volume-vsphere/938

runci/reducewaste.kerneltime is prefixed and the same commit when pushed into runci/kerneltime caused all tests to run https://ci.vmware.run/vmware/docker-volume-vsphere/939

This PR without a prefix branch lead to all tests being run. https://ci.vmware.run/vmware/docker-volume-vsphere/942


.drone.yml
```diff
@@ -38,7 +37,7 @@ build:
     when:
       event: tag
 
-  test:
+  test_dev_branch: # Identicle to test except for when it occurs
```
msterin: did you mean 'Identical' ?

kerneltime: Yes. Thank you :-)

msterin:
looks good to my untrained eye - I am afraid we do not have many experts in Drone so as long as the Ci runs for PRs pushes once, does not run for regular branches and does run for runci/branch , it's all good to go. 
Minor comment about a comment inside.;

kerneltime:
@msterin I can ping the contributors of drone or look in their code if you have a specific question you would like answered for drone ;-) For now the new behavior is that every thing is the same for PRs and master but dev branches do not get tests unless requested by naming scheme (runci/*). Drone supports skipping CI runs via commit messages but not the other way around... so naming scheme in branches is a good bet. Also, we can run all tests locally any how.

PR #528, opened 2016-07-11
Use go lang container to run gvt for vendoring.

kerneltime:
The centos image used for building the plugin has an older version
of git that does not work for the latest version of gvt.
Switch to using the golang container for running gvt.

Test:
make gvt
gvt update github.com/docker/go-plugins-helpers/volume (used to fail with cent os 6.6)


govint:
Ship it!

msterin:
lgtm

PR #530, opened 2016-07-12
Support for Docker 1.12 plugin capabilities.

govint:
1. Updated server to support a capabilities request, the server returns global scope if at least one datastore exists thats !local to the host. Capabilities in Docker 1.12 is to get the scope of volumes created by the plugin. We don't want to indicate a global scope if there aren't any !local datastores attached to the host. Return local scope if all datastores are local-only to the host. Scope is not per volume though its shown as such and even if that particular volume is on a local datastore (weird docker behavior).

2. Modified the client to add a Capabilities() method.

3. Modified mock-esx to support the capabilities request - returns local always.

Please ignore the go-plugin-helper files that are part of this PR, those will be removed once #528 is merged.

Testing done:
1. Ran create/inspect on Docker 1.12 - both mock and regular plugin modes.

mock-esx:
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume inspect clonevol_217
[
    {
        "Name": "clonevol_217",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/clonevol_217",
        "Labels": {},
        "Scope": "local"
    }
]
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume inspect clonevol_216
[
    {
        "Name": "clonevol_216",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/clonevol_216",
        "Labels": {},
        "Scope": "local"
    }
]

Regular mode of plugin:
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume inspect clonevol_212
[
    {
        "Name": "clonevol_212",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/clonevol_212",
        "Labels": {},
        "Scope": "global"
    }
]
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ =docker volume create --driver=vmdk --name=clonevol_213
No command '=docker' found, did you mean:
 Command 'docker' from package 'docker' (universe)
 Command 'kdocker' from package 'kdocker' (universe)
=docker: command not found
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ =docker volume create --driver=vmdk --name=clonevol_213^C
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume create --driver=vmdk --name=clonevol_213
clonevol_213
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume inspect clonevol_213
[
    {
        "Name": "clonevol_213",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/clonevol_213",
        "Labels": {},
        "Scope": "global"
    }
]

Docker 1.11 - there is no scope
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume create --driver=vmdk --name=clonevol_220
clonevol_220
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker volume inspect clonevol_220
[
    {
        "Name": "clonevol_220",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/clonevol_220",
        "Labels": {}
    }
]
administrator@hte-1s-eng-dhcp98:~/vpl-vol-cap/docker-volume-vsphere$ docker --version
Docker version 1.11.1, build 5604cbe


govint:
CI fails as the 6.0u2 Vsphere vmodl seems to be different and ne change to get scope of datastore is failing.

esx_service/utils/vmdk_utils.py
```diff
@@ -44,18 +46,23 @@ def get_datastores():
...
-    datastores = [(d.info.name,
+    datastores = [[d.info.name,
                    os.path.split(d.info.url)[1],
-                   os.path.join(d.info.url, 'dockvols'))
+                   os.path.join(d.info.url, 'dockvols'), True]
                   for d in ds_objects]
+    for d in ds_objects:
+       ds = [x for x in datastores if x[0] == d.info.name]
+       if d.summary.type == 'VMFS':
+          ds[0][3] = d.info.vmfs.local
```
msterin: I am not sure why "VMFS "indicates "local". What if it's just an array LUN exposed via iSCSI and formatted as VMFS (?

esx_service/vmdk_ops.py
```diff
@@ -365,6 +366,15 @@ def listVMDK(vm_datastore):
              u'Attributes': {}} \
             for x in vmdks]
 
+def getScope():
+    """
+    Returns global scope if there is at least one datastore thats not local.
+    """
+    global_ds = [x[0] for x in vmdk_utils.get_datastores() if x[3]]
+    if len(global_ds):
```
msterin: is not it simply 
```python
   if global_ds:
     return 'global'
   return 'local' 
``` ?
BTW please define global consts for the const strings.

File: vendor/github.com/docker/go-plugins-helpers/volume/api.go [hunk 3]
```diff
@@ -10,28 +10,31 @@ const (
...
+	MountID string            `json:",omitempty"`
 }
 
 // Response is the strucutre that the plugin's responses are serialized to.
 type Response struct {
-	Mountpoint string
-	Err        string
-	Volumes    []*Volume
-	Volume     *Volume
+	Mountpoint   string
```
msterin: what's the change here ? is it TAB->space ? github does not show much....

vendor/github.com/docker/go-plugins-helpers/volume/api.go
```diff
@@ -40,6 +43,11 @@ type Volume struct {
 	Mountpoint string
 }
 
+// Capability represents the list of capabilities a volume driver can return
+type Capability struct {
+	Scope string
```
msterin: nit: please add comment with choices.

msterin:
Thanks for putting it together !

overall the change looks clean, with the following comments: 
[ ] VMFS could be shared - please see inline comments for correct way to find out the "share"-ness
[ ] int the future, please separate whitespace/name fixing in a dedicated PR. It makes it much harder to focus on reviewing the actual changes when name/whitespaces are around

esx_service/utils/vmdk_utils.py
govint: The VMODL has a boolean for VMFS datastores called "local" which is
true/false depending on whether the device is local or not.

govint:
Sure, an earlier change had used "format" as a variable name while format
is an internal name in Python, hence renamed that.

esx_service/utils/vmdk_utils.py
```diff
@@ -29,11 +29,13 @@
 
 def get_datastores():
     """
-    Returns a list of (name, url-name, dockvol_path), with an element per datastore
+    Returns a list of (name, url-name, dockvol_path, boolean), with an
```
pdhamdhere: nit: boolean => isLocal?

pdhamdhere:
> server returns global scope if at least one datastore exists thats !local to the host. Capabilities in Docker 1.12 is to get the scope of volumes created by the plugin. We don't want to indicate a global scope if there aren't any !local datastores attached to the host. 

I am not sure if this is correct interpretation of "scope" capability introduced in Docker 1.12. First, capability is property of a Volume Plugin and not specific to Volume. What does it mean for same volume plugin to be local or global in Swarm/Cluster environment.

Second, determining local/global based on Datastore accessibility to ESX is incorrect. For e.g. multiple docker hosts (VMs) running on local datastore within same swarm cluster can access volumes created by other Docker Host. So, capability in this scenarios is "global".

I think we should always be advertising Docker Volume Plugin with "global" scope. Logic for determining whether DS is shared or local is redundant.

msterin:
@pdhamdhere - good point ! I was focusing on method of getting it and missed the concept mismatch :-)

govint:
I did that first but then what happens if a host has all local datastores?

pdhamdhere:
>I did that first but then what happens if a host has all local datastores?
Doesn't matter since there may be more than one VM on same local datastore. These VMs aka Docker Hosts can share docker volumes on local datastore.

govint:
This feature is used by swarm and does swarm work only inside a node. Then
yes just report global. But docker definition of scope is not clear

govint:
Closing this PR.

Issue #501, opened 2016-06-30
ESX API need idempotency or a well defined retry framework


kerneltime:
vSocket is not perfect and we can see errors on vSocker based on kernel version. Thus, requests need retrying, which in turn needs idempotent APIs or well defined behavior for plugins to behave as per their API.

pdhamdhere:
@kerneltime is this still an issue? Close?

pdhamdhere:
#1076 has more context and is tracking same issue. Closing as dupe.

PR #532, opened 2016-07-12
Use different privileged ports to bind on client side.

kerneltime:
Using the same port for all requests on the client side leads to
errors from the server if the clean up has not completed.
Use different privileged ports on the client side.
We still need sleep and retry logic in the plugin in case the client
port is busy but since this is not expected to occur in most cases
the log level has been elevated to Error.

Testing:
1. Reproduce the issue on local box
2. Add the fix and retest

No VMCI errors are seen once this fix was introduced.

### DIFF
--- a/esx_service/vmci/vmci_client.c
+++ b/esx_service/vmci/vmci_client.c
@@ -26,6 +26,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <stdint.h>
+#include <assert.h>
 
 #include "vmci_sockets.h"
 #include "connection_types.h"
@@ -51,6 +52,11 @@ typedef struct be_request {
 } be_request;
 
 #define MAXBUF 1024 * 1024 // Safety limit. We do not expect json string > 1M
+#define MAX_CLIENT_PORT 1023 // Last privileged port
+#define START_CLIENT_PORT 1000 // Where to start client port
+
+// Retry entire range on bind failures
+#define BIND_RETRY_COUNT (MAX_CLIENT_PORT - START_CLIENT_PORT)
 
 typedef struct be_answer {
    char *buf;   // calloced, so needs to be free()
@@ -151,6 +157,7 @@ dummy_get_reply(be_sock_id *id, be_request *r, be_answer* a)
    return CONN_SUCCESS;
 }
 
+
 // vsocket interface implementation
 //---------------------------------
 
@@ -161,9 +168,10 @@ dummy_get_reply(be_sock_id *id, be_request *r, be_answer* a)
 static be_sock_status
 vsock_init(be_sock_id *id, int cid, int port)
 {
+   static int round_robin = START_CLIENT_PORT; // Round robin client bind port
    int ret;
    int af;    // family id
-   int sock;     // socket id
+   int sock;  // socket id
 
    if ((af = vsock_get_family()) == -1) {
       return CONN_FAILURE;
@@ -174,16 +182,31 @@ vsock_init(be_sock_id *id, int cid, int port)
    }
 
    id->sock_id = sock;
-
    memset(&id->addr, 0, sizeof id->addr);
    id->addr.svm_family = af;
    id->addr.svm_cid = VMCISock_GetLocalCID();
-   id->addr.svm_port = port;
 
-   // Bind a port. If less than 1024 it insures the client is capable of
-   // binding a port lower than 1024 which is typically a root process or
-   // a process given capabilities by root.
-   ret = bind(sock, (const struct sockaddr *) &id->addr, sizeof id->addr);
+   int retryCount = 0;
+
+   while (retryCount++ < BIND_RETRY_COUNT) {
+      id->addr.svm_port = round_robin;
+      if (round_robin == MAX_CLIENT_PORT) {
+         round_robin = START_CLIENT_PORT;
+      } else {
+         round_robin++;
+      }
+
+      assert((round_robin >= START_CLIENT_PORT) && (round_robin <= MAX_CLIENT_PORT));
+
+      // Bind a port. If less than 1024 it insures the client is capable of
+      // binding a port lower than 1024 which is typically a root process or
+      // a process given capabilities by root.
+      ret = bind(sock, (const struct sockaddr *) &id->addr, sizeof id->addr);
+      if (ret == 0) {
+         break;
+      }
+   }
+
    if (ret != 0) {
       int old_errno = errno;
       vsock_release(id);
--- a/vmdk_plugin/vmdkops/esx_vmdkcmd.go
+++ b/vmdk_plugin/vmdkops/esx_vmdkcmd.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	log "github.com/Sirupsen/logrus"
 	"syscall"
-	"time"
 	"unsafe"
 )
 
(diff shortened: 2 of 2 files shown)


esx_service/vmci/vmci_client.c
```diff
@@ -51,6 +52,8 @@ typedef struct be_request {
 } be_request;
 
 #define MAXBUF 1024 * 1024 // Safety limit. We do not expect json string > 1M
+#define MAX_CLIENT_PORT 1023 // Last privileged port
+#define START_CLIENT_PORT 1000 // Where to start client port
```
msterin: out of curiosity - why 1000 ? 
A bigger question is "would the range of 23 be enough for the timing to be right" ?

esx_service/vmci/vmci_client.c
```diff
@@ -161,9 +165,10 @@ dummy_get_reply(be_sock_id *id, be_request *r, be_answer* a)
 static be_sock_status
 vsock_init(be_sock_id *id, int cid, int port)
 {
+   static int round_robin = START_CLIENT_PORT;
```
msterin: nit: maybe just name it client_port and add "// selected using round-robin" to the coments?

esx_service/vmci/vmci_client.c
```diff
@@ -161,9 +165,10 @@ dummy_get_reply(be_sock_id *id, be_request *r, be_answer* a)
 static be_sock_status
 vsock_init(be_sock_id *id, int cid, int port)
```
msterin: if the port is selected this way, should not we drop port flag and passing port to the code ?

esx_service/vmci/vmci_client.c
kerneltime: It is a brown range.. I could have started from 1 and work up to 1023 but felt that was excessive, also, if we run into the issue with a range of this size, maybe we need to investigate the VMCI performance with the VMCI folks. I can see why a range of 1 can lead to issues but a range of 23 should be large enough. Like I said a brown range...

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -99,11 +99,11 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 			errno = err.(syscall.Errno)
 			msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 			if retryCount >= maxRetryCount {
-				if errno == syscall.ECONNRESET {
+				if errno == syscall.ECONNRESET || errno == syscall.ETIMEDOUT {
 					msg += " Hit communication issue with ESX (vmci or ESX service)\n"
```
msterin: not a part of this fix but since you are touching the code here.., I think the message should be changed to simply " Failed to communicated with ESX. For details please refer to FAQ..."

esx_service/vmci/vmci_client.c
kerneltime: will do

msterin:
Looks good - but I think the "port" options should be dropped as it is not used anymore or at least proper comments should be made. Plus a few minor nits.

esx_service/vmci/vmci_client.c
kerneltime: That is the port for the ESX service which we need to be configurable and is based on config in service file. We initially used the same port on the client side for simplicity.

kerneltime: That is the port for the ESX service which is configured in the service file.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: Ok.

esx_service/vmci/vmci_client.c
pdhamdhere: Where is it used? I don't see "port" param used anywhere in function.

msterin: well, the code does change `id->addr.svm_port = port` to `id->addr.svm_port = round_robin`, so now I  not sure how it could have been the ESX  service port

kerneltime: line 208 https://github.com/vmware/docker-volume-vsphere/pull/532/files#diff-ccd9fe1931a983c9c566599fae404fcdR208

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -99,11 +99,11 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
 			errno = err.(syscall.Errno)
 			msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
 			if retryCount >= maxRetryCount {
```
pdhamdhere: I thought with this change, we no longer need retry logic and we don't have urgency to fix #501

kerneltime:
As I mentioned in the comments, port is still needed to tell the code which port to connect to with ESX and is configurable via service config. Nits have been addressed.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
kerneltime: True. With this fix, we are still exposed port being busy, will move the retry logic inside the attempt to bind and remove the higher level retry which needs #501 to be fixed.

esx_service/vmci/vmci_client.c
msterin: Ah, the old code was *reusing* the same port for outbound and inbound, with no related comments :-(. Yes, the change makes it better.

msterin:
LGTM

esx_service/vmci/vmci_client.c
```diff
@@ -161,9 +168,10 @@ dummy_get_reply(be_sock_id *id, be_request *r, be_answer* a)
...
 {
+   static int round_robin = START_CLIENT_PORT; // Round robin client bind port
```
pdhamdhere: I liked Mark's original suggestion to name it as "client_port ". Comment takes care of explaining port is selected in round robin fashion.

kerneltime: round_robin gets assigned to the client port.. it is not the client port..

vmdk_plugin/vmdkops/esx_vmdkcmd.go
```diff
@@ -86,36 +85,16 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-			} else {
-				log.Errorf(msg)
-				return nil, errors.New(msg)
-			}
+	_, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+	if err != nil {
+		var errno syscall.Errno
+		errno = err.(syscall.Errno)
+		msg := fmt.Sprintf("'%s' failed: %v (errno=%d).", cmd, err, int(errno))
+		if errno == syscall.ECONNRESET || errno == syscall.ETIMEDOUT {
```
pdhamdhere: Why special case these errors? There could be other valid errors like EADDRINUSE, EADDRNOTAVAIL etc

kerneltime: This is to redirect folks to the FAQ to explain the need to have tools installed. This log message was to deal with unpredictability of vm tools running on a VM

esx_service/vmci/vmci_client.c
pdhamdhere: Ok. makes sense.

vmdk_plugin/vmdkops/esx_vmdkcmd.go
pdhamdhere: I see.

pdhamdhere:
LGTM

PR #533, opened 2016-07-12
Fixed typos in comments

msterin:
Fixed a few typos found  as a result of running syntax checker in the new flashy VisualStudio Code.

Also order of imports in GO code has auto-changed to comply with recommended sorted order


pdhamdhere:
LGTM

PR #534, opened 2016-07-13
Fix how volumes are identified on the client for mount..

govint:
Modified approach to identifying device to mount:

a. The server sends the PCI slot number at which the PVSCSI controller is attached, to the client.

2. On Linux /sys/bus/pci/slot/<slot-num/address has the name of the device at that slot. we use this to create the name of the device to mount on the client.


kerneltime:
Are you planning to use the CI system for testing? Not sure what "not for review" means here.. is it WIP?

esx_service/vmdk_ops.py
```diff
@@ -569,9 +569,26 @@ def findDeviceByPath(vmdk_path, vm):
     return None
 
 
-def busInfo(unit_number, bus_number):
+def get_controller_pci_slot(vm, pvscsi, key_offset):
```
pdhamdhere: nit: Add a function header comment

esx_service/vmdk_ops.py
```diff
@@ -638,7 +655,23 @@ def disk_attach(vmdk_path, vm):
...
-    # Make sure we have a PVSCI and add it if we don't
+    # Check if this disk is already attached, and if it is - skip the disk
+    # attach and the checks on attaching a controller if needed.
+    device = findDeviceByPath(vmdk_path, vm)
+    if device:
+        # Disk is already attached.
+        logging.warning("Disk %s already attached. VM=%s",
+                        vmdk_path, vm.config.uuid)
+        setStatusAttached(vmdk_path, vm)
+        # Get that controller to which the device is configuired for
```
pdhamdhere: nit: configuired  => configured

vmdk_plugin/utils/fs/fs.go
```diff
@@ -86,47 +87,22 @@ func GetDevicePath(str []byte) (string, error) {
...
-
-	defer dirh.Close()
-
-	// Change this to do a read in slices, needed for large dirs.
-	names, err := dirh.Readdirnames(-1)
+	pciSlotAddr := fmt.Sprintf("%s/%s/address", sysPciSlots, volDev.ControllerPciSlotNumber)
 
+	fh, err := os.Open(pciSlotAddr)
 	if err != nil {
 		log.WithFields(log.Fields{"Error": err}).Warn("Get device by path: ")
```
pdhamdhere: Log "pciSlotAddr "?

pdhamdhere:
Thanks for doing this. LGTM subject to CI passing. Can you please update PR with "testing done" and enumerate which linux distros you have tested.

esx_service/vmdk_ops.py
```diff
@@ -569,9 +569,26 @@ def findDeviceByPath(vmdk_path, vm):
     return None
 
 
-def busInfo(unit_number, bus_number):
+def get_controller_pci_slot(vm, pvscsi, key_offset):
+    if pvscsi[0].slotInfo:
+       return(pvscsi[0].slotInfo.pciSlotNumber)
+    else:
+       # Slot number is got from from the VM config
+       key = 'scsi{0}.pciSlotNumber'.format(pvscsi[0].key -
```
andrewjstone: When would `pvscsi[0].slotInfo` not exist, but `pvscsi[0].key` would exist?

esx_service/vmdk_ops.py
```diff
@@ -569,9 +569,26 @@ def findDeviceByPath(vmdk_path, vm):
     return None
 
 
-def busInfo(unit_number, bus_number):
+def get_controller_pci_slot(vm, pvscsi, key_offset):
+    if pvscsi[0].slotInfo:
+       return(pvscsi[0].slotInfo.pciSlotNumber)
+    else:
+       # Slot number is got from from the VM config
```
andrewjstone: nit: got -> gotten | retrieved

esx_service/vmdk_ops.py
govint: This is what I discovered. VSphere 6.0 VMODL is different from the current
one and since we support both earlier and later VSphere releases need to
handle both cases. If the slotInfo is set then its the newer VMODL else the
old one.

andrewjstone: Cool!

andrewjstone:
LGTM

msterin:
I just skimmed the code without going deep and it looks good for the. However , please do add test matrix (Linux/ ESX versions) of  at you have  checked; LGTM contingent on this and CI passing

kerneltime:
Do we plan to expand the list of distros supported as part of this (then documentation and CI needs updating) else LGTM

esx_service/vmdk_ops.py
```diff
@@ -638,7 +655,23 @@ def disk_attach(vmdk_path, vm):
...
-    # Make sure we have a PVSCI and add it if we don't
+    # Check if this disk is already attached, and if it is - skip the disk
+    # attach and the checks on attaching a controller if needed.
+    device = findDeviceByPath(vmdk_path, vm)
+    if device:
+        # Disk is already attached.
+        logging.warning("Disk %s already attached. VM=%s",
+                        vmdk_path, vm.config.uuid)
+        setStatusAttached(vmdk_path, vm)
+        # Get that controller to which the device is configured for
```
msterin: nit: the comment is hard to read. Maybe just drop the "for" in the end?

msterin:
The fix looks good - please DO ADD some info on testing (automation may be better but here a description of what you did manually, with cut-n-paste of the  results, will be good enough)l LGTM after that

govint:
Testing matrix:

Vsphere versions - 6.0U3 and 6.5

Linux - Ubuntu 14.04.1 LTS and Photon

Centos and Redhat testing is still underway pending availability.

As discussed will merge these changes and continue to test.

PR #535, opened 2016-07-13
Reduce redundant logs and cleanup format

kerneltime:
This is to clean up just the drone part of the CI run.
This should allow identification of individual steps and reduction of redundant messages in CI runs.

Take a look at CI run here for end results.
Compare to: https://ci.vmware.run/vmware/docker-volume-vsphere/968


andrewjstone:
LGTM contingent on CI passing

kerneltime:
The CI is linking to the wrong run, it will fix once the real run ends.

msterin:
+1 (LGTM contingent on CI passing)

PR #538, opened 2016-07-15
Disable vsan info tests till VSAN ls is running slow. Show target runs.

kerneltime:


kerneltime:
Removing the tests should allow tests to pass but they still take a long time to run while the root of vsanDatastore is slow.

PR #543, opened 2016-07-18
Add single ubuntu vm for VSAN CI. Clean up makefile

kerneltime:
While the issue with Photon and Docker is being debugged, I deployed a Ubuntu VM to unblock VSAN testing (It is being used as VM1 and VM2). 
Also, some minor clean up.

Testing: The CI run is all that matters here.


vmdk_plugin/Makefile
```diff
@@ -88,6 +88,14 @@ VMDKOPS_MODULE_SRC = $(VMDKOPS_MODULE)/*.go $(VMCI_SRC)
 SRC = plugin.go main.go log_formatter.go refcnt.go \
 	utils/fs/fs.go utils/config/config.go
 
+# Canned recipe
+define log_target
+@echo
+@echo "=> Running target $@"
```
msterin: nit:  I think having "time : message" on one line is better - less clutter

msterin:
lgtm

andrewjstone:
LGTM

vmdk_plugin/Makefile
kerneltime: Time message makes it harder to grep if time is at the end, also reading wise looks cleaner (to my eyes), open to change but can do it separately.

PR #544, opened 2016-07-19
Add Photon OS on VSAN for CI

kerneltime:
Testing: CI Run should use all 4 VMs (2 VMFS, 2 VSAN, ubuntu and photon)


pdhamdhere:
Awesome. Thanks for doing this. LGTM!

PR #545, opened 2016-07-20
Add the demo scripts

kerneltime:
To see the demos head to https://github.com/vmware/docker-volume-vsphere/wiki/Demos

### DIFF
--- a/misc/demos/docker-volume-commands.yaml
+++ b/misc/demos/docker-volume-commands.yaml
@@ -1,0 +1,28 @@
+title: Run docker volume commands
+scenes:
+  - name: Docker volume commands
+    action: echo "Post install, use the plugin for docker volume operations"
+  - name: Create a volume
+    action:  docker volume create --driver=vmdk --name=MyVolume -o size=10gb
+    wait: true
+  - name: List volume
+    action: docker volume ls
+    wait: true
+  - name: Inspect volume
+    action: docker volume inspect MyVolume
+    wait: true
+  - name: Write a file
+    action: docker run --rm -v MyVolume:/vol busybox "cp /etc/hosts /vol/file"
+    wait: true
+  - name: Explain
+    action: echo "The /vol/file now exists in MyVolume and can be reused"
+    wait: true
+  - name: Read a file
+    action: docker run --rm -v MyVolume:/vol busybox "cat /vol/file"
+    wait: true
+  - name: Delete a volume
+    action: docker volume rm MyVolume
+    wait: true
+  - name: List volume
+    action: docker volume ls
+    wait: true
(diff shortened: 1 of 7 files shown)


kerneltime:
The ESX admin monitoring demo was done manually as the default width in the automated demos does not work with the admin CLI and currently i am battling the tools to figure out the automated solution. For now there is only a manual option for the admin CLI.

misc/demos/docker-volume-commands.yaml
```diff
@@ -0,0 +1,28 @@
+title: Run docker volume commands
+scenes:
+  - name: Docker volume commands
+    action: echo "Post install, use the plugin for docker volume operations"
+  - name: Create a volume
+    action:  docker volume create --driver=vmdk --name=MyVolume -o size=10gb
```
pdhamdhere: Formatting 10gb would be slow? Make it  1GB?

misc/demos/install-vm.yaml
```diff
@@ -0,0 +1,21 @@
...
+    action: echo "Download RPM or Deb from GitHub Releases. Using RPM on Photon here.."
+  - name: Curl rpm
+    action: curl -OL https://github.com/vmware/docker-volume-vsphere/releases/download/1.0.beta/docker-volume-vsphere-1.0.beta-1.x86_64.rpm
+    wait: true
+  - name: just a pause
+    pause: 1
+  - name: Copy RPM
+    action: scp docker-volume-vsphere-1.0.beta-1.x86_64.rpm root@$VM:/tmp
+    wait: true
+  - name: Install RPM
```
pdhamdhere: Add new lines between major commands?

pdhamdhere:
Nice. Like it! LGTM.

misc/demos/docker-volume-commands.yaml
kerneltime: The demo seems fine..

misc/demos/install-vm.yaml
kerneltime: Ok.

msterin:
Veeeery nice. 2 small comments - it would be nice to have '#' before comment lines in shell (when you give explanations/titles', and erasing characters and typos in scripts does not look polished. 
With these caveats, LGTM

kerneltime:
I should be able to re-write the esx admin cli demo in an automated manner (need to fix the cli first to report when columns are specified).
Also, should be able to get rid of "Echo" command, needs changes to spielbash.. 
@msterin if you see a typo please, pointing it out would be helpful, will rescan when I update the scripts.

kerneltime:
merging this in for now. There are larger changes for demos coming where I will address the not echo text and admin cli

PR #550, opened 2016-07-26
Update docker packages in the repo and fix API mismatches.

govint:
Updated docker packages in the repo as otherwise the mount ID interfaces and plumbing are missing. All docker packages are updated to ensure we have all docker-side changes in the repo. Some of the APIs are modified to accept a context parameter and those are fixed in the plugin client code as well - to match the new API definitions.

Tested out volume create/inspect/run/remove with docker 1.12 and 1.11 and both work fine with the new packages.


kerneltime:
Can you include the gvt steps followed?

msterin:
> Some of the APIs are modified to accept a context parameter and those are fixed in the plugin client code as well - to match the new API definitions.

Which APIs are modified ? It is not clear from description, and the change is too big to track it manually.

Otherwise LGMT , but please, do comment on which APIs are modified so we can review

govint:
GVT steps include:

1. make gvt
2. gvt update <docker packages> and go-plugin-helper packages

govint:
Mark,
   The APIs changes are handled in vmdk_plugin/refcnt.go - the client APIs accept a context argument now. and in vmdk_plugin/sanity_test.go.

msterin:
Thanks @govint 

LGTM

govint:
Thanks Mark, one more LGTM and I'll submit

pdhamdhere:
LGTM.

PR #567, opened 2016-08-02
Allow dash in volume names

msterin:
And generally relaxed names to have anything as long as there is no conflict with deltadisk names. 
Still prevent conflict with deltadisks by blocking
'-NNNNNN' at the end of the volume name. E.g. volume-mySuperHero-2 is a valid
volume name, but volume-Here-000012 is not.
Also removed abuse of regexps (where a simple rindex() is enough) and  tidied up volume/vmdk name parse code. 

Test: included.


andrewjstone:
LGTM. Those regexps are super clean now :)

docs/misc/VolumeNames.md
```diff
@@ -1,12 +1,19 @@
...
-Short Volume Name can contain up to a 100 of alphanumeric characters, underscores and dots.
-Examples: `MyVolume_fast.12` or `_12SuperStorage`.
+Short Volume Name can contain up to a 100 characters.
+All characters are allowed, but the last '@' will be interpreted as a separator
+between volume and datastore name (see below).
+As an exception, we do not allow volume names ending with '-NNNNNNN', where
+N is a digit (this limitation is to prevent conflicts with ESXi deltadisks used
+during VM snapshots). Examples: `MyVolume_fast.12`,  `_12SuperStorage`,
+`super-volume-200` are all valid names, while `myVolume-000012` is not valid.
+
```
govint: and "-volume-name" is also not allowed.

esx_service/utils/vmdk_utils.py
```diff
@@ -67,20 +78,24 @@ def get_volumes():
...
+    delta_disks = glob.glob("{0}/{1}{2}".format(path, vol_name, SNAP_SUFFIX_GLOB))
     if not delta_disks:
         return os.path.join(path, "{0}.vmdk".format(vol_name))
 
-    # this funky code gets the name of the youngest delta disk:
+    # this funky code gets the name of the latest delta disk:
     latest = sorted([(vmdk, os.stat(vmdk).st_ctime) for vmdk in delta_disks], key=lambda d: d[1], reverse=True)[0][0]
-    logging.debug("The youngest delta disk is %s. All delta disks: %s", latest, delta_disks)
+    logging.debug("The latest delta disk is %s. All delta disks: %s", latest, delta_disks)
     return latest
```
govint: This may be the newest delta disk created but its not necessarily the delta disk in use, if the user is moving between snaps. In which case the user can be on a delta-disk thats not the newest (by ctime) but in use. Wonder if we should really store the bus and unit number at which the disk is connected to the PVSCSI controller in the KV and when we get a request we only query the VM config to get the disk thats at that location. Then we get the exact disk thats in use no matter which snao the VM is running on.

govint: For non-create ops we should get the disk from the VM config.

esx_service/utils/vmdk_utils.py
```diff
@@ -95,33 +110,47 @@ def list_vmdks(path, volname="", show_snapshots=False):
...
+    name = file_name.lower()
+
+    # filter out all files with wrong extention
+    if not name.endswith('.vmdk'):
+        return False
+
+    # filter out -delta, -flat, -digest and -ctk VMDK files
+    if re.match(SPECIAL_FILES_REGEXP, name):
+        return False
+
```
govint: Can this be,
if not name.endswith(...) or re.match(SPECIAL....):
     return False

govint:
Looks good, had a couple of comments above.

docs/misc/VolumeNames.md
msterin: why?

esx_service/utils/vmdk_utils.py
msterin: on which VM? Are you suggesting to scan all registered VMs?

esx_service/utils/vmdk_utils.py
msterin: sure it can, but it's a matter of taste.

docs/misc/VolumeNames.md
govint: Docker allows names with '-' but not ones that start with '-', like,

administrator@hte-1s-eng-dhcp98:~$ docker volume create --name -dvol_2
Error response from daemon: create -dvol_2: "-dvol_2" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed

esx_service/utils/vmdk_utils.py
govint: But we don't need to scan all VMs each server request is from a specific VM, which is how all commands are processed (get the VM config and updated as requested). On a request we figure which bus/unit the device (volume name and hence KV) is attached at. Then read the VM config to get the device name.

msterin: That is true for detach. That is not true for 'docker volume ls' and ls' not is it true for admin CLI - they enumerate all volumes. And as far as I recall detach already handles that exactly via vm.config but it may be using the code you are pointing to. 
 I'll double  check later today. Thanks for the catch.

docs/misc/VolumeNames.md
msterin: Ah, but it only applies to local volumes. Docker leaves volume name constraints to drivers. For example, spaces in vmdk volume name work just fine with this PR

esx_service/utils/vmdk_utils.py
pdhamdhere: Mark & I played with "moving between snapshots". When you move to intermediate snapshot, it creates a new delta disk representing running point. So, logic to get latest disk would still work.

pdhamdhere:
LGTM

msterin:
//CC @kerneltime

PR #576, opened 2016-08-17
Add Admin CLI usage documentation

### LINKED ISSUE #563: Add Admin CLI documentation
This is a follow up from https://github.com/vmware/docker-volume-vsphere/pull/561
The current documentation is more of a design spec and should be moved into misc, a user facing document needs to be written with demos.

andrewjstone:
Fixes #563


docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
+[TOC]
+
+# Introduction
+In the context of the docker volume plugin for vSphere, each ESXi host manages multiple VMs, with
+each of them acting as a docker host. The docker engine on these hosts communicates with the docker
```
msterin: Docker needs to be capitalized. In all meanings of the word :-)

msterin: Also, I think we need to be clear on terminology. Docker deals with volumes, but ESXi / Vmware "virtual disks", aka VMDKs. 
So when Docker requests a Volume creation, we create a virtual disk. The admin tool provide a view into virtual disks used as Docker  volumes.
I think using "volumes" in ESXi context without explicitly defining what we are talking about, will be very confusing. At the very least , if you do not feel "virtual disk" is any more clearer, 
we should use "Docker Volumes" everywhere.

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+volume plugin in order to create, delete and mount volumes. These volumes may live on any datastore
+accessible to the ESXi host and are managed by the docker user via the docker CLI. However, the
+docker CLI is limited in what visibility it can provide to the user. Furthermore, it is desirable
+that an administrator be able to get a global view of all volumes created and in use on the host.
+For these reasons, an admin CLI has been created that runs on the ESXi host and that provides access
+to information not visible from the docker CLI. The remainder of this document will describe each
+admin CLI command and provide examples of their usage.
+
+
+# ls
```
msterin: This brings another question (not related to the doc) - do you think we should wrap it all in esxcli command? Esxcli  is easily expandable - an XML file maps esxcli subcommands to external command line

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+                    'datastore', 'created-by', 'created', 'attached-to',
+                    'policy', 'capacity', 'used']
+```
+
+### List All
+
+List all properties for all docker volumes that exist on datastores accessible to the host.
+
+```
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
```
msterin: Very nice ! Finally a document with examples !

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+Volume     Datastore   Created By VM  Created                   Attached To VM  Policy  Capacity  Used
+---------  ----------  -------------  ------------------------  --------------  ------  --------  -------
+large-vol  datastore1  Ubuntu_15.10   Sat Apr 16 13:34:12 2016  detached        N/A     1.00GB    25.00MB
+vol        datastore1  Ubuntu_15.10   Sat Apr 16 20:13:18 2016  detached        N/A     100.00MB  14.00MB
+```
+
+Note that the `Policy` column shows the VSAN storage policy. Since these example volumes live on a
```
msterin: "named VSAN storage policy, crested with the same tool (vmdkops_adim). "

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+### Create
+
+Create a VSAN storage policy.
+
+```
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy create --name some-policy --content '(("proportionalCapacity" i0)("hostFailuresToTolerate" i0)'
+Successfully created policy: some-policy
+```
+
+Note that the VSAN storage policy given with `--content` is a standard VSAN storage policy. Please
```
msterin: "storage policy string"

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+refer to the VSAN documentation for storage policy syntax.
+
```
msterin: e.g. `--policy` flag in `esxcli vsan policy setdefault`
 https://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.vcli.ref.doc%2Fesxcli_vsan.html

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+```
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy ls
+Policy Name  Policy Content                                             Active
+-----------  ---------------------------------------------------------  ------
+some-policy  (("proportionalCapacity" i0)("hostFailuresToTolerate" i0)  Unused
+```
+
+When creating a volume using `docker volume create`, the policy name should be given with the `-o`
+option such as `docker volume create --driver=vmdk --name=some-vol -o vsan-policy-name=some-policy`.
+The name of the volume will then show up in the `Active` column.
```
msterin: what if there are 100 volumes using this policy ? will they all show ? How the line length will be handed ?

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Update
+
+Update a VSAN storage policy.
+
+This command will update a VSAN storage policy for all volumes currently using this policy.
+
```
msterin: Will there be any progress report ? Or a message with count of updated volumes ?

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+```
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy update --name some-policy --content '(("proportionalCapacity" i1)'
+This operation may take a while. Please be patient.
+Successfully updated policy: some-policy
```
msterin: Would be nice to tell the count of updated volumes

docs/user-guide/admin-cli.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+### Remove
+
+Remove a VSAN storage policy.
+
+```
+[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy rm some-policy
+Successfully removed policy: some-policy
```
msterin: what happens if the policy is used ? I think the rm is blocked and issues some message, right ?

msterin:
Nice write-up ! Comments inside

docs/user-guide/admin-cli.md
andrewjstone: I agree with your assesment. Will clean it up.

docs/user-guide/admin-cli.md
andrewjstone: I'm not sure if it's necessary, but I suppose that might be the right way to go if the esx admin is already used to using esxcli.

docs/user-guide/admin-cli.md
andrewjstone: hahahah

docs/user-guide/admin-cli.md
andrewjstone: Thanks for the link! I was hoping you'd dig it out for me :)

docs/user-guide/admin-cli.md
andrewjstone: Aha, I just double checked the code. It actually just shows the number of volumes attached.
 ` active = 'In use by {0} volumes'.format(used_policies[name])`

I'll correct the text

docs/user-guide/admin-cli.md
andrewjstone: It's not currently in the code though, so not documented here. It does however show those stats if vsan volumes fail to update. 

```
return ('Successfully updated {0} volumes.\n'
           'Failed to update volumes: {0}').format(update_count, failed_updates)
```

docs/user-guide/admin-cli.md
andrewjstone: Good Point! I will update the doc.

andrewjstone: I did write about the remove being blocked, but I'm not sure there's anything that can be done about it other than deleting the virtual disks or somehow removing the policy from them. But I'm unsure how we should recommend doing the latter. It feels like we shouldn't make users go outside the admin cli for that. Maybe it can be an enhancement?

andrewjstone:
@msterin I believe I addressed all your review comments. Thanks again for the great review.

docs/user-guide/admin-cli.md
msterin: Well, "delete policy" when it is used feels wrong, so I think user  should be able to drop policy from all objects it assigned to; they would revert to defaultPolicy as defined in esxcli, and then they should be able to delete the policy. If we do not have a way to remove policy (even as long as the one above) please file a PR  and let's discuss offline . If there is way please document it

msterin:
We need to file a PR or document how to delete a policy (even if it's a multiple steps operation) . 
Other that that, LGTM

Issue #413, opened 2016-05-27
Support a volume update option in the admin CLI to update volume properties.


govint:
For now, once a volume is created its impossible to change the size of the volume. Till such time docker supports a "docker volume <update or modify>" option the admin CLI can be used for this. To start, could support changing the size of the disk.

govint:
PR #583 has created the infra to allow updating volume attributes. Currently access-mode is supported, the new capability can be extended as needed to include more volume attributes as needed. Closing this issue as #583 adds the capability requested here.

Issue #274, opened 2016-04-26
Support Read only volumes?


pdhamdhere:
This came up during meeting with TM. Does Docker support read only volumes? RD Only volume will enable sharing. What are concrete use cases this would enable?

msterin:
Note: -o readonly would mean a volume is ALWAYS readonly , on any use. It would be OK If we had a way to reclassify it as "writeable" at some point - but we don't

msterin:
Let's look at that in GA context - we have a few use cases popping up already, e.g. read-only docs distribution and immutable code libraries. Not sure if we can/should squeeze it it, but it's worth taking a look  sketching  / costing

govint:
Closed via #583

Issue #442, opened 2016-06-07
Support importing a disk into a volume repository on an ESX host


govint:
To allow existing users migrate their data (vmdks) to be managed by the docker-volume-vsphere plugin we need a way by which users can specify a single or multiple volumes to be "imported" into the volume repository managed by the plugin and hence docker.

This is assuming that docker will not accept volumes that are simply placed into the volume repo on the ESX host side unless the volume is explicitly created via docker.

So,
1. Support a "docker volume create -o "importvol=<path to vol>...." " option specifying the location (on the host) from where an existing volume must be imported into the docker volume repo on the host.
1. Create fails if the volume doesn't exist.
2. If the volume exists at the specified path at the host, then move the volume to the repo and init meta-data for it.
3. Done.

Point to consider is whether the volume must be migrated to the volume-repo of the plugin or it remains where it is (say 1TB volume) but we create a vmdk descriptor in the volume-repo and the meta-data for it while the actual volume stays elsewhere (for vsan and vvol thats the case anyway).

msterin:
what's wrong with simply moving vmdk to the right folder ESX ?

govint:
That won't be the way a user who's using VSphere VMs today to migrate to docker. The volumes need to be created thru docker so that docker knows of the volumes and the plugin can create the meta-data for the imported volumes. How will meta-data be created by moving the volumes or how will docker know of the volumes either?

ANother workflow is to be able to migrate volumes from another plugin-type into the vsphere plugin. For example, how will a volume managed by flocker-vsphere-plugin be migrated/imported by the vsphere-plugin?

msterin:
First of all that's not true - dicker uses Listvrequest to know about volumes and copy to folder works. We should simply recreate metadata if it's missing

Second, unless a customer asks for that, we should not spend time on this IMO

govint:
It looks more a hack to have users copy files into a volume repository
managed by the plugin on a specific esx host than importing via docker
which looks correct given these are docker volumes. And then expect to run
their container on a VM on that specific host. Going via docker and
ensuring metadata gets created correctly is better user experience as well.

msterin:
yes , it is better, if anyone ever will be using it. But it's not free and it's distracting from what we already *know* customers are asking for - e.g. backup story. Besides, it's an extension to existing #62 which is already slated for v.next.
So let's hear at least *some* customers asking for it before starting to worry about it and spend time on this. Parking to v.next for now

govint:
[...]

PR #583, opened 2016-08-22
Support a read-only option to vsphere volumes to mount volumes as read-only on container run.

govint:
Added an option to specify an access mode on volume creation, the volume is mounted always as read-only when attached to a container. Which is ok only as long as there is a way to attach the volume to the VM and be able to init. the volume with data. Perfect for pre-initialized volumes being imported into the volume repo on the vsphere side (issue #442 ). 

Tested as below:

1. Create a volume with access=read-only and mount it. Verify it mounts read-only on the client.

administrator@hte-1s-eng-dhcp98:~/vpl-add-ro-vols/docker-volume-vsphere$ docker volume inspect testvol-4
[
    {
        "Name": "testvol-4",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/testvol-4",
        "Status": {
            "access": "read-only",
            "capacity": {
                "allocated": "14.0MB",
                "size": "100.0MB"
            },
            "created": "Mon Aug 22 07:27:35 2016",
            "created by VM": "2",
            "datastore": "bigone",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]
administrator@hte-1s-eng-dhcp98:~/vpl-add-ro-vols/docker-volume-vsphere$ docker run --rm -it -v testvol-4:/data busybox
/ # cd /data
/data # >sdafhwef
sh: can't create sdafhwef: Read-only file system
/data #

And /proc/mounts shows the read-only mount,
/dev/disk/by-path/pci-0000:13:00.0-scsi-0:0:0:0 /mnt/vmdk/testvol-4 ext2 ro,relatime 0 0

2. Try using an invalid access type:

$ docker volume create --driver=vmdk --name=testvol-5 -o "access=access1"
Error response from daemon: create testvol-5: VolumeDriver.Create: Access type 'access1' is not supported. Valid options are: ['read=write', 'read-only']

3. Verify default behavior is to create a read-write volume

$ docker volume create -d vmdk --name testvol-6
testvol-6

$ docker volume inspect testvol-6
[
    {
        "Name": "testvol-6",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/testvol-6",
        "Status": {
            "access": "read-write",
            "capacity": {
                "allocated": "14.0MB",
                "size": "100.0MB"
            },
            "created": "Mon Aug 22 11:51:26 2016",
            "created by VM": "2",
            "datastore": "bigone",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]

### DIFF
--- a/docs/misc/vmdkops-admin-cli-spec.md
+++ b/docs/misc/vmdkops-admin-cli-spec.md
@@ -121,6 +121,20 @@ Show any interesting information about the service. This includes file paths of 
 information, and PID of running service. A simple example is shown here, although it's possible
 that the exact format may be somewhat different.
 
+# Set
+Modify attribute settings on a given volume. The volume is identified by its full path on the datastore,
+for example if the volume name is `container-vol` then the volume is specified as "/vmfs/volumes/<datastore>/dockvols/container-vol.vmdk".
+The attributes to set/modify are specified as a comma separated list as "<attr1>=<value>, <attr2>=<value>....". For example,
+a command line would look like this.
+
+```
+$ vmdkops-admin set --volume=<full path of volume> --options="<attr1>=<value>, <attr2>=<value>, ..."
+```
+
+The volume attributes are set and take effect only the next time the volume attached to a VM. The changes do not impact any VM
+thats currently using the volume. For the present, only the "access" attribute is supported to be modified via this command, and
+can be set to either of the allowed values "read-only" or "read-write".
+
 ```
 [root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
 Version: 1.0.0-0.0.1
--- a/docs/user-guide/docker-volume-cli.md
+++ b/docs/user-guide/docker-volume-cli.md
@@ -54,3 +54,11 @@ Currently the following are supported
 
 1. [persistent](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/): If the VMDK is attached as persistent it will be part of a VM snapshot. If a VM snapshot has been taken while the Docker volume is attached to a VM, the Docker volume then continues to be attached to the VM that was snapshotted.
 2. [independent_persistent](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/): If the VMDK is attached as independent_persistent it will not be part of a VM snapshot. The Docker volume can be attached to any VM that can access the datastore independent of snapshots.
+
+### access
+```
+docker volume create --driver=vmdk --name=MyVolume -o access=read-only -o diskformat=thin
+docker volume create --driver=vmdk --name=MyVolume -o access=read-write -o diskformat=thin (default)
+```
+
+The access mode determines if the volume is modifiable by containers in a VM. The access mode allows to first create a volume with write access and initialize it with binary images, libraries (for exmple), and subsequently change the access to "read-only" (via the admin CLI). Thereby, creating content sharable by all containers in a VM.
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -255,6 +255,20 @@ def commands():
         'status': {
             'func': status,
             'help': 'Show the status of the vmdk_ops service'
+        },
+        'set': {
+            'func': set_vol_opts,
+            'help': 'Edit settings for a given volume',
+            'args': {
+                '--volume': {
+                    'help': 'Full path of the volume',
+                    'required': True
+                },
+                '--options': {
+                    'help': 'Options (specifically, access) to be set on the volume.',
+                    'required': True
+                }
+            }
         }
     }
 
@@ -269,12 +283,12 @@ def create_parser():
 def add_subparser(parser, cmds_dict):
     """ Recursively add subcommand parsers based on a dictionary of commands """
     subparsers = parser.add_subparsers()
-    for cmd, attributes in cmds_dict.iteritems():
+    for cmd, attributes in cmds_dict.items():
         subparser = subparsers.add_parser(cmd, help=attributes['help'])
         if 'func' in attributes:
             subparser.set_defaults(func=attributes['func'])
         if 'args' in attributes:
-            for arg, opts in attributes['args'].iteritems():
+            for arg, opts in attributes['args'].items():
                 opts = build_argparse_opts(opts)
                 subparser.add_argument(arg, **opts)
         if 'cmds' in attributes:
@@ -485,7 +499,7 @@ def policy_ls(args):
         else:
             used_policies[policy_name] = 1
 
-    for name, content in policies.iteritems():
+    for name, content in policies.items():
         if name in used_policies:
             active = 'In use by {0} volumes'.format(used_policies[name])
         else:
@@ -513,6 +527,14 @@ def status(args):
     print("LogConfigFile: {0}".format(log_config.LOG_CONFIG_FILE))
     print("LogFile: {0}".format(log_config.LOG_FILE))
     print("LogLevel: {0}".format(log_config.get_log_level()))
+
+
+def set_vol_opts(args):
+    set_ok = vmdk_ops.edit_vol_opts(args.volume, args.options) 
+    if set_ok:
+        print('Successfully updated settings for : {0}'.format(args.volume))
+    else:
+        print('Failed to update {0} for {1}.'.format(args.options, args.volume))
 
 
 VMDK_OPSD = '/etc/init.d/vmdk-opsd'
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -202,8 +202,10 @@ def validate_opts(opts, vmdk_path):
      * vsan-policy-name - The name of an existing policy to use
      * diskformat - The allocation format of allocated disk
     """
-    valid_opts = [kv.SIZE, kv.VSAN_POLICY_NAME, kv.DISK_ALLOCATION_FORMAT, kv.ATTACH_AS]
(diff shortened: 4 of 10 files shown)


msterin:
Fixes #274

pdhamdhere:
I think I am missing something. Help me understand the proposed feature;
-  If volume is created as read-only then how does user initializes it?
- Can this volume be attached to multiple docker hosts as read-only?
- Would it support 1-writer (Docker Host) and many-readers (Docker hosts)?

msterin:
@pdhamdhere , let me taker a stub at answering per my understanding:

    If volume is created as read-only then how does user initializes it

since "read only" behavior is controlled by KV, nothing prevents us from adding an admin CLI option to flip to on and off for offline volumes. With this, the question "how do you create it" is answered.  So the PR needs the admin support too.

    Can this volume be attached to multiple docker hosts as read-only?

I think mounting to multiple dockers still won't work as VMX/VMFS interaction will still lock the vmdk and treat it as writeable. We can craft something around delta disks, but as-is this is still exclusive use.

    Would it support 1-writer (Docker Host) and many-readers (Docker hosts)?

No

msterin:
FWIW I think this PR, while the code is good,  requires Admin CLI support and test automation to be functional.

govint:
1. The admin CLI command is this one:
python /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py edit -h
usage: vmdkops_admin.py edit [-h] --volume VOLUME --options OPTIONS

optional arguments:
  -h, --help         show this help message and exit
  --volume VOLUME    Full path of the volume
  --options OPTIONS  Options (specifically, access) to be set on the volume.

options is a comma separated list jut like how options are specified when creating a volume.

2. Using the command to reset access mode on a volume

$ docker volume create -d vmdk --name testvol-11
testvol-11

$ docker volume inspect testvol-11
[
    {
        "Name": "testvol-11",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/testvol-11",
        "Status": {
            "access": "read-write",
            "attach_as": "independent_persistent",
            "capacity": {
                "allocated": "14.0MB",
                "size": "100.0MB"
            },
            "created": "Thu Aug 25 08:05:33 2016",
            "created by VM": "2",
            "datastore": "bigone",
            "diskformat": "thin",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]

On the host reset the access mode on the volume

] python /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py edit --volume=/vmfs/volumes/bigone/dockvols/testvol-11.vmdk --opti
ons="access=read-write"
Successfully updated settings for : /vmfs/volumes/bigone/dockvols/testvol-11.vmdk

$ docker volume inspect testvol-11
[
    {
        "Name": "testvol-11",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/testvol-11",
        "Status": {
            "access": "read-write",
            "attach_as": "independent_persistent",
            "capacity": {
                "allocated": "14.0MB",
                "size": "100.0MB"
            },
            "created": "Thu Aug 25 08:05:33 2016",
            "created by VM": "2",
            "datastore": "bigone",
            "diskformat": "thin",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]

esx_service/cli/vmdkops_admin.py
```diff
@@ -255,6 +255,20 @@ def commands():
         'status': {
             'func': status,
             'help': 'Show the status of the vmdk_ops service'
+        },
+        'edit': {
```
msterin: 'edit' sounds like we are editing content. 'settings' or 'config' sounds better to me . `config --options`

esx_service/cli/vmdkops_admin.py
```diff
@@ -515,6 +529,14 @@ def status(args):
     print("LogLevel: {0}".format(log_config.get_log_level()))
 
 
+def edit_vol(args):
+    edit_ok = vmdk_ops.edit_vol_opts(args.volume, args.options) 
+    if edit_ok:
+        print('Successfully updated settings for : {0}'.format(args.volume))
```
msterin: printing message here seems unneeded.  Do we do it in other places ? I think we do need a log record though .

esx_service/cli/vmdkops_admin.py
```diff
@@ -515,6 +529,14 @@ def status(args):
...
+    else:
+        print('Edit failed for {0}.'.format(args.volume))
```
msterin: will it tell why it failed? An example of output would be good.

esx_service/vmdk_ops.py
```diff
@@ -996,6 +1019,33 @@ def disk_detach_int(vmdk_path, vm, device):
     return None
 
 
+# Edit settings for a volume identified by its full path
+def edit_vol_opts(vmdk_path, options):
```
msterin: nit: an example of parsed input, in comments, would be good

msterin:
Nice change !

Comments: 'edit' in admin options sounds like we are editing the volume (not config) so I suggest 'set'. Also, I do not think we need a message on success (unless we do it in all other admin commands', and instead of that we need a log record.  A few more nits inline. 

LGTM after that.

esx_service/cli/vmdkops_admin.py
andrewjstone: I agree. `update` is also a possibility. `set` is used in the role commands, while `update` is used in the policy command.

andrewjstone:
Nice work @govint LGTM after changing 'edit' to 'set'.`

pdhamdhere:
Please update documentation with new options in same PR. /cc @kerneltime

msterin:
LGTM

docs/misc/vmdkops-admin-cli-spec.md
```diff
@@ -121,6 +121,18 @@ Show any interesting information about the service. This includes file paths of
...
 that the exact format may be somewhat different.
 
+# Set
+Modify attribute settings on a given volume. The volume is identified by its full path on the datastore,
+for example if the volume name is `container-vol` then the volume is specified as "/vmfs/volumes/<datastore>/dockvols/container-vol.vmdk".
+The attributes to set/modify are specified as a comma separated list as "<attr1>=<value>, <attr2>=<value>....". For example,
+a command line would look like this.
+
+```
+$ vmdkops-admin set --volume=<full path of volume> --options="<attr1>=<value>, <attr2>=<value>, ..."
```
pdhamdhere: Should we be using volumeName@datastore similar to what we use in docker run command. I don't think vmdkops-admin prints full volume path in any command.

docs/user-guide/docker-volume-cli.md
```diff
@@ -54,3 +54,11 @@ Currently the following are supported
...
 2. [independent_persistent](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/): If the VMDK is attached as independent_persistent it will not be part of a VM snapshot. The Docker volume can be attached to any VM that can access the datastore independent of snapshots.
+
+### access
+```
+docker volume create --driver=vmdk --name=MyVolume -o access=read-only -o diskformat=thin
+docker volume create --driver=vmdk --name=MyVolume -o access=read-write -o diskformat=thin (default)
+```
+
+The access mode determines if the volume is modifiable by containers in a VM. The access mode allows to first create a volume with write access and initialize it with binary images, libraries (for exmple), and subsequently change the access to "read-only" (via the admin CLI). Thereby, creating content sharable by all containers in a VM.
```
pdhamdhere: Add "Volume must not be in use"?

pdhamdhere: BTW, I don't see a check for whether VMDK is attached?

govint: The command only updates the KV and will take effect on the next attach
I'll update that in the doc.

docs/misc/vmdkops-admin-cli-spec.md
govint: The issue is the plugin gets to know the VM and can figure which dockvols
folder to create the volume in. But the user on the ESX always needs to
know the datastore where the volume is located. Otherwise we can't locate
the volume at all. Hence the need to give the path.

Alternative is for the user to always give volume@datastore as the format.
I can change to make it that way.

kerneltime:
If this is still a single VM attach then I don't see how it fixes https://github.com/vmware/docker-volume-vsphere/issues/274 what feature are we implementing here? Also, we need to add tests, documentation and demos.

govint:
Both documentation and tests are included in the change. It's not possible
to attach a volume to multiple VMs without making it a shared volume in
vmfs.

docs/user-guide/docker-volume-cli.md
govint: Volume can be in use. Only the kV is being changed and will l take effect
on the next mount inside the guest which can happen after a detach and
subsequent attach

docs/misc/vmdkops-admin-cli-spec.md
pdhamdhere: I think ' vmdkops-admin ls ' prints datastore. VolName@DS in addition to full path would be nice and can be separate PR.

esx_service/cli/vmdkops_admin.py
```diff
@@ -255,6 +255,20 @@ def commands():
         'status': {
             'func': status,
             'help': 'Show the status of the vmdk_ops service'
+        },
+        'set': {
+            'func': set_vol_opts,
+            'help': 'Edit settings for a given volume',
+            'args': {
```
pdhamdhere: Sure. Add a note in comment that change will be effective next time volume is mounted or something on those lines.

pdhamdhere:
LGTM after updating help

kerneltime:
fixes https://github.com/vmware/docker-volume-vsphere/issues/413

docs/misc/vmdkops-admin-cli-spec.md
kerneltime: This is the wrong place to update the user facing CLI documentation.. fixing it and posting a PR..

PR #584, opened 2016-08-23
Add support for RHEL

kerneltime:
This adds support for both Photon and RHEL in the same rpm binary. 
Dropped dependency check for docker for rpm to make the same binary work for RHEL and Photon.

CI system currently tests for regression, support for RHEL in CI is part of a separate PR.

Testing:

Single RHEL 7.2 VM with docker 1.12

1. Uninstalled docker, plugin installs on FS but does not start up.
2. Install docker and restart plugin, the plugin works fine.
3. Ran manual tests after deploy-vm on Photon and RHEL (create, multiple attaches)


msterin:
LGTM. One question though - are we going to have these checked in CI ?

kerneltime:
Yes. Not this PR though.

govint:
Can we also ensure no pvscsi controllers in the VM to start with. And each CI run should start the VM without a PVSCSI controller. We have sufficient code thats never tested when a pvscsi controller is already present.

msterin:
The code is tested every time anyone bring up a new docker vm, which is reasonable often. And are you talking about having it tested in CI (which is planned and issue # is mentioned above) or about something else ?

govint:
When developing fixes, the same VM is used for all development and test. And that VM invariably has the pvscsi controller in it (from local testing). The CI VMs should be a fresh slate without any configuration that's also created by the plugin, so any issue will be caught before a fix is merged.

msterin:
@govint , we should take the comments about pvscsi out of RHEL fix thread.

kerneltime:
@msterin Agreed. We should file a ticker for missing test cases. The problem is dual sided on one had we want machines that have been running/upgraded over a long period and we want to replicate the fresh install use case. This sounds more like a use case for missing tests in CI.

kerneltime:
Using https://github.com/vmware/docker-volume-vsphere/pull/584 to capture the need for fresh VMs as part of CI.

pdhamdhere:
LGTM

PR #588, opened 2016-08-26
Add vol stats

govint:
Added volume stats on Get()


kerneltime:
govint are there changes off a fork? For some reason the CI system cannot find the change.

govint:
Yes these are on a forked branch. How do I prevent CI from picking this up

Issue #596, opened 2016-09-07
Move filesystem creation to docker host and add fstype option #597.


brunotm:
Volumes are formatted as ext4 filesystem, but we mount them as ext2 unless when mocking the ESX server. There is a reason ?

kerneltime:
The tool is suffixed with ext4 but the fs is ext2
We plan to support other fs types if requested by users.
Do you have a specific need for a FS type and which one? Do let us know.
To double check the FS type, you can run a container image using the volume and on a different console session run the file command
```shell 
root@photon-VYb05tsNa [ ~ ]# docker volume create --driver=vmdk --name=MyVolume -o size=10gb
MyVolume
root@photon-VYb05tsNa [ ~ ]# docker run -it --rm -v MyVolume:/vol busybox
/ # 
```
And then run the following command
```shell 
root@photon-VYb05tsNa [ ~ ]# mount| grep MyVolume
/dev/sdb on /mnt/vmdk/MyVolume type ext2 (rw,relatime,block_validity,barrier,user_xattr,acl)
root@photon-VYb05tsNa [ ~ ]#  file -sL /dev/sdb
/dev/sdb: Linux rev 1.0 ext2 filesystem data (mounted or unclean), UUID=59819b66-7b3b-428a-8b4c-bcc78df08216, volume name "MyVolume" (large files)
```

kerneltime:
Changing the title

brunotm:
No specific needs.

But I think that ext4 would be a reasonable default due to data integrity and performance improvements.

brunotm:
I've done a small patch #597 changing the default filesystem to ext4, and also noticed that the default mkfs command included is somewhat outdated.

However, handling filesystem creation in the plugin on the docker host seems a more reasonable approach since wouldn't be necessary to maintain filesystem specific tools over the ESX server, and in the future filesystem choice could be determined by the available drivers and tools on the docker host.

What do you think?

kerneltime:
Taking a look.

brunotm:
It looks that would be safe to handle the formatting in plugin/Create() (create->attach->mkfs->detach), avoiding nasty verifications or unsafe conditions in plugin/Mount(). And implement Mkfs in utils/fs.

Determining which mkfs.* to call could be done with a fstype option in Request.

kerneltime:
Reassigning to Mark to look into supporting ext4

msterin:
[...]

PR #597, opened 2016-09-08
Move filesystem creation to docker host and add fstype option.

brunotm:
Make ext4 default #596


kerneltime:
Taking a look. Please sign the CLA (let me know if you run into issues).

kerneltime:
The binary that is in the repo does not support ext 4. I tested your change and the volume is still formatted with ext2 (ext4 code mounts ext2 disks)
Reassingning to Mark to look into supporting ext4
```shell
root@photon-VYb05tsNa [ ~ ]# file -sL /dev/sdb
/dev/sdb: Linux rev 1.0 ext2 filesystem data (mounted or unclean), UUID=596a1184-0ad3-40b4-a8a7-06f9a8b63e45, volume name "test-ext4" (large files)
```

msterin:
This would only work if we update the mkfs binary - the current one does not support ext4.

vmdk_plugin/plugin.go
```diff
@@ -208,11 +209,50 @@ func (d *vmdkDriver) unmountVolume(name string) error {
...
+		return volume.Response{Err: err.Error()}
+	}
+
+	device, err := fs.GetDevicePath(dev)
+	if err != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Could not find attached device ")
+		return volume.Response{Err: err.Error()}
+	}
+
+	err = fs.Mkfs(mkfscmd, r.Name, device)
```
msterin: this could be very slow for large disk sizes.
Ideally it should be a go func with some timeout for ("in progress") messages, and killable but at the very least need we need to log start/end.
Note: the current code does not handle large disk formats very well either.

vmdk_plugin/plugin.go
```diff
@@ -281,7 +321,7 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 		isReadOnly = true
 	}
 
-	mountpoint, err := d.mountVolume(r.Name, isReadOnly)
+	mountpoint, err := d.mountVolume(r.Name, status["fstype"].(string), isReadOnly)
```
msterin: I think it's better to auto-detect the FS ( `-t auto`).  Tracking FS type in volume metadats (KV) is slightly more be error-prone.

vmdk_plugin/plugin.go
```diff
@@ -208,11 +209,50 @@ func (d *vmdkDriver) unmountVolume(name string) error {
 // (until Mount is called).
 // Name and driver specific options passed through to the ESX host
 func (d *vmdkDriver) Create(r volume.Request) volume.Response {
-	err := d.ops.Create(r.Name, r.Options)
+	// Defaults to ext4 if not specified
+	if _, ok := r.Options["fstype"]; ok == false {
```
msterin: nit: would be better to have a global const with "fstype" string and "ext4" string, just to have it all in one place

vmdk_plugin/plugin.go
```diff
@@ -208,11 +209,50 @@ func (d *vmdkDriver) unmountVolume(name string) error {
...
+	device, err := fs.GetDevicePath(dev)
+	if err != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Could not find attached device ")
+		return volume.Response{Err: err.Error()}
+	}
+
+	err = fs.Mkfs(mkfscmd, r.Name, device)
+	if err != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Create filesystem failed ")
+		return volume.Response{Err: err.Error()}
```
msterin: Need to detach before return.

msterin:
Good change - thanks for taking it on !

A few notes : 

* [blocker] The CI is failing in docker-side unit test (`make testasroot`). The test is trying to create a local file and feed it to the plugin via loopback device (https://github.com/vmware/docker-volume-vsphere/blob/master/vmdk_plugin/vmdkops/mock_vmdkcmd.go)  and then do some testing.  I did not investigate but I would guess it needs adjustment for API changes. 

* [blocker] need CI tests to cover different formats. It could be a part of the above code.

* Since it breaks "all options parsing is done on ESX" approach, we need some comments on ESX-side Python to point out that the fstype list is determined by actual guest VM, and IMO guest-VM error message should include a list of available filesystems. 

* I think Admin_cli ls should should show  fstype (https://github.com/vmware/docker-volume-vsphere/blob/master/esx_service/cli/vmdkops_admin.py#L343) 

A few more comments are inline.

vmdk_plugin/plugin.go
brunotm: I have to evaluate the impact of doing it async or printing progress messages when creating the volume in other asyc operations eg. docker service create. logging is definitely needed.

vmdk_plugin/plugin.go
brunotm: syscall.Mount do not support "auto" (which is the mount command doing its own magic). In this case we would have to call the mount command externally.

I like the idea of tracking the fstype in metadata because it enforce strictness, or maybe i did not get the error-prone scenario right, can you provide an example?

vmdk_plugin/plugin.go
brunotm: Ok 👍

vmdk_plugin/plugin.go
brunotm: Sure! shouldn't we also remove the volume in this case?

brunotm:
Thanks! :)
I will rework, adjust and take care of the failing tests for a second iteration.

kerneltime:
@brunotm you can run the same CI tests locally https://github.com/vmware/docker-volume-vsphere/blob/master/CONTRIBUTING.md#testing-and-ci 

Let me know if you face any challenges.

vmdk_plugin/plugin.go
msterin: Note that the current code would not do well with the slow format either. In fact, it will block all requests on a single ESX. This change already improves this so that only a single VM is impacted - even without async or printing progress.  So it's all good , just add the logs please.

vmdk_plugin/plugin.go
msterin: In "error prone" I meant whatever bug we may have in KV management may lead to an attempt to mount a wrong FS type. Nothing specific, just a vague feeling that the FS type in KV is redundant. 

But yes, you right - I forgot we switched to syscall from invoking mkfs executable so tracking it in KV and feeding ro syscall.Mount is a good choice

vmdk_plugin/plugin.go
msterin: Yes - good point.

brunotm:
@kerneltime 
Thanks!

i have most of the issues solved already, but still have to implement tests with fstype other than fs.FstypeDefault and having admin_cli show the filesystem type.

I'll find some free time to do it later today.

brunotm:
@msterin @kerneltime 

This patch should address the issues discussed.
One thing i noticed is we may find cases where the tools for creating the filesystem X are present but the running kernel has no drivers for it, this is a pain to handle in the plugin and IMO is out of limits. Users should know how when choosing an FS or use the default ext4.
Also most of mkfs tools support -L for labels, but a few less-common or dated do not (minix,bfs...)
I think labeling is a good practice here, although is not a requirement and can be removed safely.

Probably i also will have to update the documentation somewhere.

brunotm:
The Ci test continues failing, i'll have to setup them locally. build and testasroot completes without errors.

kerneltime:
@brunotm It looks like some unit tests that are failing. 
```
Running unit tests in /tmp/vmdk_ops_unittest15791/vmdkops_admin_sanity_test.py...
F..
FAIL: test_ls (__main__.TestVmdkopsAdminSanity)
  File "/tmp/vmdk_ops_unittest15791/vmdkops_admin_sanity_test.py", line 40, in test_ls
    self.assertEqual(8, len(divider_columns))
AssertionError: 8 != 9

Ran 3 tests in 1.985s

FAILED (failures=1)
Running unit tests in /tmp/vmdk_ops_unittest15791/vmdkops_admin_test.py...
.F........................
FAIL: test_ls_no_args (__main__.TestLs)
  File "/tmp/vmdk_ops_unittest15791/vmdkops_admin_test.py", line 214, in test_ls_no_args
    self.assertEqual(8, len(header))
AssertionError: 8 != 9

Ran 26 tests in 113.748s

FAILED (failures=1)
Running unit tests in /tmp/vmdk_ops_unittest15791/vol_test.py...
VOL_TEST kv_store skippped
Running unit tests in /tmp/vmdk_ops_unittest15791/vsan_info_test.py...
..
Ran 2 tests in 8.068s

OK
Running unit tests in /tmp/vmdk_ops_unittest15791/vsan_policy_test.py...
...ERROR:root:Failed to remove test_policy policy file
  File "/tmp/vmdk_ops_unittest15791/vsan_policy.py", line 209, in delete
    os.remove(policy_path(name))
OSError: [Errno 2] No such file or directory: '/vmfs/volumes/vsan:52fe282b8b5a398c-e41bd80f9e940d46/dockvols/policies/test_policy'
..
Ran 5 tests in 5.132s

OK

=> Build + Test not successful
=> Thu Sep 15 23:35:17 UTC 2016
```

vmdk_plugin/plugin.go
```diff
@@ -208,12 +209,55 @@ func (d *vmdkDriver) unmountVolume(name string) error {
 // (until Mount is called).
 // Name and driver specific options passed through to the ESX host
 func (d *vmdkDriver) Create(r volume.Request) volume.Response {
-	err := d.ops.Create(r.Name, r.Options)
+	// Use default fstype if not specified
+	if _, ok := r.Options["fstype"]; ok == false {
```
msterin: nit: `ok == false ` looks really strange, I think the name should be 'result'. `result == false` is easy to digest :-)

vmdk_plugin/plugin.go
```diff
@@ -208,12 +209,55 @@ func (d *vmdkDriver) unmountVolume(name string) error {
...
+
+	err = d.ops.Create(r.Name, r.Options)
 	if err != nil {
 		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Create volume failed ")
 		return volume.Response{Err: err.Error()}
 	}
-	log.WithFields(log.Fields{"name": r.Name}).Info("Volume created ")
+
+	// Handle filesystem creation
+	log.WithFields(log.Fields{"name": r.Name, "fstype": r.Options["fstype"]}).Info("Attaching volume and creating filesystem ")
```
msterin: nit: this is much longer that 80 chars. makes it hard to use dual pages in an editor on  a laptop....

vmdk_plugin/vmdkops/mock_vmdkcmd.go
```diff
@@ -138,7 +138,7 @@ func createBlockDevice(label string) error {
 	if err != nil {
 		return err
 	}
-	return makeFilesystem(device, label)
+	return fs.Mkfs("/sbin/mkfs."+fs.FstypeDefault, label, device)
```
msterin: nit: spaces around '+' are missing

brunotm:
@kerneltime
It needs to be rebased against current changes in upstream master, I'll look into it.

@msterin
Forgot to write the tests :S

I will fix those later today and hopefully it will be good enough to merge.

Thanks!

brunotm:
@msterin 
I can't find a reason why the CI tests are failing during the testasroot (running make testasroot finishes successfully).

In the CI it fails with (in setupLoopbackDevice()):

`=> Running target testasroot
=> Sun Sep 18 02:15:57 UTC 2016

--- FAIL: TestCommands (0.01s)
        Error Trace:    cmd_test.go:31
	Error:  	Expected nil, but got: &errors.errorString{s:"Failed to setup loopback device node /dev/loop1001 for backing file /tmp/docker-volumes/244/myVolume: exit status 1. Output = /dev/loop1001: Operation not permitted\n"}
		
        Error Trace:    cmd_test.go:40
	Error:  	Expected nil, but got: &errors.errorString{s:"Failed to setup loopback device node /dev/loop1002 for backing file /tmp/docker-volumes/244/otherVolume: exit status 1. Output = /dev/loop1002: Operation not permitted\n"}
		
        Error Trace:    cmd_test.go:45
	Error:  	Expected nil, but got: &errors.errorString{s:"Failed to setup loopback device node /dev/loop1003 for backing file /tmp/docker-volumes/244/anotherVolume: exit status 1. Output = /dev/loop1003: Operation not permitted\n"}
		
FAIL
FAIL	github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops	0.011s
make[1]: *** [testasroot] Error 1
make: *** [testasroot] Error 2
`
I have scripted the same steps used in the tests and still can't reproduce the errors.

brunotm:
@kerneltime @msterin 

I just ran the CI tests locally against the master branch, and they fail with the same errors above. This error probably is something specific to my setup (ubuntu 16.04 + docker 12.0.1-test) and a quick investigation points to the capabilities of the container running the test, which shouldn't be a problem since the test containers have "privileged: true" in .drone.dev.yml.

The problem is that i don't know where the CI is failing or why. It is possible to have access to the CI run logs so i can track this ?

Thanks!

kerneltime:
If you run testasroot you need to be logged in as root. ```"Failed to setup loopback device node ``` to be able to create loop back device for tests you need to be root. You can alternatively run make test which runs the tests in a docker container as root.

kerneltime:
Here are the logs
```
=> Deploying root@192.168.31.81 : rpm
=> Mon Sep 19 07:23:30 UTC 2016

Preparing...                          ########################################
Updating / installing...
docker-volume-vsphere-0:0.6.0405b77-1 ########################################
  File: '/proc/1/exe' -> '/usr/lib/systemd/systemd'
Created symlink from /etc/systemd/system/multi-user.target.wants/docker-volume-vsphere.service to /usr/lib/systemd/system/docker-volume-vsphere.service.
9413

=> Running target testasroot
=> Mon Sep 19 07:23:31 UTC 2016

ok  	github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops	1.425s
ok  	github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config	0.004s

=> Running target test-esx
=> Mon Sep 19 07:23:37 UTC 2016

Copying files to 192.168.31.10:/tmp/vmdk_ops_unittest9139 and running .py unittests ...
Running unit tests in /tmp/vmdk_ops_unittest9139/cli_table_test.py...
......
Ran 6 tests in 0.000s

OK
Running unit tests in /tmp/vmdk_ops_unittest9139/vmdk_ops_test.py...
..........
Ran 10 tests in 338.340s

OK
This operation may take a while. Please be patient.
This operation may take a while. Please be patient.
Running unit tests in /tmp/vmdk_ops_unittest9139/vmdkops_admin_sanity_test.py...
F..
FAIL: test_ls (__main__.TestVmdkopsAdminSanity)
  File "/tmp/vmdk_ops_unittest9139/vmdkops_admin_sanity_test.py", line 41, in test_ls
    self.assertEqual(expected_column_count, len(divider_columns))
AssertionError: 10 != 11

Ran 3 tests in 1.787s

FAILED (failures=1)
Running unit tests in /tmp/vmdk_ops_unittest9139/vmdkops_admin_test.py...
.F..........................
FAIL: test_ls_no_args (__main__.TestLs)
  File "/tmp/vmdk_ops_unittest9139/vmdkops_admin_test.py", line 227, in test_ls_no_args
    self.assertEqual(expected_column_count, len(header))
AssertionError: 10 != 11

Ran 28 tests in 188.191s

FAILED (failures=1)
Running unit tests in /tmp/vmdk_ops_unittest9139/vol_test.py...
VOL_TEST kv_store skippped
Running unit tests in /tmp/vmdk_ops_unittest9139/vsan_info_test.py...
..
Ran 2 tests in 8.002s

OK
Running unit tests in /tmp/vmdk_ops_unittest9139/vsan_policy_test.py...
...ERROR:root:Failed to remove test_policy policy file
  File "/tmp/vmdk_ops_unittest9139/vsan_policy.py", line 209, in delete
    os.remove(policy_path(name))
OSError: [Errno 2] No such file or directory: '/vmfs/volumes/vsan:52fe282b8b5a398c-e41bd80f9e940d46/dockvols/policies/test_policy'
..
Ran 5 tests in 4.568s

OK

=> Build + Test not successful
make[2]: *** [test-esx] Error 2
make[1]: *** [test-esx] Error 2
make: *** [testremote] Error 2
=> Mon Sep 19 07:32:41 UTC 2016


=> Info ESX 192.168.31.10
=> Mon Sep 19 07:32:41 UTC 2016


VMkernel sabu-esx-01 6.0.0 #1 SMP Release build-3620759 Mar  3 2016 18:41:52 x86_64 x86_64 x86_64 ESXi

VMware ESXi 6.0.0 build-3620759
VMware ESXi 6.0.0 Update 2
```

brunotm:
@kerneltime 
I was running them as root.
It looks like something particular to my environment. I will better investigate later.

Thanks!

msterin:
`AssertionError: 10 != 11` is a test bug , since there is new column expected count needs to be adjusted....

@andrewjstone - please take a look, it is local mockesx test... `Error Trace:    cmd_test.go:40`

brunotm:
@msterin 

Those are different issues.

The mockesx test is failing in my local CI runs because due to some particularity of my setup, but that was also keeping me from tracking the test failure in the vmware CI runs.

This failure was introduced by the new "Filesystem Type" column in vmdkops_admin.py, and fixed in the last commit by adjusting the expected_column_count to the correct value.

msterin:
Yes, I realize these are different issues , sorry about not being clear in my message.  I am glad the column issue is fixed. btw, you should be able to run individual 'make test-esx' 'make test' and  'make test-vm'  targets.

I was asking @andrewstone to look at mockesx stuff, I *suspect* it may be due to difference in filesystem type (?) where the mock volumes are created.

brunotm:
It fails before any filesystem related code in my local setup due to permissions when setting up the loopback devices, but only when doing a CI run with drone. make testasroot works fine.

A quick research points me to lack of capabilities of the container running the test in my setup, but it will need further investigation with a stable docker version.

vmdk_plugin/plugin.go
```diff
@@ -208,12 +208,80 @@ func (d *vmdkDriver) unmountVolume(name string) error {
...
-	log.WithFields(log.Fields{"name": r.Name}).Info("Volume created ")
+
+	// Handle filesystem creation
+	log.WithFields(log.Fields{"name": r.Name,
+		"fstype": r.Options["fstype"]}).Info("Attaching volume and creating filesystem ")
+
+	dev, err := d.ops.Attach(r.Name, nil)
+	if err != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Attach volume failed ")
+		return volume.Response{Err: err.Error()}
```
pdhamdhere: Since this is an error in volume creation, should we delete volume/vmdk?

vmdk_plugin/plugin.go
```diff
@@ -208,12 +208,80 @@ func (d *vmdkDriver) unmountVolume(name string) error {
...
+	}
+
+	device, err := fs.GetDevicePath(dev)
+	if err != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Could not find attached device ")
+		return volume.Response{Err: err.Error()}
```
pdhamdhere: Same here. Since volume creation has failed, should volume be deleted?

vmdk_plugin/plugin.go
brunotm: Thank's @pdhamdhere 
Yes, i'll fix those.

pdhamdhere:
Thanks @brunotm Looks great!

PR #606, opened 2016-09-28
Increase the range of ports tried for binding

kerneltime:
If we have more than 24 operations performed in a tight loop we
run into the issue of ports not being cleaned up and returning
an error. This change increases the range of ports that we bind.

Minor go format change that the build picked up.

Testing:
1. Ran docker volume ls in a tight loop without issues
2. Ran code listed in https://github.com/vmware/docker-volume-vsphere/issues/436 without issues.


andrewjstone:
👍

govint:
Ship it.

I didn't do this thinking we aren't supposed to use ports from any arbitrary number less than 1023 :-). Looks good. Checked out the vmci and vsock drivers, the close() (release API in vsock driver) does release the binding rightaway. But looks like thats not even invoked when the socket is closed. REUSEADDR isn't supported at all, so no point setting it.

PR #613, opened 2016-10-06
Link VIB version to the build version.

kerneltime:
The vib version was not being set for each build or release.
As per documentation an additional release version is added.

```
Versioning VIBs

Because VIBs are intended to allow the ESXi installer to handle updates as well as initial installs, it is
important to use the versioning variables in the VIB correctly.

The package version can have a length of up to 35 characters. It includes the version number, immediately
followed by a hyphen (‐) followed by the release version, as in the following example:

5.0.0-1.0vmw

Both the version number and the release version number consist of one or more characters in the set
[a‐a‐Z0‐9], with periods separating two or more groups.
```

Testing:

Make vib and install.

Run command ```esxcli software profile get```

Output:
```
(Original Vendor):VMware, Inc.
      2016-07-06T10:45:12.595760+00:00: The following VIBs are
      installed:
        esx-vmdkops-service   0.7.14b2727-0.0.1
```

### DIFF
--- a/esx_service/Makefile
+++ b/esx_service/Makefile
@@ -26,8 +26,19 @@ PYLINT := /usr/bin/pylint
 # build places binaries here:
 BIN := ../build
 
-# VIB version is either set from outside, or includes git commit SHA
-PKG_VERSION ?= 0.1.TP-pre.$(shell git log --pretty=format:'%h' -n 1)
+# PKG_VERSION is either set externally as part of a build off a tag or
+# suffixed with a sha1 of the most recent commit to the prefix of the
+# most recent tag. Tagged builds use the externally defined version,
+# developer builds use last tagged release and sha1 of the most recent commit.
+# Format: <last tagged release>.<last commit hash> 
+PKG_VERSION ?= $(shell \
+	       git describe --tags `git rev-list --tags --max-count=1` \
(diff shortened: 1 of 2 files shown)


msterin:
What's the output of `/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status` ?

kerneltime:
```
[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
Version: 0.7.14b2727-0.0.1
Status: Running
Pid: 697796
Port: 1019
LogLevel: INFO
```

esx_service/Makefile
```diff
@@ -27,7 +27,14 @@ PYLINT := /usr/bin/pylint
 BIN := ../build
 
 # VIB version is either set from outside, or includes git commit SHA
-PKG_VERSION ?= 0.1.TP-pre.$(shell git log --pretty=format:'%h' -n 1)
+PKG_VERSION ?= $(shell \
```
msterin: nit: would be nice to add a resulting format (example) to comments

kerneltime: ok

andrewjstone:
LGTM

PR #617, opened 2016-10-08
Fix error handling when rolling back volume creation

### LINKED ISSUE #616: Plugin client must print an error message in case of create error
Recent changes to formatting of the disks on the guest side may have introduced this error. When create hits an error condition we get this,

root@photon-jyLxAJ3OK [ ~ ]# docker volume create -d vmdk --name v2
Error response from daemon: create v2: Post http://%2Frun%2Fdocker%2Fplugins%2Fvmdk.sock/VolumeDriver.Create: http: ContentLength=24 with Body length 0

brunotm:
Fixes #616

### DIFF
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -235,66 +235,61 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
 		return volume.Response{Err: msg + validfs}
 	}
 
-	err := d.ops.Create(r.Name, r.Options)
-	if err != nil {
-		log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Create volume failed ")
-		return volume.Response{Err: err.Error()}
+	errCreate := d.ops.Create(r.Name, r.Options)
+	if errCreate != nil {
+		log.WithFields(log.Fields{"name": r.Name, "error": errCreate}).Error("Create volume failed ")
+		return volume.Response{Err: errCreate.Error()}
 	}
 
 	// Handle filesystem creation
 	log.WithFields(log.Fields{"name": r.Name,
 		"fstype": r.Options["fstype"]}).Info("Attaching volume and creating filesystem ")
 
-	dev, err := d.ops.Attach(r.Name, nil)
-	if err != nil {
+	dev, errAttach := d.ops.Attach(r.Name, nil)
+	if errAttach != nil {
 		log.WithFields(log.Fields{"name": r.Name,
-			"error": err}).Error("Attach volume failed, removing the volume ")
-		err = d.ops.Remove(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Remove volume failed ")
-			return volume.Response{Err: err.Error()}
+			"error": errAttach}).Error("Attach volume failed, removing the volume ")
+		errRemove := d.ops.Remove(r.Name, nil)
+		if errRemove != nil {
+			log.WithFields(log.Fields{"name": r.Name, "error": errRemove}).Warning("Remove volume failed ")
 		}
-		return volume.Response{Err: err.Error()}
+		return volume.Response{Err: errAttach.Error()}
 	}
 
-	device, err := fs.GetDevicePath(dev)
-	if err != nil {
+	device, errGetDevicePath := fs.GetDevicePath(dev)
+	if errGetDevicePath != nil {
 		log.WithFields(log.Fields{"name": r.Name,
-			"error": err}).Error("Could not find attached device, removing the volume ")
-		err = d.ops.Detach(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Detach volume failed ")
-			return volume.Response{Err: err.Error()}
+			"error": errGetDevicePath}).Error("Could not find attached device, removing the volume ")
+		errDetach := d.ops.Detach(r.Name, nil)
+		if errDetach != nil {
+			log.WithFields(log.Fields{"name": r.Name, "error": errDetach}).Warning("Detach volume failed ")
 		}
-		err = d.ops.Remove(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Remove volume failed ")
-			return volume.Response{Err: err.Error()}
+		errRemove := d.ops.Remove(r.Name, nil)
+		if errRemove != nil {
+			log.WithFields(log.Fields{"name": r.Name, "error": errRemove}).Warning("Remove volume failed ")
 		}
-		return volume.Response{Err: err.Error()}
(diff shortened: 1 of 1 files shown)


vmdk_plugin/plugin.go
```diff
@@ -249,10 +249,10 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
...
 		log.WithFields(log.Fields{"name": r.Name,
 			"error": err}).Error("Attach volume failed, removing the volume ")
-		err = d.ops.Remove(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Remove volume failed ")
-			return volume.Response{Err: err.Error()}
+		errRm := d.ops.Remove(r.Name, nil)
+		if errRm != nil {
+			log.WithFields(log.Fields{"name": r.Name, "error": errRm}).Error("Remove volume failed ")
+			return volume.Response{Err: errRm.Error()}
```
govint: Its preferable that the error returned on a create error is the create error itself vs. a remove error (which is confusing to the user). OTOH, how did this cause the empty response from docker to the docker command line.

brunotm: >Its preferable that the error returned on a create error is the create error itself vs. a remove error (which is confusing to the user, when they issued a create command). 

Agree.

>OTOH, how did this cause the empty response from docker to the docker command line. 

err was being reassigned to nil inside the rollback block after the attach error, and at the end we would return the volume response with an error nil since we didn't got any errors from the rollback operations. Which if we had, in any case we would return from there.

Tests in plugin.go:

line 248   //dev, err := d.ops.Attach(r.Name, nil)
line 249   err = fmt.Errorf("STUB")
line 250   dev := []byte("an")

**Without the patch:**
root@vsc01:/tmp/docker-volume-vsphere# docker volume create -d vmdk --name vol$$ -o size=1gb
Error response from daemon: create vol25407: Post http://%2Frun%2Fdocker%2Fplugins%2Fvmdk.sock/VolumeDriver.Create: http: ContentLength=78 with Body length 0

**in log:**
2016-10-08 21:32:38.880136128 +0100 WEST [INFO] Attaching volume and creating filesystem fstype=ext4 name=vol25407
2016-10-08 21:32:38.880268068 +0100 WEST [ERROR] Attach volume failed, removing the volume error=STUB name=vol25407

**With the patch:**
root@vsc01:/tmp/docker-volume-vsphere# docker volume create -d vmdk --name vol$$ -o size=1gb
Error response from daemon: create vol25407: VolumeDriver.Create: STUB

**in log:**
2016-10-08 21:37:47.746007178 +0100 WEST [INFO] Attaching volume and creating filesystem name=vol25407 fstype=ext4
2016-10-08 21:37:47.746070732 +0100 WEST [ERROR] Attach volume failed, removing the volume name=vol25407 error=STUB

brunotm:
@govint
I have updated the commit to address your comment about returning from the main error.

vmdk_plugin/plugin.go
```diff
@@ -249,10 +249,9 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
 	if err != nil {
 		log.WithFields(log.Fields{"name": r.Name,
 			"error": err}).Error("Attach volume failed, removing the volume ")
-		err = d.ops.Remove(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Remove volume failed ")
-			return volume.Response{Err: err.Error()}
+		errRm := d.ops.Remove(r.Name, nil)
+		if errRm != nil {
+			log.WithFields(log.Fields{"name": r.Name, "error": errRm}).Error("Remove volume failed ")
```
msterin: Warning(). Failed to remove here is not necessarily a big issue I think.

vmdk_plugin/plugin.go
```diff
@@ -261,15 +260,13 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
 	if err != nil {
 		log.WithFields(log.Fields{"name": r.Name,
 			"error": err}).Error("Could not find attached device, removing the volume ")
-		err = d.ops.Detach(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Detach volume failed ")
-			return volume.Response{Err: err.Error()}
+		errDetach := d.ops.Detach(r.Name, nil)
```
msterin: should not it be errInGetPath , following the pattern of err<where> names ?

vmdk_plugin/plugin.go
```diff
@@ -278,15 +275,13 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
 	if err != nil {
 		log.WithFields(log.Fields{"name": r.Name,
 			"error": err}).Error("Create filesystem failed, removing the volume ")
-		err = d.ops.Detach(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Detach volume failed ")
-			return volume.Response{Err: err.Error()}
+		errDetach := d.ops.Detach(r.Name, nil)
```
msterin: errMkfs

vmdk_plugin/plugin.go
brunotm: And about other operations in rollback? Should we also log with warning? IMO those are subsequent errors.

brunotm:
@msterin 
I have updated the PR, it actually makes more sense to only log as Error the original issue.

vmdk_plugin/plugin.go
msterin: Yes. I think errors on rollback are generally warnings - like it's helpful to warn if something we expected to complete failed, but it does not impact success/failure of the actual operation requested.  We can silently ignore but I am generally not comfortable with silent failures.

vmdk_plugin/plugin.go
```diff
@@ -235,66 +235,61 @@ func (d *vmdkDriver) Create(r volume.Request) volume.Response {
...
-	if err != nil {
+	errMkfs := fs.Mkfs(mkfscmd, r.Name, device)
+	if errMkfs != nil {
 		log.WithFields(log.Fields{"name": r.Name,
-			"error": err}).Error("Create filesystem failed, removing the volume ")
-		err = d.ops.Detach(r.Name, nil)
-		if err != nil {
-			log.WithFields(log.Fields{"name": r.Name, "error": err}).Error("Detach volume failed ")
-			return volume.Response{Err: err.Error()}
+			"error": errMkfs}).Error("Create filesystem failed, removing the volume ")
```
msterin: looks like detach/remove has to be separate in a function instead of copy-paste.

brunotm:
@msterin 
If it would reutilized beyond these 2 cases, i wouldn't mind.
But this PR would be the place for it?

msterin:
@brunotm - for this PR it's not a problem, I am just generally sensitive to copy-n-paste :-). 
At any rate, since we have 2 approved I am merging the PR. Thanks !

PR #624, opened 2016-10-16
Fixed ACLs in payload bin dir [trivial]

msterin:
Now all in ./bin is executable. We sometimes failed the CI because of vmdkops_admin.py was not executable, so I used a hammer and made all in bin '+x'

Also allow to narrow down the list of tests to run on ESX.

Test - manually  run only '*admin*' tests (needed it for failure debugging), and CI in flight.


andrewjstone:
LGTM

PR #630, opened 2016-10-18
UI Plugin

marksoper:
First version of UI plugin coming into the project.


msterin:
@marksoper - the change should not come in a big bulk of original ESXUI code and deltas. 
a pure copy of esxui original code should go to vendor/ folder - at least that's what we do for GO code to simplify versions racking and updates. And the fixes should go in as a separate reviewable change with a few (desirably one) commit. Currently its very hard to understand what to review, and even harder to review it ...

Issue #629, opened 2016-10-17
Support older versions of e2fsprogs (or custom filesystem creation flags)?


brunotm:
Older versions of e2fsprogs (mkfs.ext2/3/4) asks for confirmation when creating a filesystem on a entire device and not a partition (eg. /dev/sdb) and make the filesystem creation to fail.

 This was due to a broken check which was removed in [06/2015](http://git.kernel.org/cgit/fs/ext2/e2fsprogs.git/commit/?id=97e9b970769b1afe462f82390cb5876cf037b903).

Should support for these versions be added?
If so, add them through logic in fs.Mkfs() or through a fs-config.json configuration file, allowing users to customize filesystem creation for any filesystem?

Eg.

```
{
    "ext4": {
        "flags": ["-F", "-m", "1"]
    },
    "btrfs": {
        "flags": ["-d", "single"]
    },
    "xfs": {
        "flags": ["-l", "lazy-count=1"]
    }
}
```

govint:
IOW if the docker volume create should include an option (along with the
rest that's supported today) of what FS to create the command to run and
the client simply runs that command. Allowing any FS to be used to format
the disk and for those apps that intend to use a raw disk a "no-format"
option as well.

brunotm:
@govint @msterin 

This is also needed for #612, unless there is no need to support linux distributions with older version of e2fsprogs. - we know that the VM2 used in the CI (ubuntu 14.04) has an older version.

msterin:
Resolved by #631

PR #631, opened 2016-10-18
Workaround older versions of e2fsprogs

brunotm:
Issue #629 

This commit workaround the error of creating ext/2/3/4 filesystems when using older versions of e2fsprogs. Those versions had an broken check and asks for confirmation when creating a filesystem using the whole block device (eg. /dev/sdb)

It also Broaden search path for filesystem utilities. Some distributions place filesystem utilities outside /sbin.

### DIFF
--- a/vmdk_plugin/utils/fs/fs.go
+++ b/vmdk_plugin/utils/fs/fs.go
@@ -36,6 +36,8 @@ const pciAddrLen = 10                     // Length of PCI dev addr
 
 // FstypeDefault contains the default FS when not specified by the user
 const FstypeDefault = "ext4"
+// Search paths for filesystem utilities
+var BinSearchPaths = [4]string {"/bin", "/sbin", "/usr/bin", "/usr/sbin"}
 
 // VolumeDevSpec - volume spec returned from the server on an attach
 type VolumeDevSpec struct {
@@ -72,10 +74,13 @@ func Mkfs(mkfscmd string, label string, device string) error {
 
 // MkfsLookup finds existent filesystem tools
 func MkfsLookup() map[string]string {
-	mkftools, _ := filepath.Glob("/sbin/mkfs.*")
 	supportedFs := make(map[string]string)
-	for _, mkfs := range mkftools {
-		supportedFs[strings.Split(mkfs, ".")[1]] = mkfs
+
+	for _,sp := range BinSearchPaths {
+		mkftools, _ := filepath.Glob(sp+"/mkfs.*")
+		for _, mkfs := range mkftools {
+			supportedFs[strings.Split(mkfs, ".")[1]] = mkfs
+		}
 	}
 	return supportedFs
 }
--- a/vmdk_plugin/utils/fs/fs.go
+++ b/vmdk_plugin/utils/fs/fs.go
@@ -36,8 +36,9 @@ const pciAddrLen = 10                     // Length of PCI dev addr
 
 // FstypeDefault contains the default FS when not specified by the user
 const FstypeDefault = "ext4"
-// Search paths for filesystem utilities
-var BinSearchPaths = [4]string {"/bin", "/sbin", "/usr/bin", "/usr/sbin"}
+
+// BinSearchPath contains search paths for host binaries
+var BinSearchPath = []string{"/bin", "/sbin", "/usr/bin", "/usr/sbin"}
 
 // VolumeDevSpec - volume spec returned from the server on an attach
 type VolumeDevSpec struct {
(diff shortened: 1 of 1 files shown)


brunotm:
@kerneltime it is possible to have access to the CI logs? Once more i don't get any errors in my tests.
It's probably its a hassle for you to keep providing the logs, and makes much harder for me to track any issue.

Thanks!

msterin:
@brunotm - this is CI issue, looks like password expiration. @kerneltime needs to take care of this as soon as sun rises in Palo Alto :) 

The log says: 
```
dpkg-query: no packages found matching docker-volume-vsphere
WARNING: Your password has expired.
Password change required but no TTY available.
WARNING: Your password has expired.
Password change required but no TTY available.
```

an all fails at this point.

vmdk_plugin/utils/fs/fs.go
```diff
@@ -37,6 +37,9 @@ const pciAddrLen = 10                     // Length of PCI dev addr
 // FstypeDefault contains the default FS when not specified by the user
 const FstypeDefault = "ext4"
 
+// BinSearchPaths contains search paths for host binaries
+var BinSearchPaths = [4]string{"/bin", "/sbin", "/usr/bin", "/usr/sbin"}
```
govint: This may not be needed here, why not have a "default" search path in the default config used by the plugin and that can be used as the search path for all external execs that the plugin runs. The plugin behavior can be changed any time.

vmdk_plugin/utils/fs/fs.go
```diff
@@ -62,7 +65,14 @@ func Mkdir(path string) error {
 
 // Mkfs creates a filesystem at the specified device
 func Mkfs(mkfscmd string, label string, device string) error {
-	out, err := exec.Command(mkfscmd, "-L", label, device).CombinedOutput()
+	// Workaround older versions of e2fsprogs, issue 629.
+	var err error
+	var out []byte
+	if strings.Split(mkfscmd, ".")[1][0:3] == "ext" {
```
msterin: can you add a comment with info on what pattern and where you are looking for ?

msterin:
Looks good. One request  for more comments , and it's good to go after CI is fixed and passes.

vmdk_plugin/utils/fs/fs.go
govint: Even if we want to have a list here maybe a slice can be used vs. an array - []string{"/bin", "/sbin", "/usr/bin", "/usr/sbin"}

vmdk_plugin/utils/fs/fs.go
```diff
@@ -62,7 +65,14 @@ func Mkdir(path string) error {
...
+		out, err = exec.Command(mkfscmd, "-F", "-L", label, device).CombinedOutput()
+	} else {
+		out, err = exec.Command(mkfscmd, "-L", label, device).CombinedOutput()
+	}
```
govint: Can we use "-F -L" always?

vmdk_plugin/utils/fs/fs.go
brunotm: >This may not be needed here, why not have a "default" search path in the default config used by the plugin and that can be used as the search path for all external executables that the plugin runs. The plugin behavior can be changed any time.

I do think that is probably better to have it in the config file.

Please correct me if i am wrong, but we hold no reference from the config instance created in main.go (where the configfile can be other than the default), and from fs.go we would only be able to load it with config.DefaultConfigPath, this file can potentially be different from the user specified - If this is the case, IMO the required changes are beyond the scope of this PR.

>Even if we want to have a list here maybe a slice can be used vs. an array - []string{"/bin", "/sbin", "/usr/bin", "/usr/sbin"}.

It's not really needed, but its not a problem :)

vmdk_plugin/utils/fs/fs.go
brunotm: Sure.

vmdk_plugin/utils/fs/fs.go
brunotm: >Please correct me if i am wrong

Or overcomplicating the issue! :)

msterin: it is nice to have it extensible but I think it's super low priority ,and to me it does not feel like a config item

brunotm: Agree. But BinSearchPath also looks misplaced in fs.go.
So we keep it there for now?

msterin: yes - Mkfs related stuff is ok in fs.go IMO

brunotm:
Rebased the commit to address the comments.
Thanks!

Issue #638, opened 2016-10-20
VMDK server plugin doesn't return FS type and causes container run error


govint:
This is seen on the current TOT code. The new changes to return fstype on the GET call is causing an issue when there is no fstype in the volume opts. In vmdk_ops.py vol_info() there is a new check added to see if fstype is there in the volume opts and the returned status (on GET) has the fs type only if its there, Otherwise no.

Client accesses status["fstype"] and its not there and seems to hang up. Docker daemon fails as below:
  ---
$ docker run --rm -it -w /data -v dvol_2:/data busybox
docker: Error response from daemon: Post http://%2Frun%2Fdocker%2Fplugins%2Fvmdk.sock/VolumeDriver.Mount: http: ContentLength=90 with Body length 0.
  ---

And the client plugin hangs here,
 ---
2016-10-20 03:30:14.523188855 -0700 PDT [INFO] No config file found. Using defaults.
2016-10-20 03:30:14.523621362 -0700 PDT [INFO] Docker VMDK plugin started version="VMDK Volume Driver v0.3" port=1019 mock_esx=false log_level=debug config="/etc/docker-volume-vsphere.conf"
2016-10-20 03:30:14.523838499 -0700 PDT [INFO] Getting volume data from unix:///var/run/docker.sock
2016-10-20 03:30:14.629758124 -0700 PDT [DEBUG] Docker info: version=1.12.2, root=/var/lib/docker, OS=Ubuntu 14.04.1 LTS
2016-10-20 03:30:14.630803269 -0700 PDT [DEBUG] Found 0 running or paused containers
2016-10-20 03:30:14.631533162 -0700 PDT [INFO] Discovered 0 volumes in use.
2016-10-20 03:30:14.631717317 -0700 PDT [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vmdk.sock"
2016-10-20 03:30:14.632257413 -0700 PDT [DEBUG] root group found. gid: 0
2016-10-20 03:32:08.561885931 -0700 PDT [DEBUG] vmdkOps.Get name=dvol_2
2016-10-20 03:32:09.73342527 -0700 PDT [INFO] Mounting volume name="dvol_2"
  ---

govint:
Any changes to vol_info() must always follow the approach: that if the option isn't in vol opts then fill in a default.

brunotm:
How the volume was successfully created without a fstype?

govint:
[...]

PR #639, opened 2016-10-20
Handle invalid FS type on mount.

govint:
Handle invalid FS type returned by a GET call to the plugin server and fail back to using ext2 as FS type in such a case. This handles both any corruption on the server side plus makes a best effort to mount a volume with a default FS type.

### DIFF
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -366,7 +366,15 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 		isReadOnly = true
 	}
 
-	mountpoint, err := d.mountVolume(r.Name, status["fstype"].(string), isReadOnly)
+	fstype, exists := status["fstype"].(string)
+
+	if !exists {
+		msg := fmt.Sprintf("Got invalid filesystem type for %s, attempting mount with type ext2.", r.Name)
+		log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
+		// Fail back to a default version that we can try with.
+		fstype = "ext2"
+	}
+	mountpoint, err := d.mountVolume(r.Name, fstype, isReadOnly)
 	if err != nil {
 		log.WithFields(
 			log.Fields{"name": r.Name, "error": err.Error()},


govint:
Fixes #638

esx_service/volume_kv.py
```diff
@@ -82,7 +82,7 @@
 # Filesystem type
 # This option is handled in the volume-plugin at the docker host, and tracked in volume metadata. 
 FILESYSTEM_TYPE = 'fstype'
-DEFAULT_FILESYSTEM_TYPE = ''
+DEFAULT_FILESYSTEM_TYPE = 'ext4'
```
pdhamdhere: Aren't old volumes formatted with ext2?

vmdk_plugin/plugin.go
```diff
@@ -361,6 +365,14 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 		isReadOnly = true
 	}
 
+	_, exists := status["fstype"]
+
+	if !exists {
+		log.WithFields(
+			log.Fields{"name": r.Name, "error": "Invalid filesystem type."},
+		).Error("Failed to mount ")
+		return volume.Response{Err: "Failed mount - invalid filesystem type."}
```
pdhamdhere: minor edit: "Failed to mount - Invalid filesystem type"

vmdk_plugin/plugin.go
```diff
@@ -84,14 +84,18 @@ func getMountPoint(volName string) string {
 
 // Get info about a single volume
 func (d *vmdkDriver) Get(r volume.Request) volume.Response {
+	log.Debug("calling GET ")
```
msterin: I think this debug should be one layer above, when misc. methods are called.

vmdk_plugin/plugin.go
```diff
@@ -84,14 +84,18 @@ func getMountPoint(volName string) string {
...
 	status, err := d.ops.Get(r.Name)
+	log.Debug("returned GET ")
```
msterin: ditto

vmdk_plugin/plugin.go
```diff
@@ -361,6 +365,14 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 		isReadOnly = true
 	}
 
+	_, exists := status["fstype"]
+
+	if !exists {
+		log.WithFields(
+			log.Fields{"name": r.Name, "error": "Invalid filesystem type."},
```
msterin: can you use a var for message instead of copy-n-paste here and on line 374 ?  Also I believe the text ("failed mount - invalid filesystem type" ) should be the same in both cases, and ideally include some hint on what to do next. Though I have no idea what to do next ...

govint:
Updated changes, corrected error message as below:
  ---
$ docker run --rm -it -w /data -v dvol_2:/data busybox
docker: Error response from daemon: VolumeDriver.Mount: Mount error - got invalid filesystem type for dvol_2.
See 'docker run --help'.

And logs,
2016-10-23 21:14:58.662195329 -0700 PDT [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vmdk.sock"
2016-10-23 22:08:20.16849929 -0700 PDT [INFO] Mounting volume name="dvol_2"
2016-10-23 22:08:20.183811502 -0700 PDT [ERROR] name="dvol_2" error="Mount error - got invalid filesystem type for dvol_2"

govint:
With old volumes the client plugin just appears hung (no response causing
docker "run" to error, plugin responds to new commands but the volume used
for the mount is left with a refcount thats not going away till the client
is restarted. The behavior is documented in the description for the issue.

Yes, ext2 is the default.

PR #640, opened 2016-10-21
Cli auth.liping

lipingxue:
This change includes the code change to enable AdminCLI command to support tenants.
See #620

PR #645, opened 2016-10-25
AdminCLI change to support multitenancy

lipingxue:
This is a follow up review for the following PR. 
https://github.com/vmware/docker-volume-vsphere/pull/640

#640 was closed by a merge which I did by accident. 

@msterin Mark has reviewed #640 and gave some comments. I have addressed all his comments and create this new PR for review.

For the unit test, please see the original #640.


esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+    if error_info:
+        return operation_fail(error_info)
+      
+    name = args.name 
+    description = "" 
+    default_datastore ="default_ds" 
+    default_privileges =  {}
+    privileges = []
+
+    auth.connect_auth_db()
```
msterin: it would be good to have all connect_auth_db() wrapped the same way get_tenant_from_db() is wrapping it - this will place all connection mgmt to the same software layer.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
     except:
         return NOT_AVAILABLE
 
+def get_tenant_from_db(name):
```
lipingxue: Done.

esx_service/cli/vmdkops_admin.py
lipingxue: Done

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+        return operation_fail(error_info)
+      
+    name = args.name 
+    description = "" 
+    default_datastore ="default_ds" 
+    default_privileges =  {}
+    privileges = []
+
+    auth.connect_auth_db()
+    error_info, tenant = auth._auth_mgr.create_tenant(name, description, default_datastore, default_privileges, vms, privileges)
```
lipingxue: Done

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+            remove_volumes = True
+
+    error_info = auth._auth_mgr.remove_tenant(tenant.id, remove_volumes)
+    if error_info:
+        return operation_fail(error_info)
 
 def tenant_ls(args):
-    print(NOT_IMPLEMENTED)
+    """ Handle tenant ls command """
+    header = tenant_ls_headers()
```
lipingxue: Done.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+    if args.volume_totalsize:
+        size_in_MB = convert.convert_to_MB(args.volume_totalsize)
+        privileges[auth_data_const.COL_USAGE_QUOTA] = size_in_MB
+    
+    return privileges
 
 def tenant_access_add(args):
-    print(NOT_IMPLEMENTED)
+    """ Handle tenant access command """
+    # TODO in the current privileges table, no volome_maxcount field
```
lipingxue: Drop it.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+    privileges_dict[auth_data_const.COL_MOUNT_VOLUME] = privileges[5]
+    privileges_dict[auth_data_const.COL_MAX_VOLUME_SIZE] = privileges[6]
+    privileges_dict[auth_data_const.COL_USAGE_QUOTA] = privileges[7]
+    return privileges_dict
 
 def tenant_access_set(args):
-    print(NOT_IMPLEMENTED)
-
+    """ Handle tenant access set command """
+    # TODO in the current privileges table, no volome_maxcount field
```
lipingxue: Drop it.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+
+def generate_tenant_ls_rows(tenant_list):
+    """ Generate output for tenant ls command """
+    rows = []
+    for tenant in tenant_list:
+        uuid = tenant.id
+        name = tenant.name
+        description = tenant.description
+        default_datastore = tenant.default_datastore
+        vm_list = generate_vm_list(tenant.vms)
```
pdhamdhere: Number of VMs in a tenant could be really large and obfuscate the `tenant ls` output on screen. May be introduce `tenant describe` which lists everything related to tenant including list of VMs.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+        return operation_fail(error_info)
 
-def tenant_vm_add(args):
-    print(NOT_IMPLEMENTED)
+    rows = generate_tenant_ls_rows(tenant_list)
+    print(cli_table.create(header, rows)) 
 
+def tenant_vm_add(args):
+    """ Handle tenant vm add command """
+    error_info, vms = generate_tuple_from_vm_list(args.vm_list)
```
pdhamdhere: 'tenant' validation should be first thing before doing any work.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+    
+    error_info = tenant.add_vms(auth._auth_mgr.conn, vms)
+
+    if error_info:
+        return operation_fail(error_info)
+  
 def tenant_vm_rm(args):
-    print(NOT_IMPLEMENTED)
+    """ Handle tenant vm rm command """
+    error_info, vms = generate_tuple_from_vm_list(args.vm_list)
```
pdhamdhere: Same here. Please validate if tenant parameter is valid before proceeding.

esx_service/cli/vmdkops_admin.py
lipingxue: I will file a new issue to track this enhancement.

esx_service/cli/vmdkops_admin.py
lipingxue: tenant validation is done inside get_tenant_from_db. This function will do
1. connect the auth DB
2. get tenant with the given name from DB
If any error happened in above steps, an error will be return

If no errors when connect to the DB and reading from DB, but no tenant is found with the given name, then "tenant" will be None, 
and the following code will check whether the tenant exist or not.

esx_service/cli/vmdkops_admin.py
lipingxue: Same logic as above.

esx_service/cli/vmdkops_admin.py
pdhamdhere: Yes, I am suggesting get_tenant_from_db() should be before generate_tuple_from_vm_list().

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+        return operation_fail(error_info)
+ 
 def tenant_rm(args):
-    print(NOT_IMPLEMENTED)
+    """ Handle tenant rm command """
+    error_info, tenant = get_tenant_from_db(args.name)
+    if error_info:
+        return operation_fail(error_info)
+
+    remove_volumes = False
```
lipingxue: Done.

esx_service/cli/vmdkops_admin.py
```diff
@@ -723,38 +731,341 @@ def get_version():
...
+    if args.remove_volumes:
+        if args.remove_volumes == 'True':
+            remove_volumes = True
+
+    error_info = auth._auth_mgr.remove_tenant(tenant.id, remove_volumes)
+    if error_info:
+        return operation_fail(error_info)
```
lipingxue: Done

esx_service/cli/vmdkops_admin.py
lipingxue: Done.

esx_service/cli/vmdkops_admin.py
lipingxue: Done.

lipingxue:
Addressed comments from Mark and Prashant, please take another look.

esx_service/cli/vmdkops_admin.py
lipingxue: New PR is filed to track this. https://github.com/vmware/docker-volume-vsphere/issues/651

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
 
-NOT_IMPLEMENTED = "Not implemented"
+def generate_tuple_from_vm_list(vm_list):
+    """ Generate a list of (vm_uuid, vm_name) pair """
+    vms = []
+    for vm_name in vm_list:
+        vm_uuid = vmdk_utils.get_vm_uuid_by_name(vm_name)
+        if not vm_uuid:
+            error_info =  "Cannot find vm_uuid for vm {0}, tenant_create failed".format(vm_name)
+            return error_info, None
```
pdhamdhere: Feel free to ignore. This loop gives up on first error (invalid VM name). Should we iterate through list and populate 2 lists- one with valid VM names and another with invalid names? So, tenant creation can proceed with VMs with correct name? Or User in one stroke knows which VM names are incorrect. Ideally regular expression would be nice!

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+                                             name = name, 
+                                             description = description, 
+                                             default_datastore = default_datastore, 
+                                             default_privileges = default_privileges, 
+                                             vms = vms, 
+                                             privileges = privileges)
+    if error_info:
+        return operation_fail(error_info)
+    else:
+        print "tenant create succeeded"
```
pdhamdhere: Include tenant name => "Tenant <name> created successfully". ?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    # args.remove_volumes will be None
+    if args.remove_volumes:
+        if args.remove_volumes == 'True':
+            remove_volumes = True
+
+    error_info = auth._auth_mgr.remove_tenant(tenant.id, remove_volumes)
+    if error_info:
+        return operation_fail(error_info)
+    else:
+        print "tenant rm succeeded"
```
pdhamdhere: How about "Tenant <name> removed successfully"?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    error_info, vms = generate_tuple_from_vm_list(args.vm_list)
+    if error_info:
+        return operation_fail(error_info)
+ 
+    error_info = tenant.add_vms(auth._auth_mgr.conn, vms)
+
+    if error_info:
+        return operation_fail(error_info)
+    else:
+        print "tenant vm add succeeded"
```
pdhamdhere: "Added <VM names> to Tenant <tenantName> successfully"?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    
+    if error_info:
+        return operation_fail(error_info)
+      
+    error_info = tenant.remove_vms(auth._auth_mgr.conn, vms)
+
+    if error_info:
+        return operation_fail(error_info)
+    else:
+        print "tenant vm rm succeeded"
```
pdhamdhere: "Removed <VM names> from <Tenant Name> successfully"?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+        return operation_fail(error_info)
+
+    privileges = generate_privileges(args)
+    print privileges
+    error_info = tenant.set_datastore_access_privileges(auth._auth_mgr.conn, [privileges])
+      
+    if error_info:
+        return operation_fail(error_info)
+    else:
+        print "tenant access add succeeded"
```
pdhamdhere: "Datastore access privileges for <TenantName> updated successfully"?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    privileges = [d for d in tenant.privileges if d[auth_data_const.COL_DATASTORE] == args.datastore]
+    
+    if not privileges:
+        error_info = "No privileges exist for ({0}, {1})".format(args.name, args.datastore)
+        return operation_fail(error_info)
+    
+    privileges_dict = generate_privileges_dict(privileges[0])
+    privileges_dict = modify_privileges(privileges_dict, args)
+    
+    error_info = tenant.set_datastore_access_privileges(auth._auth_mgr.conn, [privileges_dict])
```
pdhamdhere: I assume Read-After-Write allows user to selectively modify subset of permissions and doesn't overwrite existing privileges.

esx_service/utils/auth_data.py
```diff
@@ -493,6 +513,62 @@ def create_tenant(self, name, description, default_datastore, default_privileges
 
         return None, tenant
     
+    def get_tenant(self, tenant_name):
```
pdhamdhere: Tenant_name is not unique right? How do you handle 2 tenants with same name?

lipingxue:
There are some cleanup work needed, PR https://github.com/vmware/docker-volume-vsphere/issues/653 is filed to track that.

msterin:
Will merge as soon as CI is back and passes

esx_service/cli/vmdkops_admin.py
```diff
@@ -248,13 +253,14 @@ def commands():
...
                         },
+
                         'ls': {
                             'help': "list VMs in a tenant",
                             'func': tenant_vm_ls,
                             'args': {
                                 '--name': {
                                     'help': "Tenant to list the VMs for",
-                                    'required': False
+                                    'required': True
```
govint: Can't this remain false and hence allow users to use a single command to list VMs in all tenants vs doing a) tenant ls and then b) for <all tenants>; do tenant vm ls --name <tenant-name>; done?

esx_service/cli/vmdkops_admin.py
```diff
@@ -123,7 +127,7 @@ def commands():
                     'metavar': 'Col1,Col2,...'
                 },
                 '--tenant' : {
-                    'help': 'Displays VMs for a given tenant'
+                    'help': 'Displays volumes for a given tenant'
```
msterin: unfortunately -c option does not filter rows, it simple shows less columns, so there is currently no way to scope the list to a specific tenant (other that `| grep`)

esx_service/cli/vmdkops_admin.py
msterin: information on VMs and tenants is different. If you have some specific format in mind, please give example on how it would look like - FWIW I can't visualize your suggestion

govint: Not giving the name filter should allow listing all VMs in all tenants similar to what "ls" would do without "-c" or "--tenant" options.

esx_service/cli/vmdkops_admin.py
```diff
@@ -284,10 +290,6 @@ def commands():
                                     'help': 'Maximum size of the volume that can be created',
                                     'metavar': 'Num{MB,GB,TB} - e.g. 2TB'
                                 },
-                                '--volume-maxcount': {
-                                    'help': 
-                                    'Maximum number of volumes to create on the datastore for this tenant'
-                                },
                                 '--volume-totalsize': {
```
govint: Nit, not your change, but should this be like --max_limit as its a resource (capacity) entitlement for the tenant on the datastore and also --max_volume_size?

esx_service/cli/vmdkops_admin.py
```diff
@@ -453,20 +455,23 @@ def ls(args):
...
+   
     print(cli_table.create(header, rows))
 
 
-def ls_dash_c(columns):
+def ls_dash_c(columns, tenant_reg):
     """ Return only the columns requested in the format required for table construction """
     all_headers = all_ls_headers()
-    all_rows = generate_ls_rows()
+    all_rows = generate_ls_rows(tenant_reg)
```
govint: Nit, if the tenant doesn't exist or no volumes created via the tenant then do we want to print even the header or a message to the effect of no volumes found. Else the user won't know whether the command is not finding the volumes or there really aren't any volumes.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
     except:
         return NOT_AVAILABLE
 
+def get_tenant_from_db(name):
+    try:
+        auth.connect_auth_db()
+    except auth_data.DbConnectionError, e:
+        error_info = "Failed to connect auth DB({0})".format(e)
+        return error_info, None
```
govint: Can this segment be a function like connect_db() vs. repeated in each function.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+
+def get_tenant_list_from_db():
+    try: 
+        auth.connect_auth_db()
+        error_info, tenant_list = auth._auth_mgr.list_tenants()
+    except auth_data.DbConnectionError, e:
+        error_info = "Failed to connect auth DB({0})".format(e)
+        return error_info, None
+
+    error_info, tenant_list = auth._auth_mgr.list_tenants()
```
govint: This line is repeated in the try-except block above at line 762

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    error_info, tenant = auth._auth_mgr.get_tenant(name)
+    return error_info, tenant
+
+def create_tenant_in_db(name, description, default_datastore, default_privileges, vms, privileges):
+    try:
+        auth.connect_auth_db()
+    except auth_data.DbConnectionError, e:
+        error_info = "Failed to connect auth DB({0})".format(e)
+        return error_info, None
+
```
govint: Should have a validate_tenant_properties() or similar function to validate that the provided props for the tenant are .... valid. Like can the description be a large block of text or the privileges are actually valid and supported ones and that the datastore actually exists. Even name should be of some max. width allowed by the DB.

Perhaps caller must check for validity of datastore and privileges perhaps, just mentioning.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    error_info, tenant_list = auth._auth_mgr.list_tenants()
+    return error_info, tenant_list
+
+def operation_fail(error_info):
+    print error_info
+    return error_info
+
+def tenant_ls_headers():
+    """ Return column names for tenant ls command """
+    headers = ['Uuid', 'Name', 'Description', 'Default_datastore', 'VM_list']
```
govint: Default datastore can't be a column header, its one of several datastores that the tenant has entitlements to. Instead list all datastores that the tenant has access to and for the default one mark it as such in its entry.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    return headers
+
+def generate_vm_list(vms):
+    """ Generate a comma separated string with given list """
```
govint: Please note what is the format of the list.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    res = ""
+    for vm in vms:
+        # vm[1] is vm_name, vm has format (vm_uuid, vm_name)
+        res = res + vm[1]
+        res = res + ","
```
govint: This can be res = res + vm[1] + ',' instead

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+        
+    if res:
+        res = res[:-1]
+
+    return res
```
govint: res can be an empty string and caller must take care or print some thing to the effect that there aren't any VMs.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    return res    
+
+def generate_tenant_ls_rows(tenant_list):
+    """ Generate output for tenant ls command """
+    rows = []
+    for tenant in tenant_list:
+        uuid = tenant.id
+        name = tenant.name
+        description = tenant.description
+        default_datastore = tenant.default_datastore
```
govint: There can be several datastores for a tenant and those should be displayed as a list vs. just the default data store.

Or for each tenant ignore the VM list (which can be long) and default datastore and only list name, description and UUID, everythiong else will be provided via a detailed view of a tenant.

Or for each tenant mention name, description, UUID, default datastore, default privileges.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+            return error_info, None
+        vms.append((vm_uuid, vm_name))
+    
+    return None, vms
 
 def tenant_create(args):
-    print(NOT_IMPLEMENTED)
-
+    """ Handle tenant create command """
+    error_info, vms = generate_tuple_from_vm_list(args.vm_list)
```
govint: Suggest not persisting VM name any where in the DB. As #652 shows, a VM name can change and the user can get confused if old VM names are popping up in admin CLI commands. The UUID is unique and matches an instance of an existing VM.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    if error_info:
+        return operation_fail(error_info)
+      
+    name = args.name 
+    description = ""
```
govint: Shouldn't this be part of the create command args.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    if 'create' in rights:
+        privileges[auth_data_const.COL_CREATE_VOLUME] = value
+
+    if 'delete' in rights:
+        privileges[auth_data_const.COL_DELETE_VOLUME] = value
+
+    if 'mount' in rights:
+        privileges[auth_data_const.COL_MOUNT_VOLUME] = value
+
+    if 'all' in rights:
```
govint: Instead check for 'all' privileges first and set the privileges and exit the function vs. checking each privilege and finally setting all privileges.

govint: Do we expect that there won't be separate settings for each privilege?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    if args.rights:
+        set_privileges(args.rights, privileges, 1)
+    
+    if args.volume_maxsize:
+        size_in_MB = convert.convert_to_MB(args.volume_maxsize)
+        privileges[auth_data_const.COL_MAX_VOLUME_SIZE] = size_in_MB
+    
+    if args.volume_totalsize:
+        size_in_MB = convert.convert_to_MB(args.volume_totalsize)
+        privileges[auth_data_const.COL_USAGE_QUOTA] = size_in_MB
```
govint: There should be validation that the volume max size and max capacity across volumes isn't specified beyond the capacity of the datastore itself. Or larger or smaller than VSphere limits. Possibly any value larger than the datastore capacity is simply capped at the datastore capacity.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    if error_info:
+        return operation_fail(error_info)
+
+    if not tenant:
+        error_info = "Tenant {0} does not exist".format(args.name)
+        return operation_fail(error_info)
+
+    privileges = generate_privileges(args)
+    print privileges
+    error_info = tenant.set_datastore_access_privileges(auth._auth_mgr.conn, [privileges])
```
govint: Documentation must say that datastores can always be over provisioned among tenants as there are no checks performed when a new tenant is added with access on a datastore.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+        size_in_MB = convert.convert_to_MB(args.volume_maxsize)
+        privileges[auth_data_const.COL_MAX_VOLUME_SIZE] = size_in_MB
+    
+    if args.volume_totalsize:
+        size_in_MB = convert.convert_to_MB(args.volume_totalsize)
+        privileges[auth_data_const.COL_USAGE_QUOTA] = size_in_MB
+    
+    return privileges
+
+def generate_privileges_dict(privileges):
```
govint: This function should be in the tenant code as it understands the format of data read from the DB and can return a dict vs. this code trying to interpret the list returned from the DB. If the order of values is changed as returned from the DB then this code needs to change.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+        error_info = "Tenant {0} does not exist".format(args.name)
+        return operation_fail(error_info)
+
+    privileges = [d for d in tenant.privileges if d[auth_data_const.COL_DATASTORE] == args.datastore]
+    
+    if not privileges:
+        error_info = "No privileges exist for ({0}, {1})".format(args.name, args.datastore)
+        return operation_fail(error_info)
+    
+    privileges_dict = generate_privileges_dict(privileges[0])
```
govint: Shouldn't this be just privileges?

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    
+    return privileges
+
+def generate_privileges_dict(privileges):
+    # privileges is a list which is read from auth DB
+    # it has the following format
+    # (tenant_uuid, datastore, global_visibility, create_volume, delete_volume,
+    # mount_volume, max_volume_size, usage_quota)
+    privileges_dict = {}
+    privileges_dict[auth_data_const.COL_DATASTORE] = privileges[1]
```
govint: shouldn't this start at index 0?

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -74,21 +74,115 @@ def test_policy_ls(self):
...
+    def test_tenant_vm_add(self):
+        args = self.parser.parse_args('tenant vm add --name=tenant1 --vm-list vm1,vm2'.split())
+        self.assertEqual(args.func, vmdkops_admin.tenant_vm_add)
+        self.assertEqual(args.name, 'tenant1')
+        self.assertEqual(args.vm_list, ['vm1', 'vm2'])
+    
+    def test_tenant_vm_add_missing_option_fails(self):
+        self.assert_parse_error('tenant vm add')
+        self.assert_parse_error('tenant vm add --name=tenant1')
+
```
govint: Could include VM add with non-existent tenant, non-existent Vm

esx_service/utils/auth_data.py
```diff
@@ -572,30 +648,31 @@ def _remove_volumes_for_tenant(self, tenant_id):
...
-        # Delete path /vmfs/volumes/datastore_name/tenant_name
-        logging.debug("Deleting dir path%s", dir_path)
-        try:
-            os.rmdir(dir_path)
-        except os.error as e:
-            logging.error("remove dir %s failed with error %s", dir_path, e)
-            return str(e)
+                    logging.error("remove vmdk %s failed with error %s", vmdk_path, err)
+                    if err:
+                        return err
```
govint: Instead, remove as many of the volumes but keep track of any remove errors and at the end of the loop figure if there was an error in removing any of the volumes and then return error after logging.

esx_service/cli/vmdkops_admin.py
```diff
@@ -726,38 +731,382 @@ def get_version():
...
+    """ Handle tenant vm add command """
+    error_info, tenant = get_tenant_from_db(args.name)
+    if error_info:
+        return operation_fail(error_info)
+    
+    if not tenant:
+        error_info = "Tenant {0} does not exist".format(args.name)
+        return operation_fail(error_info)
+
+    error_info, vms = generate_tuple_from_vm_list(args.vm_list)
```
msterin: Will be addressed in #653

esx_service/cli/vmdkops_admin.py
msterin: how does it matter ? 'all' is exactly the same choice as (say) 'mount', so it is not clear what is changed if the suggestion is implemented

esx_service/cli/vmdkops_admin.py
msterin: That's a good point. Can you add this to #653 ?

esx_service/cli/vmdkops_admin.py
msterin: This is by default a "thin" provisioning behavior - I am not sure why we need to reiterate it again

esx_service/cli/vmdkops_admin_test.py
msterin: can you elaborate on what do you mean here?

esx_service/utils/auth_data.py
msterin: good catch. 
Besides, a single volume remove fail should not prevent other volumes from being removed, so an easier way to handle this would be to print the message and keep going, without  any other change to the code

PR #658, opened 2016-10-27
Include the ssh keys to be used in container.

kerneltime:
This the updated final fix. SSH key management is in line with what drone recommends.
SSH key to be used is part of .drone.sec and decrypted and used during build.


esx_service/Makefile
```diff
@@ -152,8 +152,8 @@ test: test-esx build
 
 # test-esx is a quick unittest for ESX Python code.
 # Deploys, runs and clean unittests (i.e. all files *_test.py) on ESX
-SSH := $(DEBUG) ssh -kTax -o StrictHostKeyChecking=no
-SCP := $(DEBUG) scp -r -q -o StrictHostKeyChecking=no
+SSH := $(DEBUG) ssh $(SSH_KEY_OPT) -kTax -o StrictHostKeyChecking=no
```
kerneltime: line 48 in drone.yml

PR #667, opened 2016-10-31
VMODL draft

msterin:
This is preliminary discussion of management interface. Please do not review/comment unless you are explicitly on //CC line 
//CC @shaominchen @pdhamdhere


msterin:
opened too early

Issue #568, opened 2016-08-03
Confusing error message from volume.create when datastore is not writable


msterin:
```
root@ubuntu-srv14:~# docker volume create -d vmdk --name=my-volume-201@pa-dbc1113
Error response from daemon: create my-volume-201@pa-dbc1113: VolumeDriver.Create: Failed to initialize volume path None
```

It actually means that pa-dbc1113 is not writable and ./dockvol creation faild

PR #677, opened 2016-11-02

BaluDontu:
This fix for #568 will ensure a meaningful error message is given to the user when a volume is created on a read only datastore. 

Test steps:
1. Used the esxcli commands to create a read only NFS datastore.
esxcli storage nfs add --host=10.162.210.32 --share=/dbc/pa-dbc1132/baominw --volume-name=nfsstore-baominw --readonly

2. Tried to create a volume using the following command
docker volume create -d vmdk --name=test-201@nfsstore-baominw

Now the error response will be a meaningful message representing that a volume cannot be created on the datastore because it is readonly. For example,
 "nfsstore-baominw is not writable and hence test-201 volume creation failed"


esx_service/vmdk_ops.py
```diff
@@ -583,7 +583,7 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
     path = get_vol_path(datastore, tenant_name)
     logging.debug("executeRequest %s %s", tenant_name, path)
     if path is None:
-        return err("Failed to initialize volume path {0}".format(path))
+        return err("%s is not writable and hence %s volume creation failed" % (datastore, full_vol_name.split('@')[0]))
```
pdhamdhere: This change assumes path == Null means Datastore is not writable. `get_vol_path` could be returning empty path due to other errors. You may want to return error from `get_vol_path`

kerneltime:
Please rebase your branch to the latest master for CI to pass.

PR #679, opened 2016-11-02
Cleanup service instance connection handling and usage.

brunotm:
Remove vmdk_ops.si global usage, better handle reconnection to service instance and cleanup.
In the current code are many code paths that calls connectLocal() directly, and can possibly overwrite the current si.

//CC @msterin


esx_service/vmdk_ops.py
```diff
@@ -619,19 +606,34 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
...
-        raise SystemExit("Failed to connect to localhost as 'dcui'.")
-
-    atexit.register(pyVim.connect.Disconnect, si)
+    global _service_instance
+    if not _service_instance:
+        try:
+            logging.info("Connecting to the local Service Instance")
+            _service_instance = pyVim.connect.Connect(host='localhost', user='dcui')
+        except Exception as e:
+            logging.exception("Failed to the local Service Instance as 'dcui': ")
```
pdhamdhere: nit: "Failed to connect to local Service Instance as 'dcui', exiting: "

esx_service/vmdk_ops.py
```diff
@@ -619,19 +606,34 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
...
+
+    atexit.register(pyVim.connect.Disconnect, _service_instance)
+
+def get_si():
+    '''
+	Return a connection to the local SI
+	'''
+    with getLock('siLock'):
+        try:
+            _service_instance.CurrentTime()
```
pdhamdhere: Will CurrentTime() indeed verify validity of _service_instance or will it return cached value? I am not an expert and [documentation](http://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.ServiceInstance.html#currentTime) isn't clear either. I wonder if we should be using RetrieveServiceContent to make sure _service_instance is still valid?

esx_service/vmdk_ops.py
```diff
@@ -619,19 +606,34 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
...
+        except:
+            connectLocalSi(force=True)
+    return _service_instance
```
govint: Sorry, may be the diff is showing the code wrongly but if we return here then the below code never executes?

esx_service/vmdk_ops.py
```diff
@@ -1038,6 +1040,7 @@ def disk_detach_int(vmdk_path, vm, device):
     spec.deviceChange = dev_changes
 
     try:
+        si = get_si()
```
govint: Is there a reason to call get_si() from within the try-except block, in the test and utils code its not called that way. Not wrong just asking.

esx_service/vmdk_ops.py
brunotm: From testing after calling  Disconnect(_service_instance) and also with a stale _service_instance, it returns:

```
vim.fault.NotAuthenticated: (vim.fault.NotAuthenticated) {
   dynamicType = <unset>,
   msg = 'The session is not authenticated.',
   faultCause = <unset>,
   faultMessage = (vmodl.LocalizableMessage) [],
   object = 'vim.VirtualDiskManager:ha-vdiskmanager',
   privilegeId = 'System.View'
}
```

The one i am not really sure about is vim.ServiceInstace.serverClock

esx_service/vmdk_ops.py
brunotm: It's a leftover. Thanks for catching this.

esx_service/vmdk_ops.py
brunotm: No reason, just easier to read. Can be moved if needed.

brunotm:
@govint 
>Looks good, but can you also add documentation that this is the way to get a connection to the ESX mgmt backend for all current and new code. Just so it becomes the way to develop code in the server.

Where the documentation should go? 

>In fact would prefer to move all of the VIM/VMOMI code into a separate module and have that expose APIs to vmdk_ops and utils/test code to do the kind of ops that the server needs. So that even the get_si() calls aren't peppered all over the code like its done today. Even this change had to touch so many places to replace a function call with another one. Something like VIM_GetVM(), VIM_ReconfigVM(), etc..

Agreed. There are many things that should be in separate modules, and in fact the actual code can be a little confusing.

My main reason to do this is correctness, begin to cleanup so it can be better maintained and reestructured more easily.

@govint @pdhamdhere 
I'll fix those and push again.

Thanks!

esx_service/vmdk_ops.py
pdhamdhere: Good catch! I am disappointed that I missed it :(

brunotm:
> Good catch! I am disappointed that I missed it :(

Me too :S

esx_service/vmdk_ops.py
```diff
@@ -619,30 +606,40 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
...
+        _service_instance = pyVim.connect.Connect(host='localhost', user='dcui')
 
     # set out ID in context to be used in request - so we'll see it in logs
     reqCtx = VmomiSupport.GetRequestContext()
     reqCtx["realUser"] = 'dvolplug'
-    logging.debug("Connect to localhost si:%s", si)
-    return si
+    atexit.register(pyVim.connect.Disconnect, _service_instance)
+
+def get_si():
```
kerneltime: Can you add a comment for SI acronym. 
```
get service instance 
```

Issue #660, opened 2016-10-28
vmdk-ops service unable to identify the newly mounted datastore on ESX host. But identifies after the service restart.


BaluDontu:
Hi guys,

I have mounted a new NFS datastore with name "NFSStore" on ESX host. 
When I execute following docker create volume command, it gives an error Invalid datastore.

docker volume create -d=vmdk --name=volume@NFSStore
_Error response from daemon: create volume@NFSStore: VolumeDriver.Create: Invalid datastore 'NFSStore'.
Known datastores: datastore1, NFSDatastore, coke (1).
Default datastore: datastore1_

I am able to see the volume create under /vmfs/volume/NFSStore on ESX host. However vmdk-ops service is unable to identify it.

msterin:
So there is 'NFSDatastore' and also "NFSStore' ? just checking. Can you shared the results of  'ls -l /vmfs/volumes` ?

BaluDontu:
Also, it was able to identify only after the service vmdk-ops restart.

[root@xxxx:/vmfs/volumes] ls -l
total 2068
drwxrwxrwx    6 99964800 201           4096 Oct 17 06:48 3bdec0ec-219e0c95
drwxr-xr-x    1 root     root             8 Jan  1  1970 4df30da5-ad6ac294-f786-80c16e6cf13a
drwxr-xr-t    1 root     root          3920 Oct 25 23:04 4df30dac-644c9042-b81a-80c16e6cf13a
drwxr-xr-x    1 root     root             8 Jan  1  1970 4df30dac-829f533e-cff2-80c16e6cf13a
drwxr-xr-x    4 999900   201           4096 Jun 10 18:35 85e097de-d1f29be0
lrwxr-xr-x    1 root     root            17 Oct 28 19:54 NFSDatastore -> f3d216d7-219d7a2b
lrwxr-xr-x    1 root     root            17 Oct 28 19:54 NFSStore -> 85e097de-d1f29be0
lrwxr-xr-x    1 root     root            17 Oct 28 19:54 Sample -> 3bdec0ec-219e0c95
drwxr-xr-x    1 root     root             8 Jan  1  1970 bbe5f970-46eca9be-484d-dd18578281fe
lrwxr-xr-x    1 root     root            38 Oct 28 19:54 coke (1) -> vvol:064ac87e656a4505-97f9297bf3c4a0da
[...]

PR #690, opened 2016-11-03
Fix for vmdk-ops service unable to identify the newly mounted datasto…

BaluDontu:
Currently vmdk-ops service is unable to identify a newly created datastore once it is up and running and after the first docker volume command is executed. Fix for #660 

vmdk-ops service caches the list of datastores accessible after the first docker volume command is executed. 
1. If the datastore is created before the first docker volume command is executed, the vmdk-ops service will identify the datastore since the list of accessible datastores are not cached till now.
2. Now after the first docker volume command is executed, the accessible datastores are not updated at all the vmdk-ops service. So the fix was update the cache of list of datastores when it couldn't find that datastore in its own cache.

Test steps:
1. Run any docker volume command. This will cache the list of datastores.
docker volume create -d=vmdk --name=testVol

2. Create an NFS datastore on ESX host.
esxcli storage nfs add --host=10.162.210.32 --share=/dbc/pa-dbc1132/bdontu/NFS2 --volume-name=nfsstore-bdontu2

3. docker volume create -d=vmdk --name=testVol2@nfsstore-bdontu2.
The above command successfully creates the volume "testVol2" on "nfsstore-bdontu2 datastore".
Without this fix, the vmdk-ops service complains that the datastore is not accessible.


pdhamdhere (commented):
@BaluDontu Logic looks good. Not your change but looks like way too may abstractions (get_datastores, known_datastores, get_latest_datastores) to check _datastore validity. How about consolidating them for e.g. 

```
def init_datstoreCache
   // populate datastoreCache

def validate_datastore(_ds)
   if datastoreCache == None
      init_datstoreCache();

   if _ds in datastoreCache
      return ds
   elif    //Cache miss
      init_datstoreCache() 

    // check again
```

esx_service/utils/vmdk_utils.py
```diff
@@ -42,18 +42,21 @@
 SNAP_SUFFIX_GLOB = "-[0-9][0-9][0-9][0-9][0-9][0-9].vmdk"
 
 
-def get_datastores():
+def get_datastores(doUpdate = False):
```
govint: Isn't it better to always update the list in case a datastore isn't found vs. one more function to decide on that.

esx_service/vmdk_ops.py
```diff
@@ -500,9 +500,13 @@ def get_vol_path(datastore, tenant_name=None):
 
 
 def known_datastores():
-    """returns names of know datastores"""
+    """returns names of known datastores"""
     return [i[0] for i in vmdk_utils.get_datastores()]
 
+def get_latest_datastores():
```
pdhamdhere: Yes, I was thinking on same line. However, handling datastore cache misses in `parse_vol_name` seems weird. So, proposed to change "get_list & validate" to "validate".

esx_service/utils/vmdk_utils.py
```diff
@@ -40,20 +40,12 @@
...
-    Returns a list of (name, url-name, dockvol_path), with an element per datastore
-    where:
-    'name' is datastore name (e.g. 'vsanDatastore') ,
-    'url-name' is the last element of datastore URL (e.g. 'vsan:572904f8c031435f-3513e0db551fcc82')
-    'dockvol-path; is a full path to 'dockvols' folder on datastore 
+    Initializes the datastore cache with the list of datastores accessible from local ESX host.
     """
-
     global datastores
     logging.debug("get_datastores: %s", datastores)
```
pdhamdhere: nit: Change get_datastores to init_datastoreCache

pdhamdhere (approved):
LGTM.

BaluDontu:
@pdhamdhere : Created a new pull request https://github.com/vmware/docker-volume-vsphere/pull/707. Please ignore this and review PR #707 .

Issue #695, opened 2016-11-03
Serialization of multiple operation requests (Create/Delete)


kerneltime:
A test run for https://github.com/vmware/docker-volume-vsphere/pull/693 failed.
The failure is not related to the PR. The sanity test post multi-threaded change seems flaky.

[build-failure.txt](https://github.com/vmware/docker-volume-vsphere/files/569687/build-failure.txt)

brunotm:
Hello @kerneltime, this is related to the kv/metadata issues and they where happening in the pre-threaded code. See #662 and #626. The parallel tests seems to trigger them more easily.

The kv creation/save terminates with error codes from disklib. It is possible to have the description of those error codes? In the log its 4008.

> 11/03/16 08:19:21 2254625 [Ubuntu.1404.vsan-volTestP14] [WARNING] Side car create for /vmfs/volumes/vsanDatastore/dockvols/volTestP14.vmdk failed - 4008

govint:
Actually, unless the KV file has been modified some way, there isn't code
to retain single KV pairs while removing all else. and the APIs used are
Python read/write APIs. Actually, till now we haven't seen a repro of the
kind of KV state  reported in #662. Nor do we have logs to look at.

That said @kerneltime can you post the error from the sanity tests?

govint:
From the test logs looks like this is a repro of #626. One thing to note is that we use Python open(), read, write() APIs to access the KV.

KV file creation is done with the ESX private library APIs and that "always" works. Because we are able to create the file with the ESX library APIs and close the KV file via the ESX private lib API immediately after creating the KV file. And that close op will close the file.

But thereafter we use Python open()/read/write APIs to load/store the KV. Now the open is done in the code and Python is expected to close the file. When that happens isn't in our control and with possibly multiple threads accessing the same file its possible that the file is still open when another thread comes in to store to the KV which can result in this scenario. But exaclty why two threads would be accessing the same file seems odd. 

Let me confirm the exact scenario from the logs.

brunotm:
@govint 
>KV file creation is done with the ESX private library APIs and that "always" works. Because we are able to create the file with the ESX library APIs and close the KV file via the ESX private lib API immediately after creating the KV file. And that close op will close the file.

The error code 4008 comes from disklib:

```
if use_sidecar_create:
       res = lib.DiskLib_SidecarCreate(dhandle, DVOL_KEY.encode(),
[...]

Issue #662, opened 2016-10-29
Corrupted metadata file should be handled gracefully and not terminate operations


msterin:
Today if metadata file is not complete (e.g. missing created-by, which may happen in diffent error paths) , getVMDK returns error. Since docker runs "get()" command before anything else, that blocks operations like `remove` or inspect.  This was found when fixing #661 

I think it would be much better if `vol_info` simply returned "N/A" or "Unknown" or "Failed to get" for fields it cannot fill for any reason. This would still check for volume existence first in getVMDK , and all admin commands (inspect/ls) would keep working.

Example of blocked operations: 

```
root@photon-1uL8XNsuB [ ~ ]# docker volume ls 
DRIVER              VOLUME NAME
vmdk                ty1
root@photon-1uL8XNsuB [ ~ ]# docker volume inspect ty1
[]
Error response from daemon: get ty1: VolumeDriver.Get: Failed to get disk details for /vmfs/volumes/datastore1/dockvols/ty1.vmdk ('created-by')
root@photon-1uL8XNsuB [ ~ ]# docker volume inspect ty1
[]
Error response from daemon: get ty1: VolumeDriver.Get: Failed to get disk details for /vmfs/volumes/datastore1/dockvols/ty1.vmdk ('created-by')
root@photon-1uL8XNsuB [ ~ ]# docker volume rm ty1
Error response from daemon: get ty1: VolumeDriver.Get: Failed to get disk details for /vmfs/volumes/datastore1/dockvols/ty1.vmdk ('created-by')
```

Assigning to @govint to take a look first, and decide the best way to handle it.

govint:
Correct, get is really to return status/meta-data about the volume and shouldn't fail unless the volume isn't there. I'll post a PR.

govint:
Taking a step back, the current behavior isn't really bad. First of all how did the meta-data get corrupted? Unless there is a code path or some sort of intentional tampering of the KV store there is no  way to corrupt the KV store file. Prefer to identify the code path or workflow that can corrupt the KV and fix that.

And if the KV store is corrupt then pretty much that volume isn't usable or shouldn't be usable thereafter.

Or we can move the volume KV to the SQL DB.

msterin:
[...]

PR #698, opened 2016-11-04
kv/disklib thread investigation.

brunotm:
This PR is intended to gather data and investigate about possibly and uncontrolled concurrent operations on same volumes metadata. Not for merging.

It will serialize and log intentions and waits for locks of kv calls with their arguments.

The topic is being discussed on #695, #662, #626 and possibly other closed issues.

### DIFF
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -25,6 +25,7 @@ import logging
 import sys
 import errno
 import time
+import threading
 
 # Python version 3.5.1
 PYTHON64_VERSION = 50659824
@@ -65,6 +66,9 @@ is_64bits = False
 # Number of times and sleep time to retry on IOError EBUSY
 EBUSY_RETRY_COUNT = 8
 EBUSY_RETRY_SLEEP = 0.5
+
+# Thread local storage
+thread_local = threading.local()
 
 class disk_info(Structure):
    _fields_ = [('size', c_uint64),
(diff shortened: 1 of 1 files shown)


msterin:
Actually, DiskLib is *not* thread-safe so this PR is a prime candidate for merge. The fix looks good though  I think the locks should be lower, in kvESX.py

//CC @govint

brunotm:
@msterin thank you! :)
The first approach was on kvESX, but as the interfaces where all exposed and used by volume_kv I moved it there. 
I will adjust and push.

msterin:
@brunotm - actually the more accurate statement about disklib is "Can't use the same handle in multiple threads, but you can use different handles in multiple threads". Basically, `dhandle` is not protected and not thread-safe, but the rest is

brunotm:
@msterin 
This seems odd. The dhandles are created inside the thread context in kvESX.vol_open_path and passed to disklib, i am missing something? :)

Pushed a change to explicitly keep in TLS the ctypes passed as pointers to disklib, but i'm not sure if placing the locks in the kvESX.create/delete/load/save/get_info/disk_lib_init and making everything else _private wouldn't be safer.

brunotm:
Closed in favor of #708

msterin:
@brunotm 
> This seems odd. The dhandles are created inside the request thread context in kvESX.vol_open_path and passed to disklib (that is loaded in the MainThread context and shared), i am missing something? :)

No, I do not think you do. dhandles are not thread safe **within** disklib - they structures backing them are not protected - but as long as dhandles are per-thread, there is no issues here. I slightly overreacted.

> Pushed a change to explicitly keep in TLS the ctypes passed as pointers to disklib, but i'm not sure if placing the locks in the kvESX.create/delete/load/save/get_info/disk_lib_init and making everything else _private, or yet, protect everything with a reentrant lock and keep as is wouldn't be safer.

As I said, I overreacted. As long as dhandles are not crossing thread boundaries, disklib is thread safe.

> what do you think?
Unfortunately, nothing. I will be looking into the code this coming week - had totally other (non-code :-)) priorities the past week.
what do you think?

PR #699, opened 2016-11-04
Handle errors on Get() call from server.

govint:
The Get() returns volume status thats optional in the docker plugin API. So, don't treat errors on status fetch as fatal. This test below had to artifically create an empty KV file because thats the only way it can be made empty or corrupted.

Tests:
```
$ docker volume create -d vmdk --name ca-vol-7
ca-vol-7

$ docker volume inspect ca-vol-7
[
    {
        "Name": "ca-vol-7",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/ca-vol-7",
        "Status": {
            "access": "read-write",
            "capacity": {
                "allocated": "13MB",
                "size": "100MB"
            },
            "created": "Fri Nov  4 11:38:53 2016",
            "created by VM": "2",
            "datastore": "bigone",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]
```
<trash the KV file - empty it>
```
$ docker volume inspect ca-vol-7
[
    {
        "Name": "ca-vol-7",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/ca-vol-7",
        "Labels": {},
        "Scope": "global"
    }
]

$ docker run --rm -it -w /data -v ca-vol-7:/data busybox
/data #
```


vmdk_plugin/plugin.go
```diff
@@ -352,22 +352,32 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
...
-		log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
-		// Fail back to a default version that we can try with.
-		fstype = "ext2"
+		// Check file system type.
+		value, exists = status["fstype"].(string)
+		if !exists {
+			msg := fmt.Sprintf("Got invalid filesystem type for %s, assuming type as ext2.", r.Name)
+			log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
+			// Fail back to a default version that we can try with.
+			value = "ext4"
```
brunotm: Same as above.

brunotm:
If we cannot rely in the metadata in those cases, maybe we could have a fallback method for mounting with exec.Command and mount -t auto instead of passing the default fstype that may not be the user specified FS ?

govint:
Yes, that sounds good but may be a separate change from this one.

vmdk_plugin/vmdkops/vmdkops.go
```diff
@@ -105,7 +105,7 @@ func (v VmdkOps) Get(name string) (map[string]interface{}, error) {
 
 	err = json.Unmarshal(str, &statusMap)
 	if err != nil {
-		return nil, err
+		log.Debugf("vmdkOps.Get failed decoding volume status for name=%s", name)
```
msterin: It should be warning , not debug. Admin cant do much about it but at least needs to know about it since it is why default would be  applied.

esx_service/vmdk_ops.py
```diff
@@ -421,7 +421,7 @@ def getVMDK(vmdk_path, vol_name, datastore):
     except Exception as ex:
         msg = "Failed to get disk details for %s (%s)" % (vmdk_path, ex)
         logging.error(msg)
-        result = err(msg)
+        result = {}
```
msterin: after this fix  `msg` is not needed and the code could be simply be as follows:
```python
  except Exception as ex:
      logging.error("Failed to get disk details for %s (%s)" % (vmdk_path, ex))
  return None
```

vmdk_plugin/plugin.go
```diff
@@ -352,22 +352,32 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
...
-	}
-
-	fstype, exists := status["fstype"].(string)
+	} else {
+		// Check access type.
+		value, exists := status["access"].(string)
+		if !exists {
+			msg := fmt.Sprintf("Got invalid access type for %s, assuming read-write access.", r.Name)
+			log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
+			isReadOnly = false
```
msterin: would be nice to actually print the offending string

vmdk_plugin/plugin.go
```diff
@@ -352,22 +352,32 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
...
-	if status["access"] == "read-only" {
-		isReadOnly = true
-	}
-
-	fstype, exists := status["fstype"].(string)
+	} else {
+		// Check access type.
+		value, exists := status["access"].(string)
+		if !exists {
+			msg := fmt.Sprintf("Got invalid access type for %s, assuming read-write access.", r.Name)
```
msterin: "Got" is not needed, the msg can simply say "Invalid access type..."

vmdk_plugin/plugin.go
```diff
@@ -352,22 +352,32 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
...
 
-	if !exists {
-		msg := fmt.Sprintf("Got invalid filesystem type for %s, attempting mount with type ext2.", r.Name)
-		log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
-		// Fail back to a default version that we can try with.
-		fstype = "ext2"
+		// Check file system type.
+		value, exists = status["fstype"].(string)
+		if !exists {
+			msg := fmt.Sprintf("Got invalid filesystem type for %s, assuming type as ext2.", r.Name)
```
msterin: "Got" is not needed, the msg can simply say "Invalid access type..."

vmdk_plugin/plugin.go
```diff
@@ -352,22 +352,32 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
...
+			log.WithFields(log.Fields{"name": r.Name, "error": msg}).Error("")
```
msterin: Would be good to print the offending string as well

msterin:
LGTM when CI passes

kerneltime:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/662

PR #705, opened 2016-11-04

BaluDontu:
Handled both scenarios as outlined below:

1. datastore is not writable
2. permission denied to create dockvols directory in datastore.

This is extension to #677 and is part of fix for #568.

osfs-mkdir -n returns different error codes when it is not able to create the dockvols directory in /vmfs/volumes/<datastore-name>. It returns error codes 13 and 30 for scenarios 1 and 2 described above. These error codes are used to give a proper errorMsg to the user.


BaluDontu:
@pdhamdhere : Can you please review it.

esx_service/vmdk_ops.py
```diff
@@ -475,24 +475,33 @@ def get_vol_path(datastore, tenant_name=None):
...
 
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores
         cmd = "{0} {1}".format(OSFS_MKDIR_CMD, dock_vol_path)
         rc, out = RunCommand(cmd)
-        if rc != 0:
-            logging.warning("Failed to create %s", dock_vol_path)
-            return None
+        errMsg = None
+        if rc == 13:
```
msterin: Please do not use literals like 13 or 30 directly. Define ERR_OSFS_MKDIR_<WATEVER>=13 and use here.

esx_service/vmdk_ops.py
```diff
@@ -475,24 +475,33 @@ def get_vol_path(datastore, tenant_name=None):
...
+            errMsg = "{0} is not writable and hence ./dockvol creation failed".format(datastore)
```
msterin: nit: A better message IMO would be '{0} creation failed - {1} not writeable).format('dockvols', datastore' (failure first, reasons next)

esx_service/vmdk_ops.py
```diff
@@ -475,24 +475,33 @@ def get_vol_path(datastore, tenant_name=None):
...
+        elif rc == 30:
+            errMsg = "Permission denied to create ./dockvol on {}".format(datastore)
```
msterin: nit: I think its "dockvols" not "dockvol". Maybe it's time to have a global var for this ?

esx_service/vmdk_ops.py
```diff
@@ -475,24 +475,33 @@ def get_vol_path(datastore, tenant_name=None):
...
+        else:
+            logging.warning("Failed to create %s", path)
+            errMsg = "Failed to initialize volume path {0}".format(path)
```
msterin: different logging and errmsg strings looks strange. Can we use the same string ?

PR #707, opened 2016-11-05
Fix for vmdk-ops service unable to identify the newly mounted datasto…

BaluDontu:
Currently vmdk-ops service is unable to identify a newly created datastore once it is up and running and after the first docker volume command is executed. Fix for #660

vmdk-ops service caches the list of datastores accessible after the first docker volume command is executed.

If the datastore is created before the first docker volume command is executed, the vmdk-ops service will identify the datastore since the list of accessible datastores are not cached till now.
Now after the first docker volume command is executed, the accessible datastores are not updated at all the vmdk-ops service. So the fix was update the cache of list of datastores when it couldn't find that datastore in its own cache.
Test steps:

Run any docker volume command. This will cache the list of datastores.
docker volume create -d=vmdk --name=testVol

Create an NFS datastore on ESX host.
esxcli storage nfs add --host=10.162.210.32 --share=/dbc/pa-dbc1132/bdontu/NFS2 --volume-name=nfsstore-bdontu2

docker volume create -d=vmdk --name=testVol2@nfsstore-bdontu2.
The above command successfully creates the volume "testVol2" on "nfsstore-bdontu2 datastore".
Without this fix, the vmdk-ops service complains that the datastore is not accessible.

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -45,7 +45,6 @@ import traceback
 import threading
 import time
 from ctypes import *
-from weakref import WeakValueDictionary
 
 from vmware import vsi
 
@@ -76,6 +75,7 @@ sys.path.insert(0, PY_LOC)
 if sys.version_info.major == 2:
     sys.path.append(PY2_LOC)
 
+import threadutils
 import log_config
 import volume_kv as kv
 import vmdk_utils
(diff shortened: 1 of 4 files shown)


esx_service/vmdk_ops.py
```diff
@@ -578,11 +573,11 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
...
         datastore = vm_datastore
-    elif datastore not in known_datastores():
+    elif not vmdk_utils.validate_datastore(datastore):
         return err("Invalid datastore '%s'.\n" \
-                   "Known datastores: %s.\n" \
-                   "Default datastore: %s" \
-                   % (datastore, ", ".join(known_datastores()), vm_datastore))
+                "Known datastores: %s.\n" \
+                "Default datastore: %s" \
+                % (datastore, ", ".join([i[0] for i in vmdk_utils.get_datastores()]), vm_datastore))
```
msterin: `[i[0] for i in vmdk_utils.get_datastores()` is cut-n-pasted twice, I think it's a small functoins (get_datastore_names_list())

BaluDontu:
New changes at - https://github.com/vmware/docker-volume-vsphere/pull/714

PR #713, opened 2016-11-07
Tenancy bug fixes

### LINKED ISSUE #710: Tenancy: volume  need to be removed from volumes table after the volume is removed
Unit test manually:
step 1: create a tenant and add privileges (all privileges) to this tenant on 'datastore1'. Set --volume-maxsize to 500MB and --volume-totalsize to 1GB

```
[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                  Name     Description  Default_datastore  VM_list    
------------------------------------  -------  -----------  -----------------  ---------  
b00870a1-e652-4b9e-93f3-7434476a43b2  tenant1               default_ds         photon-v1  

[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access add --name=tenant1 --datastore=datastore1 --rights=all --volume-maxsize=500MB --volume-totalsize=1GB

tenant access add succeeded

[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name=tenant1
Datastore   Create_volume  Delete_volume  Mount_volume  Max_volume_size  Total_size  
----------  -------------  -------------  ------------  ---------------  ----------  
datastore1  1              1              1             500.00MB         1.00GB      

```

2. create two volume 'tenant1-vol1" and "tenant1-vol2" , each with 500MB, which should succeed
```
root@photon-vOzbODUVd [ ~ ]# docker volume create --driver=vmdk --name=tenant1-vol1 -o size=500MB
tenant1-vol1
root@photon-vOzbODUVd [ ~ ]# docker volume create --driver=vmdk --name=tenant1-vol2 -o size=500MB
tenant1-vol2
```
3. create the third volume "tenant1-vol3" with 500MB, which should fail
```root@photon-vOzbODUVd [ ~ ]# docker volume create --driver=vmdk --name=tenant1-vol3 -o size=500MB
Error response from daemon: create tenant1-vol3: VolumeDriver.Create: The total volume size exceeds the usage quota

```

4. delete the first volume "tenant1-vol1" and re-create "tenant1-vol3", which should succeed 
```
root@photon-vOzbODUVd [ ~ ]# docker volume ls        
vmdk                tenant1-vol1
vmdk                tenant1-vol2
root@photon-vOzbODUVd [ ~ ]# docker volume rm tenant1-vol1
tenant1-vol1
root@photon-vOzbODUVd [ ~ ]# docker volume ls
vmdk                tenant1-vol2
root@photon-vOzbODUVd [ ~ ]# docker volume create --driver=vmdk --name=tenant1-vol3 -o size=500MB
tenant1-vol3
root@photon-vOzbODUVd [ ~ ]# docker volume ls
vmdk                tenant1-vol2
vmdk                tenant1-vol3
root@photon-vOzbODUVd [ ~ ]# 


```

### LINKED ISSUE #711: Tenancy: tenant name need to be unique
Tenant name in "tenant table" need to be unique.

lipingxue:
This change includes:
1. remove "volume" from "volumes" table when the volume is removed successfully
2. make the "tenant name" unique in "tenants" table.

Fix #710 

Fix #711


esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
...
+
+    logging.debug("remove volumes from volumes table(%s %s %s)", tenant_uuid, datastore,
+                  vol_name)
+    try:
+        _auth_mgr.conn.execute(
+                    "DELETE FROM volumes WHERE tenant_id = ? AND datastore = ? AND volume_name = ?", 
+                    [tenant_uuid, datastore, vol_name]
+            )
+        _auth_mgr.conn.commit()
+    except sqlite3.Error, e:
```
msterin: this is not Py3  compatible. See http://python-future.org/compatible_idioms.html
You should use 
```python
# Python 2 and 3:
try:
    ...
except sqlite3.Error as e:
    ...
```

esx_service/utils/auth_data.py
```diff
@@ -507,8 +514,8 @@ def create_tenant(self, name, description, default_datastore, default_privileges
                 )
             self.conn.commit()
         except sqlite3.Error, e:
```
msterin: pleas use `except <ERROR> as e:

esx_service/utils/auth_data.py
```diff
@@ -451,6 +452,12 @@ def create_tenant(self, name, description, default_datastore, default_privileges
 
         """
         logging.debug ("create_tenant name=%s", name)
+        error_info, exist_tenant = self.get_tenant(name)
+        if exist_tenant:
+            error_info = "Tenant {0} already exists".format(name)
+            logging.warning("Tenant %s already exists", name)
```
msterin: you can just use error_info here, no need to have a duplicate

esx_service/vmdk_ops.py
```diff
@@ -606,6 +606,13 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
 
     elif cmd == "remove":
         response = removeVMDK(vmdk_path)
+        # remove succeed, remove infomation of this volume from volumes table
```
msterin: this is the wrong place to manipulate the config DB. It should be done within removeVMDK

lipingxue: If a volume is created by a VM which is not associated with any tenant(the logic of __FALLBACK tenant does not exist yet), we should not insert it to the "volumes" table. Right?

esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
 
     return None
 
+def remove_volume_from_volumes_table(tenant_uuid, datastore, vol_name):
+    """ Remove volume from volumes table. """
+    _auth_mgr = get_auth_mgr()
+
+    logging.debug("remove volumes from volumes table(%s %s %s)", tenant_uuid, datastore,
```
shaominchen: 1. Should we specify the parameter name in the logging string to avoid confusion? Something like: "... (tenant=%s, datastore=%s, volume=%s)"
2. Nit: "remove volumes..." should be "remove volume..."
3. Should all log message start with capital letter?

esx_service/vmdk_ops.py
lipingxue: The reason we cannot insert it into "volumes" table is the volume is created by a VM which is not associate with any tenant.  volumes table need key like (tenant_uuid, datastore, volume_name). 

I don't understand what do you mean "how does it impact the location of DB code ?". The function to add/remove volume from "volumes" table is in auth.py.

esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
...
+                  vol_name)
+    try:
+        _auth_mgr.conn.execute(
+                    "DELETE FROM volumes WHERE tenant_id = ? AND datastore = ? AND volume_name = ?", 
+                    [tenant_uuid, datastore, vol_name]
```
shaominchen: I checked the schema of this table, I'm a little bit surprised that we are using a combination key of tenant_uuid+datastore+volume_name. Performance is not the main concern here, but I'm concerned we will run into trouble if volumes need to be referenced from other (new introduced) tables.

BTW, our auth proposal document needs to be updated to add this "volumes" table. I can take care of that.

esx_service/vmdk_ops.py
msterin: > The reason we cannot insert it into "volumes" table...

Ah, yet another reason to have _fallback_ tenant then.

> I don't understand what do you mean "how does it impact the location of DB code ?". 

My point was that the code removing volume from the table is in the wrong place. It should NOT be in executeRequest, it is not the correct layer. I was not sure how your comment was addressing mine :-)

esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
...
+            )
+        _auth_mgr.conn.commit()
+    except sqlite3.Error, e:
+        logging.error("Error %s when remove from volumes table for tenant_id %s and datastore %s",
```
shaominchen: Will this print out the whole exception stack trace or just one single exception message? If it's a stack trace, I'd suggest to print that in a new line.

esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
 
     return None
 
+def remove_volume_from_volumes_table(tenant_uuid, datastore, vol_name):
```
shaominchen: I don't think it's a good idea to include the database table name in the method name. DB schema should be isolated and decoupled from core business logic so that both can evolve independently. So I'd suggest to rename this method to "remove_volume"

esx_service/utils/auth.py
```diff
@@ -326,6 +326,25 @@ def add_volume_to_volumes_table(tenant_uuid, datastore, vol_name, vol_size_in_MB
...
+    _auth_mgr = get_auth_mgr()
+
+    logging.debug("remove volumes from volumes table(%s %s %s)", tenant_uuid, datastore,
+                  vol_name)
+    try:
+        _auth_mgr.conn.execute(
+                    "DELETE FROM volumes WHERE tenant_id = ? AND datastore = ? AND volume_name = ?", 
+                    [tenant_uuid, datastore, vol_name]
+            )
+        _auth_mgr.conn.commit()
```
shaominchen: It looks like we have quite some duplicate code in this module. We should refactor the code to add the template steps in a util class. The util class should expose some common DB operation methods to handle: 1) get connection; 2) execute statement; 3) commit (for CUD only); 4) error handling; 5) result processing (pagination support, etc.) - for R (query) only

I'm not suggesting to address this in the current review, but we should file a bug to track this improvement.

esx_service/cli/vmdkops_admin.py
```diff
@@ -845,7 +859,24 @@ def tenant_create(args):
...
+    """ Handle tenant create command """
+    error_info, tenant = get_tenant_from_db(args.name)
+    if error_info:
+        return operation_fail(error_info)
+    
+    if not tenant:
+        error_info = "Tenant {0} does not exist".format(args.name)
+        return operation_fail(error_info)
+
+    error_info = tenant.set_name(_auth_mgr.conn, args.new_name)
```
pdhamdhere: `tenant.set_name` only updates DB record. TenantName is also used as folder name under dockvols. Once tenant name is changed, list/attach/detach/get will stop working for existing volumes. Right?

lipingxue: Yes, I think we should also rename the folder name.

msterin: Is this addressed yet?

msterin:
@lipingxue -is the PR ready for review ? There were some good questions and there are merge conflicts, so I am not sure if it makes sense to review or to wait until the above is addressed.

msterin:
With refactoring work this PR is stale, and will be replaced by much shorter and more elegant code a bit later. Closing for now

PR #717, opened 2016-11-07
Pass the kill switch into docker and nested Makefiles.

kerneltime:
The kill switch is to avoid UI code from being built and packaged
by default. This kill switch will be in place till UI is ready to
be consumed. Once the UI code is ready and integrated with the
VMODL APIs, the kill switch can be removed.

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -109,6 +109,7 @@ else
   docker_socket=/var/run/docker.sock
   $DOCKER run --privileged --rm -it \
     -e "PKG_VERSION=$PKG_VERSION" \
+    -e "INCLUDE_UI=$INCLUDE_UI" \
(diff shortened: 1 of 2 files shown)


shaominchen:
The change looks straightforward, but can you add some description about the kill switch? Just for tracking purposes.

kerneltime:
Done, the commit is updated.

shaominchen:
LGTM.

msterin:
Please add something about "In order to enable the UI build, do <this and that>".
Other than that LGTM

kerneltime:
It was documented in the PR that introduced the kill switch (which stopped UI from building).
To build UI run:
```
INCLUDE_UI=true make
```
If this lands up being a permanent feature will add documentation in the relevant build instructions.

Issue #687, opened 2016-11-02
Tenancy: Add access to datastore does not add access.


kerneltime:
I had a tenant to whom I tried to give access to the datastores and it failed.

```
[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                  Name     Description  Default_datastore  VM_list                      
------------------------------------  -------  -----------  -----------------  ---------------------------  
dc17b3b1-1f4d-4f8b-a6f4-3dba064ba88f  tenant1               default_ds         photon.vsan.1,photon.vsan.2  
5afa4767-ecd8-489e-9d97-5c46ae50d3fc  DevAppX               default_ds         ubuntu                       

[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access add --name tenant1 --datastore datastore1 --rights all
{'mount_volume': 1, 'global_visibility': 0, 'max_volume_size': 0, 'create_volume': 1, 'delete_volume': 1, 'datastore': 'datastore1', 'usage_quota': 0}
tenant access add succeeded
[root@localhost:~] ls /vmfs/volumes/
57302fae-961b9cf9-2e49-000c29acb565/     57303269-7342b974-b24a-000c29acb565/     bd5a4479-14fea494-d916-6fbe178010ee/     vsan:5201dd577fe793bd-112186a6c2b42e0c/
57303269-4d223a2e-d657-000c29acb565/     b78f9f70-08216aea-97af-6952c751dfe2/     datastore1/                              vsanDatastore/
[root@localhost:~] ls /vmfs/volumes/
[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access add --name tenant1 --datastore vsanDatastore --rights all
{'mount_volume': 1, 'global_visibility': 0, 'max_volume_size': 0, 'create_volume': 1, 'delete_volume': 1, 'datastore': 'vsanDatastore', 'usage_quota': 0}
tenant access add succeeded
[root@localhost:~]  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                  Name     Description  Default_datastore  VM_list                      
------------------------------------  -------  -----------  -----------------  ---------------------------  
dc17b3b1-1f4d-4f8b-a6f4-3dba064ba88f  tenant1               default_ds         photon.vsan.1,photon.vsan.2  
[...]

PR #718, opened 2016-11-08

BaluDontu:
Handled both scenarios as outlined below:

datastore is not writable
permission denied to create dockvols directory in datastore.
This is extension to #705 and is part of fix for #568. Addressed Mark's comments over here.

+CC @pdhamdhere @msterin 

osfs-mkdir -n returns different error codes when it is not able to create the dockvols directory in /vmfs/volumes/<datastore-name>. It returns error codes 30 and 13 for scenarios 1 and 2 described above. These error codes are used to give a proper errorMsg to the user.

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -492,20 +492,22 @@ def get_vol_path(datastore, tenant_name=None):
         logging.debug("Found %s, returning", path)
         return path, None
 
-    cmd = None
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores
         cmd = "{0} {1}".format(OSFS_MKDIR_CMD, dock_vol_path)
-    if tenant_name and not os.path.isdir(path):
-        cmd = "{0} {1}".format(MKDIR_CMD, path)
-    rc, out = RunCommand(cmd)
-    if rc != 0:
-        if tenant_name:
-            errMsg = "Failed to initialize volume path {0}".format(path)
-        else:
+        rc, out = RunCommand(cmd)
+        if rc != 0:
             errMsg = "{0} creation failed - {1} on {2}".format(DOCK_VOLS_DIR, os.strerror(rc), datastore)
-        logging.warning(errMsg)
-        return None, err(errMsg)
+            logging.warning(errMsg)
+            return None, err(errMsg)
+    if tenant_name and not os.path.isdir(path):
+        # The mkdir command is used to create "tenant_name" folder inside DOCK_VOLS_DIR on "datastore"
+        cmd = "{0} {1}".format(MKDIR_CMD, path)
+        rc, out = RunCommand(cmd)
+        if rc != 0:
+            errMsg = "Failed to initialize volume path {0}".format(path)
+            logging.warning(errMsg)
+            return None, err(errMsg)
 
     logging.info("Created %s", path)
     return path, None
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -505,7 +505,7 @@ def get_vol_path(datastore, tenant_name=None):
         cmd = "{0} {1}".format(MKDIR_CMD, path)
         rc, out = RunCommand(cmd)
         if rc != 0:
-            errMsg = "Failed to initialize volume path {0}".format(path)
+            errMsg = "Failed to initialize volume path {0} - {1}".format(path, out)
             logging.warning(errMsg)
             return None, err(errMsg)
 
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -494,28 +494,20 @@ def get_vol_path(datastore, tenant_name=None):
         logging.debug("Found %s, returning", path)
         return path, None
 
+    cmd = None
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores
         cmd = "{0} {1}".format(OSFS_MKDIR_CMD, dock_vol_path)
-        rc, out = RunCommand(cmd)
-        errMsg = None
-        if rc == ERR_OSFS_MKDIR_READONLY:
-            errMsg = "{0} creation failed - {1} not writable".format(DOCK_VOLS_DIR, datastore)
-        elif rc == ERR_OSFS_MKDIR_NOPERMISSION:
-            errMsg = "{0} creation failed - Permission denied on {1}".format(DOCK_VOLS_DIR, datastore)
-        else:
-            logging.warning("Failed to initialize volume path %s", path)
-            errMsg = "Failed to initialize volume path {0}".format(path)
-        #creation of dockvols directory failed.
-        if errMsg is not None:
-            return None, err(errMsg)
     if tenant_name and not os.path.isdir(path):
         cmd = "{0} {1}".format(MKDIR_CMD, path)
-        rc, out = RunCommand(cmd)
-        if rc != 0:
-           logging.warning("Failed to initialize volume path %s", path)
-           errMsg = "Failed to initialize volume path {0}".format(path)
-           return None, err(errMsg)
+    rc, out = RunCommand(cmd)
+    if rc != 0:
+        if tenant_name:
+            errMsg = "Failed to initialize volume path {0}".format(path)
+        else:
+            errMsg = "{0} creation failed - {1} on {2}".format(DOCK_VOLS_DIR, os.strerror(rc), datastore)
+        logging.warning(errMsg)
+        return None, err(errMsg)
 
     logging.info("Created %s", path)
     return path, None
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -107,6 +107,8 @@ MAX_SKIP_COUNT = 16       # max retries on VMCI Get Ops failures
 # Error codes
 VMCI_ERROR = -1 # VMCI C code uses '-1' to indicate failures
 ECONNABORTED = 103 # Error on non privileged client
+ERR_OSFS_MKDIR_READONLY = 30 # Error on creating a dockvols directory on a non-writable datastore
+ERR_OSFS_MKDIR_NOPERMISSION = 13 # Error on creating a dockvols directory on a datastore where permissions are denied
 
 # Volume data returned on Get request
 CAPACITY = 'capacity'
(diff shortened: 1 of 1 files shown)


BaluDontu:
@pdhamdhere :  Can you verify this. All comments have been addressed.

esx_service/vmdk_ops.py
```diff
@@ -490,24 +492,33 @@ def get_vol_path(datastore, tenant_name=None):
...
-            return None
+        errMsg = None
+        if rc == ERR_OSFS_MKDIR_READONLY:
+            errMsg = "{0} creation failed - {1} not writable".format(DOCK_VOLS_DIR, datastore)
+        elif rc == ERR_OSFS_MKDIR_NOPERMISSION:
+            errMsg = "{0} creation failed - Permission denied on {1}".format(DOCK_VOLS_DIR, datastore)
+        else:
+            logging.warning("Failed to initialize volume path %s", path)
+            errMsg = "Failed to initialize volume path {0}".format(path)
+        #creation of dockvols directory failed.
```
msterin: There is no need to duplicate the message. You can simply use
```python
    errMsg = "Failed to initialize volume path {0}".format(path)
    logging.warning(errMsg)
```

esx_service/vmdk_ops.py
```diff
@@ -490,24 +492,33 @@ def get_vol_path(datastore, tenant_name=None):
...
+        if errMsg is not None:
+            return None, err(errMsg)
     if tenant_name and not os.path.isdir(path):
         cmd = "{0} {1}".format(MKDIR_CMD, path)
         rc, out = RunCommand(cmd)
         if rc != 0:
-           logging.warning("Failed to create %s", path)
-           return None
+           logging.warning("Failed to initialize volume path %s", path)
+           errMsg = "Failed to initialize volume path {0}".format(path)
```
msterin: There is no need to duplicate the message. You can simply use
```python
    errMsg = "Failed to initialize volume path {0}".format(path)
    logging.warning(errMsg)
```

nit: Besides, I recall seeing the same message a few lines ago, You may want to reuse the format string (you can define it once in the beginning of the function)

esx_service/vmdk_ops.py
```diff
@@ -490,24 +492,33 @@ def get_vol_path(datastore, tenant_name=None):
...
-            logging.warning("Failed to create %s", dock_vol_path)
-            return None
+        errMsg = None
+        if rc == ERR_OSFS_MKDIR_READONLY:
+            errMsg = "{0} creation failed - {1} not writable".format(DOCK_VOLS_DIR, datastore)
+        elif rc == ERR_OSFS_MKDIR_NOPERMISSION:
+            errMsg = "{0} creation failed - Permission denied on {1}".format(DOCK_VOLS_DIR, datastore)
+        else:
+            logging.warning("Failed to initialize volume path %s", path)
+            errMsg = "Failed to initialize volume path {0}".format(path)
```
msterin: won't it hit the 'else' clause and print this error message even when RunCommand was successful ?  I think you need to have 
```python
  elif rc == 0:
    pass
```
somewhere

esx_service/vmdk_ops.py
```diff
@@ -490,24 +492,33 @@ def get_vol_path(datastore, tenant_name=None):
...
 
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores
         cmd = "{0} {1}".format(OSFS_MKDIR_CMD, dock_vol_path)
         rc, out = RunCommand(cmd)
-        if rc != 0:
-            logging.warning("Failed to create %s", dock_vol_path)
-            return None
+        errMsg = None
+        if rc == ERR_OSFS_MKDIR_READONLY:
```
pdhamdhere: Rather than special casing error codes, why not use existing library function to translate error code to string? I am not a python expert but tried following and seems to be fine;

>>> os.strerror(13)
'Permission denied'
>>> os.strerror(30)
'Read-only file system'

msterin: good point. It also removes the need to have special case per failure

esx_service/vmdk_ops.py
BaluDontu: Sorry, I missed out on this. Only tested wrong scenarios.

esx_service/vmdk_ops.py
BaluDontu: Great point. Modified the code according to your suggestion. Using this, no special checks are required for error codes.

BaluDontu:
In the below example nfsstore-baominw1 is a read-only nfs datastore mounted on ESX host.
root@photon-a4ZgLxlWh [ ~/CNA/docker-volume-vsphere ]# docker volume create -d=vmdk --name=hello@nfsstore-baominw1
**Error response from daemon: create hello@nfsstore-baominw1: VolumeDriver.Create: dockvols creation failed - Read-only file system on nfsstore-baominw1**

In the below example nfsstore-brandonn is a nfs datastore with no access permissions.
root@photon-a4ZgLxlWh [ ~/CNA/docker-volume-vsphere ]# docker volume create -d=vmdk --name=hello@nfsstore-brandonn
**Error response from daemon: create hello@nfsstore-brandonn: VolumeDriver.Create: dockvols creation failed - Permission denied on nfsstore-brandonn**

In the below example nfsstore-bdontu2 is a nfs datastore with full access.
root@photon-a4ZgLxlWh [ ~/CNA/docker-volume-vsphere ]# docker volume create -d=vmdk --name=hello@nfsstore-bdontu2 
**hello@nfsstore-bdontu2**

On ESX host, I can see the VMDK created.
[root@promc-2n-dhcp105-41:/vmfs/volumes] cd nfsstore-bdontu2/dockvols/
**[root@promc-2n-dhcp105-41:/vmfs/volumes/6e5cea93-c7ca09c0/dockvols] ls
hello-a863f5c847c54e2d.vmfd  hello-flat.vmdk              hello.vmdk**

esx_service/vmdk_ops.py
```diff
@@ -490,24 +490,25 @@ def get_vol_path(datastore, tenant_name=None):
...
         # If the path exists then return it as is.
         logging.debug("Found %s, returning", path)
-        return path
+        return path, None
 
+    cmd = None
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores
         cmd = "{0} {1}".format(OSFS_MKDIR_CMD, dock_vol_path)
-        rc, out = RunCommand(cmd)
```
pdhamdhere: Why did you delete this code? This need to exist. See #687 
OSFS_MKDIR_CMD is used to create dock_vol_path and MKDIR_CMD is used to create Tenant folder.

BaluDontu: I didn't delete any code. It achieves the same functionality as before. I just remodified the code so that it looks simpler now.

esx_service/vmdk_ops.py
```diff
@@ -490,24 +490,25 @@ def get_vol_path(datastore, tenant_name=None):
...
-        if rc != 0:
-            logging.warning("Failed to create %s", dock_vol_path)
-            return None
     if tenant_name and not os.path.isdir(path):
         cmd = "{0} {1}".format(MKDIR_CMD, path)
-        rc, out = RunCommand(cmd)
-        if rc != 0:
-           logging.warning("Failed to create %s", path)
-           return None
+    rc, out = RunCommand(cmd)
```
BaluDontu: If it is a dockvols creation, it sets the command to OSFS_MKDIR_CMD and if it is tenant creation, it sets the command to MKDIR_CMD. Once it exits the if condition, it will take the command and execute it.

BaluDontu:
My previous review, doesn't create execute OSFS command to execute dockvols directory. It uses mkdir instead. Fixed the code to make sure dockvols is created using OSFS and tenant directory is created using mkdir.

Test steps:

First created a tenant using admin CLI
[root@promc-2n-dhcp105-41:/usr/lib/vmware/vmdkops/bin] ./vmdkops_admin.py tenant create --name T11 --vm-list photon-new-1
tenant create succeeded

[root@promc-2n-dhcp105-41:/usr/lib/vmware/vmdkops/bin] ./vmdkops_admin.py tenant access add --name T11 --datastore nfsstore-final --rights create,delete,mou
nt
{'mount_volume': 1, 'global_visibility': 0, 'max_volume_size': 0, 'create_volume': 1, 'delete_volume': 1, 'datastore': 'nfsstore-final', 'usage_quota': 0}
tenant access add succeeded

On docker host, executed the following command
root@photon-a4ZgLxlWh [ ~ ]# docker volume create -d=vmdk --name=helloT11@nfsstore-final
helloT11@nfsstore-final

It created the volume successfully. Verifying the logs I was able to confirm that first it created dockvols using OSFS command and created a tenant using mkdir command.

esx_service/vmdk_ops.py
```diff
@@ -490,24 +490,27 @@ def get_vol_path(datastore, tenant_name=None):
...
+            logging.warning(errMsg)
+            return None, err(errMsg)
     if tenant_name and not os.path.isdir(path):
+        # The mkdir command is used to create "tenant_name" folder inside DOCK_VOLS_DIR on "datastore"
         cmd = "{0} {1}".format(MKDIR_CMD, path)
         rc, out = RunCommand(cmd)
         if rc != 0:
-           logging.warning("Failed to create %s", path)
-           return None
+            errMsg = "Failed to initialize volume path {0}".format(path)
```
pdhamdhere: Why not include error string here?
errMsg = "Failed to initialize volume path {0} :  {1}".format(path, os.strerror(rc))

BaluDontu: The problem is mkdir is that it returns either 0 or 1. 0 on success and 1 on failure for any reason.

Inorder to simulate this condition, where it fails on creating a tenant and is successful on dockvol creation. I have changed the access permissions on dockvol.

For example, when I printed the **rc, out** values to log
[photon-new-1-type@nfsstore-thanks] [DEBUG  ] ****rc: 1, out: mkdir: can't create directory '/vmfs/volumes/nfsstore-thanks/dockvols/T12': Permission denied
****

Here, it clearly shows the error statement but return code is different. 
For that reason, I just settled with a normal message without the reason.

pdhamdhere: Why not then log "out" which already has error string?

BaluDontu:
Previously specific error message was not displayed on failure of tenant creation. This commit will fix it.
 
Test steps:

1. Create a tenant on a datastore which doesn't have access permissions. You can now see specific error message now.

root@photon-a4ZgLxlWh [ ~ ]# docker volume create -d=vmdk --name=vol1@nfsstore-new
Error response from daemon: create vol1@nfsstore-new: VolumeDriver.Create: Failed to initialize volume path /vmfs/volumes/nfsstore-new/dockvols/T15 - mkdir: can't create directory '/vmfs/volumes/nfsstore-new/dockvols/T15': Permission denied

This is similar to one in https://github.com/vmware/docker-volume-vsphere/blob/master/esx_service/vmdk_ops.py#L175 where the whole error message is displayed. So followed the same style. 

I can also rip out the just the error reason. for example in **mkdir: can't create directory '/vmfs/volumes/nfsstore-new/dockvols/T15': Permission denied**, I can just print the error reason - Permission denied by using code out.split(':')[-1] but sometimes it might lead to index out of range errors when error reason is not specified.  So went ahead with printing the whole error message which can be easy understood by the user.

PR #731, opened 2016-11-13
Removed trailing whitespaces

msterin:
trailing whitespaces generate lots of noise in pylint and are not recommended in PEP8. 

So I removed them using the following command: 
```bash
 for i in esx_service/*.py esx_service/*/*.py ; do echo $i ; sed -i " " -e 's/[ ]*$//' $i ; done
```

TEST: 
- git diff -b (reports no changes so it's all whitespaces)
- CI


msterin:
Note : I will wait with this merge to allow @lipingxue @govint an @brunotm to merge pending PR, to avoid conflicts with this change. I will regenerate the merge the trailing WS cleanup after that

msterin:
Note : I will wait with this merge to allow @lipingxue @govint an @brunotm to merge pending PR, to avoid conflicts with this change. I will regenerate and resubmit  the trailing WS cleanup  (and other WS cleanup) after that

PR #729, opened 2016-11-12
code refactoring for tenant management API

lipingxue:
This PR includes:
1. Refactor code to add an abstract layer provide APIs for tenant management
2. add vmdk_ops unit test for multi-tenency
3. make "name" unique in "tenants" table

PR #735, opened 2016-11-15
Add some parameter checks for multi-tenancy CLI commands

### LINKED ISSUE #673: The max_volume_size can be set a value even bigger than the total_size
The max_volume_size can be set a value even bigger than the total_size:

[root@localhost:/usr/lib/vmware/vmdkops/bin] vmdkops_admin tenant access add --name MyTenant --datastore datastore1 --volume-totalsize 1GB --rights create,delete --volume-maxsize 2GB
{'mount_volume': 0, 'global_visibility': 0, 'max_volume_size': 2048, 'create_volume': 1, 'delete_volume': 1, 'datastore': 'datastore1', 'usage_quota': 1024}
tenant access add succeeded
[root@localhost:/usr/lib/vmware/vmdkops/bin] vmdkops_admin tenant access ls --name MyTenant
Datastore   Create_volume  Delete_volume  Mount_volume  Max_volume_size  Total_size  
----------  -------------  -------------  ------------  ---------------  ----------  
datastore1  1              1              0             2.00GB           1.00GB    

We should add some parameter validation logic to prevent wrong configuration.

@lipingxue @msterin

### LINKED ISSUE #684: The behavior of "tenant access add" command is confusing
Currently the "tenant access add" command can be run any number of times to grant a datastore access for the same tenant. This may cause weird behavior which is counterintuitive. For example:

I have granted the following permissions for my tenant:

```
[root@localhost:/usr/lib/vmware/vmdkops/bin] vmdkops_admin tenant access ls --name MyTenant
Datastore   Create_volume  Delete_volume  Mount_volume  Max_volume_size  Total_size  
----------  -------------  -------------  ------------  ---------------  ----------  
datastore1  1              1              0             128.00MB         1.00GB      
```

Now I want to add the missing "mount" privilege. As the implementor, we know all the commands and we expect user to run "tenant access set" to modify the privileges. But a new user who are not familiar with all the command options might (incorrectly) do this instead:
```
[root@localhost:/usr/lib/vmware/vmdkops/bin] vmdkops_admin tenant access add --name MyTenant --datastore datastore1 --right mount
{'mount_volume': 1, 'global_visibility': 0, 'max_volume_size': 0, 'create_volume': 0, 'delete_volume': 0, 'datastore': 'datastore1', 'usage_quota': 0}
tenant access add succeeded
```
Now what happened? All the existing privilege settings are gone:

```
[root@localhost:/usr/lib/vmware/vmdkops/bin] vmdkops_admin tenant access ls --name MyTenant
Datastore   Create_volume  Delete_volume  Mount_volume  Max_volume_size  Total_size  
----------  -------------  -------------  ------------  ---------------  ----------  
datastore1  0              0              1             0B               0B          
```

This is not a correctness issue (we can argue this is a user error). But I think we can improve the usability a little bit and make the program more robust to prevent this wrong operation happening. 

Regarding the solution, I think one option is to prevent user to run "tenant access add" twice on the same tenant, and we can print a tip message to remind user to use "tenant access set" to modify the permissions.

Any other suggestions are welcome.

shaominchen:
Add parameter checks for multi-tenancy CLI commands (Fixes #673 & Fixes #684)

1. Fail the operation if "tenant access add" is called on the same tenant & datastore again
2. For "tenant access add" command:
    1) If setting both max_size and total_size, max_size should not be bigger than total_size
    2) If setting total_size only, max_size should be set to the same
3. For "tenant access set" command:
    1) If setting both max_size and total_size, max_size should not be bigger than total_size
    2) If setting max_size only, max_size should not exceed existing usage quota (if already set)
    3) If setting total_size only, and existing max_size was already set, total_size should not be smaller than that; otherwise max_size should be set to the same as total_size

Please note that the current change does not address all parameter validation scenarios - for example, volume total size should not exceed datastore capacity itself. Those scenarios will be tracked in separate issues.

### DIFF
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -998,6 +998,10 @@ def generate_privileges(args):
 
 def tenant_access_add(args):
     """ Handle tenant access command """
+    error_info = check_args(args)
+    if error_info:
+        return operation_fail(error_info)
+
     error_info, tenant = get_tenant_from_db(args.name)
     if error_info:
         return operation_fail(error_info)
@@ -1014,6 +1018,16 @@ def tenant_access_add(args):
         return operation_fail(error_info)
     else:
         print "tenant access add succeeded"
+
+def check_args(args):
+    """ Check validity of CLI arguments """
+    if (args.volume_maxsize > args.volume_totalsize):
+        error_info = "Volume max size should not exceed the total size"
+    
+    if error_info:
+        return error_info
+    else
+        return None;
 
 def modify_privileges(privileges, args):
     """ Modify privileges based on CLI argument """
@@ -1050,6 +1064,10 @@ def generate_privileges_dict(privileges):
 
 def tenant_access_set(args):
     """ Handle tenant access set command """
+    error_info = check_args(args)
+    if error_info:
+        return operation_fail(error_info)
+
     error_info, tenant = get_tenant_from_db(args.name)
     if error_info:
         return operation_fail(error_info)
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -1012,12 +1012,13 @@ def tenant_access_add(args):
         error_info = "Privilege already exists for ({0}, {1})".format(args.name, args.datastore)
         return operation_fail(error_info)
 
-    error_info = check_tenant_access_add_args(args)
+    privileges = generate_privileges(args)
+    print privileges
+
+    error_info = check_tenant_access_add_args(privileges)
     if error_info:
         return operation_fail(error_info)
 
-    privileges = generate_privileges(args)
-    print privileges
     error_info = tenant.set_datastore_access_privileges(_auth_mgr.conn, [privileges])
       
     if error_info:
@@ -1025,19 +1026,23 @@ def tenant_access_add(args):
     else:
         print "tenant access add succeeded"
 
-def check_tenant_access_add_args(args):
+def check_tenant_access_add_args(privileges):
     """ Check validity of CLI arguments for tenant access add command """
+    max_size = privileges[auth_data_const.COL_MAX_VOLUME_SIZE]
+    total_size = privileges[auth_data_const.COL_USAGE_QUOTA]
+
     # If both volume max size and volume total size are set,
     # volume max size should not exceed volume total size
-    if (args.volume_maxsize and args.volume_totalsize and (args.volume_maxsize > args.volume_totalsize)):
-        error_info = "Volume max size {0} exceeds the total size {1}".format(
-            args.volume_maxsize, args.volume_totalsize)
+    if (max_size and total_size and (max_size > total_size)):
+        error_info = "Volume max size {0}MB exceeds the total size {1}MB".format(
+            max_size, total_size)
         return error_info
-    
+
     # If volume total size is set, but volume max size is NOT set,
     # volume max size should be set to the same as volume total size
-    if (args.volume_totalsize and not args.volume_maxsize):
-        args.volume_maxsize = args.volume_totalsize;
+    if (total_size and not max_size):
+        print('Set volume max size to the same value as volume total size: {0}MB'.format(total_size))
+        privileges[auth_data_const.COL_MAX_VOLUME_SIZE] = total_size
 
     return None;
 
@@ -1105,13 +1110,8 @@ def tenant_access_set(args):
     else:
         print "tenant access set succeeded"
 
-def check_tenant_access_set_args(args, privilege):
+def check_tenant_access_set_args(args, privileges):
     """ Check validity of CLI arguments for tenant access set command """
-    print privilege
-    print privilege[auth_data_const.COL_MAX_VOLUME_SIZE]
-    print privilege[auth_data_const.COL_USAGE_QUOTA]
-    print args.volume_totalsize
-    print args.volume_maxsize
 
     # If both volume max size and volume total size are NOT set, return directly
     if (not args.volume_maxsize and not args.volume_totalsize):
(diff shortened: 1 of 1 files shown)


esx_service/cli/vmdkops_admin.py
```diff
@@ -1006,15 +1006,46 @@ def tenant_access_add(args):
         error_info = "Tenant {0} does not exist".format(args.name)
         return operation_fail(error_info)
 
+    privileges = [d for d in tenant.privileges if d[auth_data_const.COL_DATASTORE] == args.datastore]
```
msterin: nit: I think this construct used in many places.... maybe its worth to wrap in a function

msterin: And, more importantly, this seems to contradict https://github.com/vmware/docker-volume-vsphere/pull/729 and will generate conflict. I thoink the checks in VMODL and CLI should be the same, so it should be in the API layer. Please branch off #729

esx_service/cli/vmdkops_admin.py
```diff
@@ -1006,15 +1006,46 @@ def tenant_access_add(args):
...
+    
+    if privileges:
+        error_info = "Privilege already exists for ({0}, {1})".format(args.name, args.datastore)
+        return operation_fail(error_info)
```
msterin: if the requested privilege is already set, it's better to succeed as no harm is done and what is request is exactly what the end result is

esx_service/cli/vmdkops_admin.py
```diff
@@ -1063,17 +1094,63 @@ def tenant_access_set(args):
...
+
+def check_tenant_access_set_args(args, privileges):
+    """ Check validity of CLI arguments for tenant access set command """
+
+    # If both volume max size and volume total size are NOT set, return directly
+    if (not args.volume_maxsize and not args.volume_totalsize):
+        return None;
+
+    # If both volume max size and volume total size are set,
+    # volume max size should not exceed volume total size
```
msterin: Can you elaborate why it is different from check_tenant_access_add_args() which seems to have the same logic ? Can the code be consolidated?

esx_service/cli/vmdkops_admin.py
```diff
@@ -1063,17 +1094,63 @@ def tenant_access_set(args):
...
     error_info = tenant.set_datastore_access_privileges(_auth_mgr.conn, [privileges_dict])
 
     if error_info:
         return operation_fail(error_info)
     else:
         print "tenant access set succeeded"
-       
+
+def check_tenant_access_set_args(args, privileges):
+    """ Check validity of CLI arguments for tenant access set command """
```
msterin: nit: it's better to explain what "validity" is here. I.e. the text from (very clear) comments below is easier to read here IMO

esx_service/cli/vmdkops_admin.py
```diff
@@ -1063,17 +1094,63 @@ def tenant_access_set(args):
...
+    if (args.volume_maxsize and not args.volume_totalsize and existing_total_size):
+        new_max_size_in_MB = convert.convert_to_MB(args.volume_maxsize)
+        if (new_max_size_in_MB > existing_total_size):
+            error_info = "Volume max size {0}MB exceeds the existing total size {1}MB".format(
+                new_max_size_in_MB, existing_total_size)
+            return error_info
+
+    # If volume total size is set, but volume max size is NOT set,
+    # 1) if existing volume max size was already set, volume total size should not be smaller than that
+    # 2) if existing volume max size was not set, it should be set to the same as volume total size
```
msterin: the math here is clear, but I am not sure it's worth coding. 
What's wrong with having these limits unrelated and checking all of them on access control ? We'll still have to do it , so I am not sure what is it we are saving. 

 I like the ingress values control (to prevent typos essentially) but the rest seems too complicated... There does not seem to be much value in this cross-correlations, which seem to mangle the original (customer) request and now need to be maintained in sync

At any rate - why not keep it as is and simply check all on access control time  ?

esx_service/cli/vmdkops_admin.py
```diff
@@ -1063,17 +1094,63 @@ def tenant_access_set(args):
...
+    if (args.volume_totalsize and not args.volume_maxsize):
+        new_total_size_in_MB = convert.convert_to_MB(args.volume_totalsize)
+        existing_max_size = privileges[auth_data_const.COL_MAX_VOLUME_SIZE]
+        if (existing_max_size):
+            if (new_total_size_in_MB < existing_max_size):
+                error_info = "Volume total size {0}MB is smaller than the existing max size {1}MB".format(
+                    new_total_size_in_MB, existing_max_size)
```
msterin: it's a good warning but I do not think it's worth erroring out. And then testing all that

esx_service/cli/vmdkops_admin.py
shaominchen: I will review #729 and get back on this.

esx_service/cli/vmdkops_admin.py
shaominchen: Can you also take a look at issue #684? I have explained my concern there. If you still feel my concern is not valid, please just comment on that.

esx_service/cli/vmdkops_admin.py
shaominchen: Hmm...isn't it too long if we add those comments in the method documentation?

msterin: I think it's the place where I'd like to have the info. Still, it's a nit so I am fine either way

esx_service/cli/vmdkops_admin.py
shaominchen: As you can see in the code logic, check_tenant_access_add_args() is simpler - we are adding a new privilege, so all we need to check is just user input.

The check_tenant_access_set_args(), however, is different and more complicated. We need to check not only user input, but also the existing data in the database.

There's only very minimum duplicate logic in these 2 methods. I feel merging them will cause more confusion than clarity.

esx_service/cli/vmdkops_admin.py
msterin: Yes, I totally agree with your take, and the suggestion, in #684

esx_service/cli/vmdkops_admin.py
msterin: fair enough.

esx_service/cli/vmdkops_admin.py
shaominchen: Same as previous comments - my point here is: we should be stricter for the user input so that the chance of runtime errors can be reduced to the minimum. Allowing wrong configuration may cause unexpected runtime behavior and/or potential risks.

esx_service/cli/vmdkops_admin.py
msterin: agreed

msterin:
@shaominchen - unit test infra is available (we are using python unittest module), take a look at *_test.py,. Granted it can (and will) be improved, but it's there.

esx_service/cli/vmdkops_admin.py
```diff
@@ -1006,15 +1006,46 @@ def tenant_access_add(args):
...
+        return operation_fail(error_info)
+
     error_info = tenant.set_datastore_access_privileges(_auth_mgr.conn, [privileges])
       
     if error_info:
         return operation_fail(error_info)
     else:
         print "tenant access add succeeded"
 
+def check_tenant_access_add_args(privileges):
```
pdhamdhere: This validation should be part of API rather than just Admin CLI so inputs coming via API/UI can be validated as well.

msterin:
@shaominchen  - based on our offline discussion, this PR is not current anymore. Besides it needs conflict resolution :-). Do you want to close this PR for now or do you want to address the comments and do another pass of reviews ?

shaominchen:
This PR is being hold for now. I will pick it up in December.

Issue #650, opened 2016-10-26
Support drone cancel


kerneltime:
We need a post drone run webhook to clean up setup and locks.
This is needed since now the ability to cancel a build is public.
This issue will be resolved if any developer can press cancel at any stage of the test turn.

kerneltime:
https://github.com/vmware/docker-volume-vsphere/pull/736

Issue #648, opened 2016-10-26
A test run should start with clean slate


msterin:
Since we serialize test runs on the same test bed, we should start each run as a clean slate. 
This will prevent people being blocked by bug not related to current PR

We had many cases when a failed test for some PR was blocking other / unrelated PR CI runs.

The suggestion is to clean up all volumes at the end of each run. If there are some volumes to cleanup that indicates that a unit test or a test did not clean up - so we should YELL and fail a run if the cleanup was performed.

kerneltime:
Related to https://github.com/vmware/docker-volume-vsphere/issues/508

PR #736, opened 2016-11-15
Address testbed setup and concurrency.

kerneltime:
This change will use snapshot and restore on each test run.
It will also allow any ongoing build to be cancelled.

Testing:
- Ran concurrent builds, build serialized in strict build number order
- Was able to cancel a build without impacting other builds
- Each run started from the same initial state.


kerneltime:
Cancel is now supported, Restart will be blocked and not impact other builds.

kerneltime:
Restart testing https://ci.vmware.run/vmware/docker-volume-vsphere/360

kerneltime:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/650

kerneltime:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/648

PR #739, opened 2016-11-16
Run pylint as a part of the ESX build

msterin:
Pylint (in a container) is run automatically on 'make dockerbuild' (default target) and in Drone.
If you want to run it on your box with no containers, "make pylint".

Currently it will fail the build if only there are errors (e.g. wrong arg count, etc...). It ignores all the rest (esx_service/Makefile is setting 'PYLINT_IGNORE := --errors-only -d import-error,no-name-in-module')

Going forward , we will  make it more strict up to and including failing on naming convention violation, whitespaces and other PEP8 formatting exectations

testing: 
* manually introduced wrong arg counts and made sure lint fails. 
* CI


pdhamdhere:
>Pylint (in a container) is run automatically on 'make dockerbuild' (default target) and in Drone.
If you want to run it on your box with no containers, "make pylint".

Does it mean,simple "make" on my dev box won't run pylint? IMO, it should run by default and not restricted to CI runs triggered by drone.

misc/dockerfiles/Dockerfile.pylint
```diff
@@ -1,14 +1,16 @@
-FROM python:2-alpine
-MAINTAINER "EEA: IDM2 A-Team" <eea-edw-a-team-alerts@googlegroups.com>
+#
+# Dockerfile for running pylint
+#
+FROM alpine
+MAINTAINER "CNA Storage Team" <cnastorage@vmware.com>
```
pdhamdhere: cna-storage@vmware.com

esx_service/pylint_cfg.ini
```diff
@@ -0,0 +1,355 @@
+[MASTER]
```
pdhamdhere: Is it tuned to check for both Python 2.x and Python 3.x?

misc/dockerfiles/Dockerfile.pylint
msterin: oops. thanks !

msterin:
> Does it mean,simple "make" on my dev box won't run pylint? IMO, it should run by default and not restricted to CI runs triggered by drone

'make'  will run pylint because 'make' runs default target (dockerbuild) and `Pylint (in a container) is run automatically on 'make dockerbuild' (default target) `

Makefile
```diff
@@ -68,3 +68,6 @@ endif
 # redirect all
 $(TARGETS):
 	$(MAKE) --directory=vmdk_plugin $@
+
+.DEFAULT:
```
kerneltime: What is the role of this target given that we define the default target earlier in the file?

misc/dockerfiles/Dockerfile.pylint
```diff
@@ -1,14 +1,16 @@
...
+RUN wget --no-check-certificate -O - https://bootstrap.pypa.io/get-pip.py | python 
+RUN pip install --upgrade pip pylint pyvmomi pyvim 
 
 #COPY pylint.cfg /etc/pylint.cfg
-
 #ENTRYPOINT ["pylint", "--rcfile", "/etc/pylint.cfg"]
-ENTRYPOINT ["pylint"]
 
-CMD ["/code"]
+
```
kerneltime: Get ride of the empty lines.

msterin: will do

Makefile
msterin: "default target" earlier is the target which is invoked by default  :-) (i.e. on `make` command, with no targets. `.DEFAULT:` is rules of last resort - executed when target is passed but not found in Makefile. Real good for fall through.

https://www.gnu.org/software/make/manual/html_node/Special-Targets.html. 
```
.DEFAULT
The recipe specified for .DEFAULT is used for any target for which no rules are found (either
explicit rules or implicit rules). See Last Resort. If a .DEFAULT recipe is specified, every file
mentioned as a prerequisite, but not as a target in a rule, will have that recipe executed on its
behalf. See Implicit Rule Search Algorithm.

PR #754, opened 2016-11-19
VMODL API for multi-tenancy support

shaominchen:
The implementation (to integrate with auth_api) has not been implemented yet. Please review the API signatures first.


esx_service/vmodl/DatastoreAccessPrivilege.py
```diff
@@ -0,0 +1,93 @@
...
+   Indicates whether the tenant is allowed to create (or delete) a volume on this datastore
+   """
+   )
+   @Attribute(parent=_name, typ="boolean")
+   def allow_create(self):
+       pass
+
+   @JavaDocs(parent=_name, docs =
+   """
+   Max volume size allowed on this datastore
```
msterin: need units. I suggest the attribute name to have "in_bytes" or whatever at the end. Actually, in_Mib should be fine

esx_service/vmodl/DatastoreAccessPrivilege.py
```diff
@@ -0,0 +1,93 @@
...
+   def max_volume_size(self):
+       pass
+
+   @JavaDocs(parent=_name, docs =
+   """
+   Total storage usage allowed on this datastore
+   """
+   )
+   @Attribute(parent=_name, typ="string")
+   def usage_quota(self):
```
msterin: _in_MiB ?

esx_service/vmodl/DatastoreAccessPrivilege.py
```diff
@@ -0,0 +1,93 @@
...
+   """
+   )
+   
+   @DataType(name=_name, version=_VERSION)
+   def __init__(self):
+       pass
+
+   @JavaDocs(parent=_name, docs =
+   """
+   Datastore name
```
msterin: what if the datastore name changes (and url stays the same) ? Should we keep url here  ?

esx_service/vmodl/Tenant.py
```diff
@@ -0,0 +1,210 @@
...
+   Tenant uuid
+   """
+   )
+   @Attribute(parent=_name, typ="string")
+   def id(self):
+       pass
+
+   @JavaDocs(parent=_name, docs =
+   """
+   Tenant name
```
msterin: Any limitations ? Length ? Char set ? Can it be unicode ?

esx_service/vmodl/Tenant.py
```diff
@@ -0,0 +1,210 @@
...
+   def privileges(self):
+       pass
+
+   @JavaDocs(parent=_name, docs=
+   """
+   Add VMs to a tenant
+   """
+   )
+   @Method(parent=_name, wsdlName="AddVMs")
+   @Param(name="vms", typ="string[]")
```
msterin: a question abut all *VMs() operations: should we use vim.VirtualMachine MoRef instead of strings here ?

esx_service/vmodl/TenantManager.py
```diff
@@ -0,0 +1,103 @@
...
+   def CreateTenant(self, name, description=None):
+       pass
+
+   @JavaDocs(parent=_name, docs=
+   """
+   Remove a Tenant instance
+   """
+   )
+   @Method(parent=_name, wsdlName="RemoveTenant")
+   @Param(name="name", typ="string")
```
msterin: vim.vcs.Tenant instead of string here ?

esx_service/vmodl/TenantManager.py
```diff
@@ -0,0 +1,103 @@
...
+   @Param(name="remove_volumes", typ="boolean")
+   @Return(typ="void")
+   def RemoveTenant(self, name):
+       pass
+
+   @JavaDocs(parent=_name, docs=
+   """
+   Query Tenants for the given name. Return all Tenants if name is null.
```
msterin: will we support globbing (e.g. ? *) If yes needs to be documented

esx_service/vmodl/TenantManager.py
```diff
@@ -0,0 +1,103 @@
...
+   """
+   Modify the attributes of a Tenant instance
+   """
+   )
+   @Method(parent=_name, wsdlName="RemoveTenant")
+   @Param(name="name", typ="string")
+   @Param(name="new_name", typ="string", flags=F_OPTIONAL)
+   @Param(name="new_description", typ="string", flags=F_OPTIONAL)
+   @Return(typ="void")
+   def RemoveTenant(self, name):
```
msterin: needs to be ModifyTenant

esx_service/vmodl/VsphereContainerService.py
```diff
@@ -0,0 +1,65 @@
...
+         return f
+      return Decorate
+   def Internal(parent):
+      def Decorate(f):
+         return f
+      return Decorate
+
+
+# _VERSION = newestVersions.Get("vim")
+_VERSION = 'vim.version.version10'
```
msterin: I think common modules import and version should be imported from a single file instead of copy-pasted to each module

esx_service/vmodl/DatastoreAccessPrivilege.py
shaominchen: Good catch! I will use datastore URL instead. BTW, this issue already exists in the existing logic. I have filed a tracking issue: #782.

esx_service/vmodl/DatastoreAccessPrivilege.py
shaominchen: Sure, will add _in_MiB suffix.

esx_service/vmodl/DatastoreAccessPrivilege.py
shaominchen: This PR will include VMODL definitions only. In the next PR I will provide the implementation along with the unit tests.

Will add _in_MiB suffix.

esx_service/vmodl/Tenant.py
shaominchen: I will specify the length of the string. AFAIK, VMODL string supports unicode on all different language bindings: C++, Java, Python, etc. But this is implicit and we don't usually document this in the JavaDoc.

esx_service/vmodl/Tenant.py
shaominchen: Will use MoRef consistently for all references of VMs and Datastores.

esx_service/vmodl/TenantManager.py
shaominchen: Copy-paste error. Will fix.

esx_service/vmodl/TenantManager.py
shaominchen: Since all Tenant related APIs use tenant name as the unique key, I think it makes sense to also use name as the key for this API. What do you think?

esx_service/vmodl/TenantManager.py
shaominchen: I've thought about this. Globbing is typically used by shells for matching file and directory names using wildcards. Even though our Tenant name does map to a directory in the backend, this implementation detail should not be exposed to customers. Regular expressions are actually more commonly used in commands / functions for pattern matching in string text. However, supporting regex seems an overkill for our current requirements.

So, for now I think we can keep it simple, i.e. not supporting globbing or regex. In the future if there are real requirements for this, we can add this feature without breaking backward compatibility.

esx_service/vmodl/VsphereContainerService.py
shaominchen: Make sense, will do.

esx_service/vmodl/Tenant.py
```diff
@@ -0,0 +1,210 @@
...
+
+# Vmodl Names
+
+class Tenant:
+   ''' 
+   Tenant is an abstraction of a group of VMs and the associated privileges on specific datastores.
+   This class provides operations to manage VMs and privileges associated with a tenant.
+   '''
+
+   _name = "vim.vcs.TenantManager"
```
shaominchen: Thanks for catching this. Will fix.

esx_service/vmodl/Tenant.py
```diff
@@ -0,0 +1,210 @@
...
+
+   @JavaDocs(parent=_name, docs=
+   """
+   Remove VMs from a tenant
+   """
+   )
+   @Method(parent=_name, wsdlName="RemoveVMs")
+   @Param(name="vms", typ="string[]")
+   @Return(typ='void')
+   def RemoveVMsFromTenant(self, vms):
```
shaominchen: Will fix.

esx_service/vmodl/VsphereContainerService.py
shaominchen: I've decided to merge all classes into one single file/module as this seems more consistent with Python style. So no need to introduce a common module anymore.

shaominchen:
@msterin @lipingxue 

I've updated the PR based on your comments. Also redesigned some APIs. Major changes since last update:
1. Add detailed comments for all classes, methods, fields, parameters, return values and exceptions.
2. Add exception handling for all VMODL APIs
3. Use MoRef for all Datastore and VM references
4. Merge all classes into one single file/module

Exception handling is a major missing part that we haven't discussed before. I plan to introduce one or more exceptions to handle all different error conditions. Maybe we should have a discussion about this after everyone is back.

esx_service/vmodl/VsphereContainerService.py
```diff
@@ -0,0 +1,377 @@
...
+
+   @JavaDocs(parent=_name, docs=
+   """
+   Remove a Tenant instance. Remove all volumes created by this tenant
+   if remove_volumes flag is set to True.
+   """
+   )
+   @Method(parent=_name, wsdlName="RemoveTenant")
+   @Param(name="name", typ="string")
+   @Param(name="remove_volumes", typ="boolean")
```
shaominchen: Will address this.

esx_service/vmodl/VsphereContainerService.py
```diff
@@ -0,0 +1,377 @@
...
+   @Return(typ="void")
+   def ModifyTenant(self, name, new_name=None, new_description=None):
+       pass
+
+class Tenant:
+       _name = "vim.vcs.Tenant"
+
+   @JavaDocs(parent=_name, docs =
+   """
+   Tenant is an abstraction of a group of VMs and the associated access
```
shaominchen: Will clarify a little bit.

shaominchen:
@lipingxue Addressed your comments.

esx_service/vmodl/VsphereContainerService.py
```diff
@@ -0,0 +1,400 @@
...
+   @datastore If set, return privilege for this datastore only; otherwise return all privileges.
+   @return Privilege of this tenant on the given datastore.
+   @throw NotFound If this tenant has no privilege on the given datastore.
+   """
+   )
+   @Method(parent=_name, wsdlName="GetPrivileges",
+        faults=["vim.fault.NotFound"])
+   @Param(name="datastore", typ="vim.Datastore", flags=F_OPTIONAL)
+   @Return(typ='vim.vcs.storage.DatastoreAccessPrivilege[]')
+   def GetPrivileges(self, datastore):
```
msterin: nit: it feels strange that AddPrivilige is singular but GetPrivileges is plural. I suggest to stick with singular everywhere.

lipingxue:
LGTM to me.

esx_service/vmodl/VsphereContainerService.py
shaominchen: I'm actually following existing VSAN VMODL style. Basically, if the return type is an array, the name will be plural, for example: GetTenants, GetVMs, GetPrivileges. Or if the input parameter is an array, the name is also plural: AddVMs, RemoveVMs, ReplaceVMs.

Please let me know if you still have concern about this. I can change it in the next PR if needed.

Issue #608, opened 2016-10-04
Docker compatibility (post 1.12) - Do not increment reference count for a volume if the mount fails.


govint:
Fix for docker issue https://github.com/docker/docker/issues/22564 has resulted in https://github.com/docker/docker/pull/27116 to send an unmount to the plugin only if the mount actually succeeded. The plugin.go:Mount() function needs to handle the change by incrementing the refcount on a volume only if the mount succeeded. This change will be needed post 1.12 when the Docker fix becomes available.

govint:
The fix will be submitted once we have the next version >1.12 of docker where this fix should be available.

govint:
Docker PR 27116 is still open and this change can be made once thats in.

kerneltime:
link https://github.com/docker/docker/pull/27116

govint:
Docker issue has been closed and should be available in the upcoming release. Will post PR then.

govint:
Docker 1.1.13 is out and am checking with that if the fix is in it and will post the changes.

govint:
Docker version
Client:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 21:44:32 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.3
 Git commit:   6b644ec
 Built:        Wed Oct 26 21:44:32 2016
 OS/Arch:      linux/amd64

  ---

Command :  Uses -w option, docker doesn't issue an unmount on mount failure

docker run --rm -it -w /data -v mount-err:/data busybox

2016-11-22 21:56:34.903730934 -0800 PST [INFO] Attaching volume and creating filesystem name=mount-err fstype=ext4
2016-11-22 21:56:37.803787313 -0800 PST [INFO] Volume and filesystem created name=mount-err fstype=ext4
2016-11-22 21:57:28.917945144 -0800 PST [INFO] Mounting volume name=mount-err
2016-11-22 21:57:29.459062291 -0800 PST [ERROR] Failed to mount name=mount-err error="Mount error"
2016-11-22 21:57:45.748352137 -0800 PST [INFO] Mounting volume name=mount-err
2016-11-22 21:57:45.748403472 -0800 PST [INFO] Already mounted, skipping mount. name=mount-err refcount=2
[...]

PR #767, opened 2016-11-23
[WIP] Decrement reference count on error in plugin Mount().

govint:
Modified Mount() to decrement volume ref count on error. Mount() already derements ref count in one code path and in another doesn't (within Mount()).

This change is still missing one update - how do we want to handle pre Docker 1.13-0-rc1?

a. Check for Docker version < 1.13.0 - this may be redundant code over time and its expecting a certain behavior thats not documented and a bug in Docker - plugin Unmount() is called even on Mount() error.

b. Always decrement ref count in Unmount() and if less than or equal to zero then set ref count to 0 and unmount the volume.

Please comment which way we want to go with this (b) seems ok.


govint:
Fixes #608

kerneltime:
This change cannot be merged till we do not address the pre Docker 1.13 installs.

kerneltime:
Does the ref count below zero guarantee that there is no mount? I cannot see a scenario where ref count is below zero but the volume is still being accessed. I am fine with b.

kerneltime:
We should move this to https://github.com/vmware/docker-volume-vsphere/issues/608 where can finalize on options.

PR #779, opened 2016-11-27
Include hostd logs on failure only

### LINKED ISSUE #769: Capture the hostd log as part of CI run.
We need to include the hostd log from ESX as part of the run output.

kerneltime:
Due to large log files, the console output gets truncated. I am working with CI folks to bump up the limit but for successful runs, hostd log is a distraction. Thus, only dump hostd logs on failures. 
Testing:
Faked a CI failure https://ci.vmware.run/vmware/docker-volume-vsphere/505

Fixes #769

### DIFF
--- a/misc/drone-scripts/dump_log.sh
+++ b/misc/drone-scripts/dump_log.sh
@@ -22,8 +22,11 @@ HOSTD_LOGFILE="/var/log/hostd.log"
 dump_log_esx() {
   log $ESX_LOGFILE
   $SSH $USER@$1 cat $ESX_LOGFILE
-  log $HOSTD_LOGFILE
-  $SSH $USER@$1 cat $HOSTD_LOGFILE
+  if [ $INCLUDE_HOSTD == "true" ]
+  then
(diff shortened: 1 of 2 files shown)


kerneltime:
@shuklanirdesh82 @ashahi1 @lipingxue

kerneltime:
Looks like the CI console log size will not change without a CI software upgrade. The sw upgrade to 0.5 is not planned yet as it is still not GAed.

misc/drone-scripts/dump_log.sh
```diff
@@ -36,5 +36,8 @@ truncate_vm_logs() {
 
 truncate_esx_logs() {
   $SSH $USER@$1 "echo > $ESX_LOGFILE"
-  $SSH $USER@$1 "echo > $HOSTD_LOGFILE"
+  if [ $INCLUDE_HOSTD == "true" ]
+  this
```
lipingxue: Why do we need this line?

kerneltime:
Thanks, updated

Issue #709, opened 2016-11-06
Tenancy: Creating volume fails in absence of "mount" privilage


pdhamdhere:
Creating volume fails with "No mount privilege" though VM has "create" privilage.
```
# docker volume create -d vmdk --name=vol101 
Error response from daemon: create vol101: VolumeDriver.Create: No mount privilege
```
This VM is member of Tenant- T6
```
[root@localhost:~] vmdkops_admin tenant create --name=T6 --vm-list=test1
tenant create succeeded
[root@localhost:~] vmdkops_admin tenant access add --datastore=datastore1 --rights=create,delete --name=T6
{'mount_volume': 0, 'global_visibility': 0, 'max_volume_size': 0, 'create_volume': 1, 'delete_volume': 1, 'datastore': 'datastore1', 'usage_quota': 0}
tenant access add succeeded
[root@localhost:~] vmdkops_admin tenant access ls --name=T6
Datastore   Create_volume  Delete_volume  Mount_volume  Max_volume_size  Total_size  
----------  -------------  -------------  ------------  ---------------  ----------  
datastore1  True           True           False         Unset            Unset       
```

Creating a volume requires volume to be attached/detached to VM for formatting. Attaching volume to VM requires mount privilege and thus create volume fails with "No mount privilege".

CC/ @lipingxue @msterin

lipingxue:
If the user has the create privilege, does it imply he/she should have mount privileges? @pdhamdhere  @msterin ?

pdhamdhere:
It would be absurd to not have mount privilege when you are allowed to create volume. So, Create implies remove, mount, unmount privilege and `--rights` is obsolete? If datastore access is granted then VMs within that tenant has full access (create, remove, mount, unmount, list, get). Makes sense to me.

msterin:
Let's say I want to provision volumes for a tenant manually and I do not want tenant VMs to use only pre-created volumes. 

So I would say if the access is granted, still can have 2 levels
* _All_ privileges: create/delete/set/mount/unmount
* _Use_ privilege: mount/unmount (attach)

this brings a question "how this volume would be created". Ideally, 'docker volume create -o tenant=<other_tenant>. That brings another concept (not for this PR)
* _root_ (or su): a right to manipulate volumes in other tenants.

lipingxue:
Already fixed, close it.

Issue #621, opened 2016-10-14
Extra work in Tenant authorization - fallbacks and defaults


lipingxue:
1.handling "fallback" tenant (When VM does not belong to any tenant)
2. handing "fallback" privileges (no matching privileges is found in privileges table)
3. handling default datstore (for volume_name without@datastore)

msterin:
Here is  a quick design sketch for the items above, from our discussion yesterday:
# handling "fallback" tenant 
* When a VM which does not belong to any tenant issues a request to vmdk_ops, we will check in auth_db for the presense of "__FALLBACK" tenant. 
* if this tenant is present, this VM will be assumed to be in __FALLBACK tenant, and will get privileges associated with this tenant
** note that auth_db record will not be added for this VM
* __FALLBACK tenant will be automatically added on DB creation, so by default vmdk_ops will support request from any VM , thus maintaining backward compatibility and simplicity of installation 
* an admin can remove this teanant or modify privileges, thus locking down vmdk_ops to serve only explicitly configured VMs
* __FALLBACK tenant is handle just like all other tenants - all volumes are in dockvols/__FALLBACK and visible only to VMs not in other tenantsa
* if a VM is assigned to another tenant, it loses access to __FALLBACK volumes, just like in any other  VM reassignment to a different tenant

# handing "fallback" privileges 
* When a VM tries to manipulate volumes on a datastore which is not configured in auth_db  (privileges table) , the system will use privileges from "fallback" record in the `privileges` table. This record will have DB name "__FALLBACK" and Vm name '*'. 
* we will create this record on auth_db creation , thus allowing access to ANY datastore by default
* an admin can edit or remove this record, thus locking down the functionality to allow access only to explicitly configured datastores

# handling default datastore (for volume_name without@datastore)
* When a VM addresses the volume using short notation (`volume_name`, without `@datastore`), we currently use the datastore this VM config file is located on (aka "VM datastore"). We also strip `@datastore` from `docker volume ls` output for the volumes located on the VM datastore.  
** With introduction of tenants, this schema will force admin to open access to ALL VM datastores to any VM in the tenant, and will force to keep privileges table contant record for each VM's datastore. 
** as an alternative, user can alsways refer to volumes with a full name `volume@datastore` which makes docker volume user know and remember lots of irrelevant datastore names
* to address this, we will introduce "default datastore" for a tenant.
[...]

PR #788, opened 2016-12-02
Tenancy: Implement __FALLBACK tenant and __FALLBACK privilege

lipingxue:
This change solved two problems in this issue https://github.com/vmware/docker-volume-vsphere/issues/621.
- support __FALLBACK tenant
- support __FALLBACK privilege
The remain problem "support default_datastore" will be addressed in my next PR. In my next PR, I plan to  address "support default_datastore" and privilege changes (https://github.com/vmware/docker-volume-vsphere/issues/709)

This change also include necessary unit test to support __FALLBACK tenant and __FALLBACK privilege.


lipingxue:
@msterin @shaominchen

esx_service/utils/error_code.py
```diff
@@ -18,6 +18,8 @@
 VM_NOT_BELONG_TO_TENANT = "VM {0} does not belong to any tenant"
 TENANT_NOT_EXIST = "Tenant {0} does not exist"
 TENANT_ALREADY_EXIST = "Tenant {0} already exists"
+TENANT_CREATE_FAILED = "Tenant {0} create failed with err:{1}"
+TENANT_SET_ACCESS_PRIVILEGES_FAILED = "Tenant {0} set access privileges on datastore {1} failed with {2}"
```
shaominchen: Nit: Let's keep the error messages consistent: either "failed with {}" or "failed with error: {}"

esx_service/vmdk_ops.py
```diff
@@ -252,6 +252,7 @@ def cloneVMDK(vm_name, vmdk_path, opts={}, vm_uuid=None, vm_datastore=None):
         return err("Failed to initialize source volume path {0}: {1}".format(src_path, errMsg))
 
     src_vmdk_path = vmdk_utils.get_vmdk_path(src_path, src_volume)
+    logging.debug("cloneVMDK: src_path=%s, src_volume=%s, src_vmdk_path=%s", src_path, src_volume, src_vmdk_path)
```
shaominchen: The change in this file doesn't seem to be related to __FALLBACK feature. If you want to include these changes, please add some description in the review.

esx_service/utils/auth_data.py
```diff
@@ -336,6 +366,7 @@ def connect(self):
         
         if need_create_table:
             self.create_tables()
+            self.create_fallback_tenant_and_privileges()
```
shaominchen: This doesn't seem to be the right place to create fallbacks. I think we should add a flag to check if need_crreate_fallback. This way, we can handle the "upgrade" more smoothly. Also in the unit test we shouldn't have the logic to create fallbacks.

esx_service/vmdk_ops_test.py
```diff
@@ -685,7 +689,38 @@ class VmdkTenantTestCase(unittest.TestCase):
     datastore_name = None
     datastore_path = None
     
-
+    def create_fallback_tenant_and_privileges(self):
```
shaominchen: As mentioned in above comment, we should check whether fallbacks need to be created even if table already exists. This way we don't need to do this in unit tests.

lipingxue:
@msterin Please review this change.

esx_service/utils/auth.py
```diff
@@ -30,6 +32,8 @@
 CMD_DETACH = 'detach'
 
 SIZE = 'size'
+FALLBACK_TENANT = '__FALLBACK'
```
pdhamdhere: FALLBACK => _DEFAULT or _SYSTEM?

esx_service/utils/auth_data.py
```diff
@@ -308,6 +309,35 @@ def get_auth_db_path(self):
 
         """
         return AUTH_DB_PATH
+    
+    def create_fallback_tenant_and_privileges(self):
+        # create a __FALLBACK tenant
+        error_info, tenant = self.create_tenant(
+                                           name=auth.FALLBACK_TENANT, 
+                                           description="This is a fallback tenant",
```
pdhamdhere: "This is a system _default tenant"?

esx_service/utils/auth_data.py
```diff
@@ -308,6 +309,35 @@ def get_auth_db_path(self):
 
         """
         return AUTH_DB_PATH
+    
+    def create_fallback_tenant_and_privileges(self):
```
pdhamdhere: +1 on splitting into 2 separate methods.

lipingxue:
@pdhamdhere 
__FALLBACK tenant and __FALLBACK privilege are created by system. Admin can  remove the __FALLBACK tenant. Admin can also remove or modify __FALLBACK privilege.

esx_service/utils/auth.py
```diff
@@ -84,7 +149,18 @@ def get_tenant(vm_uuid):
             tenant_name = result[0]
             logging.debug("Found tenant_uuid %s, tenant_name %s", tenant_uuid, tenant_name)
 
-    return None, tenant_uuid, tenant_name
+        return None, tenant_uuid, tenant_name
+    else:
+        error_info, tenant_uuid, tenant_name = get_fallback_tenant()
```
msterin: get_fallback_tenant() seems to be doing pretty much the dame as the code above (line 136-151). I think it would be better wrap it in a function get_tenant_from_db(where_clause, key) , and use it with something like.
```python
get_tenant_from_db("id = ?", (tenant_uuid, ))
```

Basically I think copy-n-paste is evil :-)

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
+    _auth_mgr = get_auth_mgr()
+    privileges = []
+    logging.debug("get_fallback_privileges")
+    try:
+        cur = _auth_mgr.conn.execute(
+            "SELECT * FROM privileges WHERE datastore = ?",
+            (FALLBACK_DS,)
+            )
+        privileges = cur.fetchone()
+    except sqlite3.Error as e:
```
msterin: It feels the exceptions are handled too often. I think yu should handle them much higher up in the stack. Let's discuss offline

esx_service/utils/auth_data_test.py
```diff
@@ -476,19 +485,19 @@ def test_list_tenants(self):
...
-                                 tenants_list[1].default_privileges[0][auth_data_const.COL_MOUNT_VOLUME],
-                                 tenants_list[1].default_privileges[0][auth_data_const.COL_MAX_VOLUME_SIZE],
-                                 tenants_list[1].default_privileges[0][auth_data_const.COL_USAGE_QUOTA]
+        tenant2_actual_output = [tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_TENANT_ID],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_DATASTORE],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_CREATE_VOLUME],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_DELETE_VOLUME],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_MOUNT_VOLUME],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_MAX_VOLUME_SIZE],
+                                 tenants_list[tenant2_idx].default_privileges[0][auth_data_const.COL_USAGE_QUOTA]
```
msterin: Nice seeing unit tests in the PR ! 

A suggestion: I think you need to check for fallback tenant and priviledge being created in the DB as a separate test. And I think there should be a test with deleting the fallback and making sure ops from nontenant fails

esx_service/utils/vmdk_utils.py
```diff
@@ -248,6 +248,15 @@ def get_vm_uuid_by_name(vm_name):
     except:
         return None
 
+def get_vm_name_by_uuid(vm_uuid):
+    """ Returns vm_name for given vm_uuid, or None """
+    si = vmdk_ops.get_si()
+    try:
+        vm = [d for d in si.content.rootFolder.childEntity[0].vmFolder.childEntity if d.config.uuid == vm_uuid]
+        return vm[0].config.name
```
msterin: can it be more that 1 element in the list ?

esx_service/vmdk_ops_test.py
```diff
@@ -251,6 +251,7 @@ def testPolicy(self):
             err = vmdk_ops.removeVMDK(vmdk_path)
             self.assertEqual(err == None, unit[2], err)
 
+@unittest.skip("Liping remove")
```
msterin: did you want to remove it :-) ?

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
+            )
+        result = cur.fetchone()
+    except sqlite3.Error as e:
+        error_info = "Error {0} when querying from tenants table for tenant {1}".format(e, FALLBACK_TENANT)
+        logging.error(error_info)
+        return str(e), None, None
+    if result:
+        # found __FALLBACK tenant
+        tenant_uuid = result[0]
+        logging.debug("Found __FALLBACK tenant, tenant_uuid %s, tenant_name %s", tenant_uuid, FALLBACK_TENANT)
```
lipingxue: Done

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
+        return None, tenant_uuid, FALLBACK_TENANT
+    else:
+        # cannot find __FALLBACK tenant
+        logging.debug(error_code.TENANT_NOT_EXIST.format(FALLBACK_TENANT))
+        return None, None, None
+
+
+def get_fallback_privileges():
```
lipingxue: Done

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
+    privileges = []
+    logging.debug("get_fallback_privileges")
+    try:
+        cur = _auth_mgr.conn.execute(
+            "SELECT * FROM privileges WHERE datastore = ?",
+            (FALLBACK_DS,)
+            )
+        privileges = cur.fetchone()
+    except sqlite3.Error as e:
+        error_info = "Error {0} when querying privileges table for __FALLBACK privilege".format(e)
```
lipingxue: Done.

esx_service/utils/auth.py
lipingxue: Yes. This should just be one entry for fallback privilege.  But in this entry, they have different privileges (create, remove, and mount). That is why I use "privileges".

esx_service/utils/auth.py
```diff
@@ -84,7 +149,18 @@ def get_tenant(vm_uuid):
             tenant_name = result[0]
             logging.debug("Found tenant_uuid %s, tenant_name %s", tenant_uuid, tenant_name)
 
-    return None, tenant_uuid, tenant_name
+        return None, tenant_uuid, tenant_name
+    else:
+        error_info, tenant_uuid, tenant_name = get_fallback_tenant()
+        if error_info:
+            return error_info, None, None
+        if not tenant_uuid:
```
lipingxue: I will keep it as is.

esx_service/utils/auth_data.py
lipingxue: Done.

esx_service/utils/auth_data.py
lipingxue: fallback has nothing to do with upgrade.
fallback tenant and fallback privilege will be created automatically when table was first created. 
upgrade happens when we detect that there exist a DB file, but we found the version is not the latest, then we should have some code to upgrade the DB table to the latest. Those part of code is not here yet.

Since user can remove the fallback tenant and fallback privilege, in unit test, we have some test to test the behavior when fallback tenant and fallback privileges are removed. And after our test, we want to return back to the status that fallback tenant and fallback privileges are still in the DB table.

esx_service/utils/error_code.py
```diff
@@ -18,6 +18,8 @@
 VM_NOT_BELONG_TO_TENANT = "VM {0} does not belong to any tenant"
 TENANT_NOT_EXIST = "Tenant {0} does not exist"
 TENANT_ALREADY_EXIST = "Tenant {0} already exists"
+TENANT_CREATE_FAILED = "Tenant {0} create failed with err:{1}"
```
lipingxue: Done.

esx_service/vmdk_ops.py
lipingxue: Yes. This is not related to FALLBACK feature. The reason I add this debug log is I found it is very difficult when I debug FALLBACK feature, the unit test failed in cloneVMDK related test. I think the log is helpful for others to debug if they hit errors in this part.

esx_service/vmdk_ops_test.py
lipingxue: Same as above.

esx_service/utils/auth_data.py
lipingxue: I will keep it as FALLBACK.

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
     return thread_local._auth_mgr
 
+def get_fallback_tenant():
+    """ 
+        Get __FALLBACK tenant by querying the auth DB.
+        VMs which does not belong to any tenant explicitly will 
+        be assigned to __FALLBACK tenant if __FALLBACK tenant exists
+        Return value:
+        -- error_info: return None on success or error info on failure
+        -- tenant_uuid: return __FALLBACK tenant uuid on success,
```
lipingxue: Done.

esx_service/utils/auth.py
```diff
@@ -42,6 +46,67 @@ def get_auth_mgr():
...
+        -- error_info: return None on success or error info on failure
+        -- tenant_uuid: return __FALLBACK tenant uuid on success,  
+           return None on failure or __FALLBACK tenant does not exist
+        -- tenant_name: return __FALLBACK tenant name on success,
+           return None on failure or __FALLBACK tenant does not exist
+    """
+    tenant_uuid = None
+    _auth_mgr = get_auth_mgr()
+    try:
+        cur = _auth_mgr.conn.execute(
```
lipingxue: I am not going to address it in this PR, but will file an issue to track this enhancement.

esx_service/utils/auth.py
lipingxue: We have discussed offline.

esx_service/utils/auth.py
lipingxue: These two place are different. For normal tenant, from vm_uuid, we query vms table to find the tenant_uuid of the tenant which own this VM. Then we query the tenant table by id to get tenant_name.

For FALLBACK tenant, we know the tenant name is  "__FALLBACK". We then query the tenant table by name to get tenant_uuid.

esx_service/utils/auth_data_test.py
lipingxue: "test_vmdkops_on_fallback_tenant_vm" did this.
1. It test the create, attach, detach, remove command when FALLBACK tenant and FALLBACK privileges are present
2. It then test the vmdk ops after removing FALLBACK tenant and FALLBACK privilege

What do you mean "you need to check for fallback tenant and privileges being created in the DB as a separate test"? 
FALLBACK tenant and privileges are created when table is created.

esx_service/utils/auth.py
shaominchen: Seems like it's still the same?

esx_service/utils/vmdk_utils.py
lipingxue: I don't think so. I think the vm name should be unique?

esx_service/utils/auth_data_test.py
msterin: > What do you mean "you need to check for fallback tenant and privileges being created in the DB as a separate test"?
FALLBACK tenant and privileges are created when table is created.

is there a sanity check that validates it ? (low priority)

esx_service/utils/auth.py
shaominchen: I think we should differentiate the term "privilege" and "right". IIUC, in our service the term "privilege" indicates some rights (create, remove, mount, etc.) on one datastore. One tenant can have multiple "privileges" on different datastores, but there will be one and only one privilege (which includes multiple "rights") on one single datastore.

esx_service/utils/auth.py
shaominchen: Seems like it's still the same?

esx_service/utils/auth.py
shaominchen: Can you please explain a little bit?

esx_service/utils/auth_data_test.py
lipingxue: In the setUp() function of class "VmdkTenantTestCase", I have the code to check whether FALLBACK tenant and FALLBACK privilege are missing. It they are missing,  the code will try to create them. See function "create_fallback_tenant_and_privileges".

esx_service/utils/vmdk_utils.py
msterin: > I think the vm name should be unique?

no they are not

lipingxue: Then how do we handle this? Assume we already has an VM named "vm1", then I try to create another vm also named "vm1". Then how can I tell which VM is the one I just created?

lipingxue: I think in one esx, the vm name should be unique, right?

esx_service/utils/auth.py
lipingxue: I think it is different coding style. No major advantage to make the change.

lipingxue:
I have adressed all comments. Please take a look.

esx_service/utils/vmdk_utils.py
lipingxue: Made change to "create_vm" and "remove_vm" to avoid using functino "find_vm_by_name".

lipingxue:
CI test failed in "VmdkClone" test. I have added more debug info in the change to debug the CI failure. 

Please ignore the change related to that.
All other comments have been addressed. Please review it. 
@msterin @pdhamdhere

esx_service/utils/log_config.py
```diff
@@ -30,7 +30,7 @@
 # since we rely on it to locate log file name after config is loaded.
 LOG_CONFIG_FILE = "/etc/vmware/vmdkops/log_config.json"
 
-LOG_LEVEL_DEFAULT = 'INFO'
+LOG_LEVEL_DEFAULT = 'DEBUG'
```
msterin: this should be changed back before merge

PR #795, opened 2016-12-05
Update docker-volume-cli.md

shuklanirdesh82:
using 'vsan-policy-name' instead of 'policy' to keep the documentation in sync with the code and consistent with https://github.com/vmware/docker-volume-vsphere/blob/8717d753b606fad41597d074e5ee07a6893ee2f2/docs/user-guide/admin-cli.md#list-4

/CC @kerneltime

### DIFF
--- a/docs/user-guide/docker-volume-cli.md
+++ b/docs/user-guide/docker-volume-cli.md
@@ -20,13 +20,13 @@ The volume units can be ```kb, mb, gb and tb```
 
 The default volume size is 100mb
 
-### policy
+### policy (vsan-policy-name)
 
 ```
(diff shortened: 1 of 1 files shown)


docs/user-guide/docker-volume-cli.md
```diff
@@ -20,13 +20,13 @@ The volume units can be ```kb, mb, gb and tb```
 
 The default volume size is 100mb
 
-### policy
+### policy (vsan-policy-name)
```
kerneltime: Rename this to vsan-policy-name, it should reflect the CLI invocation.

Issue #786, opened 2016-11-30
Address regression in ESX 6.5 support.


kerneltime:
Recent changes part of https://github.com/vmware/docker-volume-vsphere/pull/729 caused regression in 6.5 support

kerneltime:
Fix on it's way

PR #796, opened 2016-12-05
Fix python code 6.5 support.kerneltime

kerneltime:
Will push this in later today based on approval from @msterin


msterin:
merged to unblock 6.5 work

kerneltime:
fixes https://github.com/vmware/docker-volume-vsphere/issues/786

PR #791, opened 2016-12-02
Fix to recnt_test.sh to select line for recovery tests.

govint:
Search for specific line in plugin log for recovery test.

Fixes https://github.com/vmware/docker-volume-vsphere/issues/790

PR #798, opened 2016-12-06
Add photon volumes to the plugin

govint:
Add support for Photon disks in docker vsphere volume plugin. Following changes are made:

1. Code is separated under vmdk_plugin into two drivers - Photon and Vmdk.

2. Common code is still present under vmdk_plugin and driver specific code is in vmdk_plugin/drivers.

3. Refcnt is made into a package under utils and modified to accept a generic driver interface - usable from either driver.

4. New functions added to fs-package to handle ID based device handling and scanning for devices on disk attach.

5. Misc changes to make files and test code to handle the new location of the code.

6. Code supports create, inspect, mount/unmount of 

Tests:
-------

1. CI fails with issue related refcount tests, thats handled via another PR #791.

2. Tested manual creation, inspect, run, remove of volumes and containers.

3. Verified refcount recovery
a. Start container with photon disk
b. kill plugin
c. Restart plugin and it correctly identifies containers with photon disks and re-initializes ref counts for those.

4. Test method is:
a. Deploy photon installer
b. Create Photon mgmt deployment via installer
c. Create a tenant, project and ticket in photon mgmt plane.
d. Upload a photon OS image to photon controller and create a VM with that
e. Build changes and deploy to VM (from step d)
f. Stop plugin in VM and restart as below (host is for docker-host or the VM which runs the plugin) or add to plugin config file,
/usr/local/bin/docker-volume-vsphere -driver=photon -target=http://<photon controller mgmt VM IP>:9000 -project=<photon project ID> -host=<photon VM ID>
g. Run docker volume commands

TODO:
1. Will continue running tests with multiple containers and volumes
2. Implement a deferred unmount and detach feature in the plugin. Every unmount detaches the disk from the VM and needs a fresh rescan on the next container run. Instead, defer the unmount so if a container using the same volume starts within a time interval on the same docker host the attach, rescan can be avoided.
3. CI needs a photon deployment so the current plugin tests can be verified.


vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -32,58 +32,74 @@ import (
...
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/refcount"
 	"golang.org/x/exp/inotify"
 )
 
 const (
 	devWaitTimeout   = 1 * time.Second
-	mountRoot        = "/mnt/vmdk" // VMDK block devices are mounted here
 	sleepBeforeMount = 1 * time.Second
 	watchPath        = "/dev/disk/by-path"
+	version          = "VMDK Volume Driver v0.3"
```
kerneltime: We should add a Makefile command to update this on each build?

vmdk_plugin/drivers/vmdk/vmdkops/esx_vmdkcmd.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 /*
-#cgo CFLAGS: -I ../../esx_service/vmci
+#cgo CFLAGS: -I ../../../../esx_service/vmci
```
msterin: nit: this line is getting uglier by the minute :-( ...

vmdk_plugin/main.go
```diff
@@ -91,40 +91,62 @@ func logInit(logLevel *string, logFile *string, configFile *string) bool {
...
+	log.WithFields(log.Fields{
+		"target":  *targetURL,
+		"project": *projectID,
+		"host":    *vmID,
+		"port":    *port}).Info("Plugin options - ")
+
+	if *driverName == "photon" {
+		driver = photon.NewVolumeDriver(*targetURL, *projectID,
+			*vmID, mountRoot)
+	} else if *driverName == "vmdk" {
```
msterin: nit: 
should be support "vsphere" alias for "vmdks" too ? this way it will be symmetric 
-d vsphere
-d photon
?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -265,13 +301,14 @@ func (r refCountsMap) syncMountsWithRefCounters(d *vmdkDriver) {
 			if cnt.count == 0 {
 				// Volume mounted but not used - UNMOUNT and DETACH !
 				log.WithFields(f).Info("Initiating recovery unmount. ")
-				err := d.unmountVolume(vol)
-				if err != nil {
+				resp := d.Unmount(volume.UnmountRequest{Name: vol})
+				if resp.Err != "" {
 					log.Warning("Failed to unmount - manual recovery may be needed")
 				}
+			} else if driverName == "photon" {
```
msterin: we probably should use a constant for this .

vmdk_plugin/main.go
govint: Agree, @pdhamdhere, @kerneltime?

pdhamdhere: Yes. I was always strong proponent of "vSphere". Do we have to keep "VMDk" too for backward compatibility?

msterin: We do not have, but I think we should . It's also 50℅ less typing :-)

kerneltime: We should move to vsphere, till we hit GA we can make changes as long as we document (maybe support backward compatibility if feasible, we can print a message stating the deprecation and remove it at some point the future)

misc/scripts/build.sh
```diff
@@ -115,7 +115,7 @@ then
      -v $PWD:$dir -w $dir $plug_container $MAKE_UI $2
 elif [ "$1" == "gvt" ]
 then
-  $DOCKER run --rm -it -v $PWD/..:$dir -w $dir $go_container bash -c "go get -u github.com/FiloSottile/gvt; bash"
+  $DOCKER run --rm -it -v $PWD/..:$dir -w $dir --net=host $go_container bash -c "go get -u github.com/FiloSottile/gvt; bash"
```
kerneltime: Why this change? and Why only here?

vendor/manifest
```diff
@@ -218,6 +218,14 @@
 			"branch": "master"
 		},
 		{
+			"importpath": "github.com/vmware/photon-controller-go-sdk",
```
kerneltime: We should check with photon team if they want us to use master or a different point in time

vmdk_plugin/main.go
```diff
@@ -91,40 +91,62 @@ func logInit(logLevel *string, logFile *string, configFile *string) bool {
...
+	}).Info("Starting plugin ")
+
+	log.WithFields(log.Fields{
+		"target":  *targetURL,
+		"project": *projectID,
+		"host":    *vmID,
+		"port":    *port}).Info("Plugin options - ")
+
+	if *driverName == "photon" {
+		driver = photon.NewVolumeDriver(*targetURL, *projectID,
```
kerneltime: If the params are optional, how do we validate that they have been provided?

vmdk_plugin/Makefile
```diff
@@ -83,12 +83,14 @@ GO := GO15VENDOREXPERIMENT=1 go
...
 # make sure we rebuild of vmkdops or Dockerfile change (since we develop them together)
-VMDKOPS_MODULE := vmdkops
+VMDKOPS_MODULE := drivers/vmdk/vmdkops
+VMDKOPS_TEST_MODULE := vmdkops
 VMDKOPS_MODULE_SRC = $(VMDKOPS_MODULE)/*.go $(VMCI_SRC)
 
 # All sources. We rebuild if anything changes here
-SRC = plugin.go main.go log_formatter.go refcnt.go \
-	utils/fs/fs.go utils/config/config.go utils/TestInputParamsUtil.go
+SRC = main.go log_formatter.go utils/refcount/refcnt.go \
```
govint: The test file needs to be included in the build. May be test code should get separated out later away from the product code.

vmdk_plugin/main.go
govint: This topic is still not closed and hence the way its been coded. For now if any of these params are not given the client simply fails to connect to the photon controller. Plan is that when PC deploys a VM it will create the config file for the plugin to use.

kerneltime: Do you plan to add validation?

vmdk_plugin/Makefile
kerneltime: Can this be done via wildcard and filter?

vendor/manifest
govint: Will do.

govint:
Closing this PR will open a new one.

Issue #800, opened 2016-12-07
Fix build.sh to also check for testasroot target


ashahi1:
misc/scripts/build.sh is not running make on testasroot target.

Main job of build.shell is to run containers and then do 'make Target' within those containers. 

Due to some earlier  change in build.sh, we had missed to add condition to also run 'testsasroot' target while running our tests locally. 

This issue was not discovered on drone because while running tests on drone, we do not call build.sh

PR #801, opened 2016-12-07
Modified  build.sh to also run testasroot target

ashahi1:
Fixes https://github.com/vmware/docker-volume-vsphere/issues/800

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -122,15 +122,11 @@ then
 elif [ "$1" == "pylint" ]
 then
   $DOCKER run --rm -it -v $PWD/..:$dir -w $dir $pylint_container $MAKE_ESX pylint
-elif [ "$1" == "build" ] || [ "$1" == "clean-as-root" ] || [ "$1" == "testasroot" ]
-# we might have to modify this elif in the future if we decide to add more targets
-then
+else 
   docker_socket=/var/run/docker.sock
   $DOCKER run --privileged --rm -it \
     -e "PKG_VERSION=$PKG_VERSION" \
     -e "INCLUDE_UI=$INCLUDE_UI" \
     -v $docker_socket:$docker_socket  \
     -v $PWD/..:$dir -w $dir $plug_container $MAKE $1
-else
-  echo "Error: unknown build target $1"
 fi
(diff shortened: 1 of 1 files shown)


ashahi1:
misc/scripts/build.sh is not running make on testasroot target.

Main job of build.shell is to run containers and then do 'make Target' within those containers.

Due to some earlier change in build.sh, we had missed to add condition to also run 'testsasroot' target while running our tests locally.

This issue was not discovered on drone because while running tests on drone, we do not call build.sh

kerneltime:
What was the testing done?

misc/scripts/build.sh
```diff
@@ -122,7 +122,8 @@ then
 elif [ "$1" == "pylint" ]
 then
   $DOCKER run --rm -it -v $PWD/..:$dir -w $dir $pylint_container $MAKE_ESX pylint
-elif [ "$1" == "build" ] || [ "$1" == "clean-as-root" ]
+elif [ "$1" == "build" ] || [ "$1" == "clean-as-root" ] || [ "$1" == "testasroot" ]
```
kerneltime: We should make this the default condition, unless a special rule exists run make using the $plug_container and the $MAKE cmd. Change this to ```else``` instead of ```elseif```

ashahi1:
Addressed the review comment of just having an else condition which will accept all targets.

Ran the test locally with ESX and two VMs and tests were working fine.

==========================

I could also see in the log that testasroot target is being executed  - 

../misc/scripts/build.sh testasroot
make: Entering directory `/go/src/github.com/vmware/docker-volume-vsphere/vmdk_plugin'

=> Running target testasroot Wed Dec 7 18:22:29 UTC 2016

GO15VENDOREXPERIMENT=1 go test github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops
ok      github.com/vmware/docker-volume-vsphere/vmdk_plugin/vmdkops     1.050s
GO15VENDOREXPERIMENT=1 go test github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config
ok      github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config        0.005s
make: Leaving directory `/go/src/github.com/vmware/docker-volume-vsphere/vmdk_plugin'

kerneltime:
LGTM post CI run.

PR #802, opened 2016-12-07
Add clone and single vm concurrent tests

### LINKED ISSUE #772: Plugin parallel sanity test needs to include clone and single vm concurrency.
Currently the clone ops are only covered in vmdk_ops unit test, we also need to cover clone ops in parallel sanity test.

This can be addressed with the same method or in the same parallel tests.

---
EDIT:
Updated to include tests for single vm concurrency.
Will address in #802 

//CC @msterin

brunotm:
Fixes #772

### DIFF
--- a/vmdk_plugin/sanity_test.go
+++ b/vmdk_plugin/sanity_test.go
@@ -319,6 +319,7 @@ func TestConcurrency(t *testing.T) {
 
 	// Create master volume for cloning
 	createRequest.Name = masterVolName
+	createRequest.DriverOpts["size"] = "100mb"
 
 	_, err := clients[0].client.VolumeCreate(context.Background(), createRequest)
 	if err != nil {
--- a/vmdk_plugin/sanity_test.go
+++ b/vmdk_plugin/sanity_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/docker/engine-api/types/container"
 	"github.com/docker/engine-api/types/filters"
 	"github.com/docker/engine-api/types/strslice"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config"
 	"golang.org/x/net/context"
 )
@@ -243,9 +244,9 @@ func TestSanity(t *testing.T) {
 }
 
 // Test concurrent volume operations
-// - concurrent create/delete btween different docker hosts
-// - concurrent clone/delete btween different docker hosts
+// - concurrent create/delete between different docker hosts
 // - concurrent create/delete on the same docker host
+// - concurrent clone/delete between different docker hosts
 func TestConcurrency(t *testing.T) {
 
 	clients := getClients(t)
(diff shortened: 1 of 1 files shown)


vmdk_plugin/sanity_test.go
```diff
@@ -240,7 +242,7 @@ func TestSanity(t *testing.T) {
 			"size": "1gb",
 		},
 	}
-	// Create/delete routine
+	// Create/delete goroutine
```
msterin: So the problem here - which I observed on my testbed-  is that if there is  a client  in clients more than once, the code below (go func) fails because ESX rejects reconfigure request on the same VM when another one is in flight. 

and our current Makefiles allow for VM1 and  VM2 (test clients) to be the same - in fact this is how single VM testing is done. 

I'd suggest  checking for this  and warn then skip duplicate clients. Yes, it will skip the parallel test but  will unblock single VM test.

vmdk_plugin/sanity_test.go
```diff
@@ -260,4 +262,54 @@ func TestSanity(t *testing.T) {
...
+	}
+
+	// Remove size option for clone
+	delete(createRequest.DriverOpts, "size")
+
+	// Clone goroutine
+	for idx, elem := range clients {
+		go func(idx int, c *client.Client) {
+			for i := 0; i < parallelClones; i++ {
+				volName := masterVolName + "-clone" + strconv.Itoa(idx) + strconv.Itoa(i)
```
msterin: nit: formatting with fmt. is easier to read IMO.

vmdk_plugin/sanity_test.go
```diff
@@ -260,4 +262,54 @@ func TestSanity(t *testing.T) {
...
+				_, err := c.VolumeCreate(context.Background(), createRequest)
+				results <- err
+				err = c.VolumeRemove(context.Background(), volName)
+				results <- err
+			}
+		}(idx, elem.client)
+	}
+
+	// Read the results from the channel
+	for i := 0; i < len(clients)*parallelClones*2; i++ {
```
msterin: nit: do we need spaces around * ?

vmdk_plugin/sanity_test.go
```diff
@@ -41,6 +41,7 @@ const (
 	defaultTestLogPath = "/tmp/test-docker-volume-vsphere.log"
 	// Number of volumes per client for parallel tests
 	parallelVolumes = 9
+	parallelClones  = 3
```
msterin: nit: both these constants seem prime candidates for test arguments

vmdk_plugin/sanity_test.go
brunotm: I wasn't aware of that possibility, or the fact that esx would error out. But independently of skipping the tests for that specific scenario we should serialize reconfig tasks per vm in vmdk_ops.

vmdk_plugin/sanity_test.go
brunotm: Yes :)

vmdk_plugin/sanity_test.go
brunotm: This is done by gofmt.

vmdk_plugin/sanity_test.go
brunotm: Ok

vmdk_plugin/sanity_test.go
msterin: ignore my comment about * then :-)

brunotm:
@msterin 
I was thinking about also pushing in this PR single vm concurrent tests and address your comments. The corresponding issue was updated to reflect the single vm tests. Unless you think they should go in a different one. :)

msterin:
> @msterin I was thinking about also pushing in this PR single vm concurrent tests and address your comments. The corresponding issue was updated to reflect the single vm tests. Unless you think they should go in a different one. :)

IMO either way is fine as long as there are 2 separate commits

brunotm:
@msterin 
I have included the commit from #810 here, otherwise tests would fail.
Hopefully this PR together with #810, will be the last edge for the concurrency case.

brunotm:
@msterin 
Updated the patch to address the last comment and reverted the commit included from #810.
The CI run might fail without #810.

From my part is good to go.

Thanks.

msterin:
> The CI run might fail without #810.

And it does fail: 
```
	sanity_test.go:189: Successfully connected to tcp://192.168.31.206:2375
	sanity_test.go:313: Same docker host concurrent create/delete test failed, err: Error response from daemon: create volTestP-same20: Post http://%2Frun%2Fdocker%2Fplugins%2Fvmdk.sock/VolumeDriver.Create: http: ContentLength=49 with Body length 0
FAIL
```
Until #810 is in, this one has to wait to avoid breaking the master

brunotm:
@msterin 
Definitely! :)

brunotm:
@msterin 
The concurrent tests are good to go after #810.

PR #810, opened 2016-12-09
Make the plugin thread safe

brunotm:
Update to reflect the current code in this PR.

This makes the plugin thread safe by implementing a per volume locking scheme inside refCountsMap{} and making it fully synchronized, removes the plugin wide locks in mount/umount, and protects the vmci data structures during the comand/response sequence in EsxVmdkCmd.

---
The access to vmci data structures on the plugin side needs to be serialized during the command/response sequence to avoid race conditions in high concurrency scenarios.

Found when stressing concurrency on the same client in #802.

//CC @msterin 

before:
```
=== RUN   TestConcurrency
Running concurrent tests on tcp://vsc02:2375 and tcp://vsc02:2375 (may take a while)...
Running create/delete tests...
--- FAIL: TestConcurrency (20.30s)
	sanity_test.go:190: Successfully connected to tcp://vsc02:2375
	sanity_test.go:190: Successfully connected to tcp://vsc02:2375
	sanity_test.go:287: Parallel test failed, err: Error response from daemon: create volTestP14: Post http://%2Frun%2Fdocker%2Fplugins%2Fvmdk.sock/VolumeDriver.Create: http: ContentLength=44 with Body length 0
FAIL
```

after:
```
=== RUN   TestConcurrency
Running concurrent tests on tcp://vsc02:2375 and tcp://vsc02:2375 (may take a while)...
Running create/delete tests...
Running clone parallel tests...
--- PASS: TestConcurrency (9.89s)
	sanity_test.go:190: Successfully connected to tcp://vsc02:2375
	sanity_test.go:190: Successfully connected to tcp://vsc02:2375
PASS
```

### DIFF
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -343,6 +343,12 @@ func (d *vmdkDriver) Path(r volume.Request) volume.Response {
 
 // Provide a volume to docker container - called once per container start.
 // We need to keep refcount and unmount on refcount drop to 0
+//
+// The serialization of operations per volume is assured by the volume/store
+// of the docker daemon.
+// As long as the refCountsMap is protected is unnecessary to do any locking
+// at this level during create/mount/umount/remove.
+//
 func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Mounting volume ")
 
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -45,10 +45,9 @@ const (
 )
 
 type vmdkDriver struct {
-	m          *sync.Mutex // create() serialization - for future use
 	useMockEsx bool
 	ops        vmdkops.VmdkOps
-	refCounts  refCountsMap
+	refCounts  *refCountsMap
 }
 
 // creates vmdkDriver which to real ESX (useMockEsx=False) or a mock
@@ -56,34 +55,37 @@ func newVmdkDriver(useMockEsx bool) *vmdkDriver {
 	var d *vmdkDriver
 	if useMockEsx {
 		d = &vmdkDriver{
-			m:          &sync.Mutex{},
 			useMockEsx: true,
 			ops:        vmdkops.VmdkOps{Cmd: vmdkops.MockVmdkCmd{}},
-			refCounts:  make(refCountsMap),
+			refCounts:  newRefCountsMap(),
 		}
 	} else {
 		d = &vmdkDriver{
-			m:          &sync.Mutex{},
 			useMockEsx: false,
 			ops: vmdkops.VmdkOps{
 				Cmd: vmdkops.EsxVmdkCmd{
 					Mtx: &sync.Mutex{},
 				},
 			},
-			refCounts: make(refCountsMap),
+			refCounts: newRefCountsMap(),
 		}
 		d.refCounts.Init(d)
 	}
-
 	return d
 }
-func (d *vmdkDriver) getRefCount(vol string) uint           { return d.refCounts.getCount(vol) }
-func (d *vmdkDriver) incrRefCount(vol string) uint          { return d.refCounts.incr(vol) }
+
+// Return the number of references for the given volume
+func (d *vmdkDriver) getRefCount(vol string) uint { return d.refCounts.getCount(vol) }
+
+// Increment the reference count for the given volume
+func (d *vmdkDriver) incrRefCount(vol string) uint { return d.refCounts.incr(vol) }
+
+// Decrement the reference count for the given volume
 func (d *vmdkDriver) decrRefCount(vol string) (uint, error) { return d.refCounts.decr(vol) }
 
+// Returns the given volume mountpoint
 func getMountPoint(volName string) string {
 	return filepath.Join(mountRoot, volName)
-
 }
 
 // Get info about a single volume
@@ -344,9 +346,6 @@ func (d *vmdkDriver) Path(r volume.Request) volume.Response {
 func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Mounting volume ")
 
-	d.m.Lock()
-	defer d.m.Unlock()
-
 	// If the volume is already mounted , just increase the refcount.
 	//
 	// Note: We are deliberately incrementing refcount first, before trying
@@ -409,8 +408,6 @@ func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 // unmount and detach from VM
 func (d *vmdkDriver) Unmount(r volume.UnmountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Unmounting Volume ")
-	d.m.Lock()
-	defer d.m.Unlock()
 
 	// if the volume is still used by other containers, just return OK
 	refcnt, err := d.decrRefCount(r.Name)
--- a/vmdk_plugin/plugin.go
+++ b/vmdk_plugin/plugin.go
@@ -65,8 +65,12 @@ func newVmdkDriver(useMockEsx bool) *vmdkDriver {
 		d = &vmdkDriver{
 			m:          &sync.Mutex{},
 			useMockEsx: false,
-			ops:        vmdkops.VmdkOps{Cmd: vmdkops.EsxVmdkCmd{}},
-			refCounts:  make(refCountsMap),
+			ops: vmdkops.VmdkOps{
+				Cmd: vmdkops.EsxVmdkCmd{
+					Mtx: &sync.Mutex{},
+				},
+			},
+			refCounts: make(refCountsMap),
 		}
 		d.refCounts.Init(d)
 	}
--- a/vmdk_plugin/refcnt.go
+++ b/vmdk_plugin/refcnt.go
@@ -71,6 +71,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/engine-api/client"
@@ -102,8 +103,11 @@ type refCount struct {
 	dev string
 }
 
-// volume name -> {count, mounted, device}
-type refCountsMap map[string]*refCount
+// refCountsMap struct
+type refCountsMap struct {
+	refMap map[string]*refCount // Map of refCounts
+	mtx    *sync.RWMutex        // Synchronizes refCountsMap ops
+}
 
 var (
 	// vmdk or local. We use "vmdk" only in production, but need "local" to
@@ -120,6 +124,21 @@ func init() {
 
 	// defaults: in test they can be overwritten (see sanity_test.go)
 	driverName = "vmdk"
+}
+
+// creates a new refCountsMap
+func newRefCountsMap() *refCountsMap {
+	return &refCountsMap{
+		refMap: make(map[string]*refCount),
+		mtx:    &sync.RWMutex{},
+	}
+}
+
+// Creates a new refCount
+func newRefCount() *refCount {
+	return &refCount{
+		count: 0,
+	}
 }
 
 // Init Refcounts. Discover volume usage refcounts from Docker.
@@ -150,8 +169,11 @@ func (r refCountsMap) Init(d *vmdkDriver) {
 		return
 	}
 
-	log.Infof("Discovered %d volumes in use.", len(r))
-	for name, cnt := range r {
+	// RLocks the refCountsMap
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+	log.Infof("Discovered %d volumes in use.", len(r.refMap))
+	for name, cnt := range r.refMap {
 		log.Infof("Volume name=%s count=%d mounted=%t device='%s'",
 			name, cnt.count, cnt.mounted, cnt.dev)
 	}
@@ -159,9 +181,12 @@ func (r refCountsMap) Init(d *vmdkDriver) {
 
 // Returns ref count for the volume.
 // If volume is not referred (not in the map), return 0
-// NOTE: this assumes the caller holds the lock if we run concurrently
 func (r refCountsMap) getCount(vol string) uint {
-	rc := r[vol]
+	// RLocks the refCountsMap
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+
+	rc := r.refMap[vol]
 	if rc == nil {
 		return 0
 	}
@@ -170,10 +195,14 @@ func (r refCountsMap) getCount(vol string) uint {
 
 // incr refCount for the volume vol. Creates new entry if needed.
 func (r refCountsMap) incr(vol string) uint {
-	rc := r[vol]
+	// Locks the refCountsMap
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	rc := r.refMap[vol]
 	if rc == nil {
-		rc = &refCount{}
-		r[vol] = rc
+		rc = newRefCount()
+		r.refMap[vol] = rc
 	}
 	rc.count++
 	return rc.count
@@ -183,7 +212,11 @@ func (r refCountsMap) incr(vol string) uint {
 // returns -1  for error (and resets count to 0)
 // also deletes the node from the map if refcount drops to 0
 func (r refCountsMap) decr(vol string) (uint, error) {
-	rc := r[vol]
+	// Locks the refCountsMap
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	rc := r.refMap[vol]
 	if rc == nil {
 		return 0, fmt.Errorf("decr: refcount is already 0. name=%s", vol)
 	}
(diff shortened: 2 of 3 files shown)


govint:
Good find, but can the same mutex be used that's used to serialize mount
and unmount ops. OTOH, can the vmci client be made reentrant or remove the
lock used to serialize the mount and unmount ops and keep the new one.

govint:
With this change a mount takes the lock in plugin.go, then one in
esx_command.go and then one more on the esx service side all of which
serialise ops per VM. Possibly have one lock on the client and one on the
esx service.

brunotm:
@govint 
The only locks that can serialize everything for a single docker host while being held is the attach/detach in vmdk_ops (per vm), during which the plugin will also hold the vmdkcmd.Run one (until it gets a response from esx).

As you said, It looks like we can drop the mutex in vmdkDriver which is only being held at attach/detach per volume, and is already being protected by the volume/attach/detach locks in vmdk_ops.

I'll test and update.

brunotm:
@govint @msterin 
While testing with same tests from #802 went fine, we still need a lock for the *RefCount functions that can access/modify the same map concurrently that where previously protected in mount/umout by the vmdkDriver mutex.

We can either keep that (which will synchronize more than the refcount) or drop it and use a mutex specific for the refcount. I think the later is better for being explicit.

What do you think?

brunotm:
Added a commit with a rwmutex for the refcount. We can either use only the first commit or all of them for the second approach.

brunotm:
@msterin @govint 
I have reviewed the change to better conform to the current code and allow the plugin to run as freely as possible while keeping start/stop storm scenarios safe.

From my part is ready for review/merge.

Thanks.

vmdk_plugin/plugin.go
```diff
@@ -45,7 +45,7 @@ const (
 )
 
 type vmdkDriver struct {
-	m          *sync.Mutex // create() serialization - for future use
+	m          *sync.RWMutex // create() serialization - for future use
```
govint: Nit - change the comment to match exactly how this lock is being used.

vmdk_plugin/refcnt.go
```diff
@@ -203,6 +214,16 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
+func (r refCountsMap) getVolMtx(vol string) *sync.Mutex {
+	rc := r[vol]
+	if rc == nil {
+		rc = newRefCount()
+		r[vol] = rc
+	}
+	return rc.Mtx
+}
+
 // enumberates volumes and  builds refCountsMap, then sync with mount info
```
govint: Nit and optional - could add comments in the discover and sync functions that they don't use any locking anywhere when they set the count or mount/unmount volumes. They don't need to as that happens on the init path when the plugin isn't up yet.

brunotm: Left them out for this, and mainly because the refCountsMap doesn't hold the responsibility of synchronisation. I have found another issues with this approach, comment below.

brunotm:
@govint 
After re-reading the code again i understand that exists the possibility of races or stale locks due to the fact that is possible to have a refcount removed from the map while a lock is still being held outside of the path calling decr(). Not only this, but also if we are creating and deleting a volume with the same name.

Probably the best approach is to have the refcount map fully synchronized and keep a count of volume lock holders/waiters.

I'll try to have something to test later today.

govint:
Yes, thats correct, but as long as those races don't leave the refcount in an inconsistent state and clean up ok its not a problem. Like Remove() checks for a zero ref count and if it is goes ahead and removes the volume, while a mount can race and create a refcount map entry and mount the volume. In which case the Remove() should fail. But it will at least not corrupt the refcount map.

Do you see any cases where the refcount map would be trashed in some scenario?

brunotm:
I'm aware that the possibility is low with the current code but there are a few windows open where concurrent ops could hold different locks for the same volume. In principle having a refcount being deleted while it's lock still lives elsewhere is not good.
I already have most of the code for the synchronization of the refcountmap in place, with the lock/waiter count and a locking semantics that doesn't expose the mutex to the caller. I didn't had the opportunity to test so I didn't pushed yet :) but I would like to hear your opinion on it.

brunotm:
@govint 
This last push implements the strategy mentioned before. Tests where successful including the tests from #802 with ~3sec less than the current code with the driver locked for  mount/umount.

Please review. I have a local branch with the previous code if necessary.

vmdk_plugin/refcnt.go
```diff
@@ -192,23 +234,72 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
+	rc := r.refMap[vol]
+	if rc == nil {
+		return fmt.Errorf("Volume %s has no refcount, cannot be unlocked", vol)
+	}
+
+	if rc.lockers == 0 {
+		return fmt.Errorf("Volume %s has no lockers, cannot be unlocked", vol)
+	}
+
+	rc.lockers--
```
govint: Nit, rc.lockers is incremented with the refmap lock only and without the per volume lock. Here its decremented with the per volume lock. Does lockers need the per volume lock?

vmdk_plugin/refcnt.go
```diff
@@ -192,23 +234,72 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
+	if rc.count == 0 && rc.lockers == 0 {
+		delete(r.refMap, vol)
 	}
 	return rc.count, nil
 }
 
+func (r refCountsMap) lockVolume(vol string) {
+	// Lock the refCountsMap
+	r.mtx.Lock()
+
```
govint: Defer unlock of refmap lock?

vmdk_plugin/refcnt.go
```diff
@@ -192,23 +234,72 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
 func (r refCountsMap) discoverAndSync(c *client.Client, d *vmdkDriver) error {
 	// we assume to  have empty refcounts. Let's enforce
-	for name := range r {
-		delete(r, name)
+
+	r.mtx.Lock() // Lock the refCountsMap to purge the refcounts
+	for name := range r.refMap {
+		delete(r.refMap, name)
 	}
+	r.mtx.Unlock() // Unlock.
```
govint: This segment is guaranteed to run without any other threads, do we still need the locks here.

vmdk_plugin/refcnt.go
```diff
@@ -252,7 +343,11 @@ func (r refCountsMap) discoverAndSync(c *client.Client, d *vmdkDriver) error {
 
 // syncronize mount info with refcounts - and unmounts if needed
 func (r refCountsMap) syncMountsWithRefCounters(d *vmdkDriver) {
-	for vol, cnt := range r {
+	// Lock the refCountsMap
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	for vol, cnt := range r.refMap {
```
govint: Same as above are the locks needed here?

vmdk_plugin/refcnt.go
brunotm: No, only the refCountsMap is needed. It is composed this way because if we try to lock the volume and that blocks while holding the refCountsMap lock it will block the whole refCountsMap and driver/plugin.

vmdk_plugin/refcnt.go
brunotm: It has a explicit unlock below. Same reason as the previous comment.

vmdk_plugin/refcnt.go
brunotm: With the current code no. I have protected those sections for future usage, like online recovery. If thats unlikely i can remove them (but they don't hurt anyone :)

vmdk_plugin/refcnt.go
```diff
@@ -192,23 +234,72 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
+}
+
+func (r refCountsMap) unlockVolume(vol string) error {
+	// Lock the refCountsMap
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	rc := r.refMap[vol]
+	if rc == nil {
+		return fmt.Errorf("Volume %s has no refcount, cannot be unlocked", vol)
```
govint: Caller will leave the lock held?

vmdk_plugin/refcnt.go
```diff
@@ -192,23 +234,72 @@ func (r refCountsMap) decr(vol string) (uint, error) {
...
+	}
+
+	if rc.lockers == 0 {
+		return fmt.Errorf("Volume %s has no lockers, cannot be unlocked", vol)
+	}
```
govint: Caller leaves the lock held?

vmdk_plugin/refcnt.go
brunotm: Yes, attention will be needed when touching the init functions in the future. If the online recovery ideia is unlikely we should remove the locks then.

vmdk_plugin/refcnt.go
brunotm: In this case theres noting to unlock because there is no refcount for the volume. The error is there mostly to avoid bad usage.

vmdk_plugin/plugin.go
```diff
@@ -209,6 +221,10 @@ func (d *vmdkDriver) unmountVolume(name string) error {
 // Name and driver specific options passed through to the ESX host
 func (d *vmdkDriver) Create(r volume.Request) volume.Response {
 
+	// Lock this volume for create
+	d.lockVolume(r.Name)
+	defer d.unlockVolume(r.Name)
```
govint: This will essentially create a refmap entry and lock the entry. Once create completes the lock is released and the refmap entry will be removed once create completes. Do we then need the lock for the create? Same for remove, the last unmount will destroy the refmap entry and remove creates and deletes it.

vmdk_plugin/refcnt.go
brunotm: There's noting to unlock. This is for bad usage as above - calling unlockVolume() before lockVolume.
The only way to have a scenario where the lockers == 0 and the mutex is locked is if we bypass the volumeLock() function and lock the volume directly in refMap[vol].mtx. Which wouldn't make any sense.

vmdk_plugin/plugin.go
brunotm: Agreed with the remove part.
For create i think its necessary to protect the create protocol that has several not protected steps within the driver/plugin (create/attach/mkfs/detach)

what do you think?

brunotm: Do you know which is the behaviour ifa container starts with a vol that is being removed? Mount() in parallel with Remove() ? Does the docker daemon protects from this?

vmdk_plugin/refcnt.go
brunotm: Also, calling unlock on a unlocked mutex throws a runtime error and go mutexes doesn't have a IsLocked() or TryLock()

vmdk_plugin/plugin.go
brunotm: After having a second thought, without locking on create we can race while formating the filesystem with the volume attached and another container tries to start with that volume. So locking here is really needed.

govint:
Ok, Docker may allow a Get to go thru to the volume which can find the
volume? Get() too must lock the VM.

govint:
Sorry not VM - volume.

brunotm:
If the daemon allows it, we probably do need to volume lock on Get() too.
As a test i'll try to start a container while formatting a big volume.

msterin:
so to summarize the discussion, we seem to be saying that 
(1) we need to serialize access to a volume in plugin (get/create/mount/unmount)
    - a per-volume lock replaces plugin-wide serialization of ALL volume operations
(2) we still need per-VMDK lock in the esx service side (not related to this PR code, just a note).

Is it about right? Also, talking about VMCI - where exactly it is not thread safe ?

brunotm:
From basic testing the docker daemon seems to protect from this. Trying to start a container with a volume before the formatting is done waits till Create() returns and apparently the same happens for Remove().

I'll remove the locks over create/remove.

brunotm:
@msterin 
>(1) we need to serialize access to a volume in plugin (get/create/mount/unmount)
>- a per-volume lock replaces plugin-wide serialization of ALL volume operations

No, locking will be needed on mount/umount.

> (2) we still need per-VMDK lock in the esx service side (not related to this PR code, just a note).

Yes, this is intra-docker-host.

> Is it about right? Also, talking about VMCI - where exactly it is not thread safe ?

We seem to be overwriting data over parallel requests, like the #802 error.
Synchronizing the vmdkDriver eliminates the problem from tests.

vmdk_plugin/plugin.go
brunotm: @msterin 
The remove case was a bad example.
As much as the daemon seems to be protecting us to start a container with a volume in the middle of Create() i would like to keep the locks there just to be safe.

msterin:
> No, locking will be needed on mount/umount.

sounds good

> > talking about VMCI - where exactly it is not thread safe ?
> We seem to be overwriting data over parallel requests, like the #802 error.

This is most likely due to the simple fact that VMCI code does not track requestId. So if request1 is sent and then request 2 is sent, but reply2 arrives first there is no code mapping it to the correct request.  So while serialization of mount/unmount is hiding the problem (as these are longer-running ops) I suspect there issue will still be there, just not seen often

brunotm:
@govint @msterin 
If it is ok, i would keep the per volume locks on create/mount/umount.
For the init functions i would like to keep the refCountsMap lock for possibly future usage like online recovery, but if that scenario seems unlikely we can drop it.

@msterin 
> This is most likely due to the simple fact that VMCI code does not track requestId. So if request1 is sent and then request 2 is sent, but reply2 arrives first there is no code mapping it to the correct request. So while serialization of mount/unmount is hiding the problem (as these are longer-running ops) I suspect there issue will still be there, just not seen often

Agreed.
The lock on vmdkDriver eliminates this as it serializes the command/response sequence.

brunotm:
@msterin @govint
After inspecting the docker code, they already have a locking by name scheme in the volume store (not driver wide but volume store global) similar with what I was trying to do, that already protects this situations.

So the only thing that's needed from the plugin side are the synchronized refCountsMap and the  vmdkDriver.

I'll update the PR.
Sorry about the noise, I should have done a proper investigation, but it worth the exercise.

brunotm:
@msterin 
In my previous comments about synchronizing the vmdkDriver, i meant EsxVmdkCmd.

brunotm:
@msterin @govint
Do you think this still needs changes before merging?

pdhamdhere:
@msterin Please review.

govint:
Looks good.

vmdk_plugin/refcnt.go
```diff
@@ -150,18 +169,24 @@ func (r refCountsMap) Init(d *vmdkDriver) {
...
 	}
 }
 
 // Returns ref count for the volume.
 // If volume is not referred (not in the map), return 0
-// NOTE: this assumes the caller holds the lock if we run concurrently
 func (r refCountsMap) getCount(vol string) uint {
-	rc := r[vol]
+	// RLocks the refCountsMap
+	r.mtx.RLock()
```
msterin: can't another thread change the value right after getCount returns and unlocks, but before the value is used for comparison plugin.go:Remove() ?

vmdk_plugin/refcnt.go
```diff
@@ -170,10 +195,14 @@ func (r refCountsMap) getCount(vol string) uint {
 
 // incr refCount for the volume vol. Creates new entry if needed.
 func (r refCountsMap) incr(vol string) uint {
-	rc := r[vol]
+	// Locks the refCountsMap
```
msterin: Same question. There seems to be a gap between the refcount is changed in incr() and checked in Mount().  During this time another thread can change the value

vmdk_plugin/plugin.go
```diff
@@ -340,9 +346,6 @@ func (d *vmdkDriver) Path(r volume.Request) volume.Response {
 func (d *vmdkDriver) Mount(r volume.MountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Mounting volume ")
 
-	d.m.Lock()
-	defer d.m.Unlock()
```
msterin: maybe I am missing something, but won't it be now possible to get refcnt (libe 354)m then another thread to change the actual refcount before decision on line 356 is made ? Or is there something else (e.g. Docker serializing of mounts) than would prevent it ? If there is, **please add a comment to the code explaining it** .

vmdk_plugin/refcnt.go
brunotm: The docker daemon has a reference counted resource locking by name scheme (much like what we have in the esx service) that serialize operations per volume (pkg/locker and volume/store) so we can't have more than one op on the same daemon/volume at the same time, so race on the refcount value is not possible. 
Our only concern it's to serialize reads/writes on the refcount map.

vmdk_plugin/refcnt.go
brunotm: Same as above

vmdk_plugin/plugin.go
brunotm: I'll update with comments 
Thanks!

msterin: Thanks ! LGTM / Ready to merge after the comment is added.

msterin:
Ready to merge when CI passes.

PR #811, opened 2016-12-12
Add photon volumes to volume plugin

govint:
This PR is a new version of https://github.com/vmware/docker-volume-vsphere/pull/798. The old PR had to be closed as the branch used had to be deleted and remade.


kerneltime:
This PR should also includes updates for demos that are online.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -32,58 +32,73 @@ import (
...
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/refcount"
 	"golang.org/x/exp/inotify"
 )
 
 const (
 	devWaitTimeout   = 1 * time.Second
-	mountRoot        = "/mnt/vmdk" // VMDK block devices are mounted here
 	sleepBeforeMount = 1 * time.Second
 	watchPath        = "/dev/disk/by-path"
+	version          = "VMDK Volume Driver v0.3"
```
msterin: version is long overdue for clicking up. I suggest 1.3 :-)

kerneltime: The value checked in should be irrelevant.
We need to set this from the make file similar to how we patch the vib xml. 
This is important for tagged releases.

govint: @kerneltime, is it a make generated go file with the version in it and it gets compiled along with the driver?

kerneltime: To do it properly would be more involved, but looking at the code, why do we need this field? This is not the same as a RPC version which can be something checked in and tracked in code. This field can be dropped.

govint: Assuming we don't have a problem with having a version string logged and returned by the driver, we can do it via the build vs. hardcoding it here.

Actually, the driver version is pretty much the plugin version itself, which is why its there (may not be displayed).

kerneltime: Either ways this is not related to the change here.

govint:
PR #833 was created to run the CI tests of the changes in this PR. Build vis #833 is ok.

Testing on Photon if without authentication and authentication will be added up later.

docs/user-guide/docker-plugin-drivers.md
```diff
@@ -0,0 +1,15 @@
+[TOC]
+The Docker volume plugin supports the below platforms and the corresponding drivers for those. The plugin supports all volume provisioning and managenment operations, defined by the Docker Volume plugin interface, on both platforms.
+
+1. Vmdk
```
pdhamdhere: Are we going to add/replace driver name to `vSphere'

govint: I've raised a separate issue to rename to "vsphere".

Issue #824, opened 2016-12-15
End_to_End test to verify the size of volume using admin cli and docker cli


ashahi1:
This test should verify that the size of volume created should be same as reported by vmkfstools.

PR #825, opened 2016-12-15
Add a test to verify the capacity of the volume created using vmkfstools

ashahi1:
Issue #824

Test name: test_VerifyVolumeSize

Test steps:
Create a volume of size 100 mb.
Verify the size of volume using vmkfstools.

Successfully ran the test locally using one esx and two photon vms.


esx_service/vmdk_ops_test.py
```diff
@@ -193,6 +194,20 @@ def testAttachAsOpts(self):
...
+	                          vmdk_path=self.name,
+				  vol_name=self.volName)
+	self.assertEqual(err, None, err)
+	self.assertEqual(
+	     os.path.isfile(self.name), True,
+	     "VMDK {0} is missing after create.".format(self.name))
+	response = vmdkops_admin.get_vmdk_size_info(path=self.name)
+	self.assertEqual(response["capacity"], "100.00MB",
+	                 "Expected size of volume = 100 MB. Actual size = {0}.
+			 Incorrect size for VMDK {1}".format(response["capacity"], self.name))
```
govint: This tests for the default size, but should also test that create is making a disk of the size requested by caller. Verify that a disk of the requested size has been made and verify that size conversions if any are correct.

kerneltime:
Please fix build error from CI.
```
************* Module esx_service.vmdk_ops_test
../esx_service/vmdk_ops_test.py:200: [E0001(syntax-error), ] unindent does not match any outer indentation level
Makefile:159: recipe for target '.lint_src' failed
```

esx_service/vmdk_ops_test.py
shuklanirdesh82: That's true! 

My opinion is to call vmfs tool directly instead of calling "vmdkops_admin.get_vmdk_size_info()" and compare the bytes returned from vmkfs tool with the expected one instead of going through the route of  get_vmdk_size_info & human_readable (which is consumed by 'vmdkops_admin.py ls')

kerneltime:
@ashahi1 swing by tomorrow to get the commit history sorted out.

esx_service/utils/vmdk_utils.py
```diff
@@ -305,9 +307,21 @@ def log_volume_lsof(vol_name):
                 cartel, name, ftype, fd, desc)
             logging.info("Volume open descriptor: %s", msg)
 
+def get_vmdk_capacity(path):
```
msterin: we already have get_vmdk_size_info in vmdkops_admin.py. You my want to move it to vmdk_utils and reuse from both places.

ashahi1: Alright. I will directly call get_vmdk_size_info in vmdkops_admin.py.

msterin:
> Per  @ashahi1 : 
> 
> @msterin No, we do not need #825 after this PR #831 . I wasn't sure if I can close it hence I left it open for the time being.

Closing

Issue #790, opened 2016-12-02
Refcount tests are failing against docker 1.12.2


kerneltime:
Ran into this while deploying the 6.5 testbed which has VMs with newer version of Docker. 
https://ci.vmware.run/vmware/docker-volume-vsphere/584
This issue is blocking https://github.com/vmware/docker-volume-vsphere/issues/763

Ran it manually as well against the testbed.

Sample run
```
=== RUN   TestRefCnt
--- FAIL: TestRefCnt (152.23s)
	refcnt_test.go:53: Looking for refcnt_test.sh in '/tmp/docker-volume-vsphere'
	refcnt_test.go:74: Time     Info
		
	refcnt_test.go:76: 2016-12-02T07:58:19Z Testing refcounts...
	refcnt_test.go:76: 2016-12-02T07:58:19Z Creating volume refCountTestVol and 5 containers using it
	refcnt_test.go:76: 2016-12-02T07:58:19Z refCountTestVol
	refcnt_test.go:76: 2016-12-02T07:58:19Z DRIVER              VOLUME NAME
	refcnt_test.go:76: 2016-12-02T07:58:19Z vmdk                refCountTestVol
	refcnt_test.go:76: 2016-12-02T07:58:19Z 0cf8f98beca12c6b2ca31077d989d332e8798db4474fd77d785fd274c23dfb3e
	refcnt_test.go:76: 2016-12-02T07:58:20Z 33da8d4959367d7e88742563dca7cf0c0f9261672366d24ee030b1dca40e6133
	refcnt_test.go:76: 2016-12-02T07:58:21Z c19fdd2d422f5866a07bc7b82aa49c4b3e55a2af74212f71b1d981b6c7f47172
	refcnt_test.go:76: 2016-12-02T07:58:21Z e54b95d751552342bc0a22179975200e4c25a2169ba2a1489a37a6ac3864966e
	refcnt_test.go:76: 2016-12-02T07:58:21Z f4e914254adf6f090b11ea05c1ef200112edee986acc1d67f2c5249c6e7fec66
	refcnt_test.go:76: 2016-12-02T07:58:21Z Checking volume content
	refcnt_test.go:76: 2016-12-02T07:58:22Z Found 5 files. Expected 5
	refcnt_test.go:76: 2016-12-02T07:58:22Z Checking the last refcount and mount record
	refcnt_test.go:76: 2016-12-02T07:58:22Z Checking 'docker volume rm'
[...]

PR #826, opened 2016-12-15
Reorder ref count tests to ensure correct checks on count of volume users

govint:
Had made these changes in PR #811, will submit via this PR instead as its not related to adding the photon driver.

Plus, the ref count test code is using plugin logs to figure out test result which doesn't work as expected always. These fixes ensure that the test result is deterministic.

Changes are:
1. Got errors when creating the test volume that volume size is not specified - which happens as a result of a failing check in the auth code. Passing a size of 1GB for the test volume used.

2. Moved the test for user count on the test volume (via check_files) to after the check on the users of the test volume. The test for user count itself runs a countainer to check the number of users of the ref count test volume and ends up increasing the volume user count. And the subsequent check (on user count) failed on occasion. Hence the reorder.

Test:
1. Ran the ref count test manually and it works ok.
```
sudo ./build/docker-volume-vsphere.test
Running tests on  unix:///var/run/docker.sock (may take a while)...
Running parallel tests on unix:///var/run/docker.sock and unix:///var/run/docker.sock (may take a while)...
PASS
```


govint:
Fixes #790

govint:
Those issues I've addressed in this change that actually repro'ed during testing. Pls. review and let me know if its good to merge.

Issue #836, opened 2016-12-22
vm creation is failing on vsanDatastore


shaominchen:
I have a VSAN setup where there's only one single vsanDatastore. I tried to run test-esx in this environment and some test cases failed:

Running unit tests in /tmp/vmdk_ops_unittest15141/vmdk_ops_test.py...
..E.........EEE.
======================================================================
ERROR: testAttachDetach (__main__.VmdkAttachDetachTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/vmdk_ops_unittest15141/vmdk_ops_test.py", line 482, in setUp
    datastore=self.datastore_name)
  File "/tmp/vmdk_ops_unittest15141/vmdk_ops_test.py", line 420, in create_vm
    vmdk_ops.wait_for_tasks(si, [task])
  File "/tmp/vmdk_ops_unittest15141/vmdk_ops.py", line 1497, in wait_for_tasks
    raise task.info.error
vim.fault.FileNotFound: (vim.fault.FileNotFound) {
   dynamicType = <unset>,
   dynamicProperty = (vmodl.DynamicProperty) [],
   msg = 'File  was not found',
   faultCause = <unset>,
   faultMessage = (vmodl.LocalizableMessage) [
      (vmodl.LocalizableMessage) {
         dynamicType = <unset>,
         dynamicProperty = (vmodl.DynamicProperty) [],
         key = 'com.vmware.esx.hostctl.default',
         arg = (vmodl.KeyAnyValue) [
            (vmodl.KeyAnyValue) {
               dynamicType = <unset>,
               dynamicProperty = (vmodl.DynamicProperty) [],
               key = 'reason',
               value = 'Could not get ObjectStoreFileSystem for /vmfs/volumes/vsan:52b2067d4d8fa677-de50f4c784493a59/test-vm/test-vm (File Not Found)'
            }
         ],
         message = 'Operation failed, diagnostics report: Could not get ObjectStoreFileSystem for /vmfs/volumes/vsan:52b2067d4d8fa677-de50f4c784493a59/test-vm/test-vm (File Not Found)'
      }
   ],
   file = ''
}

======================================================================
ERROR: test_vmdkops_on_default_tenant_vm (__main__.VmdkTenantTestCase)
Test vmdk life cycle on a VM which belongs to DEFAULT tenant
This operation may take a while. Please be patient.
This operation may take a while. Please be patient.
----------------------------------------------------------------------
[...]

PR #839, opened 2016-12-22
Supplying a fix for "#836: vm creation is failing on vsanDatastore"

shuklanirdesh82:
Vm creation was not complaining for vmfs as it doesn't use osfs code path and fails only for vsan as it is objectBased FS and vm-name was appending twice hence the issue.

Testing Done: Yes (both on vsan and vmfs)

vmdk_ops_test.py tests are passing locally and able to create vm successfully.

```
Running unit tests in /tmp/vmdk_ops_unittest22238/vmdk_ops_test.py...
.................
Ran 17 tests in 272.076s

OK
```
//CC @ashahi1 @shaominchen


esx_service/vmdk_ops_test.py
```diff
@@ -397,14 +397,11 @@ def create_vm(si, vm_name, datastore):
...
                                                                                        hosts, resource_pool))
-
-        datastore_path = '[' + datastore + '] ' + vm_name
-
         # bare minimum VM shell, no disks. Feel free to edit
         vmx_file = vim.vm.FileInfo(logDirectory=None,
                                    snapshotDirectory=None,
                                    suspendDirectory=None,
-                                   vmPathName=datastore_path)
+                                   vmPathName='[' + datastore + '] ')
```
msterin: Please rename create_vm param  from `datastore` to `vm_path` and it will be good to go

shuklanirdesh82: Sure, Thanks!

esx_service/vmdk_ops_test.py
```diff
@@ -479,7 +476,7 @@ def setUp(self):
         si = vmdk_ops.get_si()
         error, self.vm = create_vm(si=si, 
                                    vm_name=self.vm_name, 
-                                   datastore=self.datastore_name)
+                                   vm_path=self.datastore_name)
```
msterin: this is still weird... why self.datastore_name actually keep the vm_path ? 
I'll approve to unbock , but please investigate as a separate item

shuklanirdesh82: Actually ```create_vm``` is internally prepares vm_path (vmHome) i.e. ```vm_path = datastore_name + vmDisplayName```.

So I am going to rename it with 'datastore_name' with more meaningful name to reduce the confusion.

Issue #830, opened 2016-12-19
Tenancy: Add unit test for tenant management AdminCLI


lipingxue:
Add unit test for the following tenant managment Admin CLI command
- tenant create/update/rm/ls
- tenant vm add/rm/ls
- tenant access add/set/rm/ls

/CC @kerneltime @ashahi1

PR #850, opened 2017-01-04
Unit tests for tenant AdminCLI

lipingxue:
This change add unit tests for tenant AdminCLI.


esx_service/cli/vmdkops_admin_test.py
```diff
@@ -379,6 +410,371 @@ class TestStatus(unittest.TestCase):
...
+        """ Setup run before each test """
+        
+        if (not self.datastore_name):
+            datastores = vmdk_utils.get_datastores()
+            if datastores:
+                datastore = datastores[0]
+                self.datastore_name = datastore[0]
+                self.datastore_path = datastore[2]
+            
+                if len(datastores) >= 1:
```
shuklanirdesh82: this fails when there is only one datastore available in the inventory. Can you please provide a fix or provide an explanation if I have misunderstood?

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -379,6 +410,371 @@ class TestStatus(unittest.TestCase):
     def test_status(self):
         self.assertEqual(vmdkops_admin.status(None), None)
 
+class TestTenant(unittest.TestCase):
+    """ Test tenant functionality """
+    # tenant1 info
+    tenant1_name = "test_tenant1"
+    vm1_name = 'test_vm1'
```
shuklanirdesh82: minor: it is better to randomize the vm name to avoid subsequent failures if cleanup doesn't happen correctly. We have done such exercise recently so it would be great to keep it consistent.

esx_service/utils/auth_data_test.py
```diff
@@ -435,6 +444,7 @@ def test_remove_tenants(self):
                                                           description='Some tenant',
                                                           vms=vms,
                                                           privileges=privileges)
+
```
shuklanirdesh82: It seems this file is having only empty newline, please remove this file from the review if it is an accidental change.

esx_service/cli/vmdkops_admin_test.py
lipingxue: @shuklanirdesh82 Could you point me what is the change you made to randomize the vm name? I want to take a look. Thanks!

lipingxue: Comment addressed.

esx_service/cli/vmdkops_admin_test.py
lipingxue: It will not fail the test. Some test case need to run with at least two datastores.  Those test will be skipped if the condition is not met.

shuklanirdesh82: sorry for the confusion...

let's say we have only one datastore ```if len(datastores) >= 1:``` satisfies and failed at line:443. 

> Those test will be skipped if the condition is not met.

Can you please point me to the line which skips the TestTenant when only one datastore is found in the inventory?

Thanks!

esx_service/cli/vmdkops_admin_test.py
shuklanirdesh82: Thanks @lipingxue !

esx_service/vmdk_ops_test.py
```diff
@@ -1097,6 +1105,7 @@ def test_vmdkops_on_different_tenants(self):
     
     @unittest.skipIf(len(vmdk_utils.get_datastores()) < 2,
                      "second datastore is not found - skipping this test")
+    @unittest.skip("Liping remove")
```
msterin: this looks like something you want to remove :-)

lipingxue:
@shuklanirdesh82 
Unit test related change is in "vmdkops_admin_test.py"
Some minor changes made in other file does not related to unit test, but I still want to keep them in the PR, since those changes improve the readability.

I have already rebase with the master and there is no conflicts with the base branch.

shuklanirdesh82:
Thanks @lipingxue  for addressing comments and providing explanation. 

You may want to squash the commit as I am seeing changes from other PRs as part of your changes.

lipingxue:
Fixed #830

Issue #832, opened 2016-12-21
[driver tests] Improve vmdk_ops test


shuklanirdesh82:
need to supply unique vm/tenant name for vmdk_ops_test

Following tests are failing because of reusing the same name for vm/tenant name

OTOMH few tasks:

- [x] getting rid off static value for vm name (e.g. vm_name = 'test-vm', tests_vmX)
- [x] need to improve tearDown() method to cleanup the created VMs 
- [x] Need to improve following validation

```
        result = vmdk_ops.executeRequest(vm1_uuid, self.vm1_name, self.vm1_config_path, 'list', None, opts)
       
        self.assertEqual(3, len(result))
        self.assertEqual("tenant1_vol1", result[0]['Name'])
        self.assertEqual("tenant1_vol2", result[1]['Name'])
        self.assertEqual("tenant1_vol3", result[2]['Name'])
```
- [x] need to use ```vmdk_utils.get_datastores()``` instead of hardcoded value ('datastore1')
- [x] need to fix following code block as it is complaining when there is only one datastore in the inventory

```
               if len(datastores) >= 1:
 730                     datastore1 = datastores[1]
 731                     self.datastore1_name = datastore1[0]
 732                     self.datastoer1_path = datastore[2]
```

3 tests are failing locally consistently whereas 'testAttachDetach' passes if the created test-vm is not present in the inventory.

```
Running unit tests in /tmp/vmdk_ops_unittest3019/vmdk_ops_test.py...
..E.E........FF.
======================================================================
ERROR: testAttachDetach (__main__.VmdkAttachDetachTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/vmdk_ops_unittest3019/vmdk_ops_test.py", line 531, in testAttachDetach
    vm=vm[0])
IndexError: list index out of range

======================================================================
ERROR: testBadOpts (__main__.VmdkCreateCloneRemoveTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/vmdk_ops_unittest3019/vmdk_ops_test.py", line 272, in setUp
    path, err = vmdk_ops.get_vol_path(self.vm_datastore, auth.DEFAULT_TENANT)
  File "/tmp/vmdk_ops_unittest3019/vmdk_ops.py", line 638, in get_vol_path
    os.symlink(path, symlink_path)
OSError: [Errno 17] File exists

======================================================================
[...]

PR #866, opened 2017-01-18
Improve vmdk_ops_test 

ashahi1:
Fix for issue #832 

Tested the CI tests with one esx and two photon os VMs.


shuklanirdesh82:
@ashahi1  can you please run vmdk_ops_test when there is only vsan datastore present in the inventory?

esx_service/vmdk_ops_test.py
```diff
@@ -39,6 +39,7 @@
 import error_code
 import vmdk_utils
 import random
+import time
```
shuklanirdesh82: I guess we don't need this. Please remove.

esx_service/vmdk_ops_test.py
```diff
@@ -731,7 +730,7 @@ def setUp(self):
                 self.datastore_path = datastore[2]
                 logging.debug("datastore_name=%s datastore_path=%s", self.datastore_name,
                                                                      self.datastore_path)
-                if len(datastores) >= 1:
+                if len(datastores) >1:
```
kerneltime: Why is this change needed?

esx_service/vmdk_ops_test.py
ashahi1: I missed it. Will drop it. Thanks.

esx_service/vmdk_ops_test.py
ashahi1: old code: 
if len(datastores) >=1:
                    datastore1 = datastores[1]    >>>>>> this will cause failure if we have only one datastore in testbed
                    self.datastore1_name = datastore1[0]
                    self.datastoer1_path = datastore[2]
                    logging.debug("Found second datastore: datastore_name=%s datastore_path=%s",
                                  self.datastore1_name, self.datastore1_path)


This new condition enables that code within if block will only be executed if we have more than one datastore.

esx_service/vmdk_ops_test.py
```diff
@@ -1086,23 +1096,32 @@ def test_vmdkops_on_different_tenants(self):
...
+        result = vmdk_ops.executeRequest(vm2_uuid, self.vm2_name, self.vm2_config_path, 'list', None, opts)        
         # listVMDK return vol name with the format like vol@datastore
         self.assertEqual(3, len(result))
-        self.assertEqual("tenant2_vol1@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant2_vol2@"+self.datastore_name, result[1]['Name'])
-        self.assertEqual("tenant2_vol3@"+self.datastore_name, result[2]['Name'])
-    
+         
+        volume_names = ["tenant2_vol1@"+self.datastore_name,"tenant2_vol2@"+self.datastore_name,"tenant2_vol3@"+self.datastore_name]
+        for name in volume_names:
```
shuklanirdesh82: It would be better to move this piece of code into verification util instead of copying multiple places. 

Please address the comment into your next diff.

esx_service/vmdk_ops_test.py
```diff
@@ -100,7 +100,8 @@ class VmdkCreateRemoveTestCase(unittest.TestCase):
     valid_attach_as_opt_1 = {volume_kv.ATTACH_AS: u'persistent'}
     valid_attach_as_opt_2 = {volume_kv.ATTACH_AS: u'independent_persistent'}
     name = ""
-    vm_name = 'test-vm'
+    random_id = random.randint(0, 65536)
```
msterin: please separate this into a `generate_test_vm_name` function instead of cut-n-paste throughout the code. 
Also, replace these 2 lines in all places below the function call please

esx_service/vmdk_ops_test.py
```diff
@@ -445,6 +447,18 @@ def remove_vm(si, vm):
         task = vm.Destroy_Task()
         vmdk_ops.wait_for_tasks(si, [task])
 
+def checkIfVolumeExist(volume_names,result):
+    #checks if volume exists in the list of vmdks present in result
+    if result:
```
msterin: this would be much better (and less nesting) if it said:
```python
if not result:
   return True

for name in volume_names:
    ...   
```
also I do not understnad the login... `if not result` returns true in the code. Why ?

esx_service/vmdk_ops_test.py
```diff
@@ -731,7 +740,7 @@ def setUp(self):
                 self.datastore_path = datastore[2]
                 logging.debug("datastore_name=%s datastore_path=%s", self.datastore_name,
                                                                      self.datastore_path)
-                if len(datastores) >= 1:
+                if len(datastores) >1:
```
msterin: So if we have only one datastore, `datastore1`, `self.datastore1_name ` and  `self.datastoer1_path ` will be not set. Is it OK ? '
'also - nit: space after `>` is missing

esx_service/vmdk_ops_test.py
```diff
@@ -972,9 +981,8 @@ def test_vmdkops_on_tenant_vm(self):
         # "result"" should have two volumes :tenant1_vol2@default_datastore, and  tenant1_vol3@datastore
         # these volumes are created at default_datastore, which is self.datastore_name
         self.assertEqual(2, len(result))
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[1]['Name'])
-
+        volume_names = ["tenant1_vol2@"+self.datastore_name,"tenant1_vol3@"+self.datastore_name]
```
msterin: spaces are missing around `+` and after `,`. Please stay with with the code style: https://www.python.org/dev/peps/pep-0008/

esx_service/vmdk_ops_test.py
```diff
@@ -996,9 +1004,8 @@ def test_vmdkops_on_tenant_vm(self):
         # listVMDK return vol name with the format like vol@datastore
         # result should have two volumes :tenant1_vol2, and  tenant1_vol3
         self.assertEqual(2, len(result))
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[1]['Name'])
-
+        volume_names = ["tenant1_vol2@"+self.datastore_name,"tenant1_vol3@"+self.datastore_name]
```
msterin: please stay with the code style. 

Also, I suggest a function for constructing volume_names - instead of the copy-pasted code used now

esx_service/vmdk_ops_test.py
```diff
@@ -1086,23 +1093,19 @@ def test_vmdkops_on_different_tenants(self):
...
         opts = {}
         result = vmdk_ops.executeRequest(vm1_uuid, self.vm1_name, self.vm1_config_path, 'list', None, opts)
-       
         # listVMDK return vol name with the format like vol@datastore
         self.assertEqual(3, len(result))
-        self.assertEqual("tenant1_vol1@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[1]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[2]['Name'])
 
+        volume_names = ["tenant1_vol1@"+self.datastore_name,"tenant1_vol2@"+self.datastore_name,"tenant1_vol3@"+self.datastore_name]
```
msterin: ditto

esx_service/vmdk_ops_test.py
```diff
@@ -1086,23 +1093,19 @@ def test_vmdkops_on_different_tenants(self):
...
-       
+        result = vmdk_ops.executeRequest(vm2_uuid, self.vm2_name, self.vm2_config_path, 'list', None, opts)        
         # listVMDK return vol name with the format like vol@datastore
-        self.assertEqual(3, len(result))
-        self.assertEqual("tenant2_vol1@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant2_vol2@"+self.datastore_name, result[1]['Name'])
-        self.assertEqual("tenant2_vol3@"+self.datastore_name, result[2]['Name'])
-    
+        self.assertEqual(3, len(result)) 
+        volume_names = ["tenant2_vol1@"+self.datastore_name,"tenant2_vol2@"+self.datastore_name,"tenant2_vol3@"+self.datastore_name]
```
msterin: ditto - code style and less copy-n-paste if possible

esx_service/vmdk_ops_test.py
ashahi1: Created a separate function - generate_test_vm_name().

esx_service/vmdk_ops_test.py
ashahi1: Fixed it.

esx_service/vmdk_ops_test.py
ashahi1: Ran the vmdk_ops_test.py test only with one datastore - vmfs. Results are as follows:
================= 

Running unit tests in /tmp/vmdk_ops_unittest3474/vmdk_ops_test.py...
WARNING:root:dockvols creation failed - Operation not permitted on local.3-0
WARNING:root:dockvols creation failed - Operation not permitted on local.3-1
ss........sss....
Ran 17 tests in 204.492s

OK (skipped=5)

esx_service/vmdk_ops_test.py
ashahi1: Created a separate function for volume_names - generate_volume_names
Fixed the code style too

esx_service/vmdk_ops_test.py
```diff
@@ -972,9 +992,8 @@ def test_vmdkops_on_tenant_vm(self):
         # "result"" should have two volumes :tenant1_vol2@default_datastore, and  tenant1_vol3@datastore
         # these volumes are created at default_datastore, which is self.datastore_name
         self.assertEqual(2, len(result))
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[1]['Name'])
-
+        volume_names = ["tenant1_vol2@" + self.datastore_name, "tenant1_vol3@" + self.datastore_name]
```
shuklanirdesh82: don't we need to use ```generate_volume_names``` here?

esx_service/vmdk_ops_test.py
```diff
@@ -996,9 +1015,8 @@ def test_vmdkops_on_tenant_vm(self):
         # listVMDK return vol name with the format like vol@datastore
         # result should have two volumes :tenant1_vol2, and  tenant1_vol3
         self.assertEqual(2, len(result))
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[1]['Name'])
-
+        volume_names = ["tenant1_vol2@" + self.datastore_name, "tenant1_vol3@" + self.datastore_name]
```
shuklanirdesh82: same as above

esx_service/vmdk_ops_test.py
```diff
@@ -1086,23 +1104,21 @@ def test_vmdkops_on_different_tenants(self):
...
         opts = {}
         result = vmdk_ops.executeRequest(vm1_uuid, self.vm1_name, self.vm1_config_path, 'list', None, opts)
-       
         # listVMDK return vol name with the format like vol@datastore
         self.assertEqual(3, len(result))
-        self.assertEqual("tenant1_vol1@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant1_vol2@"+self.datastore_name, result[1]['Name'])
-        self.assertEqual("tenant1_vol3@"+self.datastore_name, result[2]['Name'])
 
+        # volume_names = ["tenant1_vol1@"+self.datastore_name,"tenant1_vol2@"+self.datastore_name,"tenant1_vol3@"+self.datastore_name]
```
shuklanirdesh82: need to remove dead code here

esx_service/vmdk_ops_test.py
```diff
@@ -1086,23 +1104,21 @@ def test_vmdkops_on_different_tenants(self):
...
-       
+        result = vmdk_ops.executeRequest(vm2_uuid, self.vm2_name, self.vm2_config_path, 'list', None, opts)        
         # listVMDK return vol name with the format like vol@datastore
-        self.assertEqual(3, len(result))
-        self.assertEqual("tenant2_vol1@"+self.datastore_name, result[0]['Name'])
-        self.assertEqual("tenant2_vol2@"+self.datastore_name, result[1]['Name'])
-        self.assertEqual("tenant2_vol3@"+self.datastore_name, result[2]['Name'])
-    
+        self.assertEqual(3, len(result)) 
+        # volume_names = ["tenant2_vol1@"+self.datastore_name,"tenant2_vol2@"+self.datastore_name,"tenant2_vol3@"+self.datastore_name]
```
shuklanirdesh82: same as above

esx_service/vmdk_ops_test.py
```diff
@@ -86,6 +86,12 @@ def test_name_parse(self):
             except vmdk_ops.ValidationError as ex:
                 self.assertFalse(expected_result, "Expected vol name parsing to succeed for '{0}'"
                                  .format(full_name))
+    
+def generate_test_vm_name():
```
shuklanirdesh82: nit: better to have method level doc/description that talks about the purpose of this helper

esx_service/vmdk_ops_test.py
```diff
@@ -445,12 +451,34 @@ def remove_vm(si, vm):
         task = vm.Destroy_Task()
         vmdk_ops.wait_for_tasks(si, [task])
 
+def generate_volume_names(tenant, datastore_name):
```
shuklanirdesh82: nit: need method's doc/description

esx_service/vmdk_ops_test.py
```diff
@@ -445,12 +451,34 @@ def remove_vm(si, vm):
...
+    """ Returns a list of volume names """
+    volNames = []
+    for x in range(3):
+        volNames.append(tenant + "_vol" + str(x + 1) + "@" + datastore_name)
+    
+    return volNames
+    
+def checkIfVolumeExist(volume_names,result):
```
shuklanirdesh82: same as above

esx_service/vmdk_ops_test.py
```diff
@@ -445,12 +451,34 @@ def remove_vm(si, vm):
...
+    #checks if volume exists in the list of vmdks present in result
+    if not result:
```
msterin: nit: the code is OK, but IMO it would be nicer to say 
```python
  if not result or not volume_names:
   return False
```

esx_service/vmdk_ops_test.py
ashahi1: @shuklanirdesh82  Test 'test_vmdkops_on_tenant_vm' creates a tenant and then attaches volume tenant1_vol1_name and then detaches the volume. And then finally checks that 'result' should have two volumes :tenant1_vol2@default_datastore, and tenant1_vol3@datastore. That's why I didn't use this method here. I have used this method at instances where code checks for all the three volumes. Suggestions?

shuklanirdesh82: This sounds confusing. By looking at the tests I can confirm that test is deleting the first created volume ```tenant1_vol1@``` hence the verification only for the rest two volumes. 

Going forward we should improve the steps as the util is in placed we should leverage it better. You may address this in the separate PR.

PR #873, opened 2017-01-23
Reduce CI run time. Add gofmt updates.

kerneltime:
This reduces CI time down by close to half.
We currently run the same target test-esx 2 times against each ESX since we blindly repeat the entire test suite against each target. This change will only run test-esx once per ESX.

Also, picked up some gofmt complaints.


shuklanirdesh82:
deploy-and-test-wrapper.sh: changes looks good to me. 

> Also, picked up some gofmt complaints.
Thanks for fixing! This one is already addressed as part of my previous checkin.

Issue #875, opened 2017-01-24
Drop DB deletion from test-esx


lipingxue:
Details from this issue 
https://github.com/vmware/docker-volume-vsphere/issues/865

Our original plan was as follows:

introduce schema version in the DB
force DB deletion on test to hide "no versioned" -> "versioned" schema update
For further schema changes, support version to version in-place upgrade
so, the auth_db deletion was a temporary measure. It should not be needed anymore, we just need to make sure we do not have really stale DBs around

Actions:

we should drop the DB delete from test set (@liping)
if you still have od DB, remove it manyully per error message

This PR is used to track the first action which drop the DB delete from test set
Going forward,  for further schema changes, we will support version to version in-place upgrade.
If incase user has any ancient DB(no version), he/she need to remove it manually.

lipingxue:
Fixed #865

PR #877, opened 2017-01-24
Drop the auth-DB remove in test-esx.

lipingxue:
This PR includes two changes:
1. drop the auth-DB remove in test-esx (fixes issue #875)
2. Fix a small bug in function "get_datastore_url" and "get_datastore_name" to handle the _DEFAULT_DS correctly.

test-esx passed. Please review this change.


shaominchen:
LGTM.

esx_service/utils/auth_data.py
```diff
@@ -210,6 +210,7 @@ def get_default_datastore(self, conn):
             return str(error_info), None
         else:
             datastore_url = result[auth_data_const.COL_DEFAULT_DATASTORE_URL]
+            logging.debug("get_default_datastore: datastore_url=%s", datastore_url)
```
msterin: nit - why this is dropped ? Looks like a useful debug info for me. 
It's a nit, feel free to merge with it

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -12,15 +12,15 @@
...
+DS='-ds=datastore1'
 unit_test_array=($TEST_URL_ARRAY)
 numServers=${#unit_test_array[@]}
 DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER:=0}
 prevBuildStatus=`drone build info vmware/docker-volume-vsphere $(( $DRONE_BUILD_NUMBER-$numServers ))`
 outArray=($prevBuildStatus)
 total_builds=`drone build list vmware/docker-volume-vsphere| wc -l`
 
-govc datastore.mkdir docker-volume-vsphere/$DRONE_BUILD_NUMBER
+govc datastore.mkdir $DS docker-volume-vsphere/$DRONE_BUILD_NUMBER
```
msterin: so on which DS it was doing mkdir before the change ?

esx_service/utils/auth_data.py
lipingxue: It is NOT dropped. I added this logging in my change.

lipingxue: It is NOT dropped, I added this log in my change :)

misc/drone-scripts/validate-and-serialize.sh
lipingxue: This change is not from me. It was made by Ritesh in master, I rebase my change with master and then you see the change. I have removed from my commit log.

esx_service/utils/auth_data.py
msterin: oops. you are right :-) !

PR #885, opened 2017-02-02
Fix issue #673

shaominchen:
1. Add logic to check the validity of privilege prameters
2. Add test cases to cover new logic
3. Refactored some test cases
4. Remove all trailing spaces in auth_api.py


esx_service/utils/auth_api.py
```diff
@@ -491,21 +491,43 @@ def check_datastore(datastore_name):
...
+def check_privilege_parameters(privilege):
+    """ Check whether the privilege parameters are invalid """
+
+    volume_maxsize = privilege[auth_data_const.COL_MAX_VOLUME_SIZE]
+    volume_totalsize = privilege[auth_data_const.COL_USAGE_QUOTA]
+
+    # If both volume max size and volume total size are set,
+    # volume max size should not exceed volume total size
+    if (volume_maxsize and volume_totalsize and (volume_maxsize > volume_totalsize)):
+        error_msg = "Volume max size {0}MB exceeds the total size {1}MB".format(
```
lipingxue: Please define a new error in error_code.py.

esx_service/utils/auth_api.py
```diff
@@ -491,21 +491,43 @@ def check_datastore(datastore_name):
...
         ds_name = vmdk_utils.get_datastore_name(p.datastore_url)
         if ds_name == datastore_name:
             return True
-    
+
     return False
-    
+
+def check_privilege_parameters(privilege):
+    """ Check whether the privilege parameters are invalid """
```
lipingxue: Add some description of the input param "privilege" and also the description of the return value of this function.

esx_service/utils/auth_api.py
```diff
@@ -491,21 +491,43 @@ def check_datastore(datastore_name):
...
+    if (volume_maxsize and volume_totalsize and (volume_maxsize > volume_totalsize)):
+        error_msg = "Volume max size {0}MB exceeds the total size {1}MB".format(
+            volume_maxsize, volume_totalsize)
+        error_info = error_code.generate_error_info(ErrorCode.INVALID_ARGUMENT, error_msg)
+        return error_info
+
+    # If volume total size is set, but volume max size is NOT set,
+    # volume max size should be set to the same as volume total size
+    if (volume_totalsize and not volume_maxsize):
+        print('Note: volume max size will be set to the same value as volume total size: {0}MB'.format(volume_totalsize))
```
lipingxue: We should not use print here. Is it a error case here? It it is, return the error, otherwise, we should log this.
I don't understand why you assign volume_maxsize to be volume_totalsize here. There two parameters are independent.

esx_service/vmodl/vmodl_test.py
```diff
@@ -497,6 +489,29 @@ def test_add_privilege(self):
         # Verify the default datastore
         self.assertEqual(result[0].default_datastore, self.datastore)
 
+    def test_add_privilege_without_volume_max_size(self):
```
lipingxue: We need a test which volume_maxsize is bigger than volume_totalsize.

esx_service/vmodl/vmodl_test.py
```diff
@@ -497,6 +489,29 @@ def test_add_privilege(self):
...
+        self.tenantMgr.AddPrivilege(tenant, privilege)
+
+        # Get the privilege
+        result = self.tenantMgr.GetTenants(name=TENANT_NAME)
+        self.assertTrue(result)
+        p = result[0].privileges
+        self.assertTrue(p)
+
+        # Verify that the volume_max_size was set to the same value as volume_total_size
+        self.assertEqual(p[0].volume_total_size, 1024)
```
lipingxue: Same as comments above. I don't understand the logic that set volume_max_size to the same value as volume_total_size.

esx_service/vmodl/vmodl_test.py
```diff
@@ -573,12 +588,31 @@ def test_add_privilege_already_exists(self):
...
+        privilege.datastore = DS_NOT_EXIST
+        privilege.allow_create = False
+        privilege.volume_max_size = 1024
+        privilege.volume_total_size = 2048
+
+        # Add the privilege to the tenant
+        with self.assertRaises(vmodl.fault.InvalidArgument):
+            self.tenantMgr.AddPrivilege(tenant, privilege)
+
+    def test_add_privilege_invalid_volume_size(self):
```
lipingxue: Name of this function is a little bit confusing.

esx_service/vmodl/vmodl_test.py
```diff
@@ -661,7 +695,82 @@ def test_update_privilege(self):
...
+        # Update the privilege
+        self.tenantMgr.UpdatePrivilege(tenant, self.datastore, volume_total_size=1024)
+
+        # Get the privilege
+        result = self.tenantMgr.GetTenants(name=TENANT_NAME)
+        self.assertTrue(result)
+        p = result[0].privileges
+        self.assertTrue(p)
+
+        # Verify that the volume_max_size was set to the same value as volume_total_size
```
lipingxue: Same comments as above.

msterin:
+1 on Liping's notes.

Also - bundling trailing space removal in the same PR makes it very hard to read. Please do separate them in the future. 

Other than that, the change looks good. Putting in an approval so when Liping approve you can merge.

esx_service/utils/auth_api.py
shaominchen: Done.

esx_service/utils/auth_api.py
shaominchen: Done.

esx_service/utils/auth_api.py
shaominchen: As discussed offline, the backend will check the totalsize even if maxsize is unset. So I agree we don't need this logic. I will remove this logic and related test cases.

esx_service/vmodl/vmodl_test.py
shaominchen: This test was already added in my PR: test_update_privilege_with_invalid_volume_size()

esx_service/vmodl/vmodl_test.py
shaominchen: Same as above.

esx_service/vmodl/vmodl_test.py
shaominchen: Same as above.

esx_service/vmodl/vmodl_test.py
shaominchen: OK, I will use this longer name instead: test_update_privilege_with_volume_maxsize_greater_than_totalsize()

Let me know if you have better suggestions.

msterin: Please don't :-) , it will be totally unreadable. 
The current one is good enough IMO, especially if you add a good docstring. 

@lipingxue - feel free to offer a better name though, with a reasonable length()

shaominchen: Adding some docstrings sounds good to me. I will do this.

PR #893, opened 2017-02-07
Added wait code for attach during create.

govint:
Added wait code for the attach() done on create() before the volume is actually formatted.

This wait code is repeated in the Mount() path and I'll see if we need to have a waitPre() and waitPost() functions to avoid the code duplication.


vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -312,6 +331,34 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 		return volume.Response{Err: errGetDevicePath.Error()}
 	}
 
+	if skipInotify {
+		time.Sleep(sleepBeforeMount)
+	}
+loop:
+	for {
+		select {
```
pdhamdhere: You can end here with skipNotify = FALSE. I assume it would break the loop?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -312,6 +331,34 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
...
+			}
+		case err := <-watcher.Error:
+			log.WithFields(
+				log.Fields{"name": r.Name, "device": device, "error": err},
+			).Error("Hit error during watch ")
+			break loop
+		case <-time.After(devWaitTimeout):
+			log.WithFields(
+				log.Fields{"name": r.Name, "timeout": devWaitTimeout, "device": device},
+			).Warning("Reached timeout while waiting for device, trying to mount anyways ")
```
pdhamdhere: "Trying to mount" => Copy-paste?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -312,6 +331,34 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
...
+			break loop
+		}
+	}
+
 	errMkfs := fs.Mkfs(mkfscmd, r.Name, device)
```
pdhamdhere: We could reach here (Mkfs) with error from Notify loop. Is that okay?

govint:
Created common code for the wait logic and using that in all places (except photon which will be addressed along with other photon changes).

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -312,6 +276,14 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 		return volume.Response{Err: errGetDevicePath.Error()}
 	}
 
+	if skipInotify {
+		time.Sleep(sleepBeforeMount)
+	}
+
+	// Wait for the attach to complete, may timeout
+	// in which case we continue creating the file system.
+	fs.DevAttachWait(watcher, r.Name, device)
```
msterin: it looks like fs.DevAttachWait() is called even if skipNotify is true. Is it by design  ?

vmdk_plugin/utils/fs/fs.go
```diff
@@ -185,46 +231,11 @@ func GetDevicePathByID(id string) (string, error) {
...
-			).Error("Hit error during watch ")
-			break loop
-		case <-time.After(devWaitTimeout):
-			log.WithFields(
-				log.Fields{"timeout": devWaitTimeout, "device": device},
-			).Warning("Reached timeout while waiting for device ")
-			break loop
-		}
-	}
+	DevAttachWait(watcher, id, device)
```
msterin: same question -  if skipNotify is true, then watcher will be not configured and DevAttachWait() probably should be called, right ?

vmdk_plugin/utils/fs/fs.go
```diff
@@ -56,6 +57,51 @@ type VolumeDevSpec struct {
 	ControllerPciSlotNumber string
 }
 
+// DevAttachWaitPrep create the wather
```
shuklanirdesh82: nit: watcher?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
govint: Thanks for the catch, fixed.

vmdk_plugin/utils/fs/fs.go
govint: Fixed, thanks for the catch.

vmdk_plugin/utils/fs/fs.go
govint: Thanks, fixed.

Issue #892, opened 2017-02-03
AdminCLI "vmdkops_admin.py ls" failed when datastore name includes white space


lipingxue:
Steps to reproduce:
1. create several volumes in datastore (for example "tmp-(datastore1)"), then run vmdkops_admin.ls
```
[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Volume  Datastore         Created By VM  Created                   Attached To VM  Policy  Capacity  Used     Disk Format  Filesystem Type  Access      Attach As               
------  ----------------  -------------  ------------------------  --------------  ------  --------  -------  -----------  ---------------  ----------  ----------------------  
vol1    tmp-(datastore1)  photon5        Mon Sep  5 01:47:03 2016  detached        N/A     100.00MB  13.00MB  thin         ext4             read-write  independent_persistent  
vol2    tmp-(datastore1)  photon5        Mon Sep  5 01:48:43 2016  detached        N/A     100.00MB  13.00MB  thin         ext4             read-write  independent_persistent  
vol3    tmp-(datastore1)  photon5        Mon Sep  5 04:08:01 2016  detached        N/A     100.00MB  13.00MB  thin         ext4             read-write  independent_persistent  

```

2. rename the datastore name by adding extra whitespace in the datastore name  ("tmp(datastore1)" -> "tmp (datastore1)")
run the same command again, get the following error

```
[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Invalid destination specification:
vmkfstools: (datastore1)/dockvols/077d0f05-f262-4720-91b2-9d55a23870ff/vol1.vmdk: No name specification expected.
Failed to stat /vmfs/volumes/tmp- (datastore1)/dockvols/077d0f05-f262-4720-91b2-9d55a23870ff/vol1.vmdk. VMDK corrupted. Please remove and then retry

```

shuklanirdesh82:
resolving duplicate of #816 (#816 should care of the reported issue)

shuklanirdesh82:
reopening this one as per offline discussion with @lipingxue

govint:
May be fixed by changes posted for #816

govint:
Closing, as #911 is changing the same function thats causing the behavior to use KV API vs. command line vmkfstools util.

[...]

Issue #818, opened 2016-12-13
Not able to create vsphere docker volume (using short volume names) after renaming datastore; running into "IndexError('list index out of range',)"


shuklanirdesh82:
Short volume name reference: https://github.com/vmware/docker-volume-vsphere/blob/master/docs/misc/VolumeNames.md#short-volume-names

Steps to reproduce:

1)  create vsphere docker volume using following command (Let's say we have created photon1 docker host on TestDatastore)
```
# docker volume create --driver=vmdk --name=MyVol -o size=1gb
MyVol
```
2) make sure volume is created successfully
```
[root@promc-2n-dhcp105-97:/var/log/vmware] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Volume  Datastore       Created By VM  Created                   Attached To VM  Policy  Capacity  Used     Disk Format  Filesystem Type  Access      Attach As               
------  --------------  -------------  ------------------------  --------------  ------  --------  -------  -----------  ---------------  ----------  ----------------------  
MyVol   TestDatastore  photon2        Tue Dec 13 21:13:49 2016  detached        N/A     1.00GB    39.00MB  N/A          ext4             read-write  independent_persistent
```
3) rename the datastore which is hosting docker VM i.e. TestDatastore => TestDatastore1
4) wait for some time ~30-60secs to propagate the name changes

```
[root@promc-2n-dhcp105-97:/var/log/vmware] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Volume  Datastore       Created By VM  Created                   Attached To VM  Policy  Capacity  Used     Disk Format  Filesystem Type  Access      Attach As               
------  --------------  -------------  ------------------------  --------------  ------  --------  -------  -----------  ---------------  ----------  ----------------------  
MyVol   **TestDatastore1**  photon2        Tue Dec 13 21:13:49 2016  detached        N/A     1.00GB    39.00MB  N/A          ext4             read-write  independent_persistent

Note: make sure 'Datastore' is reflected correctly i.e. with new name
```
5) invoke the volume creation command with different volume 
```
# docker volume create --driver=vmdk --name=MyVol1 -o size=1gb
Error response from daemon: create MyVol1: VolumeDriver.Create: dockvols creation failed - Operation not permitted on TestDatastore  <===== driver is still referring to the old name
```

**Expectation**: volume creation should not complain if underlying datastore got renamed

from ESX logs:

```
[...]

Issue #816, opened 2016-12-13
volume creation fails when datastore name is having embedded spaces e.g. "datastore (1)"


shuklanirdesh82:
Steps to reproduce:
1. rename existing datastore which is not having spaces e.g. datastore => datastore (1)
2. invoke volume creation (docker volume create --driver=vmdk --name=MyVol22334 -o size=1gb)

Expectation: volume creation should succeed

**Note**: having spaces in datastore name is a valid name in vc context; while adding ESXs to VC which is having same datastore name; VC renames datastore by appending '(N)' for later ESX automatically

Current result: volume creation is failing with following outcome

```
-----------from console---------------
# docker volume create --driver=vmdk --name=MyVol22334 -o size=1gb
Error response from daemon: create MyVol22334: VolumeDriver.Create: Failed to create /vmfs/volumes/datastore (1)/dockvols/MyVol22334.vmdk. /bin/sh: syntax error: unexpected "("

-------from docker-volume-vsphere.log -------------
2016-12-13 20:34:06.835956871 +0000 UTC [ERROR] Create volume failed error="Failed to create /vmfs/volumes/datastore (1)/dockvols/MyVol22334.vmdk. /bin/sh: syntax error: unexpected \"(\"\n" name=MyVol22334 
```

on ESX side:
```
----------from vmdk_ops.log ----------
12/13/16 20:34:06 2663252 [photon2-[datastore (1)] dockvols/MyVol22334.vmdk] [INFO   ] executeRequest 'get' completed with ret={u'Error': 'Volume MyVol22334 not found (file: /vmfs/volumes/datastore (1)/dockvols/MyVol22334.vmdk)'}
12/13/16 20:34:06 2663252 [photon2-[datastore (1)] dockvols/MyVol22334.vmdk] [INFO   ] *** createVMDK: /vmfs/volumes/datastore (1)/dockvols/MyVol22334.vmdk opts = {u'size': u'1gb', u'fstype': u'ext4'}
12/13/16 20:34:06 2663252 [photon2-[datastore (1)] dockvols/MyVol22334.vmdk] [INFO   ] executeRequest 'create' completed with ret={u'Error': 'Failed to create /vmfs/volumes/datastore (1)/dockvols/MyVol22334.vmdk. /bin/sh: syntax error: unexpected "("\n'}
```

shuklanirdesh82:
[...]

PR #905, opened 2017-02-10
Fixes issue 816 (space in DS names) and add tenant name validation during tenant create

lipingxue:
This PR includes:
1. Fix issue 816(space in DS names)
2. validate tenant name during tenant create, tenant name are only allowed to be [a-zA-Z0-9_][a-zA-Z0-9_.-]*


esx_service/vmdk_ops.py
```diff
@@ -854,11 +854,14 @@ def findDeviceByPath(vmdk_path, vm):
         # to match with the given volume name.
         # Filename format is as follows:
         #   "[<datastore name>] <parent-directory>/tenant/<vmdk-descriptor-name>"
-        logging.debug("d.backing.fileName %s", d.backing.fileName)
-        backing_disk = d.backing.fileName.split(" ")[1]
+        # Assume there is one space between these two parts "[<datastore name>]" and
+        # <parent-directory>/tenant/<vmdk-descriptor-name>
+        logging.debug("findDeviceByPath: d.backing.fileName %s", d.backing.fileName)
+        backing_disk = d.backing.fileName.rsplit(" ", 1)[1]
```
msterin: Can  `<parent-directory>/tenant/<vmdk-descriptor-name>` hasve a space ? The current code is guaranteed to fail. Maybe it's better to look for ']' ? Something like
```python
    ds, file  = d.backing.fileName.rsplit("]", 1)[1]
    datastore = ds[1:]
    backing_disk = file[1:]
```
That raises a couple of additional things to check:
(1) can we create a tenant with a space 
(2) can we create a volume with a space in the name.

shaominchen: IMO, we shouldn't allow spaces in tenant & volume names. It should be OK to have spaces in tenant description though.

esx_service/vmdk_ops.py
```diff
@@ -854,11 +854,14 @@ def findDeviceByPath(vmdk_path, vm):
         # to match with the given volume name.
         # Filename format is as follows:
         #   "[<datastore name>] <parent-directory>/tenant/<vmdk-descriptor-name>"
-        logging.debug("d.backing.fileName %s", d.backing.fileName)
-        backing_disk = d.backing.fileName.split(" ")[1]
+        # Assume there is one space between these two parts "[<datastore name>]" and
```
shaominchen: This comment is not very clear to me. I'd suggest to give an example in the comment so that it's easier to understand the logic of ....rsplit(" ", 1)[0]

esx_service/vmdk_ops.py
```diff
@@ -854,11 +854,14 @@ def findDeviceByPath(vmdk_path, vm):
...
+        # <parent-directory>/tenant/<vmdk-descriptor-name>
+        logging.debug("findDeviceByPath: d.backing.fileName %s", d.backing.fileName)
+        backing_disk = d.backing.fileName.rsplit(" ", 1)[1]
 
         # datastore='[datastore name]'
```
shaominchen: Same as above: suggest to add an example in the comment to explain the logic below

esx_service/vmdk_ops.py
lipingxue: I don't know space in "volume names" is allowed in vsphere or not? "datastore" names can have space.

msterin: That was the point - please check it :-). Directory names (i.e. tenant) and volume names (i.e. files) will accept spaces unless we block them on creation. So please check if we do block it on creation.  I agree with @shaominchen that it's better to block spaces in tenant names and volume names. on creation.
With that said, i still think the code I suggested is slightly more resilient than looking for spaces.

lipingxue: Yes. I agree with you.  I will modify the code as you suggested. Currently, I think we don't check for tenant name and volume name during creation. I will add the check logic.

lipingxue: I did some test.
Vsphere allows volume name including space since I can create a VM with name including space.

But if I create a volume using local driver, it does not allow volume name to include space, and I got the following error.

govint:
@lipingxue can you also include a test for the scenario in #892, if spaces are disallowed then that issue shouldn't be able to occur.

What about is a VSphere has already created a datastore with spaces in it. Then will that datastore be prevented from being used in any tenant. Ideally, user should be disallowed from adding that datastore.

esx_service/utils/vmdk_utils.py
```diff
@@ -232,7 +232,7 @@ def list_vmdks(path, volname="", show_snapshots=False):
 def vmdk_is_a_descriptor(path, file_name):
     """
     Is the file a vmdk descriptor file?  We assume any file that ends in .vmdk,
-    does not have -delta or -flat or he likes at the end of filename, 
+    does not have -delta or -flat or he likes at the end of filename,
```
govint: seems a typo, sentence must be corrected - "or he likes at the end of filename"

esx_service/vmdk_ops.py
lipingxue: Added code to check tenant name and volume name during creation.

esx_service/vmdk_ops.py
lipingxue: Done.

esx_service/vmdk_ops.py
lipingxue: Done.

esx_service/utils/vmdk_utils.py
lipingxue: @govint We only disallow space in tenant name and volume name during creation. So with this fix, issue #892 can still happen. We cannot control the datastore name. In VSphere, spaces in datastore name are allowed.

lipingxue: Typo in comment is fixed too.

lipingxue:
Add code to validate tenant_name and volume name during tenant create and volume create.
1. create a tenant with invalid name
```
[root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=.abc*d
Tenant name .abc*d is invalid, only [a-zA-Z0-9_][a-zA-Z0-9_.-]* is allowed

```

2. create a volume with invalid name
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name="abc*   edf"
Error response from daemon: create abc*   edf: VolumeDriver.Create: Volume name abc*   edf is invalid, only [a-zA-Z0-9][a-zA-Z0-9_.-]* is allowed
```

lipingxue:
@msterin @shaominchen @govint I have addressed your comments, change is repushed.  Please review it again.

esx_service/utils/vmdk_utils.py
```diff
@@ -100,8 +100,11 @@ def get_datastores():
     'url-name' is the last element of datastore URL (e.g. 'vsan:572904f8c031435f-3513e0db551fcc82')
     'dockvol-path; is a full path to 'dockvols' folder on datastore
     """
-    if datastores == None:
-        init_datastoreCache()
+    # if datastores == None:
+    #     init_datastoreCache()
+    # we should not use cached value, user may change datastore name on the fly
```
msterin: the problem is that init_datastoreCache is a an  expensive operation and it is called way too many times, often many times in the same call chain. The correct way to fix it would have been passing datastore list, but it was a very invasive change so the cache was introduced.  Let's discuss how to test and what are the other ways to fix the issue, rather than refetch all DS list every time

esx_service/utils/vmdk_utils.py
```diff
@@ -100,8 +100,11 @@ def get_datastores():
     'url-name' is the last element of datastore URL (e.g. 'vsan:572904f8c031435f-3513e0db551fcc82')
     'dockvol-path; is a full path to 'dockvols' folder on datastore
     """
-    if datastores == None:
-        init_datastoreCache()
+    # if datastores == None:
+    #     init_datastoreCache()
```
msterin: Why this change is introduced ? It seems to irrelevant to #816

esx_service/utils/auth_api.py
```diff
@@ -24,6 +24,10 @@
 import logging
 from error_code import ErrorCode
 from error_code import ErrorInfo
+import re
+
+# regex for valid tenant name
+VALID_TENANT_NAME_REG = "[a-zA-Z0-9_][a-zA-Z0-9_.-]*"
```
msterin: Need to add documentation for this, a one liner

esx_service/utils/vmdk_utils.py
```diff
@@ -100,8 +100,11 @@ def get_datastores():
...
+    # we should not use cached value, user may change datastore name on the fly
+    # need to repull the info
+    init_datastoreCache()
```
shaominchen: With this change, we are losing the purpose of introducing datastore cache. Have we filed an issue to fix the datastore cache implementation?

esx_service/utils/auth_api.py
```diff
@@ -239,9 +243,22 @@ def get_default_datastore(name):
         error_info = error_code.generate_error_info(ErrorCode.INTERNAL_ERROR, error_msg)
     return error_info, default_datastore
 
+def is_valid_tenant_name(name):
+    """ Check the given name is a valid tenant name """
+    pattern = "^" + VALID_TENANT_NAME_REG + "$"
```
msterin: nit: why not pre-compile it once and keep the compiled regexp in a global ? It's not a bit deal since the operation is rare, but generally it's a waste.

esx_service/utils/auth_api.py
```diff
@@ -239,9 +243,22 @@ def get_default_datastore(name):
         error_info = error_code.generate_error_info(ErrorCode.INTERNAL_ERROR, error_msg)
     return error_info, default_datastore
 
+def is_valid_tenant_name(name):
+    """ Check the given name is a valid tenant name """
```
shaominchen: Nit: "Check..." => "Check if..."

esx_service/utils/auth_api.py
```diff
@@ -239,9 +243,22 @@ def get_default_datastore(name):
...
+    pattern = "^" + VALID_TENANT_NAME_REG + "$"
+    valid_name_reg = re.compile(pattern)
+    if valid_name_reg.match(name):
```
shaominchen: Nit: we can simplify it a little bit by removing the if-else block:

   return valid_name_reg.match(name)

esx_service/vmdk_ops.py
```diff
@@ -153,12 +157,24 @@ def RunCommand(cmd):
...
+        return True
+    else:
+        return False
 
 # returns error, or None for OK
 # opts is  dictionary of {option: value}.
 # for now we care about size and (maybe) policy
 def createVMDK(vmdk_path, vm_name, vol_name, opts={}, vm_uuid=None, tenant_uuid=None, datastore=None):
     logging.info("*** createVMDK: %s opts = %s", vmdk_path, opts)
+    if not is_valid_volume_name(vol_name):
```
msterin: I think we already have  code to parse volume names - please check if can re-use

esx_service/vmdk_ops.py
```diff
@@ -153,12 +157,24 @@ def RunCommand(cmd):
 
     return (s, o)
 
+def is_valid_volume_name(name):
+    """ Check the given name is a valid volume name """
```
shaominchen: Nit: "Check..." => "Check if..."

esx_service/vmdk_ops.py
```diff
@@ -153,12 +157,24 @@ def RunCommand(cmd):
...
+    pattern = "^" + VALID_VOLUME_NAME_REG + "$"
+    valid_name_reg = re.compile(pattern)
+    if valid_name_reg.match(name):
```
shaominchen: Nit: Same as above - we can simplify it a little bit:

   return valid_name_reg.match(name)

esx_service/utils/vmdk_utils.py
lipingxue: This change is needed. Think about this case:
step 1: assume we have datastore named "datastore1", which is the default_datastore, we create a volume "vol1"
step 2: rename "datastore1" to "datastore 1" (including space in the name)
step3: create another vol "vol2", which will fail.  

It will hit the error same as #818 . The root cause is "vmdk_utils.get_datastores" returns cached value which is outdated.

esx_service/utils/vmdk_utils.py
lipingxue: I discussed with Mark offline, we decided to remove this cache related change from this PR.

esx_service/utils/vmdk_utils.py
lipingxue: As we discussed offline, will remove this change from this PR.

esx_service/vmdk_ops.py
lipingxue: We have a util function parse_vol_name, which check whether the vol_name is match r"^.*-[0-9]{6}$" . So there is no limit on which character will be included in the vol_name. But docker local driver limit vol_name can only be "[a-zA-Z0-9][a-zA-Z0-9_.-]" . Do we want have the same limit as docker local driver does?

lipingxue: @msterin

esx_service/vmdk_ops.py
lipingxue: Done. Remove this function and introduce a global variable "valid_volume_name_reg", use valid_volume_name_reg.match() to check directly.

esx_service/vmdk_ops.py
lipingxue: Done.

esx_service/vmdk_ops.py
```diff
@@ -129,6 +130,9 @@
 # For managing resource locks.
 lockManager = threadutils.LockManager()
 
+# regex for valid volume name
+VALID_VOLUME_NAME_REG = "[a-zA-Z0-9][a-zA-Z0-9_.-]*"
```
pdhamdhere: Why do we need this one? We already validate volume name in `parse_vol_name`.

esx_service/utils/vmdk_utils.py
pdhamdhere: Are you going to remove `init_datastoreCache()`

esx_service/vmdk_ops.py
lipingxue: docker local driver has this limitation on volume names during volume create. I think we should apply the same check.  We can discuss if you think differently. @msterin

govint:
When trying out with a vol@<data store name with spaces> figured the below:

1. Run command to mkdir - failed because of spaces, quoting the path name
passed to mkdir worked.

2. find_devicebypath() in vmdk_ops.py has some code to get the datastore
name from a vmdk path and thats splitting the path incorrectly when the
datastore name has spaces in it.

Basically, fix an issue and then it breaks elsewhere. Keep fixing the
issues till the entire use case (create/get/remove/attach/detach) works
fine.

With (1) above fixed create is ok, attach/detach fail in (2) above.

esx_service/vmdk_ops.py
msterin: We do not care what local diver does, so unless there is a good reason for  the change allowed volnames, please remove this code. 

And, as Prashant has mentioned, the parsing is happening in `parse_vol_name` so any changes should be done there.

esx_service/vmdk_ops.py
msterin: parse_vol_name does more that that - it also separated DS name and handles them differnetly.
And no, I see no reason to mimic local local driver, we have deviated from it already by adding `@` to full volume name.

esx_service/utils/vmdk_utils.py
lipingxue: No, we will keep init_datastoreCache().

lipingxue:
@msterin I have addressed your comments:
1. remove "whitespace cleanup" from this PR
2. remove the volume name checking during volume creation

Please review it.

lipingxue:
Since Sam is on PTO, @msterin @govint @pdhamdhere please review it.

esx_service/utils/auth_api.py
```diff
@@ -241,7 +247,12 @@ def get_default_datastore(name):
 
 def _tenant_create(name, description="", vm_list=None, privileges=None):
     """ API to create a tenant """
+    global valid_tenant_name_reg
     logging.debug("_tenant_create: name=%s description=%s vm_list=%s privileges=%s", name, description, vm_list, privileges)
+    if not valid_tenant_name_reg.match(name):
+        error_info = error_code.generate_error_info(ErrorCode.TENANT_NAME_INVALID, name, VALID_TENANT_NAME_REGEXP)
+        return error_info, None
+
```
govint: Can this be a function like is_tenant_name_valid() vs. repeated in multiple locations.

esx_service/vmdk_ops_test.py
```diff
@@ -715,20 +716,27 @@ def create_default_tenant_and_privileges(self):
...
+            if error_info:
+                logging.warning(error_info.msg)
 
         # create DEFAULT privilege if needed
-        error_info, privileges = auth.get_default_privileges()
+        error_msg, privileges = auth.get_default_privileges()
+        if error_msg:
+            logging.warning(error_msg)
+            return
+
```
govint: This action and error handling must be done before tenant create, its leaving a new tenant created and left in an unknown state (line 726 above).

docs/user-guide/admin-cli.md
```diff
@@ -39,17 +39,17 @@ optional arguments:
 ### Create
 A tenant named "_DEFAULT" will be created automatically post install.
 
-Creates a new named tenant and optionally assigns VMs.
+Creates a new named tenant and optionally assigns VMs. Valid tenant name is only allowed to be "[a-zA-Z0-9_][a-zA-Z0-9_.-]*"
```
shuklanirdesh82: @lipingxue  a quick question: PR description says ```Fixes issue 816 (space in DS names).``` can you please help to understand why some of the fixes are related with ```TENANT_NAME_INVALID``` or ```VOLUME_NAME_INVALID```?

esx_service/vmdk_ops_test.py
```diff
@@ -715,20 +716,27 @@ def create_default_tenant_and_privileges(self):
         """ Create default tenant and privilege if not exist"""
 
         # create DEFAULT tenant if needed
-        error_info, tenant_uuid, tenant_name = auth.get_default_tenant()
+        error_msg, tenant_uuid, tenant_name = auth.get_default_tenant()
+        if error_msg:
+            logging.warning(error_msg)
```
shuklanirdesh82: don't we need to fail the test here ( not able to create default tenant)?

esx_service/vmdk_ops.py
```diff
@@ -855,11 +856,10 @@ def findDeviceByPath(vmdk_path, vm):
         # Filename format is as follows:
         #   "[<datastore name>] <parent-directory>/tenant/<vmdk-descriptor-name>"
         logging.debug("d.backing.fileName %s", d.backing.fileName)
-        backing_disk = d.backing.fileName.split(" ")[1]
-
-        # datastore='[datastore name]'
-        datastore = d.backing.fileName.split(" ")[0]
-        datastore = datastore[1:-1]
+        ds, disk_path = d.backing.fileName.rsplit("]", 1)
```
shuklanirdesh82: can you please confirm about having a test for this code snip?

esx_service/utils/auth_api.py
```diff
@@ -24,6 +24,12 @@
 import logging
 from error_code import ErrorCode
 from error_code import ErrorInfo
+import re
+
+# regex for valid tenant name
+VALID_TENANT_NAME_REGEXP = "[a-zA-Z0-9_][a-zA-Z0-9_.-]*"
```
shuklanirdesh82: 1. Is it something we are introducing new?
2. What happens if customer is having tenant name that violates this rule? I guess new tenant creates enforces the restriction but would like to make sure here.

esx_service/utils/auth_api.py
```diff
@@ -241,7 +247,12 @@ def get_default_datastore(name):
 
 def _tenant_create(name, description="", vm_list=None, privileges=None):
     """ API to create a tenant """
+    global valid_tenant_name_reg
     logging.debug("_tenant_create: name=%s description=%s vm_list=%s privileges=%s", name, description, vm_list, privileges)
+    if not valid_tenant_name_reg.match(name):
```
shuklanirdesh82: 1. need to add testing done for this code snip
2. can you confirm about having tests exercising this new change?

esx_service/utils/auth_api.py
lipingxue: Done.

esx_service/vmdk_ops_test.py
lipingxue: Addressed your comments. Now the logic is
1. check default_tenant exists or not
2. if default_tenant exists, do nothing, continue
3. if default_tenant does not exist, try to create, assert the create should not return error, since the test need to start with default_tenant exist

Similar logic also apply to default_privilege

docs/user-guide/admin-cli.md
lipingxue: This tenant name validation fix is not related to issue 816 itself.  Sam pointed out earlier that in his comment to suggest that we should give some limitation to the tenant name during creation. For example, no space allowed in the tenant name, which will avoid some potential issues that may occur like issue 816 if user specify tenant name with spaces and other weird characters.

esx_service/vmdk_ops_test.py
lipingxue: Yes. I addressed this. See above comments I replied to Govindan's comments.

esx_service/utils/auth_api.py
lipingxue: Yes, the tenant create with invalid name should fail.

docs/user-guide/admin-cli.md
shuklanirdesh82: It would be helpful if you can add some useful information in the PR title as well as in the first comment. 

Such information will be useful for future reference while searching from title.  Current title proposing the ```Fixes issue 816 (space in DS names).```

lipingxue:
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol2
vol2
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls                              
vmdk                vol1@datastore    1
vmdk                vol2@datastore    1

lipingxue:
Test to make sure vol can be created on a datastore which name includes spaces.
Configuration: 
only _DEFAULT tenant is configured, VM "photon5" is created on "datastore1"
Step1: From VM"photon5", create a volume "vol1"
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol1
vol1
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore1
```

Step2:
rename datastore "datastore1" to "datastore    1" (to insert spaces)
restart /etc/init.d/vmdk-opsd service. This is to avoid hitting the issue caused by stale datastore info returned by datastore cache(issue 818, which I am working on).  

Step3: create another volume "vol2"
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore    1
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol2
vol2
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls                              
vmdk                vol1@datastore    1
vmdk                vol2@datastore    1

```

lipingxue:
Test to make sure create tenant with invalid name will fail

create a tenant with invalid name like "tenant     abc" failed  as expect.
```
[root@localhost:/vmfs/volumes/57473702-8b550986-70c0-000c290799f4/dockvols] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
------------------------------------  --------  ------------------------  -----------------  -------  
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default tenant                              
[root@localhost:/vmfs/volumes/57473702-8b550986-70c0-000c290799f4/dockvols] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name="tenant     abc"
Tenant name tenant     abc is invalid, only [a-zA-Z0-9_][a-zA-Z0-9_.-]* is allowed

```

docs/user-guide/admin-cli.md
lipingxue: OK. Change the title.

esx_service/vmdk_ops.py
lipingxue: Yes. I added test for this.

govint:
@lipingxue can you also add a test that creates the volume on a datastore with spaces in it like "docker volume create --driver=vmdk --name=vol1@data\ \ \ store"

lipingxue:
@govint Added the test you request
Configuration: only _DEFAULT tenant is configured, VM "photon5" is created on datastore1
Step1: create two volumes "vol1" and "vol2"
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol1
vol1
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol2
vol2
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore1
vmdk                vol2@datastore1

```
Step2:
rename datastore "datastore1" to "datastore   1"
restart /etc/init.d/vmdk-opsd service. This is to avoid hitting the issue caused by stale datastore info returned by datastore cache(issue 818, which I am working on).

Step3: create a volume "vol3@datastore   1"
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name="vol3@datastore   1"
vol3@datastore   1

```

Step4: create a volume "vol4@datastore\ \ \ 1" with the following command
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol4@datastore\ \ \ 1
vol4@datastore   1
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore   1
vmdk                vol2@datastore   1
vmdk                vol3@datastore   1
vmdk                vol4@datastore   1

```

govint:
@lipingxue thanks and looks good.

esx_service/vmdk_ops.py
```diff
@@ -87,6 +87,7 @@
 import error_code
 import auth_api
 from error_code import ErrorCode
+import re
```
shuklanirdesh82: trivial: I guess this needs to be removed as well.

shuklanirdesh82: never mind .. not sure why diff is saying it is added as part of this change. Please ignore above comment. Sorry for the confusion.

Thanks!

lipingxue: Done. Thanks for review.

PR #906, opened 2017-02-13
Added refcount decrement for mount errors.

govint:
Minor fix to add refcount decrement on all errors when mounting a volume.

For release 0.12, the release notes must reference the docker issue and mention that with Docker 1.12 mount error may cause the volume to be unusable (owing to an outstanding reference count) unless the plugin is restarted (which anyway won't affect existing running containers.


vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -241,6 +241,10 @@ func (r RefCountsMap) Decr(vol string) (uint, error) {
 	}
 
 	rc.count--
+
+	if rc.count < 0 {
+		log.Warning("Decr: refcnt is negative, name=%s refcnt=%d", vol, rc.count)
```
msterin: So the corrupted slot will stay in the map ? 
I think it would be better to simply change to ` if rc.count <=0` (and update message) on line 234. This will also delete the map entry.

govint: Agree, fixed.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -241,8 +241,13 @@ func (r RefCountsMap) Decr(vol string) (uint, error) {
 	}
 
 	rc.count--
+
+	if rc.count < 0 {
+		log.Warning("Decr: refcnt is negative deleting,
```
msterin: this message does not make sense to me , can you rephrase ?  Something like `"Decr: internal error - refcnt is negative. Trying to recover - deleting the counter. name = ..., refct=...")...`.

govint: True, updated the error message.

PR #910, opened 2017-02-15
[TRIVIAL] Reading testbed environment vars in correct order

shuklanirdesh82:
Reading testbed env variable in sequence to avoid confusion while triaging failures.

//CC @lipingxue @pshahzeb


shuklanirdesh82:
My bad for not adding enough information. 

>  -e "VM1=$4" \
 > -e "VM2=$3" \

with the above change VM1 was getting IP address of VM2 (environment variable) and vice versa that leads to confusion.

PR #912, opened 2017-02-15
Resolving some go build warnings

shuklanirdesh82:
This PR is addressing to resolve following build warnings.

From CI (https://ci.vmware.run/vmware/docker-volume-vsphere/1178):

```
utils/fs/fs.go:67:9: if block ends with a return statement, so drop this else and outdent its block
utils/fs/fs.go:77:1: exported function DevAttachWait should have comment or be unexported
```

with the fix 

from latest CI run: https://ci.vmware.run/vmware/docker-volume-vsphere/1183

```
$ make -s build
Staging in ../build/payloads
GENERATING VMODL PAYLOAD...
```


vmdk_plugin/utils/fs/fs.go
```diff
@@ -60,20 +60,23 @@ type VolumeDevSpec struct {
 // DevAttachWaitPrep create the watcher.
 func DevAttachWaitPrep(name string, devPath string) (*inotify.Watcher, bool) {
 	watcher, errWatcher := inotify.NewWatcher()
+	needToSkipInotify := false
```
pdhamdhere: I think rather than introducing new variable, follow suggestions in error;
```
	if errWatcher != nil {
		log.WithFields(log.Fields{"Name": name}).Error("Failed to create watcher, skip inotify ")
		return nil, true
	} 
	err := watcher.Watch(devPath)
	if err != nil {
		log.WithFields(log.Fields{"Name": name}).Error("Failed to watch /dev, skip inotify ")
                return nil, true
	}
        err := watcher.Watch(devPath)
         if err != nil {
                   log.WithFields(log.Fields{"Name": name}).Error("Failed to watch /dev, skip inotify ")
                   return nil, true
         }
```

shuklanirdesh82: Done! reverted the proposed changed and move out the else part code to resolve the warning.

PR #913, opened 2017-02-15
Adding retry mechanism to get VM IP

shuklanirdesh82:
This work is under progress. 

https://ci.vmware.run/vmware/docker-volume-vsphere/1181 ran into infinite loop, I will update the PR when it is ready for the review.

with the proposed changed you may see following improvements

```
Resetting testbed
Waiting for revert to complete
ESX 6.5
192.168.31.62
ESX 6.0
192.168.31.19
Reset complete
Wait for VM to get ready
govc: Post https://192.168.31.62/sdk: dial tcp 192.168.31.62:443: getsockopt: no route to host
govc: Post https://192.168.31.62/sdk: dial tcp 192.168.31.62:443: getsockopt: no route to host
govc: Post https://192.168.31.62/sdk: dial tcp 192.168.31.62:443: getsockopt: no route to host
govc: Post https://192.168.31.62/sdk: dial tcp 192.168.31.62:443: getsockopt: no route to host
192.168.31.124
192.168.31.85
Resume testing
```


misc/drone-scripts/setup.sh
```diff
@@ -33,8 +33,20 @@ do
     sleep 1;
 done
 
-echo "Reset complete"
-sleep 5;
+echo "Reset complete\nWaiting for VM to get IP..."
+
+until govc vm.ip photon.vmfs
```
pdhamdhere: I am always scared with unbounded sleeps. Please add some timeout and fail the run.

shuklanirdesh82: Yeah that's correct. I will work on the retry mechanism with timeout and generate the diff again.

Thanks!

misc/drone-scripts/setup.sh
```diff
@@ -19,22 +19,28 @@ govc snapshot.revert -vm $ESX_6_0 init
 
 echo "Waiting for revert to complete";
 
+DIR=$(dirname ${BASH_SOURCE[0]})
+. $DIR/../scripts/wait_for.sh
+
+timeout=30
```
pdhamdhere: minutes, seconds. Add it in a comment?

shuklanirdesh82: sure thing! I have renamed the variable and supplied the inline comment too.

pdhamdhere:
How long each retry?

shuklanirdesh82:
> How long each retry?

every 1 seconds and waits fro 30 seconds before timing out the operation.

misc/drone-scripts/setup.sh
```diff
@@ -19,22 +19,29 @@ govc snapshot.revert -vm $ESX_6_0 init
 
 echo "Waiting for revert to complete";
 
+DIR=$(dirname ${BASH_SOURCE[0]})
+. $DIR/../scripts/wait_for.sh
+
+# Threshold to time out
```
govint: is this 30 second timeout.

misc/drone-scripts/setup.sh
```diff
@@ -19,22 +19,29 @@ govc snapshot.revert -vm $ESX_6_0 init
...
-done
+wait_for "govc vm.ip $ESX_6_0" $retryCount
 
 echo "Reset complete"
-sleep 5;
+
+export GOVC_URL=$GOVC_URL_6_5
+export GOVC_USERNAME=$GOVC_USERNAME_ESX
+export GOVC_PASSWORD=$GOVC_PASSWORD_ESX
+
```
govint: Do we need to set GOVC_DATACENTER also? The VMs will be configured in one in VC.

misc/drone-scripts/setup.sh
```diff
@@ -19,22 +19,29 @@ govc snapshot.revert -vm $ESX_6_0 init
...
+echo "Wait for VM to get ready"
+
+wait_for "$GOVC_GET_IP photon.vmfs" $retryCount
+wait_for "$GOVC_GET_IP Ubuntu.16.10" $retryCount
+
```
govint: Could add comments here and as appropriate just so if another VM is added then user knows how to update this code. Or loop over a set of VMs and do the wait in sequence.

misc/drone-scripts/setup.sh
shuklanirdesh82: wait_for method check every 1 seconds and threshold has been set for 30 time i.e. 30 seconds maximum to time out.

misc/drone-scripts/setup.sh
shuklanirdesh82: This is sufficient to login to ESX 6.5 host through govc bindings.

govint: Ok, is it possible to verify VC with more than one datacenter configured and with same VM names in each.

misc/drone-scripts/setup.sh
shuklanirdesh82: Sure, I will add some inline comments for the future reference. Thanks for pointing out!

> Or loop over a set of VMs and do the wait in sequence.

Could do that. Based on my experience, the issue has been observed against ESX 6.5 as it is the first one to run tests against so it would be OK for user not to add any more vms to the list unless replacing them with new one.

misc/drone-scripts/setup.sh
shuklanirdesh82: I believe it is achievable but in our scenario we don't have VC.

PR #919, opened 2017-02-16
Rename vmdk driver as vsphere.

govint:
Updated docs and the plugin init code to expose a "vsphere" driver.

plugin logs

2017-02-16 01:42:49.690079924 -0800 PST [INFO] No config file found. Using defaults.
2017-02-16 01:42:49.690279065 -0800 PST [INFO] Starting plugin log_level=debug config="/etc/docker-volume-vsphere.conf" driver=vsphere
2017-02-16 01:42:49.690325892 -0800 PST [WARNING] Failed to load config file %s: %v/etc/docker-volume-vsphere.confopen /etc/docker-volume-vsphere.conf: no such file or directory
2017-02-16 01:42:49.690346301 -0800 PST [INFO] Plugin options - port=1019
2017-02-16 01:42:49.690389849 -0800 PST [INFO] Getting volume data from unix:///var/run/docker.sock
2017-02-16 01:42:49.763661689 -0800 PST [DEBUG] Docker info: version=1.13.0-rc1, root=/var/lib/docker, OS=Ubuntu 14.04.5 LTS
2017-02-16 01:42:49.764895552 -0800 PST [DEBUG] Found 0 running or paused containers
2017-02-16 01:42:49.765255637 -0800 PST [INFO] Discovered 0 volumes in use.
2017-02-16 01:42:49.765306372 -0800 PST [INFO] Docker Vsphere plugin started version="VSphere Volume Driver v0.3" port=1019 mock_esx=false
2017-02-16 01:42:49.765423389 -0800 PST [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vsphere.sock"
2017-02-16 01:42:49.765706329 -0800 PST [DEBUG] root group found. gid: 0

Docker reports the new driver.

$ docker volume ls
vsphere             newvol1@bigone
vsphere             newvol2@bigone
vsphere             newvol3@bigone
vsphere             newvol@bigone

### DIFF
--- a/docs/user-guide/docker-plugin-drivers.md
+++ b/docs/user-guide/docker-plugin-drivers.md
@@ -1,21 +1,24 @@
 [TOC]
-The Docker volume plugin supports the below platforms and the corresponding drivers for those. The plugin supports all volume provisioning and managenment operations, defined by the Docker Volume plugin interface, on both platforms.
+The Docker volume plugin supports the below platforms and the corresponding drivers for those. The plugin supports all volume provisioning and managenment operations, defined by the Docker Volume plugin interface, on both platforms. Both drivers allow users to provision and use VMDK backed volumes for containers in Docker.
 
-1. Vmdk
+1. vSphere
 2. Photon
 
 <script type="text/javascript" src="https://asciinema.org/a/80417.js" id="asciicast-80417" async></script>
 
-## Docker Vmdk volume driver 
-The Docker Vmdk volume driver supports provisioning and managing docker volumes on a standalone or cluster of ESX servers via a service (ESX service) that's installed and runs on each server. Docker volumes are created and managed via publicly available VIM (Virtual Infrastructure Management) APIs on the ESX host.
+## Docker vsphere volume driver
+The Docker vsphere volume driver supports provisioning and managing docker volumes on a standalone or cluster of ESX servers via a service (ESX service) that's installed and runs on each server. Docker volumes are created and managed via publicly available VIM (Virtual Infrastructure Management) APIs on the ESX host.
 
-## Docker Photon volume driver 
-The Docker Photon Volume driver supports provisioning and managing docker volumes on a Photon platform consisting of a cluster of ESX hosts managed via a Photon controller instance. Docker volumes are created and managed entirely via the open Photon platform API via the Photon controller.
+## Docker photon volume driver
+The Docker photon volume driver supports provisioning and managing docker volumes on a Photon platform consisting of a cluster of ESX hosts managed via a Photon controller instance. Docker volumes are created and managed entirely via the open Photon platform API via the Photon controller.
 
 The Docker Volume plugin can support either or both types of volumes, as required, on a given Docker host.
 
 ## Configuring the Docker Volume Plugin
-The docker volume plugin is designed to load run time options and values from a json configuration file (default /etc/docker-volume-vsphere.conf) on the host. The user can also provide a configuration file, via the "--config" option, specifying the full path of the file. The file contains the values for run time options used by the plugin. Options that are currently recognized include the below set. Options provided on the command line will override those in the configuration file.
+The docker volume plugin loads runtime options and values from a json configuration file (default /etc/docker-volume-vsphere.conf) on the host. The user can override the default configuration by providing a different configuration file, via the "--config" option, specifying the full path of the file. Options that are currently recognized include the below set. Options passed on the command line override those in the configuration file.
+
+### Selecting the driver to handle volume operations
+The docker volume plugin supports two drivers, namely, "photon" and "vsphere" for the Photon and vSphere platforms respectively. The "vsphere" driver was earlier named as "vmdk" and the plugin still supports both names. The "vmdk" driver name can be used in place of "vsphere" for now, but will be deprecated in a later release. The choice of driver is specified as below in the sample configuration. The plugin uses "vsphere' as the default driver, which is overriden via the configuration file for the Photon platform.
 
 ### Options for the photon volume driver
 Target    - URL at which to contact the Photon Controller
@@ -28,8 +31,9 @@ LogPath       - location where plugin log fils are created
 MaxLogSizeMb  - max. size of the plugin log file
 MaxLogAgeDays - number of days to retain plugin log files
 
-## Sample plugin configuration, the "Target", "Project" and "Host" options are for photon only.
+## Sample plugin configuration
 {
+	"Driver": "<driver name - vsphere/vmdk/photon>"
 	"MaxLogAgeDays": 28,
 	"MaxLogSizeMb": 100,
 	"LogPath": "/var/log/docker-volume-vsphere.log",
--- a/docs/user-guide/docker-volume-cli.md
+++ b/docs/user-guide/docker-volume-cli.md
@@ -8,33 +8,33 @@ The service works with existing Docker volume commands.
 4. [Docker volume rm](https://docs.docker.com/engine/reference/commandline/volume_rm/)
 
 
-The Docker volume commands are supported for both the Vmdk and Photon platforms with minor differences in capabilities. Features that are specific to either of the platforms are mentioned explicitly below.
+The Docker volume commands are supported for both the vSphere and Photon platforms with minor differences in capabilities. Features that are specific to either of the platforms are mentioned explicitly below.
 <script type="text/javascript" src="https://asciinema.org/a/80417.js" id="asciicast-80417" async></script>
 
 ## Docker volume create options
 ### size
 
 ```
-docker volume create --driver=<vmdk/photon> --name=MyVolume -o size=10gb
+docker volume create --driver=<vsphere/photon> --name=MyVolume -o size=10gb
 ```
 
 The volume units can be ```mb, gb and tb```
 
 The default volume size is 100mb
 
-### vsan-policy-name (Vmdk only)
+### vsan-policy-name (vSphere only)
 
 ```
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o vsan-policy-name=allflash
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o vsan-policy-name=allflash
 ```
 
 Policy needs to be created via the vmdkops-admin-cli. Once policy is created, it can be addressed during create by passing the ```-o vsan-policy-name``` flag.
 
-### diskformat (Vmdk only)
+### diskformat (vSphere only)
 ```
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o diskformat=zeroedthick
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o diskformat=thin
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o diskformat=eagerzeroedthick
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o diskformat=zeroedthick
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o diskformat=thin
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o diskformat=eagerzeroedthick
 ```
 
 Docker volumes are backed by VMDKs. VMDKs support multiple [types](https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1022242)
@@ -45,10 +45,10 @@ Currently the following are supported
 2. Thin Provision ([thin]((https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1022242)))
 3. Thick Provision Eager Zeroed ([eagerzeroedthick]((https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1022242)))
 
-### attach_as (Vmdk only)
+### attach_as (vSphere only)
 ```
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o attach_as=independent_persistent
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o attach_as=persistent
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o attach_as=independent_persistent
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o attach_as=persistent
 ```
 Docker volumes are backed by VMDKs. VMDKs are attached to the VM in which Docker is requesting for a volume during Docker run. VMDKs can be attached in [different modes.](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/)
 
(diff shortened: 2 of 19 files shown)


docs/misc/plugin-lifecycle.md
```diff
@@ -22,7 +22,7 @@ A plugin's lifecycle includes.
 Note: Upgrade of the plugin is done as an uninstall of the older version and an install of the newer version.
 
 1. Install: If it is the first install Docker can be running while the plugin is installed. During installs that are part of an upgrade, the admin must insure that docker is stopped and is started after the plugin starts up.
-2. Uninstall: During uninstall docker should be stopped. Before uninstall, make sure no containers still use the VMDK volumes, stop docker engine, and then uninstall the plugin. Docker will generate timeouts on docker start if there are still VMDK volumes after the plugin is uninstalled.
+2. Uninstall: During uninstall docker should be stopped. Before uninstall, make sure no containers still use the Vsphere volumes, stop docker engine, and then uninstall the plugin. Docker will generate timeouts on docker start if there are still Vsphere volumes after the plugin is uninstalled.
```
pdhamdhere: Vsphere => "vSphere" that seems to standard vmware convention. Please find & replace at other places in change.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -41,8 +41,8 @@ const (
 	devWaitTimeout   = 1 * time.Second
 	sleepBeforeMount = 1 * time.Second
 	watchPath        = "/dev/disk/by-path"
-	version          = "VMDK Volume Driver v0.3"
-	driverName       = "vmdk"
+	version          = "VSphere Volume Driver v0.3"
+	driverName       = "vsphere"
```
pdhamdhere: I think we should keep, "vmdk" for backward compatibility. If possible warn user that "vmdk" driver type is deprecated and use "vSphere" instead.

vmdk_plugin/main.go
```diff
@@ -38,7 +38,7 @@ const (
 	pluginSockDir = "/run/docker/plugins"
 	mountRoot     = "/mnt/vmdk" // VMDK and photon volumes are mounted here
 	photonDriver  = "photon"
-	vmdkDriver    = "vmdk"
+	vmdkDriver    = "vsphere"
```
pdhamdhere: Looks like driverName is case sensitive. 
```
docker volume create -d VMDK --name=v1@bad
Error response from daemon: create v1@bad: create v1@bad: Error looking up volume plugin VMDK: plugin not found
```

"vSphere" with uppercase S is error prone. Can we register multiple times - "vsphere", "vSphere", "VSPHERE". Or a PR to Docker to make plugin name case insensitive?

govint:
Can we instead use "vsphere", because none of the docker command options
are upper case and an incorrect name will return error. Plus Photon driver
is also specified as "photon", or we'll have to change that to Photon.

pdhamdhere:
>Can we instead use "vsphere", because none of the docker command options

Sure.

shaominchen:
I agree with Mark and Prashant that we should keep backward compatibility for the "vmdk" driver name, and warn user it is deprecated. As for the case sensitivity of the plugin name, I would suggest to favor local consistency - since docker is using lower case consistently, we should just follow that.

govint:
Thanks for reviewing. Updated diffs based on comments.

1. Corrected to use vSphere when logging driver init.
2. Updated docs to indicate the two names for the vsphere volume plugins and that "vmdk" will be deprecated.
3. Supporting both names for now. But catch is at some point user must stop using the vmdk name.

Release notes for 0.12 release must mention the change name and plan to deprecate by 0.13.

docs/user-guide/docker-volume-cli.md
```diff
@@ -8,33 +8,33 @@ The service works with existing Docker volume commands.
 4. [Docker volume rm](https://docs.docker.com/engine/reference/commandline/volume_rm/)
 
 
-The Docker volume commands are supported for both the Vmdk and Photon platforms with minor differences in capabilities. Features that are specific to either of the platforms are mentioned explicitly below.
+The Docker volume commands are supported for both the Vsphere and Photon platforms with minor differences in capabilities. Features that are specific to either of the platforms are mentioned explicitly below.
```
shaominchen: Vsphere => vSphere. For documents/comments, I think we should use "vSphere".

docs/user-guide/docker-volume-cli.md
```diff
@@ -8,33 +8,33 @@ The service works with existing Docker volume commands.
...
-docker volume create --driver=<vmdk/photon> --name=MyVolume -o size=10gb
+docker volume create --driver=<vsphere/photon> --name=MyVolume -o size=10gb
 ```
 
 The volume units can be ```mb, gb and tb```
 
 The default volume size is 100mb
 
-### vsan-policy-name (Vmdk only)
+### vsan-policy-name (Vsphere only)
```
shaominchen: Same as above.

docs/user-guide/docker-volume-cli.md
```diff
@@ -8,33 +8,33 @@ The service works with existing Docker volume commands.
...
 
 ```
-docker volume create --driver=vmdk --name=MyVolume -o size=10gb -o vsan-policy-name=allflash
+docker volume create --driver=vsphere --name=MyVolume -o size=10gb -o vsan-policy-name=allflash
 ```
 
 Policy needs to be created via the vmdkops-admin-cli. Once policy is created, it can be addressed during create by passing the ```-o vsan-policy-name``` flag.
 
-### diskformat (Vmdk only)
+### diskformat (Vsphere only)
```
shaominchen: Same as above.

docs/user-guide/docker-volume-cli.md
```diff
@@ -45,10 +45,10 @@ Currently the following are supported
 2. Thin Provision ([thin]((https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1022242)))
 3. Thick Provision Eager Zeroed ([eagerzeroedthick]((https://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1022242)))
 
-### attach_as (Vmdk only)
+### attach_as (Vsphere only)
```
shaominchen: Same as above.

docs/user-guide/docker-volume-cli.md
```diff
@@ -57,33 +57,33 @@ Currently the following are supported
 1. [persistent](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/): If the VMDK is attached as persistent it will be part of a VM snapshot. If a VM snapshot has been taken while the Docker volume is attached to a VM, the Docker volume then continues to be attached to the VM that was snapshotted.
 2. [independent_persistent](http://cormachogan.com/2013/04/16/what-are-dependent-independent-disks-persistent-and-non-persisent-modes/): If the VMDK is attached as independent_persistent it will not be part of a VM snapshot. The Docker volume can be attached to any VM that can access the datastore independent of snapshots.
 
-### access (Vmdk only)
+### access (Vsphere only)
```
shaominchen: Same as above.

docs/user-guide/docker-volume-cli.md
```diff
@@ -57,33 +57,33 @@ Currently the following are supported
...
-docker volume create --driver=<vmdk/photon> --name=MyVolume -o size=10gb -o fstype=xfs
-docker volume create --driver=<vmdk/photon> --name=MyVolume -o size=10gb -o fstype=ext4 (default)
+docker volume create --driver=<vsphere/photon> --name=MyVolume -o size=10gb -o fstype=xfs
+docker volume create --driver=<vsphere/photon> --name=MyVolume -o size=10gb -o fstype=ext4 (default)
 ```
 
 Specifies which filesystem will be created on the new volume. vSphere Docker Volume Service will search for a existing /sbin/mkfs.**fstype** on the docker host to create the filesystem, and if not found it will return a list of filesystems for which it has found a corresponding mkfs. The specified filesystem must be supported by the running kernel and support labels (-L flag for mkfs). Defaults to ext4 if not specified. 
 
-### clone-from (Vmdk only)
+### clone-from (Vsphere only)
```
shaominchen: Same as above.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -15,7 +15,7 @@
 package vmdk
 
 //
-// VMWare VMDK Docker Data Volume plugin.
+// VMWare Vsphere Docker Data Volume plugin.
```
shaominchen: Same as above.

vmdk_plugin/main.go
```diff
@@ -151,7 +152,7 @@ func main() {
 		}
 		driver = photon.NewVolumeDriver(*targetURL, *projectID,
 			*vmID, mountRoot)
-	} else if *driverName == vmdkDriver {
+	} else if *driverName == vmdkDriver || *driverName == vsphereDriver {
```
shaominchen: Can we print a message to warn user that "vmdk" driver is deprecated (print a warning message, but still continue to run the command)?

govint: Thanks, fixed the comments and the docs and pushed again.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -41,8 +41,8 @@ const (
 	devWaitTimeout   = 1 * time.Second
 	sleepBeforeMount = 1 * time.Second
 	watchPath        = "/dev/disk/by-path"
-	version          = "VMDK Volume Driver v0.3"
-	driverName       = "vmdk"
+	version          = "vSphere Volume Driver v0.3"
```
msterin: I think version is overdue for the change

govint: Notes in docs/user-guide/docker-plugin-drivers.md is updated to mention the change and deprecation of the vmdk driver name. 

For version number, which number should we use?

govint:
Updated the change to retain "vmdk" as the default for now (when no args are provided). This is so existing users can continue to use the current name till the "vmdk" name is deprecated. Users can still use "vsphere" if they start the plugin with the "-driver=vsphere" option. Later we will remove the vmdk name all together and make "vsphere" the default.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pdhamdhere: v0.4 with this change?

vmdk_plugin/main.go
```diff
@@ -97,8 +98,8 @@ func main() {
 	var driver volume.Driver
 
 	// Define command line options
-	driverName := flag.String("driver", "vmdk", "Volume driver")
-	logLevel := flag.String("log_level", "debug", "Logging Level")
+	driverName := flag.String("driver", vmdkDriver, "Volume driver")
```
pdhamdhere: This change expects user to either configure or pass command line argument --driver=vsphere.

Any reason why we can't register both "vmdk" and "vsphere" pointing to same handlers? Ideally, with NO changes in installation process, following commands should work;

`docker volume create -d=vmdk v1`
Not sure if we could postfix warning that vmdk is deprecated and use vsphere instead.
`docker volume create -d=vsphere v1`

This can potentially open volume naming issues which I believe Docker tracks per Volume Driver. However, we don't expect user to mix & match vmdk/vsphere but rather to stick to one.  So, we should be fine.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
msterin: I'd suggest 0.11 to catch up with the release #. But if you prefer 0.4 to stay sequential it's good too

vmdk_plugin/main.go
govint: True, we can run both instances in parallel, only that adds more code that will need to be removed later. Docker on startup associates a driver to each volume that each driver reports. A volume created by a vmdk driver can get reported by a vsphere driver instance later (after docker and plugin restart) and Docker would associate that volume as a vsphere volume.

Only issue is with scripts - test scripts and user scripts to create a docker app stack that specifies the volume driver to use.
I can make driver also a part of the configuration so user can change the driver name there vs. change the startup scripts.

msterin: I also do not think running 2 process is a good idea - but can we just register the driver with the same socket twice ? Or can we register it twice and listen on both sockets from one driver ? Restarting the driver service to switch between supporting `-d vmdk` and `-d vsphere` seems pointless, if that's the only choice then I think we (anbd customers) would be better off if we  make a breaking change and stick with `vsphere`

govint: Its possible to start two listeners on two separate sockets, the sockets are named after the driver - vmdk.sock, vsphere.sock. One instance of the plugin process runs with the two listeners in this case.

govint: I'll try out having one listener run as a goroutine and check that its working ok.

BUt do we need this capability at all?

govint:
OK, tried out having both vmdk and vsphere drivers registered (and listening) in a single instance of the plugin. Volume creation is no problem, volume reporting certainly is. The way volume reporting happens is Docker asks each registered plugin what its volumes are and in our case since both vsphere and vmdk instances are registered, both end up reporting the same volumes.

I'm removing the dual driver registration and keeping "vmdk" the default while vsphere can be used if the user wishes.

$ docker volume create --driver=vsphere --name=vsp1
vsp1

$ docker volume create --driver=vmdk --name=vsp2
vsp2

$ docker volume ls
vmdk                vsp1@bigone
vsphere             vsp1@bigone
vmdk                vsp2@bigone
vsphere             vsp2@bigone
vmdk                vsp@bigone
vsphere             vsp@bigone

msterin:
> I'm removing the dual driver registration and keeping "vmdk" the default while vsphere can be used if the user wishes.

So 'vmdk' will be by default, and for 'vsphere' admins would need to do something (what is this that they need to do, btw ? )  Is it what we wanted ? I think if we are chaning to vsphere, we should make it default. @pdhamdhere

pdhamdhere:
Installation time Config option to choose driver type (vmdk/vsphere) doesn't make sense. If there is no easy way to support both drivers simultaneously (without manual intervention), then let's just switch to "vsphere" and release note in BOLD, CAPS, RED that vmdk is deprecated. 

You will have to combine this with test changes.

govint:
Agreed, I'll post the updates.

govint:
In which case to reduce impact can we announce this in the 0.12 release as an "up coming change" and actually make it for 0.13.

@pdhamdhere

pdhamdhere:
>In which case to reduce impact can we announce this in the 0.12 release as an "up coming change" and actually make it for 0.13.

Not sure how useful would be 30-day advance notice. It's going to be breaking change whenever we do it. We better do it sooner than later.

govint:
@pdhamdhere the last updates before this push was to announce that the vmdk
driver deprecated. Are we changing that again

govint:
@pdhamdhere are these changes ok the vmdk driver option is deprecated from
earlier updates and driver version has been update to 0.4

pdhamdhere:
Sorry for confusion. How about following?

By default plugin registers & starts with "vsphere" driver type.

If user wants to use "VMDk" then he can configure driver name in config file or pass as startup CLI argument. We need to document this in release note so let's add "how to" in PR description.

govint:
That was the code a couple of iterations back. Now will add it back agan.

docs/user-guide/docker-plugin-drivers.md
```diff
@@ -43,3 +43,5 @@ Note:
 1. The current version of the Photon volume plugin doesn't support authentication as yet, this will be added up in a subsequent release.
 2. The "target port" is set to "9000" if authentication isn't enabled in photon controller, else port "443" is used.
 3. Both project and VM IDs mentioned above can be obtained via the "photon" CLI, see https://github.com/vmware/photon-controller/wiki/Compile-Photon-Controller-CLI.
+4. The plugin driver supports both "vsphere" and "vmdk" as driver names and both can be used interchangeably for now. The "vmdk" driver name will be deprecated in a later release, allowing users to use only "vsphere" as the driver name.
+5. The "vmdk" driver is however the default driver name that is used when the plugin is installed.
```
msterin: I did not find the explanation of how to switch between vmdk and vsphere (i.e `edit xxx file; restart the service` or something along these lines

govint:
Updated diff please see last diffs. Added Driver as an option for user to specify in the config file. The default in the code is "vsphere", which can be overridden with the "Driver" setting in the config file (if provided or present at default location) and the command line parameter overrides the conf. file setting. This helps us with the Photon platform as well where user can set the Driver option in the config file and the code will handle that. Service start up scripts should not be modified by the user and hence left those as is.

Updated documentation also.

vmdk_plugin/main.go
```diff
@@ -97,17 +99,31 @@ func main() {
...
 
+	// Load the configuration if one was provided.
+	c, err := config.Load(*configFile)
+	if err != nil {
+		log.Warning("Failed to load config file %s: %v", *configFile, err)
+	}
+
+	// Default driver, overridden by the config file
+	name := vsphereDriver
+	if c.Driver != "" {
```
pdhamdhere: I am not a go expert. Would dereferencing "Driver" (`c.Driver`) crash if "c" is NULL?

vmdk_plugin/main.go
```diff
@@ -97,17 +99,31 @@ func main() {
...
+		name = c.Driver
+	}
+
+	// Driver specified on the command line overrides option in the config file
+	driverName := flag.String("driver", name, "Volume driver")
+
 	// photon driver options
 	targetURL := flag.String("target", "", "Photon controller URL")
```
pdhamdhere: Not related to this change but if user starts Plugin with "photon" driver then we should fail to start if required parameters are missing in CLI or config file.

vmdk_plugin/main.go
```diff
@@ -120,13 +136,7 @@ func main() {
...
 
-	// Load the configuration if one was provided.
-	c, err := config.Load(*configFile)
-	if err != nil {
-		log.Warning("Failed to load config file %s: %v", *configFile, err)
-	}
-
-	// The vmdk driver doesn't depend on the config file for options.
+	// The vSphere driver doesn't depend on the config file for options
 	if *driverName == photonDriver && err == nil {
```
pdhamdhere: It's awkward to chck "err == nil" which is set earlier in the function. I think code can be restructured a bit to do validation in beginning.

vmdk_plugin/main.go
govint: Reordered the code to select the driver if no error in config file, logs below,

2017-02-27 20:54:44.414369896 -0800 PST [INFO] No config file found. Using defaults.
2017-02-27 20:54:44.415016096 -0800 PST [WARNING] Failed to load config file %s: %v/etc/docker-volume-vsphere.confopen /etc/docker-volume-vsphere.conf: no such file or directory
2017-02-27 20:54:44.415259471 -0800 PST [INFO] Starting plugin log_level=info config="/etc/docker-volume-vsphere.conf" driver=vsphere
2017-02-27 20:54:44.415492463 -0800 PST [INFO] Plugin options - port=1019
2017-02-27 20:54:44.415743608 -0800 PST [INFO] Getting volume data from unix:///var/run/docker.sock
2017-02-27 20:54:44.437056322 -0800 PST [INFO] Discovered 0 volumes in use.
2017-02-27 20:54:44.437362054 -0800 PST [INFO] Docker VMDK plugin started mock_esx=false version="vSphere Volume Driver v0.4" port=1019
2017-02-27 20:54:44.437833239 -0800 PST [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vsphere.sock"

vmdk_plugin/main.go
govint: Yes, there is a verification step for the photon driver to check the options given and exit.

vmdk_plugin/main.go
govint: Done.

govint:
Actually, looks much better now than earlier :-)

PR #920, opened 2017-02-16
Auto find ips

govint:
Modified photon driver to search for tenant/project/VM IDs from photon given the target URL (photon controller IP). The plugin doesn't need any more config except the target URL.

Testing is in progress but I'd like comments on the approach.


pdhamdhere:
@luomiao can you please review?

govint:
Ping for a review.

luomiao:
vmdk_plugin/drivers/photon/.photon_driver.go.swp

Remove the temporary file?

luomiao:
This seems adds a big loop in order to get the project and VM IDs. Is there a reason to do so?

govint:
Basically go thru all tenants and figure the tenant, project and VM IDs for
the VM. I'm not making the assumption that there is a only a single project
in the photon controller config. User only needs to give the target URL.

luomiao:
But it seems this is only for getting the project ID and VM ID for the local node?
Do you think it's possible that we can get this information from the node instead of checking with the Photon endpoint?
For example, if Photon created some file including the project/VM ID and stored it on this VM, then there is not need to talk to remote Photon endpoint and go through this expensive loop to get this local information. We can check with Photon team to verify if there is already some file on the VM. And if it's not there, maybe we can check if they can add this feature.

Using a multi-level loop in order to achieve some local information seems not very efficient.

govint:
Yes, in fact this PR changes the assumption that there is a file on the
node that had this information. And yes this code is definitely not optimal
but it's more or less temporary till the meta data service of photon is
available. I'd love to have a whoami() type API from photon. But after
discussion with the photon team this was the approach advised given the
existing APIs

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -57,50 +58,112 @@ const (
...
+		projects, err := d.client.Tenants.GetProjects(tenant.ID, nil)
+		if err != nil {
+			log.WithFields(log.Fields{"target": d.target}).Warning("Invalid target, unable to get list of projects, exiting.")
+			return
+		}
+		for _, project := range projects.Items {
+			vms, err := d.client.Projects.GetVMs(project.ID, nil)
+			if err != nil {
+				log.WithFields(log.Fields{"target": d.target}).Warning("Invalid target, unable to fetch list of VMs, exiting.")
+				return
```
luomiao: Even there is an error from GetVMs here, this project.ID may not be the one that the current local node is belong to. So maybe continue this loop instead of return here?

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -57,50 +58,112 @@ const (
...
+			for _, vm := range vms.Items {
+				var taskErr error
+				netTask, err := d.client.VMs.GetNetworks(vm.ID)
+				if err == nil {
+					taskErr = d.taskWait(netTask.ID)
+				}
+				if err != nil || taskErr != nil {
+					log.WithFields(log.Fields{"target": d.target, "error": err, "task-error": taskErr}).
+						Warning("Invalid target, unable to get VM properties, exiting.")
+					return
```
luomiao: Similar to the comment above. Choose to continue instead of return?

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -57,50 +58,112 @@ const (
...
 	}
+
 	// Use default timeout of thirty seconds and retry of three
 	d.client = photon.NewClient(targetURL, nil, nil)
 
-	err := d.verifyTarget()
-	if err != nil {
-		log.WithFields(log.Fields{"target": targetURL, "project-id": projectID}).Warning("Invalid target and or project ID, exiting.")
+	// Get host IP address
+	hostIP := d.getHostIP()
```
luomiao: It seems this hostIP can be an empty string if there is no valid IPv4 address on this node. Should we add a check and stop here if it's empty? because it won't match in identifyHost function. And even it matches, it could be the wrong one.

pdhamdhere:
@govint Are you still working on this PR? Can we close and post when it is ready?

govint:
Currently I'm prioritizing the docker issue (mem. corruption and plugin crash). I need to address @luomiao comments for this PR.

msterin:
@govint Can you close the PRs which you do not want to be reviewed/merged yet ? It's confusing to have them , I keep wasting time on checking if they need review. 

You can always reopen when you are ready.

govint:
Closing for now, will reopen once Photon supports the mgd. plugin.

PR #926, opened 2017-02-17
Fix for issue 193.

lipingxue:
This change is branched from 0.11 release, and no need to merge back to master.


lipingxue:
@msterin @shaominchen Please review this PR.

msterin:
On the first glance looks good; need to review further. 
A few words about testing and new behavior would be good.

esx_service/utils/auth.py
```diff
@@ -34,6 +34,7 @@
 
 SIZE = 'size'
 DEFAULT_TENANT = '_DEFAULT'
+DEFAULT_TENANT_UUID = '_DEFAULT_TENANT_UUID'
```
pdhamdhere: So, this would create a folder "_DEFAULT_TENANT_UUID" and soft link _DEFAULT? Should we be using some constant magic number?

esx_service/utils/auth_data.py
```diff
@@ -514,15 +531,30 @@ def connect(self):
...
                 # when schema changes
                 raise DbConnectionError(self.db_path)        
+            
+            error_msg, tenant = self.get_tenant(auth.DEFAULT_TENANT)
+            if error_msg:
+                raise DbConnectionError(self.db_path)
+            if not self.is_valid_default_tenant_uuid(tenant.id):
+                error_msg = """
+                            Your ESX installation seems to be using config db created by previous version of
+                            Vsphere Docker Volume Service, and requires upgrade.
```
pdhamdhere: Vsphere => vSphere

esx_service/utils/auth_data.py
```diff
@@ -514,15 +531,30 @@ def connect(self):
...
+                            Please download script from {0} and run the script on each ESX where you are using
```
pdhamdhere: "Please download script from {0} and run the script on each ESX where you are using the product." =>
"Please download script from {0} and run the script on each ESX where vSphere Docker Volume Service is installed"

esx_service/utils/auth.py
lipingxue: Yes, this would create a folder "_DEFAULT_TENANT_UUID" and soft link _DEFAULT.
For the second question, I have no preference, either way looks good to me.

esx_service/utils/auth_data.py
lipingxue: Done.

esx_service/utils/auth.py
pdhamdhere: If team accepts a proposal to rename TENANT => Namespace then folder _DEFAULT_TENANT_UUID will out of place. I propose we pick a magic string. Also test if randomly generated UUID happen to match with magic string then collisions are handled correctly.

esx_service/utils/auth_data.py
```diff
@@ -514,15 +531,30 @@ def connect(self):
...
+        self.conn.row_factory = sqlite3.Row
+
         if not need_create_table:
             if os.path.isfile(self.db_path) and self.need_upgrade_db():
                 # Currently, when need upgrade, raise exception
                 # TODO: add code to handle DB schema upgrade 
                 # when schema changes
                 raise DbConnectionError(self.db_path)        
+            
+            error_msg, tenant = self.get_tenant(auth.DEFAULT_TENANT)
```
shaominchen: Can we extract line 544-555 to a separate method? This piece of logic is trying to verify whether the default tenant already exists or not. It's better to keep it modulized.

lipingxue:
Test is done.
The following are the testing I have done.

Test1:
Two ESX (ESX1 and ESX2) which share a common datastore (vSanDatastore).
create VM1 on ESX1 and VM2 on ESX2, both VMs are created in the shared datastore.
all ESX and VM are fresh install
1. From VM1, run  "docker volume create" to create two volumes named "vol1" and "vol2", those volumes will be created at shared datastore (vSanDatastore).
2. From VM2,  run "docker volume ls", those two volumes which were created from VM1 are displayed correctly.

Test2:
1.  Create a new named tenant "tenant1", and add VM1 to this tenant
2. From VM1, run "docker volume create" to create two volumes named "tenant1-vol1" and "tenant1-vol2"
3. From VM1, run "docker volume ls", it shows two volumes "tenant1-vol1" and "tenant1-vol2"
4. From VM2, run "docker volume ls", it shows two volumes "vol1" and "vol2"
5. Remove VM1 from tenant1, and from VM1, run "docker volume ls", it shows two volumes "vol1" and "vol2"

esx_service/utils/auth_data.py
```diff
@@ -410,7 +422,8 @@ def create_default_tenant(self):
                                            name=auth.DEFAULT_TENANT, 
                                            description="This is a default tenant", 
                                            vms=[], 
-                                           privileges=[])
+                                           privileges=[],
+                                           tenant_uuid=auth.DEFAULT_TENANT_UUID)
```
shaominchen: We can deduce the default tenant uuid based on its name. So I'd suggest not to change the logic here. Instead, we can handle this inside create_tenant() method.

esx_service/utils/auth.py
shaominchen: Personally I feel using some special string that can be easily identified (so as to differ from the auto-generated uuid string) will be helpful for debugging.

esx_service/utils/auth_data.py
```diff
@@ -66,6 +68,16 @@ def __init__(self, db_path):
     def __str__(self):
         return "DB connection error %s" % self.db_path
 
+class DbAccessError(Exception):
+   """ An exception thrown when access to a sqlite database fails. """
```
lipingxue: Yes.

esx_service/utils/auth_data.py
lipingxue: Currently, this change is just for DEFAULT_TENANT. But my change can be easily extended to named tenant if later we plan to do similar change.

shaominchen: Even if we may (I guess not) introduce some other special tenants in the future, we can still follow the same approach, i.e. handle the special tenant_name to tenant_uuid mapping transparently inside the create_tenant() method, hiding this detail from the client code.

lipingxue: I don't think it is a good approach. Basically, what I am doing is create_tenant has a new parameter, "tenant_uuid", if caller choose to parse this param, we will create a tenant using the passed "tenant_uuid", otherwise, we will auto-generate a uuid for this tenant. Caller of "create_tenant" should has more knowledge and more control for this.

esx_service/utils/auth_data.py
lipingxue: Done.

esx_service/utils/auth_data.py
```diff
@@ -514,15 +531,30 @@ def connect(self):
...
         if not need_create_table:
             if os.path.isfile(self.db_path) and self.need_upgrade_db():
                 # Currently, when need upgrade, raise exception
                 # TODO: add code to handle DB schema upgrade 
                 # when schema changes
                 raise DbConnectionError(self.db_path)        
+            
+            error_msg, tenant = self.get_tenant(auth.DEFAULT_TENANT)
+            if error_msg:
+                raise DbConnectionError(self.db_path)
```
lipingxue: Done.

lipingxue:
Test for error case.
1. testbed with 0.11 release builds, in my setup, I have one ESX named "ESX1", and create a VM named "vm1"
2. install the build with this fix
3.  run "docker volume ls" from "vm1", get the following error
```
root@photon-E5AoXvvmq [ ~ ]# docker volume ls
list vmdk: VolumeDriver.List: DB access error at /etc/vmware/vmdkops/auth-db (
                        Your ESX installation seems to be using config db created by previous version of
                        vSphere Docker Volume Service, and requires upgrade.
                        Please download script from TBD and run the script on each ESX where vSphere Docker
                        Volume Service is installed.  (_DEFAULT_UUID = ec123f5c-b94e-424f-931b-29d9c05ba433, expected = 11111111-1111-1111-1111-111111111111)
                     )

```
4. User need to following the instruction in the error message to apply upgrade using the script. Right now, the place of script is "TBD", and it will be fixed when the script is ready.

esx_service/utils/auth_data.py
```diff
@@ -58,14 +63,26 @@ def get_version_str(major_ver, minor_ver):
...
+    """
 
     def __init__(self, db_path):
         self.db_path = db_path
 
     def __str__(self):
         return "DB connection error %s" % self.db_path
 
+class DbAccessError(Exception):
+   """ An exception thrown when when a client tries to run a SQL using an established connection.
```
shaominchen: Nit: remove a duplicate "when"

esx_service/utils/auth_data.py
shaominchen: I don't agree, because if we ever need to use any special tenant uuid, we will always introduce corresponding special tenant name. So the uuid handling can always be encapsulated in the method itself.

Anyway, I will let you make the decision since this is not really a big concern.

msterin:
Al the changes from this PR have already made it to master via  #944 (cherry-picked)

Issue #898, opened 2017-02-08
[test automation] Automate negative test scenarios related to tenancy


shuklanirdesh82:
Negative test: Tenant creation with Invalid VM name

Objective: Test to verify tenant creation by passing non-existing vm

1. create tenant "t1" by passing nonExisting VM
2. expects an error/exception

-----------------------------
Negative test: Tenant creation with the same name multiple times

Objective: Test to verify creation of two tenants with same name

1. Create a tenant with name 'tenant_abc'
2. Create another tenant with the same name
3. Expects an error message for the step#2 as the same named tenant is present.
4. Then list all the tenant created
5. Test should only see one tenant with

tusharnt:
This can be pushed out of GA if we are short on time.

ashahi1:
- Negative test: Tenant creation with Invalid VM name.

Operation to add a non-existent vm to a tenant failed - which is expected as vm does not exists. 

```
[root@sc-rdops-vm01-dhcp-0-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=SampleTenant --vm-list=photon4
Cannot find vm photon4
[root@sc-rdops-vm01-dhcp-0-159:~]

```

- Negative test: Tenant creation with the same name multiple times.

Was not able to create two tenants with the same name - as expected.
```
[root@sc-rdops-vm01-dhcp-0-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=SampleTenant --vm-list=photon4
Cannot find vm photon4
[root@sc-rdops-vm01-dhcp-0-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=SampleTenant
tenant create succeeded
[root@sc-rdops-vm01-dhcp-0-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=SampleTenant
Tenant SampleTenant already exists
[root@sc-rdops-vm01-dhcp-0-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                  Name          Description               Default_datastore  VM_list
------------------------------------  ------------  ------------------------  -----------------  -------
11111111-1111-1111-1111-111111111111  _DEFAULT      This is a default tenant
b797ca92-6cef-4104-bbb3-fdc02ff8af48  SampleTenant

[...]

PR #914, opened 2017-02-16
Negative test to add non-existent vm to a tenant and creating two tenants with same name.

ashahi1:
This PR covers following two negative scenarios specified in Issue #898 :
1. To verify that we are not able to add a non-existent vm to a tenant.
2. To verify that we are not able to create a new tenant with the same name as an existing tenant.

Performed testing locally with one ESX and two vms setup.

Test Output: 
```
=== RUN   TestAddNonEsxtntVmToTenant
2017/02/25 00:03:43 Creating a tenant named - tenant1_134jhcb3 with a non-existent vm - vm1_134jhcb3
2017/02/25 00:03:44
 As expected, error while creating a tenant named - tenant1_134jhcb3 by passing a non-existent vm - vm1_134jhcb3
--- PASS: TestAddNonEsxtntVmToTenant (2.67s)
=== RUN   TestCreatingSameNameTenant
2017/02/25 00:03:46
 Successfully created a  tenant named - tenant1_134jhcb5
2017/02/25 00:03:46 Creating another tenant with the same name -  tenant1_134jhcb5
--- PASS: TestCreatingSameNameTenant (3.88s)
PASS
```

PR #928, opened 2017-02-17
Issue create tenants with same name.ashahi1

ashahi1:
This PR addresses the negative test scenario mentioned in Issue #898 - Create two tenants with the same name.

Objective: Test to verify tenant creation by trying to create two tenants with same name
Steps:

1. create tenant.
2. Try creating another tenant with same name.
3. Use tenant ls command to verify we have only one tenant with same name.

Performed testing locally with one ESX and two vms setup.


vmdk_plugin/E2E_Tests/neg_CreateTenantsWithSameName_test.go
```diff
@@ -0,0 +1,84 @@
...
+/*
+Test steps:
+1. Create a new tenant.
+   Command: /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=<tenantName>
+2. Create another tenant with same name.
+   Tenant creation will fail as we had already created tenant with same name
+3. Do a list of tenants and check if we have only one with same name  present.
+   Command: /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
+*/
+func TestCreateTwoTenantsWithSameName(t *testing.T) {
```
shuklanirdesh82: It should be clubbed into the existing negative testcase file rather than keep creating separate files. It is better to have all trivial negative tests into the same file.

ashahi1:
This is being addressed in a separate PR: https://github.com/vmware/docker-volume-vsphere/pull/914

Hence closing this PR

PR #932, opened 2017-02-18
Changes to err message 

msterin:


lipingxue:
LGTM

msterin:
the fix is already in all correct branches, Closing the PR

Issue #929, opened 2017-02-17
Tenant ls command does not works if vm that was added to tenant is deleted.


ashahi1:
Steps to reproduce:

1) Create a tenant and associate a vm to it:
  /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=tenant1 --vm-list=photon4
2) Tenant ls command gives following output:
```
[root@sc-rdops-vm06-dhcp-186-90:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                            Name                   Description               Default_datastore  VM_list
------------------------------------  --------  ------------------------  -----------------  -------
41242167-6019-4f3f-a384-66a82cf9997d   _DEFAULT   This is a default tenant
5aa44411-0eba-4591-bacf-c4cfc27beb85    tenant1                                                      Photon4
```
3) Delete the vm that was added to the tenant.
4) Tenant ls command gives following output - 

[root@sc-rdops-vm06-dhcp-186-90:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Traceback (most recent call last):
  File "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py", line 1004, in <module>
    main()
  File "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py", line 51, in main
    args.func(args)
  File "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py", line 862, in tenant_ls
    rows = generate_tenant_ls_rows(tenant_list)
  File "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py", line 810, in generate_tenant_ls_rows
    vm_list = generate_vm_list(tenant.vms)
  File "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py", line 790, in generate_vm_list
    res = res + vm_name
TypeError: Can't convert 'NoneType' object to str implicitly

============
Ideally the tenant ls command should give the list of all the tenants with deleted vm being no longer part of the VM_list for that particular tenant.

govint:
PR #937 addresses a small fix to prevent this issue.

The larger issue is the tenant DB needs to be updated to remove VMs (uuids) that have since been deleted. Possibly separate issue for that.

govint:
[...]

PR #937, opened 2017-02-20
Check vm

govint:
Check for null VM name string before adding to list of vm names in generate_vm_list().

### DIFF
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -790,8 +790,12 @@ def generate_vm_list(vms_uuid):
     res = ""
     for vm_uuid in vms_uuid:
         vm_name = vmdk_utils.get_vm_name_by_uuid(vm_uuid)
-        res = res + vm_name
-        res = res + ","
+        # If the VM name cannot be resolved then its possible
+        # the VM has been deleted or migrated off the host,
+        # skip the VM in that case.
+        if vm_name:
+            res = res + vm_name
+            res = res + ","
 
     if res:
         res = res[:-1]


govint:
Fixes #929

esx_service/cli/vmdkops_admin.py
```diff
@@ -788,13 +788,15 @@ def generate_vm_list(vms_uuid):
...
     # vms_uuid is a list of vm_uuid
     # example: vms_uuid=["vm1_uuid", "vm2_uuid"]
-    # the return value is a string like this vm1,vm2
+    # the return value is a string like this vm1,vm2.
     res = ""
     for vm_uuid in vms_uuid:
         vm_name = vmdk_utils.get_vm_name_by_uuid(vm_uuid)
-        res = res + vm_name
+        if vm_name:
+            res = res + vm_name
```
pdhamdhere: Add comment that VM may be deleted/not found.

esx_service/cli/vmdkops_admin.py
```diff
@@ -788,13 +788,15 @@ def generate_vm_list(vms_uuid):
...
+        else:
+            res = res + vm_uuid
```
pdhamdhere: This should never happen since tenant are ESX-local and only VMs on same ESX can be added to Tenant.

govint: But the VM can migrate out of the host or be deleted. In the former case we still have the UUID from the KV but the VM name can't be resolved. Is there any way to keep the VMs of a tenant on the same host?

govint:
Updated PR to ignore VMs in a tenant whose name can't be resolved.

esx_service/cli/vmdkops_admin.py
```diff
@@ -788,13 +788,13 @@ def generate_vm_list(vms_uuid):
...
     # example: vms_uuid=["vm1_uuid", "vm2_uuid"]
-    # the return value is a string like this vm1,vm2
+    # the return value is a string like this vm1,vm2.
     res = ""
     for vm_uuid in vms_uuid:
         vm_name = vmdk_utils.get_vm_name_by_uuid(vm_uuid)
-        res = res + vm_name
-        res = res + ","
-
+        if vm_name:
```
pdhamdhere: I would add a comment- "If VM is already deleted or migrated, vm_name could be NULL and will be skipped".

lipingxue: I agree with Prashant.  Please add a comment here. Other LGTM.

govint:
Added comment and below is the behavior with the fix. VM name is "1" and its ID is "2" below. First its added to the default tenant and shows up in "ls" and later unregistered and the command skips the VM 
when retried.

vmdkops_admin.py tenant ls
------------------------------------  --------  ------------------------  -----------------  -------
cb15582e-7963-4d67-95ed-d1f5de21bb25  _DEFAULT  This is a default tenant                     1

vim-cmd vmsvc/unregister 2
(vim.fault.NotFound) {
   faultCause = (vmodl.MethodFault) null,
   faultMessage = <unset>
   msg = "Unable to find a VM corresponding to "2""
}
vmdkops_admin.py tenant ls
------------------------------------  --------  ------------------------  -----------------  -------
cb15582e-7963-4d67-95ed-d1f5de21bb25  _DEFAULT  This is a default tenant

Issue #934, opened 2017-02-18
vmdkops_admin output is not consistent when more than 2 ESX are present and accessing shared datastore (attached-to-vm field is inconsistent)


shuklanirdesh82:
vmdkops_admin output is not consistent when more than 2 ESX are present and accessing shared datastore e.g. vsan, shared vmfs etc (attached-to-vm field is inconsistent)

Test bed scenario: 2 ESX having shared datastore (vsan/shared vmfs). One docker host per ESX created on the the shared datastore (let's say vm1 & vm2)

Note: both VMs are part of _DEFAULT tenant

Steps to reproduce
1. create volume from vm1 & vm2 on shared datastore
```
VM1: docker volume create --driver=vmdk --name=vol_231_vsan
VM2: docker volume create --driver=vmdk --name=vol_99_vsan
```
2. make sure both volumes are accessible from either VMs (docker volume ls from both VMs)
3. attach container from one of the VM 
```
VM1: docker run --rm -it -v vol_231_vsan@vsanDatastore:/mnt/myvol2 --name cnt_2311 busybox
```
4. invoke docker inspect from another VM and make sure it says volume is ```attached```
```
root@photon-TX4FfW2ET [ ~ ]# docker volume inspect vol_231_vsan@vsanDatastore
[
    {
        "Name": "vol_231_vsan@vsanDatastore",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/vol_231_vsan@vsanDatastore",
        "Status": {
            "access": "read-write",
            "attach-as": "independent_persistent",
            "capacity": {
                "allocated": "80MB",
                "size": "100MB"
            },
            "clone-from": "None",
            "created": "Sat Feb 18 06:16:43 2017",
            "created by VM": "VM84_vsan",
            "datastore": "vsanDatastore",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "attached"
        },
        "Labels": {},
        "Scope": "global"
    }
]
```

5. On ESX side invoke vmdkops_admin

```
ESX1:

[root@sc-rdops-vm17-dhcp-0-84:/vmfs/volumes] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Volume          Datastore      Created By VM  Created                   Attached To VM  Policy          Capacity  Used     Disk Format  Filesystem Type  Access      Attach As               
[...]

PR #939, opened 2017-02-20
Use VM uuid if the VM name can't be got locally.

govint:
Use the VM uuid if the VM name can't be got on the local ESX host. The VM could be on a different ESX host in that case. Its possible that the VM doesn't exist even, but the volume KV may still record the fact. In which case a subsequent attach of the volume on the other ESX host will clear the stale attached-to entry in the volume KV. From a remote node there's nothing much that can be done in this case except use the data thats in the KV.

### DIFF
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -563,7 +563,7 @@ def get_creation_info(metadata):
 
 
 def get_attached_to(metadata):
-    """ Return which VM a volume is attached to based on its metadata """
+    """ Return which VM a volume is attached to based on its metadata. """
     try:
         vm_name = vmdk_ops.vm_uuid2name(metadata[kv.ATTACHED_VM_UUID])
         if not vm_name:
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -528,7 +528,7 @@ def ls_dash_c(columns, tenant_reg):
 def all_ls_headers():
     """ Return a list of all header for ls -l """
     return ['Volume', 'Datastore', 'Created By VM', 'Created',
-            'Attached To VM', 'Policy', 'Capacity', 'Used',
+            'Attached To VM (name/uuid)', 'Policy', 'Capacity', 'Used',
             'Disk Format', 'Filesystem Type', 'Access', 'Attach As']
 
 def generate_ls_rows(tenant_reg):
--- a/esx_service/cli/vmdkops_admin.py
+++ b/esx_service/cli/vmdkops_admin.py
@@ -567,7 +567,10 @@ def get_attached_to(metadata):
     try:
         vm_name = vmdk_ops.vm_uuid2name(metadata[kv.ATTACHED_VM_UUID])
         if not vm_name:
+            if metadata[kv.ATTACHED_VM_UUID]:
+                return metadata[kv.ATTACHED_VM_UUID]
             return kv.DETACHED
+
         return vm_name
     except:
         return kv.DETACHED


govint:
Fixes #934.

esx_service/cli/vmdkops_admin.py
```diff
@@ -563,11 +563,14 @@ def get_creation_info(metadata):
 
 
 def get_attached_to(metadata):
-    """ Return which VM a volume is attached to based on its metadata """
+    """ Return which VM a volume is attached to based on its metadata. """
     try:
         vm_name = vmdk_ops.vm_uuid2name(metadata[kv.ATTACHED_VM_UUID])
         if not vm_name:
+            if metadata[kv.ATTACHED_VM_UUID]:
+                return metadata[kv.ATTACHED_VM_UUID]
```
pdhamdhere: Should we prefix with "VM UUID:"? Or change admin command header to "VM Name/ UUID"?

pdhamdhere: On second thought, I propose we bring vm_name back in KV. It may be stale in rare cases when vm_name is changed.

if (vm_name = vm_uuid2name(...))
    return vm_name
if (metadata[kv.ATTACHED_VM_NAME])
   vm_name = vm_name + "*"
   return vm_name
if (metadata[kv.ATTACHED_VM_UUID])
   return metadata[kv.ATTACHED_VM_UUID]

Log message in vmdkops_admin that 
`* Volume attached to VM running on another ESX Host.`

Other things we could do;
- Include ESX Host Name in KV.
- Always log both VM Name and UUID.

CC /@msterin

msterin: i think logging bitgh vmname and uuid and resetting vmname to "" if it's not avail is a way to go since it adds predictibility.

On ESX name - I am not sure which ESX ? where it's attached ? Where it was created ?

govint:
@msterin the vm name isn't known because the vm isn't on the host. we can
invoke handle_stale_attach() from vmdk_ops.py to check and reset the
attached vm state. But it's not correct if the VM had migrated out of the
host and still using the volume on a shared data store.

esx_service/cli/vmdkops_admin.py
govint: VM name was remvoed earlier to avoid the issue with a VM's name changing, which is why the name is retrieved each time. I'd vote to change the header to VM (name/uuid) for this issue.

govint:
This is what the new header looks like,

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls
Volume                       Datastore  Created By VM  Created                   Attached To VM (name/uuid)  Policy  Capacity  Used  Disk Format  Filesystem Type  Access      Attach As             
---------------------------  ---------  -------------  ------------------------  --------------------------  ------  --------  ----  -----------  ---------------  ----------  ----------------------
60635dd205d5067be93aaeade..  bigone     2              Wed Feb 15 10:36:11 2017  detached                    N/A     100MB     13MB  thin         ext4             read-write  independent_persistent
newvol                       bigone     2              Wed Feb 15 10:36:43 2017  detached                    N/A     100MB     13MB  thin         ext4             read-write  independent_persistent
newvol1                      bigone     1              Wed Feb 15 11:40:52 2017  detached                    N/A     100MB     13MB  thin         ext4             read-write  independent_persistent
newvol2                      bigone     1              Wed Feb 15 11:55:02 2017  detached                    N/A     100MB     13MB  thin         ext4             read-write  independent_persistent
newvol3                      bigone     2              Wed Feb 15 11:55:18 2017  detached                    N/A     100MB     13MB  thin         ext4             read-write  independent_persistent


I'm not sure we want to have separate vm-name and vm-uuid columns as mostly the vm-uuid column would be empty save the occasional case where the VM is missing on the local host

Issue #942, opened 2017-02-21
VMDK Driver - VolumeDriver.Mount: Server rejected


brockz:
Hello 

i have 2 Worker hosts (photon os based) running in swarm mode.
When i create a Service with an mounted Volume via vmdk Driver and the Docker Host fails . The vmdk volume doesnt get unmounted. The Instance is getting restarted on the other Worker but  the volume is not mounted:
cswy7i9te7688920nh6wvjbu3   \_ nginx.1  nginx  id-nb-pht-wrk01  Shutdown       Rejected 24 minutes ago  "VolumeDriver.Mount: Server re…"

pdhamdhere:
@brockz We will need some more details about error and your setup. Can you please share some more details;

- Which release you are using?
- Setup: Can you please share more details on Swarm setup, VMs & ESX Hosts.
- Command that you used to create volume and/or start service
- Logs from Docker Hosts (/var/log/docker-volume-vsphere.log)
- Logs from ESX Host (/var/log/vmware/vmdk_ops.log)

brockz:
Hello pdhamdhere.
Here are the Infos.

vSphere: 
4 Node vSphere Cluster running ESX 6.0.0 Build 4600944
vCenter 6.0.0 4541948
vSan Datastore

Docker/Swarm:
4 VMS running Photon OS 1.0, Revision 2  (2 Swarm Manager and 2 Swarm Worker=
docker -v
Docker version 1.12.1, build 23cf638

Building Swarm:
Manager Node1: docker swarm init --advertise-addr 10.10.105.221
Manager Node2: docker swarm join     --token xxxxxxxxxxxxxxxxxxx    10.10.105.221:2377
Worker Node1: docker swarm join     --token xxxxxxxxx   10.10.105.221:2377
Worker Node2: docker swarm join     --token xxxxxxxxx   10.10.105.221:2377

 Create Volume:
 docker volume create --driver=vmdk --name=MyVolume -o size=10gb

Created Service:
 docker service create --name nginx  --mount type=volume,source=MyVolume@shrk-vSAN01,target=/usr/share/nginx/html -p 80:80 nginx

Service started on Worker 2 succesfully:
root@id-nb-pht-ma01 [ ~ ]# docker service ps nginx
ID                         NAME     IMAGE  NODE             DESIRED STATE  CURRENT STATE          ERROR
[...]

PR #946, opened 2017-02-21
Chery-pick to 0.11 servicing - "Fix for #942 - set uuid correctly" 

msterin:
Chery-pick of commit 5903d71 "Fix for #942 - set uuid correctly"  from master to 0.11 service branch


msterin:
merging without CI. I expect @shuklanirdesh82 to cherry pick a fix to CI next, so we will save a CI pass.

PR #950, opened 2017-02-22
Initial change to deprecate "vmdk" driver name in favor or "vsphere" driver.

govint:
This is the same as #919  but after sync'ing with the current master. #919 is reporting 20 files as changed and seems to have taken files from other commits. I'll be merging this one. Pls. see #919 for review and test results.


govint:
Closing. Have corrected issues with commits in #919

PR #959, opened 2017-02-23
Fixes issue #818 by forcing to refresh datastore cache 

### LINKED ISSUE #818: Not able to create vsphere docker volume (using short volume names) after renaming datastore; running into "IndexError('list index out of range',)"

lipingxue:
Fixes #818

When we are trying to get datastore name from VM configure path , it may return wrong datastore name if the datastore was renamed after vmdk_opsd service is started. Specifically, it returns the old name of the datastore - and this name  may no longer exist.

The issue is that the list of datastores is caches on start, and never refreshed. 

For example:
Configuration: only _DEFAULT tenant is configured, VM "photon5" is created on "datastore1"

step 1: create volume "vol1"  - `docker volume create --driver=vmdk --name=vol1`
A volume "vol1" will be created on  "datastore1"

step2: rename  "datastore1" to "datastore   1" (or to any other name)
step3: create another volume "vol2" by running `docker volume create --driver=vmdk --name=vol1`
The  command fails with  "IndexError('list index out of range',)".

Function  `get_datastore_name_from_config_path()` is called to get the datastore where the VM was created. In our case, it should return the new datastore name (i.e. the one after rename) .
However `get_datastore_name_from_config_path()` returns incorrect (old, cached)  `datastore1`.

**The fix** is to force refresh datastore cache in function  `get_datastore_name_from_config_path()`.

In this PR we add a flag to to `get_datastores()` telling it to refresh the DS list, and we force the refresh when calling from `get_datastore_name_from_config_path()`.


lipingxue:
Testing is done to verify this fix.
Configuration: only _DEFAULT tenant is configured. VM "photon5" is created on datastore "datastore1"
Step1:  run "docker volume ls", which shows 4 volumes 
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore1
vmdk                vol2@datastore1
vmdk                vol3@datastore1
vmdk                vol4@datastore1
```

Step2:
rename datastore "datastore1"  to "datastore   1"

Step3:
create another volume "vol5", it succeeded.
```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name=vol5              
vol5
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore   1
vmdk                vol2@datastore   1
vmdk                vol3@datastore   1
vmdk                vol4@datastore   1
vmdk                vol5@datastore   1

```

esx_service/utils/vmdk_utils.py
```diff
@@ -331,9 +336,27 @@ def get_datastore_name(datastore_url):
...
     # Query datastore name from VIM API
+    logging.debug("get_datastore_name: ds_objects=%s", get_datastore_objects())
     res = [d.info.name for d in get_datastore_objects() if d.info.url == datastore_url]
     return res[0]
 
+def get_datastore_name_from_config_path(config_path):
+    """Returns datastore NAME in config_path (not url-name which may be used in path)
+    """
+    # path is always /vmfs/volumes/<datastore>/... , so extract datastore:
+    config_ds_name = config_path.split("/")[3]
```
shuklanirdesh82: minor: better to check config_path is not null/empty

esx_service/utils/vmdk_utils.py
```diff
@@ -331,9 +336,27 @@ def get_datastore_name(datastore_url):
...
+    # get_datastore() return the cached value
+    # if datastore is changed outside
+    # ds_name may have a stale value
+    # therefore, force to refresh datastore cache by calling get_datastores()
+    # with need_refresh=True
+    ds_name = [x[0] for x in get_datastores(need_refresh=True) \
+                    if x[0] == config_ds_name or x[1] == config_ds_name ]
+    if len(ds_name) != 1:
+        logging.error("get_datastore_name_from_config_path: found more than one match: %s" % ds_name)
```
shuklanirdesh82: I see the code block is just a move from vmdk_ops .. Is there any plan to get rid off duplicates from the cache value?

pdhamdhere:
Does following work;
`docker volume create -d vmdk --name='vol@datastore 1'`

esx_service/utils/vmdk_utils.py
```diff
@@ -331,9 +336,27 @@ def get_datastore_name(datastore_url):
...
+    """Returns datastore NAME in config_path (not url-name which may be used in path)
+    """
+    # path is always /vmfs/volumes/<datastore>/... , so extract datastore:
+    config_ds_name = config_path.split("/")[3]
+    # get_datastore() return the cached value
+    # if datastore is changed outside
+    # ds_name may have a stale value
+    # therefore, force to refresh datastore cache by calling get_datastores()
+    # with need_refresh=True
+    ds_name = [x[0] for x in get_datastores(need_refresh=True) \
```
pdhamdhere: This will force refresh on every access. We should first check with get_datastores() and if not found then call again with get_datastores(need_refresh=True)

esx_service/vmdk_ops.py
```diff
@@ -708,7 +695,10 @@ def executeRequest(vm_uuid, vm_name, config_path, cmd, full_vol_name, opts):
     Returns None (if all OK) or error string
     """
     logging.debug("config_path=%s", config_path)
-    vm_datastore = get_datastore_name(config_path)
+    error_info, vm_datastore = vmdk_utils.get_datastore_name_from_config_path(config_path)
```
pdhamdhere: config_path received here is with new datastore name or old name?

lipingxue:
@pdhamdhere 
Please see my answer to your question

Does following work;
docker volume create -d vmdk --name='vol@datastore 1'

[Liping] Yes, it works

```
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore1
vmdk                vol2@datastore1
vmdk                vol3@datastore1
vmdk                vol4@datastore1
vmdk                vol5@datastore1
```
rename datastore "datastore1" to "datastore   1"

```
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore   1
vmdk                vol2@datastore   1
vmdk                vol3@datastore   1
vmdk                vol4@datastore   1
vmdk                vol5@datastore   1
```
create a new volume "vol6@datastore   1"

```
root@photon-eZ5ILvREQ [ ~ ]# docker volume create --driver=vmdk --name="vol6@datastore   1"
vol6@datastore   1
root@photon-eZ5ILvREQ [ ~ ]# docker volume ls
vmdk                vol1@datastore   1
vmdk                vol2@datastore   1
vmdk                vol3@datastore   1
vmdk                vol4@datastore   1
vmdk                vol5@datastore   1
vmdk                vol6@datastore   1
```

esx_service/utils/vmdk_utils.py
lipingxue: This will not work. Since get_datastore() use cache value, it will return the old datastore name here instead of "not found". Unless we do refresh, we don't know whether the old datastore name is still valid.

esx_service/vmdk_ops.py
lipingxue: config_path received here has the format like this "/vmfs/volumes/datastore_url/..."

esx_service/utils/vmdk_utils.py
```diff
@@ -331,9 +336,27 @@ def get_datastore_name(datastore_url):
...
+    logging.debug("get_datastore_name: ds_objects=%s", get_datastore_objects())
     res = [d.info.name for d in get_datastore_objects() if d.info.url == datastore_url]
     return res[0]
 
+def get_datastore_name_from_config_path(config_path):
+    """Returns datastore NAME in config_path (not url-name which may be used in path)
+    """
+    # path is always /vmfs/volumes/<datastore>/... , so extract datastore:
+    config_ds_name = config_path.split("/")[3]
+    # get_datastore() return the cached value
```
shaominchen: Nit: get_datastore() => get_datastores()

esx_service/utils/vmdk_utils.py
shaominchen: I have the same concern as Prashant's - this will force refresh on every access, which actually invalidate the whole purpose of introducing datastore cache. Please see my comments below.

esx_service/vmdk_ops.py
shaominchen: I think Prashant is asking what's the value (not the format) of the config_path here - does it contain the old datastore name or the new one? I think it should contain the new (i.e. renamed) datastore name. If that's the case, our datastore cache should not contain this name. We should NOT force to refresh the cache unless the cache is missed.

lipingxue: There is no "datastore" name in the "config_path". Only "datastore_url" in the config_path. We tried to figure out the datastore name from datastore_url. If we don't refresh the cache, for the datastore_url, it still map to old datastore name.

shaominchen: I see. If that's the case, I agree with you we don't seem to have better choice for this.

lipingxue:
I discussed with Mark offline. The fix I did in this PR is not the best one since it refresh datastore cache in each operation.
Based on our discussion, we decide there are two steps to fix #818 
step1:  force datastore cache to refresh for listVMDK. This fix is simple and also provide a workaround for the error hit in #818. When user see "IndexError('list index out of range',)"  he/she need to run "docker volume ls" which will force datastore cache to refresh. After that, "docker volume create" will work.

step2:  We checked the trace of this exception, and find that it caused by function "get_datastore_url" during the operation authorize call. After analyze the code, we found that function "get_datastore_url" is not needed, and we can remove it by refactoring operation authorize related code. 

I will close this PR for now.

PR #968, opened 2017-02-24
Handling case insensitivity for tenant access allow-create option

### LINKED ISSUE #918: Lower case input "true" passed to "vmdkops_admin.py tenant access set --allow-create ALLOW_CREATE" is misinterpreted as false.
```vmdkops_admin.py tenant access set --allow-create ALLOW_CREATE``` fails to validate input value for ```ALLOW_VALUE```; while setting a datastore-tenant access with such value operation always succeed. 

Tested with latest from master

```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
Version: 0.10.e7acd33-0.0.1
```

Steps to reproduce:
1. Create Tenant T1 
```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name T1
tenant create succeeded
```
2. Validate Tenant is created successfully
```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
```
3. make sure there is no access set for the tenant
```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name T1
Datastore  Allow_create  Max_volume_size  Total_size  
---------  ------------  ---------------  ----------
```
4. Add valid datastore access to T1
```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access add --allow-create --name=T1 --datastore=TestDatastore
tenant access add succeeded
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name T1
Datastore      Allow_create  Max_volume_size  Total_size  
-------------  ------------  ---------------  ----------  
TestDatastore  True          Unset            Unset     
```
5. Perform following steps

```
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access set --allow-create true --name T1 --datastore=TestDatastore
tenant access set succeeded
[root@promc-2n-dhcp105-97:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name T1
Datastore      Allow_create  Max_volume_size  Total_size  
-------------  ------------  ---------------  ----------  
TestDatastore  False         Unset            Unset      
``` 

While performing the above steps, step#5 leads to the confusion, it says ```tenant access set succeeded``` and sets ```--allow-create``` to ```False``` even though value has been passed at ```true```. 

If ```ALLOW_CREATE``` is case sensitive then ```tenant access set``` invocation should be failed rather saying ```succeeded``` in our scenario.

### PR DESCRIPTION
**Pull request by pshahzeb**
1. Accepting case insensitive true/false values for allow-create option to
   set access for tenants. But throwing error for invalid values for allow-create option

Testing:
Added test cases for case insensitive true/false values as well as invalid value
Manual testing done.

```
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name test_tenant
tenant create succeeded
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access add --allow-create --name=test_tenant --datastore=vsanDatastore
tenant access add succeeded
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name test_tenant
Datastore      Allow_create  Max_volume_size  Total_size
-------------  ------------  ---------------  ----------
vsanDatastore  True          Unset            Unset

[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access set --allow-create false --name test_tenant --datastore=vsanDatastore
tenant access set succeeded
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name test_tenant
Datastore      Allow_create  Max_volume_size  Total_size
-------------  ------------  ---------------  ----------
vsanDatastore  False         Unset            Unset

[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access set --allow-create TRUE --name test_tenant --datastore=vsanDatastore
tenant access set succeeded
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name test_tenant
Datastore      Allow_create  Max_volume_size  Total_size
-------------  ------------  ---------------  ----------
vsanDatastore  True          Unset            Unset

[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access set --allow-create INVALIDVAL --name test_tenant --datastore=vsanDatastore
Invalid value for --alow-create INVALIDVAL
[root@sc-rdops-vm04-dhcp-126-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant access ls --name test_tenant
Datastore      Allow_create  Max_volume_size  Total_size
-------------  ------------  ---------------  ----------
vsanDatastore  True          Unset            Unset
```

fixes #918

### CODE REVIEW

File: esx_service/cli/vmdkops_admin.py [hunk 1]
```diff
@@ -955,9 +958,13 @@ def tenant_access_set(args):
     if args.volume_totalsize:
         volume_totalsize_in_MB = convert.convert_to_MB(args.volume_totalsize)
 
+    if not validate_create_value(str(args.allow_create)):
+        msg = "Invalid value for --alow-create {0}".format(args.allow_create)
```
pdhamdhere: typo alow => allow

esx_service/utils/auth_api.py
```diff
@@ -190,7 +190,8 @@ def modify_privileges(privileges, allow_create, volume_maxsize_in_MB, volume_tot
     logging.debug("modify_privileges: allow_create=%s, volume_maxsize_in_MB=%s, volume_totalsize_in_MB=%s",
                   allow_create, volume_maxsize_in_MB, volume_totalsize_in_MB)
     if allow_create is not None:
-        if allow_create == "True":
+        allow_create = str(allow_create)
```
shuklanirdesh82: oh wait! this doesn't solve the problem fully. 

auth_api._tenant_access_set() just checks for ```true``` and fails to filter out ```Invalid``` as mentioned the testcase (vmdkops_admin_test.py#739) falls to ```false``` rather than complaining.

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -735,30 +735,56 @@ def test_tenant_access(self):
...
-                                                 allow_create="True",
-                                                 volume_maxsize_in_MB=volume_maxsize_in_MB,
-                                                 volume_totalsize_in_MB=volume_totalsize_in_MB
-                                             )
-        self.assertEqual(None, error_info)
 
-        error_info, privileges = auth_api._tenant_access_ls(self.tenant1_name)
-        self.assertEqual(None, error_info)
+        for val in ["False", "FALSE", "false", "Invalid"]:
+            error_info = auth_api._tenant_access_set(name=self.tenant1_name,
```
shuklanirdesh82: ```auth_api._tenant_access_set``` should not be invoked rather call goes through vmdkops_admin the proposed test steps fails to exercise new code.

esx_service/cli/vmdkops_admin.py
```diff
@@ -922,7 +922,10 @@ def tenant_vm_ls(args):
     rows = generate_tenant_vm_ls_rows(vms)
     print(cli_table.create(header, rows))
 
-
+# Allowing upper/lower case true, false values but not
+# any random strings
+def validate_create_value(allow_create):
```
shaominchen: Suggest to rename this method to validate_allow_create or validate_allow_create_value.

esx_service/cli/vmdkops_admin.py
```diff
@@ -955,9 +958,13 @@ def tenant_access_set(args):
     if args.volume_totalsize:
         volume_totalsize_in_MB = convert.convert_to_MB(args.volume_totalsize)
 
+    if not validate_create_value(str(args.allow_create)):
```
shaominchen: 1. We have two different interfaces: VMODL API and Admin CLI - both will call auth_api methods. So we should add the validation logic in auth_api, so that we don't need to repeat the logic in two different interfaces.

2. We should add the validation logic in both tenant_access_add and tenant_access_set methods.

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -735,30 +735,56 @@ def test_tenant_access(self):
...
-                                                 datastore=self.datastore_name,
-                                                 allow_create="True",
-                                                 volume_maxsize_in_MB=volume_maxsize_in_MB,
-                                                 volume_totalsize_in_MB=volume_totalsize_in_MB
-                                             )
-        self.assertEqual(None, error_info)
 
-        error_info, privileges = auth_api._tenant_access_ls(self.tenant1_name)
-        self.assertEqual(None, error_info)
+        for val in ["False", "FALSE", "false", "Invalid"]:
```
shaominchen: I don't follow the logic here: the first 3 values are valid input, but the last one is invalid. However, line 746 assume the returned error_info will be None, which is not correct for the last invalid input.

esx_service/cli/vmdkops_admin.py
pshahzeb: I agree with your first point.
Regarding second point, for the tenant_access_add method, the command line behavior is such that mentioning --allow-create option itself means the option is true. Not mentioning it means it is false. So there isn't a scope for invalid input to be received by tenant_access_add method through Admin CLI interface.
But I am not sure when this api is called through VMODL API; should we assume the caller of tenant_access_add() through VMODL gives a valid value for allow-create option; or is there a scope to receive a wrong value for it?

shaominchen: For AdminCLI, --allow-create without explicitly specifying the value is just a special case for simplifying user input. The CLI should always pass an explicit value to auth_api. 

VMODL API will always pass an explicit value to auth_api.

pshahzeb: Ok. And this explicit value from VMODL API can in wrong/invalid not a boolean value always. So we need a validation in tenant_access_add as well. Am I correct?

shaominchen: Correct. Since Python itself is not type safe, it's possible that non-boolean or other invalid values are passed in.

The AdminCLI commands (including both "tenant access add" and "tenant access set") accept any user input, which may contain invalid values because of typo. So we need to check both methods in auth_api.

pshahzeb: Got it. Thanks.

esx_service/utils/auth_api.py
```diff
@@ -190,8 +195,8 @@ def modify_privileges(privileges, allow_create, volume_maxsize_in_MB, volume_tot
     logging.debug("modify_privileges: allow_create=%s, volume_maxsize_in_MB=%s, volume_totalsize_in_MB=%s",
                   allow_create, volume_maxsize_in_MB, volume_totalsize_in_MB)
     if allow_create is not None:
-        allow_create = str(allow_create)
-        if allow_create.lower() == "true":
+        # allow_create has been valided. It is either "true" or "false"
```
shaominchen: Nit: valided => validated.

esx_service/vmodl/vmodl_test.py
```diff
@@ -690,7 +690,7 @@ def test_update_privilege_with_invalid_volume_size(self):
 
         # Update the privilege with invalid volume size
         with self.assertRaises(vmodl.fault.InvalidArgument):
-            self.tenantMgr.UpdatePrivilege(tenant, self.datastore, volume_max_size=2048, volume_total_size=1024)
+            self.tenantMgr.UpdatePrivilege(tenant, self.datastore, allow_create=True, volume_max_size=2048, volume_total_size=1024)
```
shaominchen: Why do we need to make this change? Same for the changes below.

pshahzeb: Since python VMODL for VCS UpdatePrivilege is using None value for allow_create parameter if it is not provided. This is invalid; unless we have a default behavior when the value for allow_create option is None or not provided during tenant_access_set procedure.

shaominchen: This is incorrect. The auth_api._tenant_access_set() method should be able to handle the case where the allow_create parameter is not set (i.e. set to None). This API is meant to handle update request - it's fair enough that user may just need to update some other properties, but leave existing allow_create unchanged.

esx_service/utils/auth_api.py
```diff
@@ -190,7 +195,8 @@ def modify_privileges(privileges, allow_create, volume_maxsize_in_MB, volume_tot
     logging.debug("modify_privileges: allow_create=%s, volume_maxsize_in_MB=%s, volume_totalsize_in_MB=%s",
                   allow_create, volume_maxsize_in_MB, volume_totalsize_in_MB)
     if allow_create is not None:
-        if allow_create == "True":
+        # allow_create has been valided. It is either "true" or "false"
+        if allow_create == "true":
```
msterin: Looks like allow_create is a boolean. in this case  str(allow_create) (like 640) is not needed, you can do `allow_create = allow_create.lower()` in L640

esx_service/vmodl/vmodl_test.py
pshahzeb: Oh. Got it. Handling this and reverting this changes to tests.

esx_service/utils/auth_api.py
```diff
@@ -190,7 +195,8 @@ def modify_privileges(privileges, allow_create, volume_maxsize_in_MB, volume_tot
     logging.debug("modify_privileges: allow_create=%s, volume_maxsize_in_MB=%s, volume_totalsize_in_MB=%s",
                   allow_create, volume_maxsize_in_MB, volume_totalsize_in_MB)
     if allow_create is not None:
```
pshahzeb: ```if allow_create :``` means it is not None and also it is not boolean False. We have different actions for these; i.e. don't change the privilege if it is None and change the privilege to False if allow_create is False.

msterin: Ah... well, this is confusing to read. I'd recommend initializing allow_create so is never None

pshahzeb: Isn't None a valid value (don't change the privilege)?
boolean, string case insensitive true/false values mean to change the privilege. Rest everything is invalid.

esx_service/utils/auth_api.py
pshahzeb: Not always. Consuming tenant_access_set through admin cli gives allow_create as string, and through VMODL it might be boolean or string.

esx_service/utils/auth_api.py
```diff
@@ -631,10 +637,20 @@ def _tenant_access_set(name, datastore, allow_create=None, volume_maxsize_in_MB=
         error_info = ErrorInfo(err_code, err_msg)
         return error_info
 
+    allow_create_val = str(allow_create)
+    allow_create_val = allow_create_val.lower()
```
pshahzeb: Using ```allow_create_val = str(allow_create).lower()```
Preserving original allow_create value to log error and return it in case of invalid input.

esx_service/utils/auth_api.py
```diff
@@ -631,10 +637,20 @@ def _tenant_access_set(name, datastore, allow_create=None, volume_maxsize_in_MB=
...
+
+    if not validate_allow_create_value(allow_create_val):
+        err_code = ErrorCode.PRIVILEGE_INVALID_ALLOW_CREATE_VALUE
+        err_msg = error_code.error_code_to_message[err_code].format(allow_create)
+        logging.error(err_msg)
+        error_info = ErrorInfo(err_code, err_msg)
```
pshahzeb: Done

esx_service/utils/auth_api.py
msterin: Well, it's even worse then. Please make sure it's type is consistent and not fluctuates between invocations.

I'd suggest to keep as boolean always - i..e one entry call `validate_bool_string` or `conert_to_bool` (built-in bool() won't work for you)  to make sure it's True or False (boolean)  and use str() when needed (though Python will do it for you on print

if you want it to be string always, fine as well. (use `type(var) == bool` to check incoming stuff from vmodl.)

esx_service/utils/auth_api.py
msterin: so the logic seems to be
- check if it's  a valid true/false string (case unsensitive)
- convert to bool
- pass to modify_privileges

in the code would be 
```python
  allow_create_bool, valid = convert_to_bool(allow_create)
 if not valid: 
   ...error msg... and return
 

 ...  pass allow_create_bool to modify privilege
```

right ? 

You code works ok, but reading it requires more effort :-)

pshahzeb: Yes this looks better. Strings can be converted to bools; but the end value of allow_create needs to be True, False or None ( and not just True or False) to distinguish between set privilege, unset privilege and don't change privilege.

msterin: ok.  Please add it to the comments then.

esx_service/utils/auth_api.py
```diff
@@ -161,6 +161,25 @@ def set_privileges(allow_create, privileges, value):
     privileges[auth_data_const.COL_ALLOW_CREATE] = value
     return privileges
 
+# Validating case insensitive true, false strings
```
shaominchen: Please add some comments to explain the return values.

esx_service/vmdk_ops_test.py
```diff
@@ -1005,6 +1005,7 @@ def test_vmdkops_on_tenant_vm(self):
         volume_totalsize_in_MB = convert.convert_to_MB('2GB')
         error_info = auth_api._tenant_access_set(name=self.tenant1_new_name,
                                                  datastore=self.datastore_name,
+                                                 allow_create=True,
```
pshahzeb: Reverting this change. Not needed.

esx_service/utils/auth_api.py
```diff
@@ -189,8 +211,12 @@ def modify_privileges(privileges, allow_create, volume_maxsize_in_MB, volume_tot
     """ Modify privileges based on input params """
     logging.debug("modify_privileges: allow_create=%s, volume_maxsize_in_MB=%s, volume_totalsize_in_MB=%s",
                   allow_create, volume_maxsize_in_MB, volume_totalsize_in_MB)
+
+    # If None, don't change the privilege
+    # If not None, change accordingly
     if allow_create is not None:
-        if allow_create == "True":
+        # allow_create has been validated. It is either True or False
+        if allow_create:
```
pshahzeb: the allow_create is bool here, we can have the explicit bool True check here instead of ```if allow_create:```

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -735,30 +735,70 @@ def test_tenant_access(self):
...
+                            "False",
+                            "500.00MB",
+                            "1.00GB"]
+            actual_output = [rows[0][0],
+                            rows[0][1],
+                            rows[0][2],
+                            rows[0][3]]
+            self.assertEqual(expected_output, actual_output)
+
+        for val in ["True", "TRUE", "true"]:
```
shuklanirdesh82: I would prefer to iterate through all kind of input and on that basis test should compare with expected outcome instead of repeating line#742-765 again.

As discussed offline please make sure to design test steps in future PRs that increases readability.

esx_service/utils/auth_api.py
```diff
@@ -161,6 +161,27 @@ def set_privileges(allow_create, privileges, value):
     privileges[auth_data_const.COL_ALLOW_CREATE] = value
     return privileges
 
+# Validating case insensitive true, false strings
+# returning boolean value of arguement if it is valid
+# else returning it as invalid
+def validate_string_to_bool(allow_create):
+    is_valid = True
```
msterin: nit: Please add dosctring (can just move the comments to docstring)

esx_service/utils/auth_api.py
```diff
@@ -161,6 +161,27 @@ def set_privileges(allow_create, privileges, value):
     privileges[auth_data_const.COL_ALLOW_CREATE] = value
     return privileges
 
+# Validating case insensitive true, false strings
+# returning boolean value of arguement if it is valid
+# else returning it as invalid
```
msterin: I think it now returns more than that

esx_service/utils/auth_api.py
```diff
@@ -554,6 +579,17 @@ def _tenant_access_add(name, datastore, allow_create=None, default_datastore=Fal
         error_info = error_code.generate_error_info(ErrorCode.PRIVILEGE_ALREADY_EXIST, name, datastore)
         return error_info
 
+    if allow_create is not None:
```
msterin: nit: please add comment '# note - the value can be None (no changes), or True/False ("change to this value)")

pshahzeb:
Thanks @msterin. Updated the comments and removing redundant test code.
Also updated the help for admin tenant access set --allow-create option.

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -735,30 +735,53 @@ def test_tenant_access(self):
...
+
+        for val in ["INVALID", ""]:
+            command = ("tenant access set --name={0} ".format(self.tenant1_name))
+            command += ("--datastore={0} ".format(self.datastore_name))
+            command += ("--allow-create={0} ".format(val))
+            command += ("--volume-maxsize=500MB --volume-totalsize=1GB")
+
+            args = self.parser.parse_args(command.split())
+            error_info = vmdkops_admin.tenant_access_set(args)
+            self.assertNotEqual(None, error_info)
```
shuklanirdesh82: would be better to make sure ```vmdkops_admin.tenant_access_set``` throws ```ErrorCode.PRIVILEGE_INVALID_ALLOW_CREATE_VALUE```

PR #970, opened 2017-02-24
In test code, use the constant UUID to create _DEFAULT tenant.

lipingxue:
In test code (vmdk_ops_test.py), during testing for multi-tenency, we need to test the behavior when _DEFAULT tenant and _DEFAULT privilege are not present. So in our code, we have the code to remove _DEFAULT tenant and _DEFAULT privilege. After that particular test is done, we want to get back to the initial state (both _DEFAULT tenant and _DEFAULT privilege are present), so we have code to recreate the _DEFAULT tenant and _DEFAULT privilege.

In that code, we should make sure the _DEFAULT tenant is created with constant UUID as well.

This will not affect any user behavior unless they run "make test-esx".
Please review this PR. @shuklanirdesh82 @shaominchen


esx_service/vmdk_ops_test.py
```diff
@@ -722,15 +722,21 @@ def create_default_tenant_and_privileges(self):
 
         if not tenant_uuid:
             logging.debug("create_default_tenant_and_privileges: create DEFAULT tenant")
-            error_info, tenant = auth_api._tenant_create(
+            error_info, auth_mgr = auth_api.get_auth_mgr_object()
+            if error_info:
+                err = error_code.TENANT_CREATE_FAILED.format(auth.DEFAULT_TENANT, error_info.msg)
```
shuklanirdesh82: it is better to say "Error occurred while getting auth manager object" instead of TENANT_CREATE_FAILED at this point that help in failure triaging.

esx_service/vmdk_ops_test.py
```diff
@@ -722,15 +722,21 @@ def create_default_tenant_and_privileges(self):
...
+                logging.warning(err)
```
shuklanirdesh82: sorry not following much here .. don't we need to fail the test at this point  if error_info is set?

esx_service/vmdk_ops_test.py
lipingxue: That error hit while getting auth manager object will be captured in error_info.msg, and will be included in the error message. The error message will be like "_DEFAULT tenant creates fail with err ...."

esx_service/vmdk_ops_test.py
lipingxue: good catch. I addressed your comments.

PR #988, opened 2017-03-03
Make sure _DEFAULT created after remove is created with constant UUID

lipingxue:
This PR includes the following
Make sure _DEFAULT created after remove is created with constant UUID(issue 986).

Fix the issue that volume create fails on a datastore whose name includes spaces (issue 916).

Please review it.


esx_service/utils/auth_api.py
```diff
@@ -93,10 +93,18 @@ def create_tenant_in_db(name, description, vms, privileges):
         error_info = error_code.generate_error_info(ErrorCode.TENANT_ALREADY_EXIST, name)
         return error_info, None
 
-    error_msg, tenant = auth_mgr.create_tenant(name=name,
-                                                description=description,
-                                                vms=vms,
-                                                privileges=privileges)
+    if name == auth.DEFAULT_TENANT:
```
msterin: I would actually prefer
```python 
if name == auth.DEFAULT_TENANT:
   tenant_uuid = auth.DEFAULT_TENANT_UUID
else:
   tenant_uuid = None

error_msg, tenant = auth_mgr.create_tenant(name=name, ..., tenant_uuid=tenant_uuid)
```

esx_service/vmdk_ops.py
```diff
@@ -683,7 +683,7 @@ def get_vol_path(datastore, tenant_name=None):
 
     if tenant_name and not os.path.isdir(path):
         # The mkdir command is used to create "tenant_name" folder inside DOCK_VOLS_DIR on "datastore"
-        cmd = "{0} {1}".format(MKDIR_CMD, path)
+        cmd = "{0} '{1}'".format(MKDIR_CMD, path)
```
msterin: nice catch!

esx_service/utils/auth_api.py
shaominchen: I'd like to repeat my previous comment - I do think we should handle the tenant uuid inside the create_tenant() API. IMO, uuid should not be handled by the API users. It's the API's responsibility to take care of the appropriate uuids.

If you recall, exposing uuid to the API users has caused multiple unnecessary changes.

msterin: so would you suggest "create_tenant" and "create_default_tenant" APIs, or just put 
```python 
if name == DEFAULT_TENANT_NAME
  uuid = DEFAULT_UUID
else: 
  uuid = generate_uuid
```
into the create_tenant ? (I guess the latter, right ?)

shaominchen: Right. I'm suggesting the latter.

msterin: agreed

lipingxue:
@msterin @shaominchen I have addressed your comments.

Issue #930, opened 2017-02-17
[Tenant] Improve error message thrown while trying to add same vm to another tenant 


ashahi1:
Steps to repro:
1)  Create a tenant and add vm to it:
/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=tenant1 --vm-list=TestVM-1

2) Create another tenant and try adding same vm to the new tenant:
/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name=tenant2 --vm-list=TestVM-1

Error message returned: Internal Error(UNIQUE constraint failed: vms.vm_id)

=====
Error message should be such that its easier to understand for the user that adding same vm to two tenants is not allowed.

PR #992, opened 2017-03-04
Handling tenant vm add, delete, replace corner cases

pshahzeb:
1. Ensuring VM already in a tenant cannot be added to another tenant through tenant_create, tenant_vm_add and tenant_vm_replace methods
2. Ensuring VM list in tenant_vm_add, tenant_vm_replace, tenant_vm_rm cannot be empty and doesn’t contain duplicates
3. Ensuring existing VM in a tenant cannot be replaced back into same tenant
4. Sequenced parameter processing and error message simplification. Appropriate error messages instead of using the database error

Testing:
Updated tenant_vm tests to evaluate this case

Output after fixes:
```
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
------------------------------------  --------  ------------------------  -----------------  --------
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default tenant
1c6b257f-98ff-478a-8bf1-e2a4d1705a8f  test1                                                  1_photon

[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm replace --name test2 --vm-list 1_photon
Tenant test2 does not exist
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name test2 --vm-list 1_photon
VM '1_photon' has already been associated with another tenant, can't add it to tenant 'test2'
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name test2
tenant create succeeded
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm add --name test2 --vm-list 1_photon,2_photon
VM '1_photon' has already been associated with another tenant, can't add it to tenant 'test2'
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm add --name test2 --vm-list 2_photon
tenant vm add succeeded
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm replace --name test2 --vm-list 2_photon
VM '2_photon' has already been associated with tenant 'test2', cannot add it again
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm replace --name test2 --vm-list 1_photon
VM '1_photon' has already been associated with another tenant, can't add it to tenant 'test2'
```

```
[root@sc-rdops-vm17-dhcp-12-47:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name test6 --vm-list 1_photon,1_photon
VMs ['1_photon', '1_photon'] contain duplicates, they should be unique
```
fixes (#930 #994 )


esx_service/utils/auth_api.py
```diff
@@ -299,8 +299,20 @@ def _tenant_create(name, description="", vm_list=None, privileges=None):
     if not description:
         description = ""
 
-    logging.debug("_tenant_create: vms=%s", vms)
+    error_info, tenant_list = get_tenant_list_from_db()
```
shaominchen: 1. This logic should not be triggered if vm-list is None.
2. The parameter processing sequence in this method is a little bit messy. We should process the parameters one by one clearly: name, vm_list and description.

msterin: This comment is not for the code change, but for the code structure itself. 
tenant_create() is really 2 steps  (1) create empty tenant (2) add VMs if needed. 
so it should simply call _tenant_vm_add instead of creating another path of doing the same. 
//CC @lipingxue

esx_service/utils/auth_api.py
```diff
@@ -299,8 +299,20 @@ def _tenant_create(name, description="", vm_list=None, privileges=None):
...
+
     vms_uuid_list = [(vm_id) for (vm_id, vm_name) in vms]
+
+    if vms_uuid_list and any_vm_part_of_another_tenant(vms_uuid_list, tenant_list):
+        error_info = error_code.generate_error_info(ErrorCode.VM_IN_ANOTHER_TENANT, vm_list, name)
+        logging.error("_tenant_create: %s", error_info.msg)
+        return error_info, None
+
+    logging.debug("_tenant_create: vms=%s", vms)
+
```
msterin: the new code is cut-n-pasted twice. Please change it to use a function, an you wont even need a separate one for `any_vm_part_of_another_tenant` since it will be wrapped up there.

esx_service/utils/error_code.py
```diff
@@ -69,6 +70,7 @@ class ErrorCode:
     ErrorCode.REPLACE_VM_EMPTY : "Replace VM cannot be empty",
     ErrorCode.VM_ALREADY_IN_TENANT : "Some VMs in {0} have already been associated with tenant '{1}', cannot add them again",
     ErrorCode.VM_NOT_IN_TENANT : "Some VMs in {0} have not been associated with tenant '{1}', cannot remove them",
+    ErrorCode.VM_IN_ANOTHER_TENANT : "Some VMs in {0} have already been associated with another tenants, can't add them to tenant '{1}'",
```
shaominchen: Nit:　“another tenants" => "other tenants" or "another tenant"

esx_service/utils/auth_api.py
```diff
@@ -432,13 +454,24 @@ def _tenant_vm_add(name, vm_list):
         error_info = error_code.generate_error_info(ErrorCode.VM_ALREADY_IN_TENANT, vm_list, name)
         return error_info
 
+    error_info, tenant_list = get_tenant_list_from_db()
```
shaominchen: Same as above - this logic should not be triggered if vm-list is None. For this method, however, I think we should return right away if vm-list is None at the very first beginning (after line 434) to avoid unnecessary cost, because the whole purpose of this method is to add VMs to a tenant.

esx_service/utils/auth_api.py
pshahzeb: @msterin  I think tenant_create() and adding VMs to it (if specified in options) should be treated as an atomic operation since it was issued in the same command.  If adding VMs to the empty tenant fails, we ll have to cleanup the empty tenant. So creating empty tenant and adding VMs (2 db queries) and then a db commit looks more safe; unless we opt for cleaning the empty tenant in case of VM adding failed. What do you think?

esx_service/utils/auth_api.py
pshahzeb: @shaominchen  makes sense. 

So when the vmlist is empty for vm add operation, do we return without error? Or should we return an error enforcing the user to specify list of VMs when triggering vm add operation to a tenant?

shaominchen: I think we should return an error - we probably need to introduce a new error code here.

esx_service/utils/auth_api.py
lipingxue: +1 on  this.

esx_service/utils/auth_api.py
lipingxue: +1 on this. Function "any_vm_part_of_another_tenant" does not need to pass in param @tenant_list.

esx_service/utils/auth_api.py
```diff
@@ -398,36 +410,57 @@ def _tenant_ls(name=None):
     error_info, tenant_list = get_tenant_list_from_db(name)
     return error_info, tenant_list
 
-def any_vm_already_exist(existing_vms, vms):
+def vm_already_exist(name, vms):
```
shaominchen: The name of this method and the "vm_in_tenant" method below may cause some confusion. To clarify the purpose of these two methods, I'd suggest to make the following changes:
1. vm_already_exist(name, vms) => vm_already_in_tenant(name, vms)
2. vm_in_tenant(name, vms) => vm_in_any_tenants(vms)

Please note that I have removed the "name" parameter from the 2nd method because we don't really need the tenant name.

esx_service/utils/auth_api.py
```diff
@@ -398,36 +410,57 @@ def _tenant_ls(name=None):
...
+        if vm_id in existing_vms:
+            error_info = error_code.generate_error_info(ErrorCode.VM_ALREADY_IN_TENANT,
+                                                        vm_name, name)
+            logging.error(error_info.msg)
+            return error_info
+
+    return None
 
-def any_vm_not_exist(existing_vms, vms):
+def vm_not_exist(name, vms):
```
shaominchen: Same as above - suggest to rename to "vm_not_in_tenant"

esx_service/utils/auth_api.py
pshahzeb: Thanks. Actually, the name is used for generating the error message.

shaominchen: I understand that. However, for the purpose of this method, we don't need the tenant name parameter. Actually for the error message I think we should log the tenant which contain this VM instead of the tenant to be added.

pshahzeb: Got it.

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -639,6 +640,15 @@ def test_tenant_vm(self):
         actual_output = rows
         self.assertEqual(expected_output, actual_output)
 
+        # Trying to create tenant with duplicate vm names
+        error_info, tenant_dup = auth_api._tenant_create(
+                                                    name="Tenant_dup",
```
shuklanirdesh82: nit: Tenant_dup => Tenant_add_dupVms. I guess test is expecting ```error_code.ErrorCode.VM_DUPLICATE```

lipingxue:
LGTM

Issue #965, opened 2017-02-24
[Tenant] "vmdkops_admin tenant vm ls" is not showing VMs belongs to _DEFAULT tenant


shuklanirdesh82:
It would be helpful and great to have such information like how many docker_hosts are participating into ```_DEFAULT``` tenant just like vmdkops_admin showing for other tenants.

```
[root@sc-rdops-vm17-dhcp-0-84:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant ls
Uuid                                  Name      Description               Default_datastore  VM_list    
------------------------------------  --------  ------------------------  -----------------  ---------  
594dd43b-b93b-41c8-82bb-1e65374b8648  vsan                                                   VM84_vsan  
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default tenant                                

[root@sc-rdops-vm17-dhcp-0-84:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm ls --name=vsan
Uuid                                  Name       
------------------------------------  ---------  
422a93bc-62ac-e90d-6c7c-a028c6568a45  VM84_vsan  

[root@sc-rdops-vm17-dhcp-0-84:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm ls --name=_DEFAULT
Uuid  Name  
----  ----  

[root@sc-rdops-vm17-dhcp-0-84:~] 
```

//CC @msterin @pdhamdhere

lipingxue:
Why do we need that info? Any VM which are not part of any user created tenant is associated with _DEFAULT tenant.

msterin:
I agree with @shuklanirdesh82 , it is confusing when a specific request ("show me vms in default tenant") shows nothing - but it has nothing to show. We can  print something like "all vms not participating in other tenants are a part of _DEFAULT tenant". this is a low priority though

Issue #922, opened 2017-02-17
The command behavior of "tenant rm" and "policy rm" are inconsistent


shaominchen:
When we run "vmdkops_admin policy rm MyPolicy", we don't need to specify the parameter name "--name". However, to run "vmdkops_admin tenant rm --name MyTenant", we must specify "--name".

I think we should keep the command behavior consistent. My suggestion is to fix the "tenant rm" command, because the behavior of "policy rm" command is more natural, and it's also consistent with other CLI tools, such as "docker volume rm" command.

tusharnt:
Even tests need to be fixed.

msterin:
The suggested change will leave 'tenant <command>' set inconsistent. 
Please do not do it - instead, *if* it is an issue, we need to do a pass on all commands and make them consistent in using --name vs positional name

shaominchen:
@msterin I don't think this will leave "tenant set" inconsistent. For CUD operations, only D (rm) does not need "--name". Add & set still need "--name". This is just a convention for most command lines tools.

pshahzeb:
From https://docs.docker.com/engine/reference/commandline/docker/
I think many, if not all, docker commands use positional name instead of --name for all CUD operations. (only volume create requires --name)

I think we can go ahead with something consistent for all commands. Right now, only policy rm command takes positional name. All other commands require --name.

msterin:
> This is just a convention for most command lines tools.
I do not think it's true for "most" CLIs - can you give us examples ? 
At any rate, It is certainly  not true for most Docker subcommands, and it is super confusing IMO as the user has to think every time what is the command format. 

A natural way is to use positionals. I am fine with using `--name` also (as Docker user to do but seems to have moved away from) as long as it's consistent across all commans.

So I agree with @pshahzeb  here.  And if we do it, we also need to change the related doc and test, and do it in 0.13 or GA , it will be more impacful later.

shaominchen:
@msterin Some examples:
"docker create" and "docker volume create" use "--name", but "docker rm" and "docker volume rm" use positional name;
"vmdkops_admin policy create" and "vmdkops_admin policy update" use "--name", but "vmdkops_admin policy rm" use positional name

Anyway, I agree that keep consistency (especially local consistency) is very important.

pshahzeb:
@shaominchen @msterin 
[...]

PR #993, opened 2017-03-04
Admincli improvement

pshahzeb:
1. Info message for default tenant vm ls command

2. Moving default tenant and privilege constants to auth_data_const.py
    Related changes for using these constants

3. Making use of --name option for policy rm

Testing:
Updated testcases

Resolving: #965 
```
[root@sc-rdops-vm17-dhcp-30-74:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm ls --name _DEFAULT
_DEFAULT tenant contains all VMs which were not added to other tenants
```

Resolving: #922 
```
[root@sc-rdops-vm17-dhcp-30-74:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy ls
Policy Name  Policy Content                                             Active
-----------  ---------------------------------------------------------  ------
test-policy  (("proportionalCapacity" i0)("hostFailuresToTolerate" i0)  Unused

[root@sc-rdops-vm17-dhcp-30-74:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy rm --name test-policy
Successfully removed policy: test-policy
[root@sc-rdops-vm17-dhcp-30-74:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy ls
Policy Name  Policy Content  Active
-----------  --------------  ------
```


esx_service/cli/vmdkops_admin.py
```diff
@@ -914,6 +913,13 @@ def generate_tenant_vm_ls_rows(vms):
 
 def tenant_vm_ls(args):
     """ Handle tenant vm ls command """
+
+    # Handling _DEFAULT tenant case separately to print info message
+    # instead of printing empty list
+    if (args.name == "_DEFAULT"):
```
shaominchen: Can we use a constant instead of hard-coded string? I'd suggest to move all the constants below from auth.py to auth_data_const.py so that they can be shared by other modules:

DEFAULT_TENANT = '_DEFAULT'
DEFAULT_TENANT_UUID = '11111111-1111-1111-1111-111111111111'
DEFAULT_DS  = '_DEFAULT'
DEFAULT_DS_URL = DEFAULT_DS + "_URL"

esx_service/cli/vmdkops_admin.py
```diff
@@ -914,6 +913,13 @@ def generate_tenant_vm_ls_rows(vms):
...
+        print("All VMs that do not belong to any tenant belong to _DEFAULT tenant")
```
msterin: the message is confusing IMO. Should be something "'__DEFAULT' tenant contains all VMs which were not added to other tenants"

esx_service/cli/vmdkops_admin.py
msterin: +1 on this

esx_service/cli/vmdkops_admin.py
```diff
@@ -236,9 +236,8 @@ def commands():
                     'func': tenant_rm,
                     'help': 'Delete a tenant',
                     'args': {
-                        '--name': {
+                        'name': {
```
msterin: why is it changed for 'tenant rm' ? I see 'tenant create' and 'tenant update' are stil using `--name` and 'tenant rm' now uses positional 'name'. This is inconsistent. 

Please drop this change, and instead use #922 to enumerate all changes to --name or name whihc are needed to make all consistent. See #922 for more comments

shaominchen: Let's discuss this in #922.

shaominchen: @lipingxue We all agree that local consistency is more important. For AdminCLI commands, the only inconsistent command is "policy rm". So we should just fix that command to use "--name" instead of fixing all other commands to use positional name.

esx_service/cli/vmdkops_admin.py
```diff
@@ -159,8 +159,9 @@ def commands():
                     'func': policy_rm,
                     'help': 'Remove a storage policy',
                     'args': {
-                        'name': {
-                            'help': 'Policy name'
+                        '--name': {
```
shuklanirdesh82: sorry I am late. Can you make sure admin_cli information is in sync with the proposed changes?

e.g. https://github.com/vmware/docker-volume-vsphere/blob/master/docs/user-guide/admin-cli.md#remove-3

PR #996, opened 2017-03-06
Skip exiting the volume service on connect failure.

govint:
Modified connectLocalSi() in vmdk_ops.py to return on connect errors. If unable to connect on service startup the service still starts and subsequent commands may get an error till the service is able to connect to the host agent on the host (logged in the service logs).


govint:
Resolving test errors, will update for review.

govint:
Closing as testing more scenarios.

PR #997, opened 2017-03-06
Allow to use latest Docker (v 17.xx and later) in build

msterin:
The version checking scheme in build.sh was not handling new Docker major versions (i.e. 17....)
well so the build was failing with "Docker 1.8 or later is required".
Fixed the version checking scheme to work properly.
We probably do not need this at all since Docker version < 1.8 is fairly ancient, but keeping in the code out of pure paranoia.


misc/scripts/build.sh
```diff
@@ -58,14 +58,14 @@ then
    exit 2
 fi
 
-min_ver=`echo $ver | grep -o '\.[0-9]*\.' | sed 's/\.//g'`
-if [ $min_ver -lt 8 ]
+maj_ver=`echo $ver | sed 's/\..*$//'`
+min_ver=`echo $ver | sed 's/[0-9]*\.//' | sed 's/\..*$//'`
+if [ $maj_ver -lt 17 -a $min_ver -lt 8 ]
```
pdhamdhere: Hmm Is this condition correct? maj_ver < 17 && min_ver < 8? Why?

I think we can get rid of this code altogether and just document.

msterin: yes, this condition means 1.7 and below :-). See description - yes, we can drop it and document or we can leave it and not document - my (paranoid) preference is to leave it to handle the case and not touch any docs.

pdhamdhere:
LGTM

PR #999, opened 2017-03-07
Properly handle successful Run() with stale errno

msterin:
Sometimes vmci_server.c does not set errno when returning Success.
In this cases stale errno (e.g. from priorly failed syscall) might be
accidentally picked up. The commit fixes this issue, and also  improves
error printing for inotify failures

@govint- it is a light modification of a fragment of your PR. I do not want to wait for the whole PR to come in but it is blocking me, so sending this piece as a separate PR 

TEST: manual test on Alpine only (using it for managed plugin conversion). Before the test ALL commands in my runs (including "docker volume ls") where failing with "file not found" which was a leftover from something else. After the fix the issues disappeared. 

//CC @govint @pdhamdhere


vmdk_plugin/drivers/vmdk/vmdkops/esx_vmdkcmd.go
```diff
@@ -93,22 +93,28 @@ func (vmdkCmd EsxVmdkCmd) Run(cmd string, name string, opts map[string]string) (
...
-			if i < maxRetryCount {
-				log.Warnf(msg + " Retrying...")
-				time.Sleep(time.Second * 1)
-				continue
-			}
-
-			if errno == syscall.ECONNRESET || errno == syscall.ETIMEDOUT {
-				msg += " Cannot communicate with ESX, please refer to the FAQ https://github.com/vmware/docker-volume-vsphere/wiki#faq"
+		ret, err = C.Vmci_GetReply(C.int(EsxPort), cmdS, beS, ans)
+		if ret != 0 {
```
pdhamdhere: Are we doubly sure that ret == 0 && err != nil will never happen?

pdhamdhere: Oops. Re-read the PR description. LGTM.

msterin: it happens all the time on alpine ::-), it simply means the errno is not properly reset (by someone else) and we may see the stale one. I did see it when working with the managed plugin on Alpine distro. It is explicitly a non-event for us since return 0 means we are totally good

msterin:
Aha ! the CI picked up the new error message: 
```
creating volume [abc_1488855065] on VM[192.168.31.176]---
FAIL: TestVolumeCreationFirstTime (6.95s)
	createVolume_test.go:58: 
		Error has occurred [Error response from daemon: create abc_1488855065: VolumeDriver.Create: Internal issue: ret != 0 but errno is not set. Cancelling operation.
		] while creating volume [abc_1488855065] very first time: err -> exit status 1
```
@govint - that seems to confim the conclusion about "we return err, do not set errno, this fooling the code into improperly handle memory in vmci exchange with Go.

I will go ahead and merge, we should expect more CI failures before you adjust and merge #941. What's the ETA on this one ?

Issue #1002, opened 2017-03-07
Corrupted reply mesage to a typo in `docker volume create` command


shuklanirdesh82:
Getting ```Valid options and defaults: <zip object at 0x431f1f6b48>``` when there is a typo in volume creation command input
```
root@photon-xVmYMbyTn [ ~ ]# docker volume create --driver=vsphere --name abc334@vsanDatastore -o size=1MB -o vsan-policy-nam=abc     
Error response from daemon: create abc334@vsanDatastore: VolumeDriver.Create: Invalid options: ['vsan-policy-nam'] 
Valid options and defaults: <zip object at 0x431f1f6b48>
```
---------docker-volume-vsphere.log---------
```
2017-03-07 07:23:57.338452077 +0000 UTC [ERROR] Create volume failed error="Invalid options: ['vsan-policy-nam'] \nValid options and defaults: <zip object at 0x431f1f7248>" name="abc334@vsanDatastore"
```
---------vmdk_ops.log---------
```
03/07/17 07:23:53 1321926 [Thread-71] [INFO   ] cmd create with opts {'vsan-policy-nam': 'abc', 'size': '1MB', 'fstype': 'ext4'} on tenant_uuid 42b1a76c-65df-4cf6-b18d-9dd17b684a9b datastore vsanDatastore is allowed to execute
03/07/17 07:23:53 1321926 [vmFolderPhoton-[vsanDatastore] dockvols/42b1a76c-65df-4cf6-b18d-9dd17b684a9b/abc334.vmdk] [INFO   ] *** createVMDK: /vmfs/volumes/vsanDatastore/dockvols/42b1a76c-65df-4cf6-b18d-9dd17b684a9b/abc334.vmdk opts = {'vsan-policy-nam': 'abc', 'size': '1MB', 'fstype': 'ext4'}
03/07/17 07:23:53 1321926 [vmFolderPhoton-[vsanDatastore] dockvols/42b1a76c-65df-4cf6-b18d-9dd17b684a9b/abc334.vmdk] [INFO   ] executeRequest 'create' completed with ret={'Error': "Invalid options: ['vsan-policy-nam'] \nValid options and defaults: <zip object at 0x431f1f7248>"}
```

Plugin/driver version information:
On VM: ```docker-volume-vsphere-0.12.3f64805-1.x86_64```

on ESX:
```
[...]

PR #1004, opened 2017-03-07
Correct printing of error info message for docker volume create command

pshahzeb:
zip() function in python3 returns an iterator. So creating a list from it for
proper error info message printing for docker volume create command

Resolves: #1002

After fix:
```
root@photon-xVmYMbyTn [ ~ ]# docker volume create --driver=vsphere --name abc334@vsanDatastore -o size=1MB -o vsan-policy-na=abc
Error response from daemon: create abc334@vsanDatastore: VolumeDriver.Create: Invalid options: ['vsan-policy-na']
Valid options and defaults: [('size', '100mb'), ('vsan-policy-name', '[VSAN default]'), ('diskformat', 'thin'), ('attach-as', 'independent_persistent'), ('access', 'read-write'), ('fstype', 'ext4'), ('clone-from', 'None')]
```


pshahzeb:
@pdhamdhere Yes it does.

PR #1005, opened 2017-03-08
Detach after mount failure, and a small fix for logs

msterin:
### Make sure we detach after fresh volume mount failure (before returning failure to Docker)
- on create, volume is created, then attached , then mounted for formatting
-  If the mount fails for whatever reason, `Mount()`  used to drop the refcount and immediately return failure to upstairs - without detaching. 

### Fix some warning to print correctly
 * We need to use warning_f_ for formatted strings

### Testing
Manual:
* validated the log, e.g. `Failed to load config file /etc/docker-volume-vsphere.conf`. (it used to be `Failed to load config file %s /etc/docker-volume-vsphere.conf`
* during mount failure testing, observed that after failed mount I see `Detaching vol2 - it is not used anymore` message and VMDK is detached from the VM


vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -400,7 +400,11 @@ func (d *VolumeDriver) Mount(r volume.MountRequest) volume.Response {
 			log.Fields{"name": r.Name, "error": err.Error()},
 		).Error("Failed to mount ")
 
-		d.decrRefCount(r.Name)
+		refcnt, _ := d.decrRefCount(r.Name)
+		if refcnt == 0 {
+			log.Infof("Detaching %s - it is not used anymore", r.Name)
+			d.ops.Detach(r.Name, nil) // try to detach before failing the request for volume
```
pdhamdhere: Should you be calling "d.UnmountVolume(r.Name)" like in Unmount?

msterin: Logically we do not want to unmount since we know the mount has failed. We just want to clean up the attach.
Technically UnmountVolume does a set of useless calls (i.e. tries to locate the name and unmount volume) which we know will fail  and  guarantee confusing messages about failing to unmount, and  then it will do the only useful step (d.ops.Detach) which we can do directly.  

I guess that was a long winded version of "no. we need to call Detach"  :-)

pdhamdhere: 😀 yes, makes sense.

Issue #886, opened 2017-02-03
Ship guest docker volume plugin as managed plugin


pdhamdhere:
Docker managed plugins are out of experimental starting from Docker 1.13. We should consider shipping both vSphere as well as Photon as docker managed plugin.

Reference: https://docs.docker.com/engine/extend/#/developing-a-plugin

The following are PRs related to this feature in docker 1.13 release notes.

Move plugins out of experimental #28226
Add --force on docker plugin remove #25096
Add description in docker plugin ls #25556
Add -f/--format to docker plugin inspect #25990
Add docker plugin create command #28164
Split docker plugin install into two API call /privileges and /pull #28963

REX_Ray seems have implemented the managed plugin system, this blog is a good reference https://blog.codedellemc.com/2017/02/02/deep-dive-docker-1-13-storage-plugins-propagated-mounts/

shaominchen:
Since this feature needs investigation and there are quite some uncertainties, we should start to look into this ASAP to make sure this can be delivered in GA.

shaominchen:
Per offline discussion, Mark will take care of this.

msterin:
Update:
#1006 introduces the managed plugin built. 

Issues still to address: 
* We cannot simply use it from build/CI since it pushes to registry and multiple dev builds/CIs will be stepping on each other. So we need to discuss and adjust build/CI  
* Photon OS seems to have issues - 'mount' consistently fails from managed plugin, but works fine in rpm/deb installation and fine in other distros. So we'd need to keep rpm/deb around for a while 
* minor: we need to remove a mount for /etc from the managed plugin mounts, and instead place config gile in /etc/vmware OR simply use an env to point to it's location (and add env to plugin config.json)

msterin:
This is what we plan to do to address the points in the comment above:
* Devs (or whoever builds of git) will have to use  their own dockerhub accounts. We will try to construct it from DOCKER_HUB_USER env or `whoami`. We will rely on dev having done proper `docker login`, and we will provide help message if the `docker plugin push` fails. W3 will use hardcoded tag (as today) 
 * change to plugin/Makefile (tag->dev) 
 * change to vmdk_plugin Makefile 'make deploy-vm'
* CI build will use cnastorage account. Since CI serializes the build, we will add TAG=ci 
 * change to build scripts
** on release, we will manually push to vmvare/xxx and tag with the release number

//CC @pdhamdhere @shuklanirdesh82

PR #1006, opened 2017-03-08
Support for building DVS as Docker Managed Plugin.

msterin:
See issue #886 
Docker managed plugins are currently documented [here](https://docs.docker.com/engine/extend/)

This PR introduces new folder `./plugin`. Running `make all` in this folder will assume that
the DVS is pre-built and dockerhub is authenticared with `docker login`, then it will  package DVS
as a Managed Plugin, and push to dockerhub.

**Note that this is not integrated with Build or CI - since we do Build under Docker, a bit more work is needed in central Makefile and drone script - I will do it in a separate PR if this one will be approved/merged soon, or in an extension to this one if it will still be in review.** 

See Makefile for name/tag and location.

Assuming the Makefile defines names as 'cnastorage/docker-volume-vsphere:0.12', the plugin  could
be installed on Docker (1.13+) as follows:

* no question asked, and pretend the plugin name is 'vsphere' (can be used in `volume create` and `plugin rm`)
```
docker plugin install --grant-all-permissions --alias vsphere cnastorage/docker-volume-vsphere:0.12
```
* vanilla interactive install as disabled plugin (will require `docker plugin enable `  to be operational
```
docker plugin install --disable cnastorage/docker-volume-vsphere:0.12
```

The `docker volume` command behaves as before but the plugin name in `docker create` is either what was passed as `--alias` or the full plugin name , i.e. `cnastorage/docker-volume-vsphere:0.12`


plugin/Makefile
```diff
@@ -0,0 +1,57 @@
...
+USER := cnastorage
+
+PLUGIN_NAME=$(USER)/docker-volume-vsphere
+PLUGIN_TAG=0.12
+
+# Binaries we want to pick up
+BINARY := docker-volume-vsphere
+BIN_LOC := ../build
+
+# Tmp docker image used to constuct rootfs + our banaries
```
pdhamdhere: typo banaries => binaries

plugin/config.json
```diff
@@ -0,0 +1,43 @@
+{
+	"Description": "VMWare vSphere Docker Volume plugin",
+	"Documentation": "http://vmware.github.io/docker-volume-vsphere/documentation",
+	"Entrypoint": ["/usr/bin/docker-volume-vsphere", "--log_level", "info"],
+	"PropagatedMount": "/mnt/vmdk",
+	"Mounts": [
```
pdhamdhere: What about "/etc/" for `/etc/docker-volume-vsphere.conf`

plugin/Makefile
```diff
@@ -0,0 +1,57 @@
...
+# Makefile for Docker data volume *managed* plugin construction
+# This Makefile assumes it's in ./plugin in the tree and the binaries are already built
+# Eventually we will throw away deb/rpm build from ../vmdk_plugin/Makefile and will use only
+# managed plugin
+
+# User component of plugin name (and dockerhub group to push to)
+USER := cnastorage
+
+PLUGIN_NAME=$(USER)/docker-volume-vsphere
+PLUGIN_TAG=0.12
```
pdhamdhere: Later we will have to find a way to separate dev builds from release tags.

plugin/config.json
msterin: yeah. I did not want to open up all important folders. I think I will change it to /etc/vmware/<config> and open /etc/vmware. Comments ?

PR #1015, opened 2017-03-09
Allow setting log level from `docker plugin install`

msterin:
This PR allows setting log level of managed plugin  to debug (or other level).

This is done by adding `DVV_LOG_LEVEL=debug` (or other level) to the end of `docker plugin install...` command.

tested: installed with and without the option and checked the log shows correct level . 
doc: plugin/Readme.md


plugin/config.json
```diff
@@ -48,6 +47,13 @@
 		"Capabilities": ["CAP_SYS_ADMIN"],
 		"Devices": null
 
+	},
+	"Env": [ {
+      "name": "DVV_LOG_LEVEL",
```
shaominchen: How about just "LOG_LEVEL"?

Issue #961, opened 2017-02-23
[VSAN] Incorrect 'Used' size reported for volume of diskformat type - zeroedthick on vsan datastore


ashahi1:
Created a volume of size 1gb but 'Used' size is reported as 2gb.

Steps:
1. Created a volume of size 1gb with diskformat set to zeroedthick
`docker volume create --driver=vmdk --name=vsanVolm -o size=1gb -o diskformat=zeroedthick`

2. Did docker volume inspect for created volume:
```
root@photon-TX4FfW2ET [ ~ ]# docker volume inspect vsanVolm
[
    {
        "Name": "vsanVolm",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/vsanVolm",
        "Status": {
            "access": "read-write",
            "attach-as": "independent_persistent",
            "capacity": {
                "allocated": "2GB",
                "size": "1GB"
            },
            "clone-from": "None",
            "created": "Thu Feb 23 21:58:53 2017",
            "created by VM": "Worker-1-216",
            "datastore": "vsanDatastore",
            "diskformat": "zeroedthick",
            "fstype": "ext4",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]
```
3. Admin cli command output on ESX
```
[root@sc2-rdops-vm02-dhcp-89-216:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls -c volume,datastore,capacity,used
Volume    Datastore      Capacity  Used
--------  -------------  --------  ----
vol1      vsanDatastore  10GB      8MB
vsanVolm  vsanDatastore  1GB       2GB  <<<<<<<
```
Did same steps on shared-vmfs datastore. 'Used' and 'Capacity' are same.

Output is as follows:
```
[root@sc2-rdops-vm02-dhcp-90-19:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py ls -c volume,datastore,capacity,used
Volume    Datastore     Capacity  Used
--------  ------------  --------  ----
vsanVolm  sharedVmfs-0  1GB       1GB

```
--------------------------
Attached logs:
[docker-volume-vsphere.txt](https://github.com/vmware/docker-volume-vsphere/files/797885/docker-volume-vsphere.txt)
[vmdk_ops.txt](https://github.com/vmware/docker-volume-vsphere/files/797886/vmdk_ops.txt)

pshahzeb:
[...]

PR #1030, opened 2017-03-13
Documentation update for admin policy rm and docker volume inspect commands

### LINKED ISSUE #922: The command behavior of "tenant rm" and "policy rm" are inconsistent

pshahzeb:
1. Usage of --name option instead of positional name for admin policy rm command

2. Note for docker volume inspect command in case of zeroedthick and eagerzeroedthick disk formats. It shows allocated size to be total size plus the size of replicas

Fixes #922 #961


docs/user-guide/docker-volume-cli.md
```diff
@@ -127,6 +127,7 @@ docker volume inspect MyVolume
     }
 ]
 ```
+Note: For disk formats zeroedthick and zeroedthick, the allocated size would be total size plus the size of replicas.
```
shuklanirdesh82: don't we need to add this at known issue?

pshahzeb: IMO, it is not an issue. The disk format's nature results in this behavior.

msterin:
lgtm

Issue #962, opened 2017-02-23
Failed to create filesystem


pdhamdhere:
Guest: Ubuntu 
Plugin & ESX: 0.11.2

```
docker volume create -d vmdk test
Error response from daemon: create test: VolumeDriver.Create: Failed to create filesystem on /dev/disk/by-path/pci-0000:0b:00.0-scsi-0:0:0:0: exit status 1. Output = mke2fs 1.42.13 (17-May-2015)
The file /dev/disk/by-path/pci-0000:0b:00.0-scsi-0:0:0:0 does not exist and no size was specified.
```

Plugin Log

```
vi /var/log/docker-volume-vsphere.log 

2017-02-23 23:30:36.775775667 +0000 UTC [INFO] No config file found. Using defaults.
2017-02-23 23:30:36.779514188 +0000 UTC [INFO] Starting plugin driver=vmdk log_level=debug config="/etc/docker-volume-vsphere.conf"
2017-02-23 23:30:36.779532548 +0000 UTC [INFO] Plugin options - target= project= host= port=1019
2017-02-23 23:30:36.779587945 +0000 UTC [INFO] Getting volume data from unix:///var/run/docker.sock
2017-02-23 23:30:36.881349688 +0000 UTC [DEBUG] Docker info: version=1.13.1, root=/var/lib/docker, OS=Ubuntu 16.04.1 LTS
2017-02-23 23:30:36.885583808 +0000 UTC [DEBUG] Found 1 running or paused containers
2017-02-23 23:30:36.892998512 +0000 UTC [DEBUG]   Mounts for [/ng_nginx.1.l5ofaz04vxbv57yptgv2e3v20]
2017-02-23 23:30:36.893479953 +0000 UTC [INFO] Discovered 0 volumes in use.
2017-02-23 23:30:36.893535381 +0000 UTC [INFO] Docker VMDK plugin started mock_esx=false version="VMDK Volume Driver v0.3" port=1019
2017-02-23 23:30:36.893624281 +0000 UTC [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vmdk.sock"
2017-02-23 23:30:36.893795439 +0000 UTC [DEBUG] root group found. gid: 0
2017-02-23 23:34:00.878984855 +0000 UTC [DEBUG] vmdkOps.List
2017-02-23 23:34:24.853644321 +0000 UTC [DEBUG] vmdkOps.Get name=test
[...]

Issue #12, opened 2016-02-08
The current code has TODOs sprinkled, they should be closed or moved into issues


kerneltime:
There is a lot of work pending. We should either close the minor TODOs or move them into issues.

PR #1040, opened 2017-03-16
Use a longer delay in case the short timeout is insufficient for device attach

govint:
When adding a PVSCSI contro;ller on first vmdk attach, at times the process takes around 5 seconds by which time the plugin timesout and fails the create operation. At times this takes just a second.

To ensure that the client is flexible to handle these scenarios first we try with a 1 secind timeout (default) and if the device is still not present then wait for a longer timeout of 5 seconds before giving up.

In the below sample adding a pvscsi controller is done in ~1 second,
03/16/17 13:25:34 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk] [INFO   ] Attaching /vmfs/volumes/bigone/dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk as independent_persistent
....
03/16/17 13:25:34 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk] [DEBUG  ] backing_disk=Ubuntu-14.04-64-LTS-IMG/Ubuntu-14.04-64-LTS-IMG-000001.vmdk virtual_disk=dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk
03/16/17 13:25:34 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk] [INFO   ] Adding a PVSCSI controller
...
03/16/17 13:25:35 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk] [INFO   ] executeRequest 'attach' completed with ret={'ControllerPciSlotNumber': '224', 'Unit': '0'}
03/16/17 13:25:35 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz19.vmdk] [DEBUG  ] lib.vmci_reply: VMCI replied with errcode 0

And in this case takes two seconds,
03/16/17 13:16:00 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz14.vmdk] [INFO   ] Attaching /vmfs/volumes/bigone/dockvols/11111111-1111-1111-1111-111111111111/zzzz14.vmdk as independent_persistent

...

03/16/17 13:16:00 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz14.vmdk] [INFO   ] Adding a PVSCSI controller

....

03/16/17 13:16:02 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz14.vmdk] [INFO   ] executeRequest 'attach' completed with ret={'ControllerPciSlotNumber': '224', 'Unit': '0'}
03/16/17 13:16:02 104884 [2 (1)-[bigone] dockvols/11111111-1111-1111-1111-111111111111/zzzz14.vmdk] [DEBUG  ] lib.vmci_reply: VMCI replied with errcode 0


govint:
Fixes #962

govint:
I tried that but the event handling logic waits for that much time for
every single attach meaning every container always starts with that delay.
With the proposed fix it typically takes one second for attach and max five
to six seconds

govint:
The watcher code doesn't implement any timeouts on its own which is why we have a timeout explicitly when looping for events. And that timeout is what always runs. Setting to a higher value causes all attach requests to run for that much time.

msterin:
>  the event handling logic waits for that much time for every single attach meaning every container always starts with that delay.

that is not how it is supposed to work, so it seems broken. I do not think we should do a patch before understanding what is broken . We do not need sleep. We need timeout

govint:
Watcher is working fine, but the pvscsi controller add takes minimum 1 second and at least a max of 5  seconds.

The wather pkg (inotify and now fsnotify) doesn't implement a timeout based watch logic. A watcher is instantiated and then the logic is to loop waiting for events. This wait has to have an external timer as the watcher doesn't implement one. And the timer is to have a sleep. The way that works is a) check for the file events, b) if none sleep (timeout) c) break from watcher loop.

Above logic works is ok to handle device attach delays, except that (as mentioned ) pvscsi ctlr add is anywhere from 1 - 5 seconds. So, try a small timeout and then try a long timeout, typically the small timeout is what happens.

msterin:
We are not waiting for adding pvscsi controller, we are waiting for Linux kernel to discover the device after it was.succcessfully added to the VM, aren't we ? Are you saying it takes Linux kernel 1-5 sec to discover a new scsci adapter with a device ?

Also, what you describe is not iNotify but a polling mechanism. iNotify package we are using seems to have been depreciated and we need to switch to gopkg.in/fsnotify.V0 or later

govint:
Have seen the current fsnotify pkg and the API there isn't using timeouts either. Do we need to still update the pkg? Correct, the delay is for the in Linux kernel, and thats mostly 1 second but not always.

govint:
Another thing with using these watchers, we can watch a directory and any change there alone is recorded. We watch only the /dev/disk/by-path as we don't know which host is being used for the disk attach. So when we have a controller and a disk added we only can get a notification of the host (controller) add, nothing in it (the disk). A sleep and a retry is more than sufficient for our purpose.

msterin:
> Have seen the current fsnotify pkg and the API there isn't using timeouts eithe

there is no need for explicity timeout on an interface, golang allows simple and generic timeout handling 
e.g. https://blog.golang.org/go-concurrency-patterns-timing-out-and

> Do we need to still update the pkg? 
We need to have something that does not require us to do polling and returns when dev path really shows up.  If we need to kick off scsci rescan , we should also do it. but we should **not** block on polling intervals.

govint:
Here  we see the Linux kernel logs taking around 5 seconds to detect the pvscsi controller and the one disk (request that attached both to the VM).

Mar 20 01:12:47 hte-1s-eng-dhcp98 kernel: [327843.288814] pciehp 0000:00:17.0:pcie04: Button pressed on Slot(224)
Mar 20 01:12:47 hte-1s-eng-dhcp98 kernel: [327843.291040] pciehp 0000:00:17.0:pcie04: Card present on Slot(224)
Mar 20 01:12:47 hte-1s-eng-dhcp98 kernel: [327843.308102] pciehp 0000:00:17.0:pcie04: PCI slot #224 - powering on due to button press.
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.430585] pci 0000:13:00.0: [15ad:07c0] type 00 class 0x010700
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.430727] pci 0000:13:00.0: reg 0x10: [io  0x0000-0x0007]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.430868] pci 0000:13:00.0: reg 0x14: [mem 0x00000000-0x00007fff 64bit]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.431109] pci 0000:13:00.0: reg 0x30: [mem 0x00000000-0x0000ffff pref]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.431594] pci 0000:13:00.0: PME# supported from D0 D3hot D3cold
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434452] pci 0000:13:00.0: disabling ASPM on pre-1.1 PCIe device.  You can enable it with 'pcie_aspm=force'
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434606] pci 0000:13:00.0: BAR 6: assigned [mem 0xeba00000-0xeba0ffff pref]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434611] pci 0000:13:00.0: BAR 1: assigned [mem 0xfd300000-0xfd307fff 64bit]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434705] pci 0000:13:00.0: BAR 0: assigned [io  0x6000-0x6007]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434734] pcieport 0000:00:17.0: PCI bridge to [bus 13]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434751] pcieport 0000:00:17.0:   bridge window [io  0x6000-0x6fff]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434791] pcieport 0000:00:17.0:   bridge window [mem 0xfd300000-0xfd3fffff]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.434819] pcieport 0000:00:17.0:   bridge window [mem 0xeba00000-0xebafffff 64bit pref]
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.435264] vmw_pvscsi 0000:13:00.0: enabling device (0000 -> 0003)
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.437781] vmw_pvscsi: using 64bit dma
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.442198] vmw_pvscsi 0000:13:00.0: vmw_pvscsi: host->max_id: 16
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.443236] vmw_pvscsi 0000:13:00.0: irq 76 for MSI/MSI-X
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.443368] vmw_pvscsi: using MSI-X
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.443501] scsi12 : VMware PVSCSI storage adapter rev 2, req/cmp/msg rings: 8/8/1 pages, cmd_per_lun=64
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.444217] vmw_pvscsi 0000:13:00.0: VMware PVSCSI rev 2 host #12
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.444402] scsi 12:0:0:0: Direct-Access     VMware   Virtual disk     1.0  PQ: 0 ANSI: 2
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446116] sd 12:0:0:0: [sdb] 204800 512-byte logical blocks: (104 MB/100 MiB)
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446142] sd 12:0:0:0: [sdb] Write Protect is off
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446146] sd 12:0:0:0: [sdb] Mode Sense: 61 00 00 00
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446170] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446173] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446739] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.446742] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.447066]  sdb: unknown partition table
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.447216] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.447220] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.447243] sd 12:0:0:0: [sdb] Attached SCSI disk

msterin:
the issue is not with 5 sec for controller + dick discovery. The issue with OTHER cases that should not be waiting 5 seconds. 

Also. looking at the log, I see only 1 operation (slot power on) taking 5 sec, the rest is fine. 
```
Mar 20 01:12:47 hte-1s-eng-dhcp98 kernel: [327843.308102] pciehp 0000:00:17.0:pcie04: PCI slot #224 - powering on due to button press.
Mar 20 01:12:52 hte-1s-eng-dhcp98 kernel: [327848.430585] pci 0000:13:00.0: [15ad:07c0] type 00 class 0x010700
``` 
I am wondering what could be taking that long... (still, the main issue is that OTHER ops should not be impacted by the slowest one)

govint:
Only the first attach waits for 5 seconds the subsequent attaches complete
in a second or less.

govint:
If the pvscsi controller is already attached the attach time is <= 1 second.

System logs:
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.972576] vmw_pvscsi: msg type: 0x0 - MSG RING: 4/3 (5)
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.972581] vmw_pvscsi: msg: device added at scsi0:0:0
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.972725] scsi 12:0:0:0: Direct-Access     VMware   Virtual disk     1.0  PQ: 0 ANSI: 2
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.972976] sd 12:0:0:0: [sdb] 204800 512-byte logical blocks: (104 MB/100 MiB)
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.973010] sd 12:0:0:0: [sdb] Write Protect is off
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.973013] sd 12:0:0:0: [sdb] Mode Sense: 61 00 00 00
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.973046] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.973049] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.974964] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.974969] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.975186]  sdb: unknown partition table
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.975358] sd 12:0:0:0: [sdb] Cache data unavailable
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.975362] sd 12:0:0:0: [sdb] Assuming drive cache: write through
Mar 20 01:27:23 hte-1s-eng-dhcp98 kernel: [328719.975389] sd 12:0:0:0: [sdb] Attached SCSI disk


Plugin logs

2017-03-20 01:27:23.211267759 -0700 PDT [INFO] Attaching volume and creating filesystem name=yyyy7 fstype=ext4
2017-03-20 01:27:23.789967093 -0700 PDT [INFO] Starting dev wait name=yyyy7 fstype=ext4
2017-03-20 01:27:24.790304311 -0700 PDT [WARNING] Exceeded timeout while waiting for device attach to complete timeout=1s device="/dev/disk/by-path/pci-0000:13:00.0-scsi-0:0:0:0"


About a second to complete the dev wait

The five seconds is for just the first time that a pvscsi controller is added

msterin:
> About a second to complete the dev wait
> The five seconds is for just the first time that a pvscsi controller is added

they I do not understand the reasons for 1 sec timeout (and honestly, for 5 sec timeout as well). 
Just out of paranoia we can add 15 sec and hope it will never get hit (and fail the op if it does) but the resr seems to be already event driven. 
As a side note, I am surprised it's that slow in the kernel . any way to accelerate it (e.g. kick of scsi discovery ?)

govint:
Earlier had added a function to do start discovery for Photon OS (with
Photon controller). That needed more delays than what we don with Vsphere.

Unfortunately the *notify pkgs don't have any timeout in themselves

msterin:
> Unfortunately the *notify pkgs don't have any timeout in themselves. 

there is no need as there is a generic mechanism in golang for timeouts

> The select-case construct typically (on most attach-waits) ends up doing the sleep(). A large timeout results in sleeping for that much time on all waits.

Then  something is wrong as you just said than in majority of cases the kernel discovers the new devices much faster than 1 seconds. In this cse timeout should not matter since the wacher gets woken up by this event. 

Per your explanations, we should then have one very large (20sec+) timeout make sure we recover from (unlikely) case of kernle faiing to power on the new controller, and all other cases we should get event and not hit the timeout.

govint:
The proposed fix seems fine in that case, it's the original code plus a max
five second timeout in case the default doesn't help. Say best in the case
where the device is not discovered the create is failed.

pdhamdhere:
>The proposed fix seems fine in that case, i

No, if I understand correctly changing devWaitTimeout to 10secs should resolve. If decide is discovered in less than 10Secs then iNotify will break otherwise will timeout after 10Secs. Right?

govint:
Can we make this change to 10 seconds in the existing code and try it out? I've done that. The seleect-case typically doesn't ever find the device (or a file system event) when its called and around 99% of the time hits the time.sleep() case and does the sleep for the specified duration. Changing to ten seconds will cause every single attach to wait for ten seconds even if the device was identified in a some hundred milliseconds (when the system runs its periodic discovery loop).

msterin:
As we already discussed above  this is NOT an event driven behavior and this is not what we want. 
This is also not what iNotify is doing. If you see this behavior, then our code using inotify is wrong and needs to be fixed. 

Timeout 10 sec should be ALMOST NEVER Hit, all regular ops should complete within 1 sec (as you've mentioned above) or 5 sec for pvsci adapter (as you've mentioned above) - basically, in time it takes LInux kernel to find  a device and expose it in the /dev/disk/by*

govint:
"Timeout 10 sec should be ALMOST NEVER Hit, all regular ops should complete within 1 sec (as you've mentioned above) or 5 sec for pvsci adapter (as you've mentioned above) - basically, in time it takes LInux kernel to find a device and expose it in the /dev/disk/by*"

Ok, so this change is good then - it takes 1s or max 5s.

msterin:
>Ok, so this change is good then - it takes 1s or max 5s.

Please do not merge. The issue may have been resolved with #1058

pdhamdhere:
Fixed with #1058

PR #1052, opened 2017-03-20
removed .pyc files [trivial]

msterin:


msterin:
These are python compile files wrongly checked in   by some old PR

Issue #1080, opened 2017-03-24
Close the watcher instance in the plugin after event is received.


govint:
When running a stress test (500 volumes created concurrently between two docker hosts and within a single one) the plugin logs show below error in creating the watcher instance.

2017-03-24 06:40:04.419947253 +0000 UTC [INFO] Attaching volume and creating filesystem name=volTestP1349 fstype=ext4
2017-03-24 06:40:04.420212472 +0000 UTC [ERROR] Failed to create watcher. Skipping inotify Name=volTestP1349

lsof shows the plugin has a lot of open files via inotify
docker-vo  4856                  root    6u     unix 0x0000000000000000      0t0     985592 /run/docker/plugins/vsphere.sock type=STREAM
docker-vo  4856                  root    7r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root    8r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root    9r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   10r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   11r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   12r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   13r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   14r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   15r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   16r  a_inode               0,10        0         14 inotify
docker-vo  4856                  root   17r  a_inode               0,10        0         14 inotify
.....

watcher.Close() needs to be called to ensure cleanup is done per watcher instance.

govint:
Closed via #1081

PR #1081, opened 2017-03-24
Close watcher after event is received.

govint:
Fixed and stress test shows no watcher creation errors and lsof has no fiiles opened for inotify.


govint:
Fixes #1080

PR #1082, opened 2017-03-24
[No review] Print extended debug stack for plugin panic

govint:
This PR is to have the plugin print all possible stack frames when a panic happens, seen with the concurrency tests in CI on Photon OS.


govint:
Running the plugin with efence=1,gcrescanstacks=1,scavenge=1 to see if the corruption repros.

govint:
Closing.

PR #1084, opened 2017-03-24
Make ESX Ops idempotent 

### LINKED ISSUE #1076: Make ESX OPs resilient to retries
With #1049, Plugin retries operation in case of VMCI errors. However, operation on ESX may have been successful on first attempt. We need to gracefully handle "retry" requests on ESX. For e.g.

For Remove requests, 2nd attempt to removeVMDK fails with "File not found" sicne it is already deleted on first attempt. We should simply return success.

Similarly, we should check CREATE: return success if VMDK already exists, ATTACH/DETACH: return success if vmdk is already attached/detached.

See https://ci.vmware.run/vmware/docker-volume-vsphere/1797 where remove is retried due to "Bad Magic" error and it fails with "file not found"

pdhamdhere:
Make ESX Ops idempotent to tolerate plugin retries in case of VMCI communication errors.

Testing Done: Changed Plugin code to retry every operation by injecting failure on 1st attempt. And then tried docker volume create/remove and docker run.

Fixes #1076


shuklanirdesh82:
I guess few tests needs to be fixed as well.

esx_service/vmdk_ops.py
```diff
@@ -1318,10 +1325,11 @@ def disk_detach(vmdk_path, vm):
...
        # and docker will insist to sending "unmount/detach" which also fails.
-       msg = "*** Detach failed: disk={0} not found. VM={1}".format(
-       vmdk_path, vm.config.uuid)
-       logging.warning(msg)
-       return err(msg)
+       # Or Plugin retrying operation due to socket errors #1076
+       # Return success since disk is anyway not attached
+       logging.warning("*** Detach disk={0} not found. VM={1}".format(
+                       vmdk_path, vm.config.uuid))
+       return None
```
shuklanirdesh82: this is where some tests failed. Please add affected tests as part of this PR.

pdhamdhere: Fixed in latest commit.

esx_service/vmdk_ops.py
```diff
@@ -166,7 +166,9 @@ def createVMDK(vmdk_path, vm_name, vol_name, opts={}, vm_uuid=None, tenant_uuid=
     logging.info("*** createVMDK: %s opts = %s", vmdk_path, opts)
 
     if os.path.isfile(vmdk_path):
-        return err("File %s already exists" % vmdk_path)
+        # We are mostly here due to race or Plugin VMCI retry #1076
+        logging.warning("File %s already exists", vmdk_path)
+        return None
```
govint: Lets at least check that the volume opts in the KV match whats provided. Two create requests with say different VM sizes should be caught, unless the two volumes would be having the exact same properties.

esx_service/vmdk_ops.py
```diff
@@ -1318,10 +1325,11 @@ def disk_detach(vmdk_path, vm):
     if not device:
        # Could happen if the disk attached to a different VM - attach fails
        # and docker will insist to sending "unmount/detach" which also fails.
```
govint: Not this change, Docker doesn't send unmount if mount fails anymore. But this service doesn't haven't to assume such behavior. Possibly remove all mention docker in the service code.

esx_service/vmdk_ops.py
```diff
@@ -1318,10 +1325,11 @@ def disk_detach(vmdk_path, vm):
...
-       msg = "*** Detach failed: disk={0} not found. VM={1}".format(
-       vmdk_path, vm.config.uuid)
-       logging.warning(msg)
-       return err(msg)
+       # Or Plugin retrying operation due to socket errors #1076
```
govint: Just as a coding convention can skip mentioning PR number in code.

esx_service/vmdk_ops.py
govint: Test is also failing that the volume is in use when remove is issued without a detach. That path is not addressable as we can;t auto detach the volume. Or can we?

esx_service/vmdk_ops_test.py
```diff
@@ -245,9 +245,9 @@ def testPolicy(self):
                                             volume_kv.DISK_ALLOCATION_FORMAT: unit[3]})
             self.assertEqual(err == None, unit[2], err)
 
-            # clean up should fail if the created should have failed.
+            # clean up would succeed with #1084.
```
govint: As mentioned PR/issue numbers can be skipped in comments.

msterin: we actually do refer to github issues and PRs in different place, so while it may be good or bad, it is certainly consistent (there are 30 cases like that in 17 files in our repo, including ./vendor)

esx_service/vmdk_ops.py
msterin: we actually do refer to github issues and PRs in different places, so while it may be good or bad, it is certainly consistent (there are 30 cases like that in 17 files in our repo, including ./vendor). BTW , what/where  is this coding convention asking to skip PR numbers ?

esx_service/vmdk_ops.py
msterin: Good point. Actually if they are different we may want to reject the request. It's not really idempotent if requests are different :-)

msterin: I think we should go ahead and merge (after fixing the tests) and open an issue to fix create with diff.params - it is super rear now (actually, never)

esx_service/vmdk_ops.py
pdhamdhere: Failing tests are fixed. Auto-detach is outside scope of this change.

esx_service/vmdk_ops.py
pdhamdhere: Filed a new issue #1089. Yes, it's a rare case where Volume Create is requested from 2 separate nodes at the same time. Addressing it would require some code cleanup since some default options are set in KV (diskformat, fstype) but some are not set (e.g. size).
@govint Are you okay addressing this in separate issue?

esx_service/vmdk_ops.py
pdhamdhere: I added PR# to give future reader some context why this special cases can occur. Alternatively, I can update function header comment without PR.

Issue #1092, opened 2017-03-27
[Tenant] Volume creation fails after renaming the default tenant.


ashahi1:
Steps:
```
1. Make sure you only see _DEFAULT tenant
2. Create a volume.
3. Rename the _DEFAULT tenant.
4. Again try creating a volume. 
    Operation failed -  Error response from daemon: create TestVol2: VolumeDriver.Create: VM    ubuntu-VM0.3 does not belong to any vm-group
```

-----

Steps and their output are as follows:

1. Only default tenant is present
```
[root@sc-rdops-vm16-dhcp-235-79:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vm-group ls
Uuid                                  Name      Description                 Default_datastore  VM_list
------------------------------------  --------  --------------------------  -----------------  -------
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default vm-group
```

2. Created a volume
```
root@sc-rdops-vm02-dhcp-52-237:~# docker volume create --driver=vsphere --name=TestVol1 -o size=400mb
TestVol1
root@sc-rdops-vm02-dhcp-52-237:~# docker volume ls
DRIVER              VOLUME NAME
vsphere             TestVol1@sharedVmfs-0
root@sc-rdops-vm02-dhcp-52-237:~#


[root@sc-rdops-vm16-dhcp-235-79:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume    Datastore     VM-Group  Capacity  Used  Filesystem  Policy  Disk Format  Attached-to  Access      Attach-as               Created By    Created Date
--------  ------------  --------  --------  ----  ----------  ------  -----------  -----------  ----------  ----------------------  ------------  ------------------------
TestVol1  sharedVmfs-0  _DEFAULT  400MB     23MB  ext4        N/A     thin         detached     read-write  independent_persistent  ubuntu-VM0.3  Mon Mar 27 01:58:09 2017

[root@sc-rdops-vm16-dhcp-235-79:~]

```

3.  Renamed the _DEFAULT tenant
```
[root@sc-rdops-vm16-dhcp-235-79:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vm-group update --name=_DEFAULT --description="Renaming default tenant to vm-group1" --new-name=new-vm-group1
vm-group modify succeeded
[...]

PR #1099, opened 2017-03-28
Disallow renaming the default tenant.

govint:
Verified as below,
/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vm-group update --name=_DEFAULT --descr
iption="Renaming default tenant to vm-group1" --new-name=new-vm-group1
Vm-group name _DEFAULT is invalid, only rename of vm-groups other than _DEFAULT is allowed

### DIFF
--- a/esx_service/utils/auth_api.py
+++ b/esx_service/utils/auth_api.py
@@ -29,6 +29,7 @@ import re
 # regex for valid tenant name
 VALID_TENANT_NAME_REGEXP = "[a-zA-Z0-9_][a-zA-Z0-9_.-]*"
 VALID_TENANT_NAMES = 'rename of vm-groups other than _DEFAULT'
+
 global valid_tenant_name_reg
 valid_tenant_name_reg = re.compile("^" + VALID_TENANT_NAME_REGEXP + "$")
 
@@ -348,10 +349,6 @@ def _tenant_create(name, description="", vm_list=None, privileges=None):
 
 def _tenant_update(name, new_name=None, description=None, default_datastore=None):
     """ API to update a tenant """
-    if name == auth_data_const.DEFAULT_TENANT:
-        error_info = error_code.generate_error_info(ErrorCode.TENANT_NAME_INVALID, name, VALID_TENANT_NAMES)
-        return error_info
-
     error_info, tenant = get_tenant_from_db(name)
     if error_info:
         return error_info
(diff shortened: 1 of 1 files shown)


govint:
Fixes #1092

esx_service/utils/auth_api.py
```diff
@@ -347,6 +348,10 @@ def _tenant_create(name, description="", vm_list=None, privileges=None):
 
 def _tenant_update(name, new_name=None, description=None, default_datastore=None):
     """ API to update a tenant """
+    if name == auth_data_const.DEFAULT_TENANT:
+        error_info = error_code.generate_error_info(ErrorCode.TENANT_NAME_INVALID, name, VALID_TENANT_NAMES)
```
shuklanirdesh82: If I understand this code path correctly, this change stops updating description, default datastore etc for the ```_DEFAULT``` tenant where as our intention is only to stop renaming ```_DEFAULT``` tenant, please correct me otherwise.

esx_service/utils/auth_api.py
```diff
@@ -28,6 +28,7 @@
 
 # regex for valid tenant name
 VALID_TENANT_NAME_REGEXP = "[a-zA-Z0-9_][a-zA-Z0-9_.-]*"
+VALID_TENANT_NAMES = 'rename of vm-groups other than _DEFAULT'
```
shuklanirdesh82: nit: can you please improve the error message?

esx_service/utils/auth_api.py
pshahzeb: Correct. I think this check should be moved after the if check on line 368

govint:
Will move the check. Can you say what change needed to the error message.
It is part of another error message defined for invalid group name

esx_service/utils/auth_api.py
govint: The error msg comes out like this,
Vm-group name _DEFAULT is invalid, only _rename of vm-groups other than _DEFAULT_ is allowed

shuklanirdesh82: I would like to suggest something like ```renaming of _DEFAULT vm-group is not allowed```

govint: In the message the highlited parts are fixed for the error string define for TENANT_NAME_INVALID
## Vm-group name _DEFAULT is invalid, only _rename of vm-groups other than DEFAULT ## is allowed

shuklanirdesh82:
@govint Proposed changes looks OK to me.

It would be great if you can add unit test to cover proposed change.

esx_service/utils/auth_api.py
shuklanirdesh82: Thanks for the clarification.

PR #1105, opened 2017-03-28
Add kubernetes examples for blog

abrarshivani:
This PR adds yaml files that are needed for VMware blog. Reader can directly use the links for these yamls and create Kubernetes resources. This files cannot go in github.com/vmware/kubernetes since if it is rebase with upstream or someone creates branch with same name the files will be removed.

cc @pdhamdhere @tusharnt @BaluDontu @divyenpatel @msterin


shuklanirdesh82:
How about locking the branch and using its reference rather than merging to vDVS master? It leads to confusion in the future keep adding non vDVS stuff to ```vDVS/misc``` folder.

misc/kubernetes-examples/blog1/mysql-deployment-vsphere.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: v1
```
shuklanirdesh82: How about locking the branch and using its reference rather than merging to vDVS master? It leads to confusion in the future keep adding non vDVS stuff to vDVS/misc folder.

abrarshivani: @shuklanirdesh82 Thanks for mentioning about this. I have created branch on vmware/kubernetes and protected the branch.

abrarshivani:
I moved examples to vmware/kubernetes therefore we don't need this change to go in this repo. Closing this PR.

misc/kubernetes-examples/blog1/mysql-deployment-vsphere.yaml
shuklanirdesh82: Thanks @abrarshivani !

PR #1108, opened 2017-03-29
Fix for slow "vmdkops_admin vm-group ls"

### LINKED ISSUE #1104: Admin cli vm-group ls takes extremely long time to list all the tenants.
Issue: Admin cli vm-group ls takes extremely long time to list all the tenants after VMs were added to the tenants. 

For 100 tenants and 2-3 VMs per tenant, ```admin cli vm-group ls``` took 40 mins approx.

---------------------

Test setup, test steps and time duration for each operation is as follows:

Setup:
```
One ESX - 312 VMs (80 VMs are powered-on and rest are powered-off) 
Docker vm (photon) is on sharedVmfs-0
```

Test Steps:

```
1. Created 500 Volumes of 100 mb each sequentially. 
2. After all volumes are created, did docker volume ls  and admin cli volume ls.
3. Created 100 tenants sequentially.
4. After all tenants are created, did admin cli vm-group ls.
5. Added 2-3 VMs per tenant.
6. After VMs were added to each tenants, did admin cli vm-group ls.
7. Added access rule for each tenant 
8. After access rules were added for each tenant, did admin cli vm-group ls.

```

Time duration for each operation:
```
1. One volume creation took between 2-3 seconds
2. 500 volumes creation took ~27 mins (1656 seconds)
3. Docker cli volume ls ~2 seconds
4. Admin cli volume ls ~30 seconds
5. One tenant creation took ~2 seconds
6. 100 tenants creation took ~300 seconds
7. Admin cli vm-group ls ~3 seconds
8. Adding one vm to each tenant took ~12 seconds
9. After adding 2-3 VMs per tenant, Admin cli vm-group ls took 45 mins
10.Adding access rule to each tenant took ~2 seconds
11. After adding access rules per tenant, Admin cli vm-group ls took 42 mins

```

Product Logs:
[docker-volume-vsphere.txt](https://github.com/vmware/docker-volume-vsphere/files/876809/docker-volume-vsphere.txt)
[vmdk_ops.log.txt](https://github.com/vmware/docker-volume-vsphere/files/876810/vmdk_ops.log.txt)
[vmdk_ops.txt](https://github.com/vmware/docker-volume-vsphere/files/876811/vmdk_ops.txt)

Test Log:
[Tenant_Run2.txt](https://github.com/vmware/docker-volume-vsphere/files/876813/Tenant_Run2.txt)

pdhamdhere:
Use SearchIndex and Find* methods for VM-Name <-> UUID conversion.

**Testing Done**:
Manually tested on Anup's setup. 
Without this change:  3+ minutes
With this change: 5 Secs!

Fixes #1104 

This will also improve CI tests.


esx_service/utils/vmdk_utils.py
```diff
@@ -274,26 +277,25 @@ def get_vm_uuid_by_name(vm_name):
     """ Returns vm_uuid for given vm_name, or None """
     si = vmdk_ops.get_si()
     try:
-        vm = [d for d in si.content.rootFolder.childEntity[0].vmFolder.childEntity if d.config.name == vm_name]
-        return vm[0].config.uuid
+        vm = FindChild(GetVmFolder(), vm_name)
```
govint: Is GetVmFolder() defined in the ESX service, searched but not able to find the description for this. Which VM folder does this return data for?

esx_service/utils/vmdk_utils.py
```diff
@@ -274,26 +277,25 @@ def get_vm_uuid_by_name(vm_name):
...
     except:
         return None
 
 def get_vm_name_by_uuid(vm_uuid):
     """ Returns vm_name for given vm_uuid, or None """
     si = vmdk_ops.get_si()
     try:
-        vm = [d for d in si.content.rootFolder.childEntity[0].vmFolder.childEntity if d.config.uuid == vm_uuid]
-        return vm[0].config.name
+        return vmdk_ops.vm_uuid2name(vm_uuid)
```
govint: May be refactoring issue, but get VM UUID from name and get name from UUID should be in the vmdk_utils.py code. In this file we have the get UUID from name and the reverse is in vmdk_ops.py. Instead keep both variants here in vmdk_utils.py.

Not this issue but code is mixed.

From vmdk_ops,.py:
# Return VM managed object, reconnect if needed. Throws if fails twice.
def findVmByUuid(vm_uuid):
    si = get_si()
    vm = si.content.searchIndex.FindByUuid(None, vm_uuid, True, False)
    return vm

def vm_uuid2name(vm_uuid):
    vm = findVmByUuid(vm_uuid)
    if not vm or not vm.config:
        return None
    return vm.config.name

esx_service/utils/vmdk_utils.py
pdhamdhere: No, it's from `pyVim.invt`. See L33.
It's Host Agent VM Folder;
```
>>> c = GetVmFolder()
>>> print c
'vim.Folder:ha-folder-vm'
```

esx_service/utils/vmdk_utils.py
pdhamdhere: Good catch! Yes, I noticed that too. Rather code in vmdk_ops.py triggered light bulbs! I wanted to separate bug fix from code cleanup so didnt touch vm_uuid2name

PR #1098, opened 2017-03-28
[No review] DEbug mem alloc.

govint:

PR #1117, opened 2017-03-30
[No review] Use malloc vs. calloc for resp. buffer and run all concurrent tests.

govint:
PR #1098 is gettng green runs with CI after reverting the response buffer to use malloc() vs. calloc(). Creating a fresh source tree and a) restored to use malloc in vmci_client. c and b) enabled all of the concurrent tests to verify if the panic was caused by the calloc'ed response buffers in C.


govint:
Issue debugged, closing PR.

Issue #1109, opened 2017-03-29
Add a unit test case to verify that the default vm group name (_DEFAULT) can't be changed.


govint:
#1092 ensures the default vm-group isn't rename'able.

PR #1119, opened 2017-03-30
Added rename tests for default vm-group.

govint:
Added tests to verify default vm-group can't be renamed but description can be changed.

### DIFF
--- a/esx_service/cli/vmdkops_admin_test.py
+++ b/esx_service/cli/vmdkops_admin_test.py
@@ -29,6 +29,7 @@ import log_config
 import logging
 import convert
 import error_code
+import auth_data_const
 
 # Number of expected columns in ADMIN_CLI ls
 EXPECTED_COLUMN_COUNT = 13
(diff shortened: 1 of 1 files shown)


govint:
Fixes #1109

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -603,6 +612,11 @@ def test_tenant(self):
                                               new_name=self.tenant1_new_name)
         self.assertEqual(None, error_info)
 
+	# verify default tenant can't be renamed
+        error_info  = auth_api._tenant_update(name=DEFAULT,
+                                              new_name=self.tenant1_new_name)
+        self.assertNotEqual(None, error_info)
```
pshahzeb: nit: I think a better way would be to match the error code

esx_service/cli/vmdkops_admin_test.py
```diff
@@ -44,6 +44,9 @@
 
 VMGROUP = 'vm-group'
 
+# Default vm-group
+DEFAULT = '_DEFAULT'
```
lipingxue: Nit: can use "DEFAULT_TENANT" defined in auth_data_const.py.

esx_service/cli/vmdkops_admin_test.py
govint: Agree, but the error info is a string and changing the string in the code wll require changing n the test as well.

esx_service/cli/vmdkops_admin_test.py
govint: Done.

Issue #1127, opened 2017-03-31
[Multi-ESX] VM's belonging to the same tenant are not visible across ESXs


ashahi1:
Steps:
1. Initialized MultiNode config db on both the ESXs.
2. Created a tenant from ESX-1.
3. Added vm to the tenant from ESX-1.
4. Ran ```admin cli vm-group ls``` from ESX-1 and ESX-2.
VM_list field is empty for ```admin cli vm-group ls``` from ESX-2.

-------------------

Steps and their output are as follows:

1. Ran ```config init``` on both the ESXs.

```
ESX:1

[root@sc-rdops-vm06-dhcp-163-208:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py config init --datastore=sharedVmfs-0
Creating new DB at /vmfs/volumes/sharedVmfs-0/dockvols/vmdkops_config.db
Creating a symlink to /vmfs/volumes/sharedVmfs-0/dockvols/vmdkops_config.db at /etc/vmware/vmdkops/auth-db

ESX-2: 
[root@sc-rdops-vm06-dhcp-161-190:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py config init --datastore=sharedVmfs-0
Creating a symlink to /vmfs/volumes/sharedVmfs-0/dockvols/vmdkops_config.db at /etc/vmware/vmdkops/auth-db

```

2. Config information from both the ESXs:
``` 
ESX-1:
[root@sc-rdops-vm06-dhcp-163-208:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
=== Service:
Version: 0.12.9e2d6ad-0.0.1
Status: Running
Pid: 71898
Port: 1019
LogConfigFile: /etc/vmware/vmdkops/log_config.json
LogFile: /var/log/vmware/vmdk_ops.log
LogLevel: INFO
=== Authorization Config DB:
DB_LocalPath: /etc/vmware/vmdkops/auth-db
DB_Mode: MultiNode (local symlink pointing to shared DB)
DB_SharedLocation: /vmfs/volumes/sharedVmfs-0/dockvols/vmdkops_config.db
[root@sc-rdops-vm06-dhcp-163-208:~]

ESX-2:
[root@sc-rdops-vm06-dhcp-161-190:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status
=== Service:
Version: 0.12.9e2d6ad-0.0.1
[...]

PR #1151, opened 2017-04-06
Persisting VM name in config DB

### LINKED ISSUE #1130: Add VM_Name in config DB
Current Admin CLI logic is to persist the VM uuid and query VM name on demand. With shared multi-tenancy, user can run `vmdkops_admin vmgroup vm ls` and command must display VMs running on another ESX host as well.

1. VM1 belongs to ESX1, VM2 belongs to ESX2. ESX1 and ESX2 are configured to a shared datastore.
2. Admin creates T1 on ESX1 and adds VM1 to T1
3. Admin runs “vmdkops_admin tenant vm ls” on ESX2 – BROKEN!

Proposed Solution:
Add VM_Name in DB. If FindVMByUuid() fails then use VM_Name in DB which may be stale in rare cases. Once we have property collector, this issue will go away.

Fix should be backward compatible i.e. for existing entries in DB with no VM_Name, it should log UUID.

Making it P0 since it is must-fix for cross-esx tenant configuration.

CC/ @shaominchen @lipingxue

pshahzeb:
1. Persisting VM name along with VM id and tenant id in config db
when vm becomes a part of tenant. Updating the db manager api to accept
list of tuple of vm id and name instead of just vm id

2. tenant object contains the list of tuple of vm id and names of VMs that
are part of the tenant. Doing this saves trip to db while verifying params
for tenant operations

3. Bumping the db version to 1.2.
Handling db schema upgrade from previous db version for vms table.
Leaving the vm name None in db for VMs for which VM name couldn't be
resolved during upgrade. Handling VM names which are None gracefully when
printing them

Testing:
Updating related test cases for auth_data manager, vmdkops and admin cli

After fix:
1. Normal workflow
```
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name tenant1 --vm-list ubuntu-VM0.0,ubuntu-VM1.0
vmgroup 'tenant1' is created.  Do not forget to run 'vmgroup vm add' and 'vmgroup access add' commands to enable access control.
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm ls --name tenant1
Uuid                                  Name
------------------------------------  ------------
564de69f-7590-f463-8b55-989dab583f35  ubuntu-VM0.0
564dfcb3-78c6-bcd7-bf86-497c4c306615  ubuntu-VM1.0
```

2. Create tenant with vms when db version 1.1. Retain db and deploy newer vmdkops with db version 1.2 (upgrade scenario). Name is populated during db connect when upgrade is required is detected.
```
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name old_schema_tenant --vm-list ubuntu-VM0.0,ubuntu-VM1.0
vmgroup 'old_schema_tenant' is created.  Do not forget to run 'vmgroup vm add' and 'vmgroup access add' commands to enable access control.
```
After deploying newer vmdkops and retaining the authdb
```
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm ls --name old_schema_tenant
Uuid                                  Name
------------------------------------  ------------
564de69f-7590-f463-8b55-989dab583f35  ubuntu-VM0.0
564dfcb3-78c6-bcd7-bf86-497c4c306615  ubuntu-VM1.0
```

3. Create tenant with vms when db version 1.1. Retain db and deploy newer vmdkops with db version 1.2 (upgrade scenario). Migrating one VM out of ESX. Name won't be populated during upgrade because name won't be resolved.

```
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name old_schema_tenant --vm-list ubuntu-VM0.0,ubuntu-VM1.0
vmgroup 'old_schema_tenant' is created.  Do not forget to run 'vmgroup vm add' and 'vmgroup access add' commands to enable access control.
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm ls --name old_schema_tenant
Uuid                                  Name
------------------------------------  ------------
564de69f-7590-f463-8b55-989dab583f35  ubuntu-VM0.0
564dfcb3-78c6-bcd7-bf86-497c4c306615  ubuntu-VM1.0
```
Migrating ubuntu-VM1.0 out of ESX. Upgrading to newer vmdkops which uses db version 1.2
```
[root@sc2-rdops-vm02-dhcp-92-255:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm ls --name old_schema_tenant
Uuid                                  Name
------------------------------------  ------------
564de69f-7590-f463-8b55-989dab583f35  ubuntu-VM0.0
564dfcb3-78c6-bcd7-bf86-497c4c306615  N/A
```

4. Using VDVS 0.11 (db version 1.0) and creating tenants
```
[root@sc2-rdops-vm02-dhcp-74-64:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant create --name tenant_0.11 --vm-list ubuntu-VM0.1,ubuntu-VM1.0
tenant create succeeded
[root@sc2-rdops-vm02-dhcp-74-64:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py tenant vm ls --name tenant_0.11
Uuid                                  Name
------------------------------------  ------------
564dad0c-536c-955c-6147-dec1ef446041  ubuntu-VM0.1
564dfcb3-78c6-bcd7-bf86-497c4c306615  ubuntu-VM1.0
```
deploy newer vmdkops with db version 1.2 (upgrade scenario)
```
[root@sc2-rdops-vm02-dhcp-74-64:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm ls --name tenant_0.11
Internal Error(DB upgrade error at /etc/vmware/vmdkops/auth-db: Upgrade is not supported for auth-db schema version 1.0 to 1.2. Refer to vDVS release versions)
```

5. Multi ESX: Shared DB mode:
Create a named tenant and add vm from each of the two esx to the same tenant
```
ESX1:
/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vm-group create --name=vmgroup1
vmgroup 'vmgroup1' is created. Do not forget to run 'vmgroup vm add' and 'vmgroup access add' to enable access control.

[root@sc-rdops-vm05-dhcp-147-24:~] usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm add --name=vmgroup1 --vm-list=ubuntu-VM0.1
vmgroup vm add succeeded

ESX2:
[root@sc-rdops-vm05-dhcp-138-173:~] usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm add --name=vmgroup1 --vm-list=ubuntu-VM0.0
vmgroup vm add succeeded
```

vmgroup ls output:
```
ESX1:
[root@sc-rdops-vm05-dhcp-147-24:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
------------------------------------  --------  -------------------------  -----------------  -------------------------
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default vmgroup
c531f0cb-d4b3-4a14-b98b-273f4aafa4a3  vmgroup1                                                ubuntu-VM0.1,ubuntu-VM0.0

ESX2:
[root@sc-rdops-vm05-dhcp-138-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
------------------------------------  --------  -------------------------  -----------------  -------------------------
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default vmgroup
c531f0cb-d4b3-4a14-b98b-273f4aafa4a3  vmgroup1                                                ubuntu-VM0.1,ubuntu-VM0.0
```
Fixes #1130 #1127


esx_service/cli/vmdkops_admin.py
```diff
@@ -807,20 +807,23 @@ def tenant_ls_headers():
...
-    # vms_uuid is a list of vm_uuid
-    # example: vms_uuid=["vm1_uuid", "vm2_uuid"]
-    # the return value is a string like this vm1,vm2
+def generate_vm_list(vm_list):
+    """ Generate vm names with given list of (vm_uuid, vm_name) from db"""
+    # vm_list is a list of (vm_uuid, vm_name) from db
+    # the return value is a comma separated string of VM names like this vm1,vm2
     res = ""
-    for vm_uuid in vms_uuid:
+    for vm_uuid, vm_db_name in vm_list:
```
lipingxue: I don't understand the logic here. In your passed in ```vm_list```, name of the vm is already in ```vm_db_name```, then why still need to call vmdk_utils.get_vm_name_by_uuid() to get the vm name from host for every vm in ```vm_list```? Should we  only call vmdk_utils.get_vm_name_by_uuid()  if the vm_name is not available in DB?

esx_service/cli/vmdkops_admin.py
```diff
@@ -930,11 +933,17 @@ def tenant_vm_ls_headers():
 def generate_tenant_vm_ls_rows(vms):
     """ Generate output for tenant vm ls command """
     rows = []
-    for vm in vms:
-        # vm has the format like this (vm_uuid)
-        uuid = vm
-        name = vmdk_utils.get_vm_name_by_uuid(uuid)
-        rows.append([uuid, name])
+    for vm_uuid, vm_name in vms:
+        name = vmdk_utils.get_vm_name_by_uuid(vm_uuid)
```
lipingxue: Same question as above.

esx_service/utils/auth_data.py
```diff
@@ -35,10 +35,10 @@
 AUTH_DB_PATH = '/etc/vmware/vmdkops/auth-db'
 
 # DB schema and VMODL version
-# Bump the DB_MINOR_VER to 1.1
+# Bump the DB_MINOR_VER to 1.2
 # in DB version 1.1, _DEFAULT_TENANT will be created using a constant UUID
```
lipingxue: Could you add a comment about what is the DB schema change in version 1.2?

esx_service/utils/auth_data.py
```diff
@@ -538,18 +538,80 @@ def handle_upgrade_db_from_ver_1_0_to_ver_1_1(self):
...
         if DB_MAJOR_VER == 1 and DB_MINOR_VER == 1:
             # DB version read from auth-DB is 1.0, upgrade is needed
             if major_ver == 1 and minor_ver == 0:
                 self.handle_upgrade_db_from_ver_1_0_to_ver_1_1()
 
+        if DB_MAJOR_VER == 1 and DB_MINOR_VER == 2:
+            if major_ver == 1 and minor_ver == 1:
+                return self.handle_upgrade_db_from_ver_1_1_to_ver_1_2()
+            else:
+                error_msg = "handle_upgrade_db: failed to upgrade auth-db schema"
```
lipingxue: Could you change the error msg to say that "upgrade is not support for auth-db schema schema version {} to schema version {}"?

esx_service/utils/auth_data.py
```diff
@@ -576,7 +638,9 @@ def connect(self):
 
         if not need_create_table:
             if os.path.isfile(self.db_path) and self.need_upgrade_db():
-                self.handle_upgrade_db()
+                error = self.handle_upgrade_db()
+                if error:
+                    raise DbConnectionError(self.db_path)
```
lipingxue: Could you also add the test for the case that existing DB schema is before 1.1? I want some test cover the error case here.

esx_service/cli/vmdkops_admin.py
pshahzeb: The idea is to not make the vm name in db the first choice, because it could be stale. Always try to resolve it from the host first, if couldn't resolve, then use one from db.

esx_service/cli/vmdkops_admin.py
```diff
@@ -930,11 +933,17 @@ def tenant_vm_ls_headers():
...
+        # If the VM name cannot be resolved then use one from db
+        # If it is not available from db then mark it as NOT_AVAILABLE
+        if not name:
+            if vm_name:
+                name = vm_name
+            else:
+                name = NOT_AVAILABLE
```
govint: Can this be like - if not name and vm_name:
                                 name = vm_name
                            else
                                name = NOT.....

esx_service/utils/auth_api.py
```diff
@@ -335,12 +335,11 @@ def _tenant_create(name, description="", vm_list=None, privileges=None):
             return error_info, None
 
         logging.debug("_tenant_create: vms=%s", vms)
-        vms_uuid_list = [(vm_id) for (vm_id, vm_name) in vms]
 
     error_info, tenant = create_tenant_in_db(
         name=name,
         description=description,
-        vms=vms_uuid_list,
+        vms=vms,
```
govint: Is it ok to init vms as [] vs. None? Code thats called from here to create the tenant use vms as a list and don't ever check the list is None. While its done in a try-except block in auth_data.py:create_tenant(), the logs may seem ambigous why an exception happens on tenant create.

esx_service/utils/auth_api.py
```diff
@@ -527,8 +529,7 @@ def _tenant_vm_add(name, vm_list):
         return error_info
 
     logging.debug("_tenant_vm_add: vms=%s", vms)
-    vms_uuid_list = [(vm_id) for (vm_id, vm_name) in vms]
-    error_msg = tenant.add_vms(auth_mgr.conn, vms_uuid_list)
+    error_msg = tenant.add_vms(auth_mgr.conn, vms)
```
govint: add_vms is expecting only a list of VM uuids vs. a tuple of (uuid, name), is it ok to pass vms here?

esx_service/utils/auth_data.py
```diff
@@ -538,18 +538,80 @@ def handle_upgrade_db_from_ver_1_0_to_ver_1_1(self):
...
         raise DbAccessError(self.db_path, error_msg)
 
+    def handle_upgrade_db_from_ver_1_1_to_ver_1_2(self):
+        """
+        Upgrade the db from version 1.1 to 1.2
+        In 1.1 the vms table had two columns namely vm_id and tenant_id
+        In 1.2 the vms table has three columns namely vm_id, tenant_id and vm_name
+        In the process of upgrading existing db from 1.1 to 1.2, we need to populate
+        the vm names for existing records. We try to populate them and keep None for vms
+        for which the name couldn't be found. The vmdk_ops admin code which tries to use
```
govint: Isn't this entirely dependent on which node the upgrade is done from. Can we say how this works with a X-ESX env.

Depending where this code runs first, VMs on that host alone will be identified.

I'm missing this part but how is the synch. achieved across nodes thats running this code?

Do we want to use this approach of in place upgrades of DB across revisions? What happens if user wants to rollback because of some bug.

esx_service/utils/auth_data.py
```diff
@@ -123,15 +123,15 @@ def create_datastore_access_privileges(privileges):
...
 def create_vm_list(vms):
     """
-        Return a list of vm_uuid with given input
-        @Param vms: a list of tuple, each tuple has format like this (vm_uuid,)
+        Return a list of (vm_uuid, vm_name) with given input
+        @Param vms: a list of sqlite3.Row tuple, each tuple has format like this (vm_uuid, tenant_id, vm_name)
         Example:
-        Input: [(u'564d6857-375a-b048-53b5-3feb17c2cdc4',), (u'564dca08-2772-aa20-a0a0-afae6b255fee',)]
-        Output: [u'564d6857-375a-b048-53b5-3feb17c2cdc4', u'564dca08-2772-aa20-a0a0-afae6b255fee']
+        Input: [(u'564d6857-375a-b048-53b5-3feb17c2cdc4', u'88d98e7a-8421-45c8-a2c4-7ffb6c437a4f', 'ubuntu-VM0.1'), (u'564dca08-2772-aa20-a0a0-afae6b255fee', u'88d98e7a-8421-45c8-a2c4-7ffb6c437a4f', 'ubuntu-VM1.1')]
```
govint: Is this the input, below code creates the entries like - (vm-uuid, vm-name, tenant-uuid).

esx_service/utils/auth_data.py
```diff
@@ -576,7 +638,9 @@ def connect(self):
 
         if not need_create_table:
             if os.path.isfile(self.db_path) and self.need_upgrade_db():
-                self.handle_upgrade_db()
+                error = self.handle_upgrade_db()
+                if error:
```
govint: Not this change, but shouldn;t the DB upgrade check happen once on service startup and service falls back to existing version if upgrade fails or service refuses to start which is catastrophic - data loss (no volumes can be used).

Believe this change is really not so critical to prevent using data volumes for an upgrade failure and auth_* code should be able to simply fall back to using VM name from UUID if available or no-name. auth_* code should be able to work with a 1.1 DB also if required. Simply, service can't prevent using user data!

govint: Pls. add a test case where volumes (user data) are present and upgrade fails (simulate).

esx_service/utils/auth_data.py
```diff
@@ -624,6 +688,8 @@ def create_tables(self):
                 vm_id TEXT PRIMARY KEY NOT NULL,
                 -- id in tenants table
                 tenant_id TEXT NOT NULL,
+                -- name of the VM being added to the tenant
+                vm_name TEXT,
                 FOREIGN KEY(tenant_id) REFERENCES tenants(id)
```
govint: Just asking, is the order like vm-uuid, vm-name, tenant-id or vm-uuid, tenant-id, vm-name - does it make any diff? Because everywhere the inserts are done as (vm-uuid, vm-name, tenant-id) :-)

pshahzeb: the order is  vm-uuid, tenant-id, vm-name. For inserts, it takes the order of columns specified in query is used. So doesn't make any difference

esx_service/utils/auth_data.py
pshahzeb: Thanks for mentioning this. It makes sense. 
CC @lipingxue

esx_service/utils/auth_data.py
pshahzeb: I don't think we have a use case where user is upgrading from version 1.1 in X-ESX env. With the multi esx feature, the shared db would be from 1.2 version, and later on versions above. 

This is hit when the releases henceforth are deployed in single node mode and upgrades are done.

Does it answer your question?

esx_service/utils/auth_api.py
pshahzeb: It does check. Line 801?

esx_service/utils/auth_data.py
lipingxue: @govint, for the first comment to change the upgrade check on service startup, I think it is out of the scope of this PR, we can file a issue for this and later to evaluate it whether we should make the change or not.
For the second part, if upgrade failed, you suggest that we should fallback to existing version. If will only work for the specific change we make since we can use the vm UUID if no vm name is available. But if later, we made some big schema change, then fallback to the existing version is not an option unless in the code we maintain different handling logic for old schema and new schema. 
CC @msterin

esx_service/utils/auth_data.py
msterin: or (much better) refer to vDVS release versions

esx_service/utils/auth_data.py
msterin: Since we open , use and close the DB for each request from Docker then no, we should not check at the start only, it does not really improve anything

if the upgrade fails, there is no data loss. Its a corner case (i.e. very rare, preferably never) and the workaround is to remove the DB. I do not think we should carry forward almost-useless-code reading both versions (and writing both versions), and then test it too.

esx_service/cli/vmdkops_admin.py
pshahzeb: if not name and vm_name:
name = vm_name
else
name = NOT

won't be correct, since NOT AVAILABLE is used only when vm name wasn't resolved by get_vm_name_by_uuid as well as it wasn't found in DB.

esx_service/utils/auth_data.py
```diff
@@ -123,15 +123,15 @@ def create_datastore_access_privileges(privileges):
...
         Example:
-        Input: [(u'564d6857-375a-b048-53b5-3feb17c2cdc4',), (u'564dca08-2772-aa20-a0a0-afae6b255fee',)]
-        Output: [u'564d6857-375a-b048-53b5-3feb17c2cdc4', u'564dca08-2772-aa20-a0a0-afae6b255fee']
+        Input: [(u'564d6857-375a-b048-53b5-3feb17c2cdc4', u'88d98e7a-8421-45c8-a2c4-7ffb6c437a4f', 'ubuntu-VM0.1'), (u'564dca08-2772-aa20-a0a0-afae6b255fee', u'88d98e7a-8421-45c8-a2c4-7ffb6c437a4f', 'ubuntu-VM1.1')]
+        Output: [(u'564d6857-375a-b048-53b5-3feb17c2cdc4', 'ubuntu-VM0.1'), (u'564dca08-2772-aa20-a0a0-afae6b255fee', 'ubuntu-VM0.1')]
     """
     vm_list = []
     for v in vms:
-        vm_list.append(v[0])
+        vm_list.append((v[0], v[2]))
```
pshahzeb: tenant uuid

esx_service/utils/auth_data.py
```diff
@@ -538,18 +538,80 @@ def handle_upgrade_db_from_ver_1_0_to_ver_1_1(self):
...
+                "INSERT INTO vms(vm_id, vm_name, tenant_id) VALUES (?, ?, ?)",
+                vms
+            )
+
+            # update the version info for the db
+            self.conn.execute(
+                "UPDATE versions SET major_ver = ?, minor_ver = ?", (DB_MAJOR_VER, DB_MINOR_VER)
+            )
+
+            self.conn.commit()
```
pshahzeb: Not sure how to combine the above queries under single executemany(), since they are of different types.
Did you mean [executescript] ?(https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executescript)

msterin: yes, executescript - sorry about the confusion. 
we have our own example too https://github.com/vmware/docker-volume-vsphere/blob/master/esx_service/cli/vmdkops_post_update.11.1.py :-)

esx_service/utils/auth_data.py
msterin: ah, makes sense, Can you add a brief comment here abut what [0] [1] and [2] are in there ?

esx_service/utils/auth_data.py
govint: As DB changes happen we will want to be able to handle DB versions with separate "handler" code. Right now its the same code thats only able to work with the newest version of the DB. A better approach will be to make the existing code into an interface and a handler that supports the interface for a specific DB version. And we could do the DB upgrade outside of the service code itself.

As users have already started deploying the volume driver and using it for their workloads its impossible to ask them to delete the DB if an upgrade were to fail. A failure to upgrade shouldn't cause the service to disallow using any of the volumes present. Which would be data loss (no access to data) to the user if it did.

May not be for the change done now but going forward we can't take the approach of deleting the DB, the DB is perfectly fine and only the upgrade is failing.

esx_service/utils/auth_data.py
govint: Ok, I thought the X-ESX support is merged already.

pdhamdhere:
@pshahzeb Please rebase and resolve conflicts.

esx_service/utils/auth_data.py
pdhamdhere: Does this result in stopping the service?

esx_service/utils/auth_data.py
msterin: Plus, I take back the comment on commit. Sqlite3 seems to start each execute with begin transaction, but leave commit out.

pshahzeb: Yes. But executemany is needed for INSERT INTO query as it runs the sql on every tuple in the list of updated records. Cannot use list executescript.
Also the commit at the end makes this change transactional.

msterin: I actually expected "ALTER table; SELECT rows with no name, fill them in and up and REPLACE them back"
I am also wondering if it could be a single executescript() which uses connection.custom_function() to define a SQL function which fills up name_from_uuid. Can you take a look ? The current code is correct, it just does not feel sql-ish or python-ish enough :-)

esx_service/utils/auth_data.py
pshahzeb: No. These exceptions are captured and printed neatly to the user as the command output. Mentioning one such example in PR description.

pshahzeb:
@pdhamdhere @msterin @lipingxue @govint 
I have accommodated your suggestions in latest commits.

esx_service/utils/auth_data.py
pshahzeb: I agree with your point. We can have a separate issue to streamline the db upgrades.

pdhamdhere:
@pshahzeb Can you please check & fix CI failures?

pshahzeb:
Yes. Working on it.

pshahzeb:
@pdhamdhere  Fixed it

esx_service/utils/auth_api.py
```diff
@@ -526,8 +525,10 @@ def vm_not_exist(name, vms):
     if error_info:
         return error_info
 
+    existing_vm_uuids = [(vm_id) for (vm_id, _) in existing_vms]
```
msterin: why do you need () around (vm_id) ?

esx_service/utils/auth_api.py
```diff
@@ -545,7 +546,8 @@ def vm_in_any_tenant(vms):
 
     for tenant in tenant_list:
         for vm_id, vm_name in vms:
-            if vm_id in tenant.vms:
+            tenant_vm_uuids = [(cur_vm_id) for (cur_vm_id, _) in tenant.vms]
```
msterin: nit: ( ) around cur_vm_id seem useless

esx_service/utils/auth_api.py
```diff
@@ -545,7 +546,8 @@ def vm_in_any_tenant(vms):
...
+            if vm_id in tenant_vm_uuids:
```
msterin: Line 549-550 why not simply 
```python
if vm_id in dict(tenant.vms):
   error_info= ...
```

esx_service/utils/auth_data.py
msterin: it looks like the whole function is doing the followig
```python
   return [(v[0], v[2] for v in vms]
```
right ?

esx_service/utils/auth_data.py
```diff
@@ -220,7 +226,7 @@ def remove_vms(self, conn, vms):
     def replace_vms(self, conn, vms):
         """ Remove vms from the vms table for this tenant. """
```
msterin: comment is wrong - cut-n-pasted obviously. 
Also, do we really need "replace" or do we need "update" ?

esx_service/utils/auth_data.py
```diff
@@ -229,7 +235,7 @@ def replace_vms(self, conn, vms):
             )
 
             conn.executemany(
-                "INSERT INTO vms(vm_id, tenant_id) VALUES (?, ?)",
+                "INSERT INTO vms(vm_id, vm_name, tenant_id) VALUES (?, ?, ?)",
```
msterin: nit : if it's really "update", then instead of DELETE then INSERT it could be a snigle INSERT OR REPLACE.
Does not really matter in our case , but generally a good practice to avoid unneeded rescans

msterin:
A few nits/comments (please take a look) but generally LGTM

esx_service/utils/auth_data.py
pshahzeb: Yes.  should be done in one line.

esx_service/utils/auth_data.py
pshahzeb: This replaces the existing VMs that belong to a tenant with the new list of VMs.
Single INSERT OR REPLACE could work; but it again require removal of records of earlier VMs that don't belong to tenant (or update their tenant_id to empty).

PR #1156, opened 2017-04-10
Changed lock names to make the log a bit more readable

msterin:
Made logs mode a bit more readable. (below) 
Also aligned a few statemtns to avoid lint nagging my IDE.

Fixes:
1. The lock name is a part of a thread name which is a part of any log record.
So we had full VMDK names in prefix to lots of log records.
This commit changes it , and it goes from something like this:
```
4/14/17 01:01:18 71262 [umini-16.10-HW11-1-[datastore1] 
dockvols/11111111-1111-1111-1111-111111111111/vol1.vmdk] [DEBUG  ] 
add to volumes table(11111111-1111-1111-1111-111111111111 
/vmfs/volumes/56da6dd5-ba4d7517-7616-000c2951bd44 vol1 100)
```
to something like this:
```
03/14/17 01:35:13 73795 [umini-16.10-HW11-1-DEFAULT.datastore1.vol2] [DEBUG  ] 
add to volumes table(11111111-1111-1111-1111-111111111111
 /vmfs/volumes/56da6dd5-ba4d7517-7616-000c2951bd44 vol1 100)
```
so at least we do not have full lines twice.

2. a few places we we can use tenant name, not UUID, use it
tested: manual


esx_service/utils/auth.py
```diff
@@ -202,7 +202,7 @@ def get_privileges(tenant_uuid, datastore_url):
     # TBD return _auth_mgr.get_defaut_privileges_dict() when both params matchm otherwise return std err
 
     privileges = []
-    logging.debug("get_privileges tenant_uuid=%s datastore_url=%s", tenant_uuid, datastore_url)
+    #logging.debug("get_privileges tenant_uuid=%s datastore_url=%s", tenant_uuid, datastore_url)
```
pdhamdhere: Remove?

esx_service/utils/auth.py
```diff
@@ -202,7 +202,7 @@ def get_privileges(tenant_uuid, datastore_url):
     # TBD return _auth_mgr.get_defaut_privileges_dict() when both params matchm otherwise return std err
```
lipingxue: This comment need to remove or we will handle it later?

esx_service/vmdk_ops.py
```diff
@@ -766,7 +766,7 @@ def get_full_vol_name(vmdk_name, datastore):
     Forms full volume name from vmdk file name an datastore as volume@datastore
     """
     vol_name = vmdk_utils.strip_vmdk_extension(vmdk_name)
-    logging.debug("get_full_vol_name: %s %s", vmdk_name, datastore)
+    #logging.debug("get_full_vol_name: %s %s", vmdk_name, datastore)
```
lipingxue: Do we want to remove this comment?

msterin: I does not seem to contribute as the same info is available pretty much everywhere, and it makes the log very hard to read.

esx_service/utils/auth.py
msterin: yes, thanks - the commend needs to go. Done

PR #1157, opened 2017-04-10
First shot at shared in-memory DB cache (PREVIEW) 

msterin:
### Status
This is not ready for merge (more testing needed) but I'd love to get first feedback. 

### Request 
Can you take a look  ?  

The key changes are in the new DBCacheManager and some new methods in AuthDataManager
Also, test suggestions are welcomed.

@pdhamdhere @andrewjstone - I am wondering if it's clear enough,.and if there are some obvious races and suspicious code pieces.

Do not not pay attention to individual commits (it will be fixed), just the total diff.
The diff also picks up a couple of formatting changes and log clan up changes (sorry). 

### Issue 

multiple ESX hosts lock the db file (VMFS lock), We currently fail but will need to retry - but in any kind of multi-node load the retries are bound to create lots of delays for user commands , since each delay is easily 2-3 secs and they all accumulate in MultiNode since the same DB is accessed. 
  
### Solution

vmdk_ops really need readonly access to DB, so if we can grab the DB and cache it in-memory for 10-15 min, we will guarantee smooth sailing for this period, and even on refreshes we will have very low probability of file being locked, since only refresh will lock and regular requests do not.

This PR introduces shared in-memory config DB cache. Each 15 min (default, can be change via env variable) the code will grab the current state of the config DB and safe in in read-only in-memory sqliteDB.

All connection requests from this point will be provided with a connection to the cache. When it's time to refresh the DB content from on-disk DB, the code will block new requests, wait until all requests using the current cache complete, refresh the cache and allow new requests. 

On the bad side, when refresh is happening we need to wait until all "in-flight" requests to complete, so requests coming in at this moment may have a shot hiccup

## TBD
add retry of connect() to real DB fails .
Potential improvement (not done) : do stat() on the file and skip the refresh if it's not changed

### Test
test: only basic 'make test-esx' and some interactive playing with different configurations.

### DIFF
--- a/esx_service/utils/auth_data.py
+++ b/esx_service/utils/auth_data.py
@@ -423,80 +423,103 @@ class DockerVolumeTenant:
 class DBCacheManager(object):
     """Support for thread--shared in-memory RO cache."""
 
+    """TBD change to factory:
+    - factory keeps a connectoin to in-mem DB to prevent it from cleaning up static)
+    -  on first enable, check we are not enabled, populate the DB
+    - gets NEW connection when asked: conn = cache.new_connection()
+    - the rest is the same.
+    - Row can be set on each conenctoin, not cache
+    - on refresh, drain in flight, drop global connection (that will deleted the DB) and recreate cache
+
+    """
+
+    # this is in-memory shared DB URL
+    MEMDB_NAME = "file::memory:?cache=shared"
+
     # - How often to refresh memdb from on-disk DB.  Default 15 min.
     REFRESH_SEC = int(os.getenv("VMDKOPS_MEMDB_REFRESH_SEC", default=900))
 
+
     def __init__(self):
-        # Connection object.
-        # We populate it every MEMDB_REFRESH_SEC interval and
-        # we use it mainly from for esx vmdk_ops service request authorization.
+        # Init connection to in-memory DB.
+        # We keep one connection open to assure the DB is not garbage collected too early
+        # We refresh it every MEMDB_REFRESH_SEC i. When we refresh, we first drain all in-flight
+        # AuthernticatoinDataManagers to make sure nobody is using the copy.
+        #
         # 'None' means the feature is disabled
         self._memdb_con = None
 
-        # Barrier
+        # Init Barrier
         # We need to avoid refresh (which could be scheduled by Timer on a different thread)
-        # from changing connection for auth. mgrs in flight. So  we acquire a barrier,
+        # from messing DB content for auth. mgrs in flight. So  we acquire a barrier,
         # and wait until all auth.managers in flight land.
         self._barrier = threading.Condition()
         self._in_flight = 0
 
-        # Timer for periodic refreshes
+        # Init timer for periodic refreshes
         self._timer = None
 
+    def __del__(self):
+        pass
 
-    @property
-    def conn(self):
-        'Gets shared connectoin object. DO NOT close() from outside !'
-        return self._memdb_con
+    def _drain_and_close(self):
+        """
+        Drains Auth. managers in fight and closes last memdb connection.
+        *** EXPECTS to be run under _barrier lock.
+        """
+        logging.debug("_Drain and close")
+        # drain the in-flight users
+        self._memdb_con.close()
+        self._memdb_con = None
+        while self._in_flight > 0:
+            logging.debug("Wating for inflight %d", self._in_flight)
+            self._barrier.wait()
+        logging.debug("DONE")
 
-    def _drain_and_reset_cache(self):
-        """
-        Drains auth.managers in flight and closes cache connection.
-        """
-        with self._barrier:
-            while self._in_flight > 0:
-                self._barrier.wait()
-            if self._memdb_con:
-                self._memdb_con.close()
-                self._memdb_con = None
 
-    def incr_manager(self):
-        """
-        Increment "in-flight'count.
-        No need to notify since we may wait only on decreasing counts
-        """
-        with self._barrier:
-            self._in_flight += 1
-
-    def decr_manager(self):
-        """
-        Decrease in-flight count and notify waiting cache refresh, if any
-        """
-        with self._barrier:
-            self._in_flight -= 1
-            self._barrier.notify()
-
-    def memdb_enable(self):
+    def enable(self):
         """
         Enables in-memory ReadOnly cache DB for auth DB.
         This includes opening on-disk DB and populating in-memory DB from it.
         Note this will force the connection to treat the DB as read only.
         """
-        if self._memdb_con:
-            logging.debug("memdb_enable is already  enabled, ignoring")
-            return
-        logging.debug("memdb_enable Starting refresh")
-        self.memdb_refresh_periodically(frequency_sec=type(self).REFRESH_SEC)
+        logging.debug("memdb_enable")
+        if not self.is_enabled():
+            logging.debug("memdb_enable Starting refresh")
+            # No need to drain anything as nobody was using the cache so far
+            # Proceed to creation os a new cache
+            self._memdb_refresh_periodically(frequency_sec=self.__class__.REFRESH_SEC)
 
-    def memdb_disable(self):
+    def disable(self):
         """
-        Drains managers in flight and disables the DB cache.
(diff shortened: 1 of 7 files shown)


andrewjstone:
I haven't looked at the code yet, but does this mean that any new configuration changes have to wait upwards of 15 min before they become visible to users? If so, perhaps we have a way to bust the cache, so updates are visible sooner.

msterin:
Thanks for taking a look.
Yes they would have to wait. This is a payment for not needing to bust the cache on multiple esxs when DB is changed on one of them. I am open to better ideas, short of waiting for distributed KV store on esx :-(

andrewjstone:
Gotcha. Hopefully we can get you that KV store in a non-glacial timeframe.

msterin:
I am also looking to use stat()  to avoid needless refreshes and maybe iNotify() to initiate ?

esx_service/utils/auth_data.py
```diff
@@ -417,14 +420,223 @@ def remove_datastore_access_privileges(self, conn, datastore_url):
...
 
+class DBCacheManager(object):
+    """
+    Support for thread--shared in-memory RO cache.
+
+    - uses sqlite3 in -memory shared DB. One connection per thread is expected.
+    - keeps a connetion to in-mem DB to prevent it from cleaning up static cache
+    - on first enable, check we are not enabled, populate the DB
+    - gets NEW connection when asked: conn = cache.new_connection()
+    - on refresh, drains in flight Auth managers using the hshared cache,
```
andrewjstone: typo: hshared

esx_service/utils/auth_data.py
```diff
@@ -417,14 +420,223 @@ def remove_datastore_access_privileges(self, conn, datastore_url):
...
+
+    def is_enabled(self):
+        'True if memdb cache is enabled'
+        with self._barrier:
+            ret = self._memdb_con is None
+        return ret
+
+    @staticmethod
+    def _make_connection_readonly(con):
+        con.executescript("PRAGMA query_only = 1; PRAGMA read_uncommitted = 1;")
```
andrewjstone: I'm not sure read_uncommitted is what you want. It could lead to readers seeing configurations from transactions that get aborted.

esx_service/utils/auth_data.py
```diff
@@ -417,14 +420,223 @@ def remove_datastore_access_privileges(self, conn, datastore_url):
...
+            self._memdb_refresh_periodically(frequency_sec=self.__class__.REFRESH_SEC)
+
+    def disable(self):
+        """
+        Disables the DB cache and closes the connection. The  DB will be deleted when all
+        AuthorizationDataManagers using this cache land and close their connection.
+        """
+        logging.debug("memdb_disable")
+        with self._barrier:
+            if self.is_enabled():
```
andrewjstone: This call tries to re-acquire the lock: `with self._barrier`. I'm not sure if these locks are re-entrant. If not, this is a deadlock.

andrewjstone: from the docs, it appears threading.condition() defaults to an RLock, so this *should* be safe.

esx_service/utils/auth_data.py
msterin: FWIW since *all* connections to the memdb after it's created are readonly (query_only = 1) there should be no transactions to abort :-)

So I've added out of sheer paranoia, in case sqlite3 has some internal checks we can skip.

esx_service/utils/auth_data.py
msterin: yes, I explicitly checked it when putting this code together. Still, based on the fact we are discussing it, I think a comment is needed to explain that it's Rlock (reentrant under itself as long as the inner one released by the same thread which created it) . I'll add a comment

esx_service/utils/auth_data.py
```diff
@@ -417,14 +420,223 @@ def remove_datastore_access_privileges(self, conn, datastore_url):
...
+
+
+    def enable(self):
+        """
+        Enables in-memory ReadOnly cache DB for auth DB.
+        This includes opening on-disk DB and populating in-memory DB from it.
+        Note this will force the connection to treat the DB as read only.
+        """
+        logging.debug("memdb_enable")
+        if not self.is_enabled():
```
andrewjstone: I don't believe this is safe. Multiple threads can determine that the cache is not enabled, before the call to `self._memdb_refresh_periodically(frequency_sec=self.__class__.REFRESH_SEC)` and therefore multiple threads can call that function. I believe you want to call the refresh function with the lock held.

esx_service/utils/auth_data.py
```diff
@@ -417,14 +420,223 @@ def remove_datastore_access_privileges(self, conn, datastore_url):
...
+    def _drain_and_close(self):
+        """
+        Drains Auth. managers in fight and closes last memdb connection.
+        *** EXPECTS to be run under _barrier lock.
+        """
+        logging.debug("_Drain and close")
+        # drain the in-flight users
+        self._memdb_con.close()
+        self._memdb_con = None
+        while self._in_flight > 0:
```
andrewjstone: I'm not sure this loop works as intended, if multiple threads can call this function. For instance, let's say that there is only 1 in flight auth mgr. Does that mean that there is only one thread that can wait on this barrier? If not, one of the threads waiting on the barrier may never be woken up. I'm not sure this is a possibility with the way the code is intended to work, but I figured I'd bring it up.

andrewjstone:
I looked through the code, and while I believe I understand what is going on, it isn't that easy to get a grasp of. Of course this is pretty much true of all multithreaded, lock based code. I think my main concern is that there are many checks that happen outside of locks that can lead to race conditions. Ideally there would be less functions that deal with locks, and any coordination would occur with the locks held. It may make more sense for the in_flight stuff to be a semaphore instead of a regular counter. It's possible that doing that would eliminate the need for all the other checks. What I mean is when the semaphore is at 0, the cache can be enabled by the first in line.

msterin:
Yes, I agree with your concern, that's why I asked to check for glaring holes ...

> What I mean is when the semaphore is at 0, the cache can be enabled by the first in line.

yes, in-flight count is essentially a 'reverse' semaphore with green on 0. I am not sure how to craft that our of threading.semaphore or threading.bondedsemaphore without  a condition and a counter :-) 
An example would be appreciated.

I guess I need to add the following to the comments: 
There are 4 places dealing with locks: 
- new authmanager wating barrier in the call to DBCacheManager.is_shared()
- authmanager adding or removng itself from 'in flight' count on _connect() and _close() 
- DBManager refresh  - drain and refresh 
- DBCacheManager disable() - drain and close

esx_service/utils/auth_data.py
msterin: Yes, good point. This is not intended as a fuly generic API. Currently it's only intended to use as essentially a singleton, one instance per AuthorizationDataManager class.
Yes, multiple threads can wait on the barrier (any new request from Docker will check the barrier first and wait if there is upgrade in progress), but one and only one can be  draining (i.e. waiting in this function)

esx_service/utils/auth_data.py
msterin: (1) enable() (and thus starting the refresh cycle) is supposed to happen on initializing of the whole process, before anyone had a chance to ask is_enabled().
I will add comment about this 
(2) Generally the design is that getting "not enabled" erroneously is not a problem.
Even if a thread gets "not enabled"  due to a race, it simply misses the cache.

andrewjstone:
> yes, in-flight count is essentially a 'reverse' semaphore with green on 0. I am not sure how to craft that out of threading.semaphore or threading.bondedsemaphore without a condition and a counter :-)
An example would be appreciated.

I think you are right here because you don't actually want to block when incrementing in-flight. You just want to say: "here is another connection that is still working". You could do this with a bounded semaphore, except for the fact that you don't know the number of connections in total. it also doesn't provide much value over what you have. Carry on.

msterin:
So while this would would, this would actually prevent us from being able to support x-ESX quotas, even as simple as admission control for the amount of virtual storage space allocated. 
So we mothball this PR for a while and either resurrect or bury in a a few weeks

PR #1165, opened 2017-04-14
Add lock to init_datastoreCache

lipingxue:
init_datastoreCache and a couple of it's callers use global 'datastores' with no lock, which may cause race. This PR is to add lock to init_datastoreCache() to avoid the race.


esx_service/vmdk_ops.py
```diff
@@ -644,7 +645,7 @@ def listVMDK(tenant):
     Each volume name is returned as either `volume@datastore`, or just `volume`
     for volumes on vm_datastore
     """
-    vmdk_utils.init_datastoreCache()
+    vmdk_utils.init_datastoreCache(force=True)
```
msterin: Why do we need to force cache refresh here ?

lipingxue: I think it need to make sure the ```listVMDK``` return the correct value even if the "datastore" name is changed outside. Cache need to be forced refresh to pickup the new datastore name.

msterin: Ah, makes sense

msterin:
LgTM

.gitignore
```diff
@@ -1,3 +1,4 @@
+.gitignore
```
govint: Is this file to be added?

esx_service/utils/vmdk_utils.py
```diff
@@ -152,13 +165,13 @@ def get_volumes(tenant_re):
                 error_info, tenant_name = auth_api.get_tenant_name(sub_dir_name)
                 if not error_info:
                     logging.debug("get_volumes: path=%s root=%s sub_dir_name=%s tenant_name=%s",
-                                path, root, sub_dir_name, tenant_name)
+                                  path, root, sub_dir_name, tenant_name)
```
govint: Nit - white space

.gitignore
lipingxue: I think we have this file in the master branch too.

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,57 @@
...
-    global datastores
-    if datastores == None:
-        init_datastoreCache()
+    init_datastoreCache()
     if datastore in [i[0] for i in datastores]:
         return True
     else:
-        init_datastoreCache()
+        init_datastoreCache(force=True)
         if datastore in [i[0] for i in datastores]:
```
govint: What happens is one thread is doing this check here (looping thru the datastores while another thread is running init_datastoreCache() and reconstructing the datastores list?

govint: All uses of the datastores list must be protected with the lock - whether re-creating the list or looping over it.

esx_service/utils/vmdk_utils.py
```diff
@@ -109,21 +122,21 @@ def get_datastores():
     'url' is datastore URL (e.g. '/vmfs/volumes/vsan:572904f8c031435f-3513e0db551fcc82')
     'dockvol-path; is a full path to 'dockvols' folder on datastore
     """
-    if datastores == None:
-        init_datastoreCache()
+    init_datastoreCache()
     return datastores
```
govint: Another thread may in parallel be in init_datastoreCache() and set datastores to None (by force option). May be get_datastores() should be returning a copy of the datastores list.

esx_service/utils/vmdk_utils.py
lipingxue: I think we should avoid using global variable "datastores" to loop through. We should call function "get_datastores" instead.

lipingxue: @msterin

esx_service/utils/vmdk_utils.py
lipingxue: Why? I think if another thread is in "init_datastoreCache()", this thread try to call "init_datastoreCache()", they it should wait, isn't it?

esx_service/utils/vmdk_utils.py
govint: get_datastores() returns the global, not sure how Python passes globals across modules in this case between vmdk_utils.py and vmdk_ops.py

lipingxue: I think we don't use global "datastores" in "vmdk_ops.py". Am I missing anything here?

govint: We call get_datastores() in vmdk_ops.py: get_datastore_names_list()

esx_service/utils/vmdk_utils.py
govint: No, thread A has finished init_datastoreCache() and is going to return "datastores" and thread B calls init_datastoreCache() and sets datastores to None. Thread A will be returning None.

lipingxue: In this case, even "get_datastores" returns a copy of global variable "datastores" is not right. Since now the copy of "datastores" can be None too.  We should let the init_datastoreCache() to return "datastores". Then funtion "get_datastores" just return "init_datastoreCache()".

govint: Like I mentioned all uses of datastores should be with the lock held. Either hold the lock and create a copy or like you mentioned let init return a copy of datastores list.

lipingxue:
@govint Could you explain more what is the test you suggested? I am not quite clear.

govint:
Ok, modify init_datastoreCache() to always do a forced init - meaning datastores will be set to None and re-initialized on each call and run a workload of test-esx on say >=2 VMs on the host. Verify no exceptions/errors in the ESX service.

msterin:
To add to the discussion above. The main gap is that datastores flip to new value is not transactional for anyone not acquiring the lock. It is easily healed by accumulating the.liat in a local object (local var) and flipping in one op at the end of init (datastores = tmp_datastores). This way one a traversal of stores started in a thread, init in another thread will not impact the object in use. Forcing init defeats the purpose of cache though it's good for testing. 

If there is further discussion, let's take it to telegram

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,57 @@
...
+    and force init
     """
     global datastores
     logging.debug("init_datastoreCache: %s", datastores)
 
-    si = vmdk_ops.get_si()
+    with lockManager.get_lock("init_datastoreCache"):
+        if datastores:
+            if force:
+                datastores = None
```
msterin: this is what can cause issues, as @govint has mentioned ,on a non-locked consumption
it should simply say
```python
  if datastores and not force:
    return
```
and the code down the line should replaced datastores with tmp_datastores (local name bound to a new list)
And the last statement should simply be `datastores = tmp_datasores` which will re-bind `datastores` name to the new object. Note that the old object (list of datastores from the previous init()) may be in use by other threads at the same time and Python will simply garbage collects it when it's not used anymore

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,57 @@
...
-    #  We are connected to ESX so childEntity[0] is current DC/Host
-    ds_objects = \
-      si.content.rootFolder.childEntity[0].datastoreFolder.childEntity
-    datastores = []
+        #  We are connected to ESX so childEntity[0] is current DC/Host
+        ds_objects = si.content.rootFolder.childEntity[0].datastoreFolder.childEntity
+        datastores = []
+
+        for datastore in ds_objects:
+            dockvols_path, err = vmdk_ops.get_vol_path(datastore.info.name)
```
msterin: L89 -92 could be simply:
```python
    tmp_ds = []
    for datastore si.content.rootFolder.childEntity[0].datastoreFolder.childEntity:
        ... same operation, using tmp_ds instead of datastore
   datastores = tmp_ds
```

esx_service/utils/vmdk_utils.py
msterin: @govint   get_datastores() does not return global. In fact, it does not return anything and we use global 'datastres' name to access the current list-of-datastores object. So the "datastores" global   name is bound to a list object created and filled in init_cache(). , When we use the name, python will increase the list object  refcount and use the object (not the name). If during this usage a new list object is created (`[]` in init_cache()) and  bound to 'datastores' global name, it should not impact the old object-in-use. Due to GIL (global interpreter lock) an op like 'datastores' name binding to a new object is atomic - see my comment on this in the main thread. Of course it's still worth testing :-)

.gitignore
msterin: This file is there and controls what git should ignore for all... but I think this specific changbe to the file does not much sense as it removes itself from .git control and freezes the old version. My bad, I suggested this. @lipingxue - you may want to reverse this file.

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,55 @@
...
+
+
+def init_datastoreCache(force=False):
     """
-    Initializes the datastore cache with the list of datastores accessible from local ESX host.
+    Initializes the datastore cache with the list of datastores accessible
+    from local ESX host. force=True will force it to ignore current cache
+    and force init
     """
     global datastores
```
msterin: Nit: this can be moved under 'with lockManager'. Not a big difference , just a more precise scope

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,55 @@
...
 
-    for datastore in ds_objects:
-        dockvols_path, err = vmdk_ops.get_vol_path(datastore.info.name)
-        if err:
-            continue
-        datastores.append((datastore.info.name,
+        for datastore in ds_objects:
+            dockvols_path, err = vmdk_ops.get_vol_path(datastore.info.name)
+            if err:
+                continue
```
msterin: I'd suggest adding a logging.error here too, as it is not supposed to happen at all

lipingxue:
@msterin @govint I have addressed your comments and run the test that Govindan suggested(no exception is found in ESX). Please review it.

esx_service/utils/vmdk_utils.py
```diff
@@ -109,21 +120,21 @@ def get_datastores():
...
+    # volumes for "tenant1" are in /vmfs/volumes/datastore1/dockervol/tenant1
+    # volumes for "tenant2" are in /vmfs/volumes/datastore1/dockervol/tenant2
     # volumes does not belongs to any tenants are under /vmfs/volumes/dockervol
-    # tenant_re = None : only return volumes which does not belong to any tenant
+    # tenant_re = None : only return volumes which do not belong to a tenant
     # tenant_re = "tenant1" : only return volumes which belongs to tenant1
-    # tenant_re = "tenant*" : return volumes which belongs to tenant1 or tenant2
-    # tenant_re = "*" : return all volumes under /vmfs/volumes/datastore1/dockervol
+    # tenant_re = "tenant*" : return volumes which belong to tenant1 or tenant2
+    # tenant_re = "*" : return all volumes under /vmfs/volumes/datastore1/dockervol TODO: fix the comment
```
msterin: please fix the or drop TODO before merge

.gitignore
govint: Ok, could add .swp to it.

esx_service/utils/vmdk_utils.py
govint: Nit, can remove the white spaces.

govint: Could add a log saying this DS is being ignored as the dockvols path can't be created on it.

esx_service/utils/vmdk_utils.py
```diff
@@ -394,18 +417,22 @@ def get_datastore_name(datastore_url):
...
     res = [d[0] for d in get_datastores() if d[1] == datastore_url]
+    logging.debug("get_datastore_name: res=%s", res)
     return res[0] if res else None
 
+
 def get_datastore_url_from_config_path(config_path):
     """Returns datastore url in config_path """
     # path can be /vmfs/volumes/<datastore_url_name>/...
     # or /vmfs/volumes/datastore_name/...
     # so extract datastore_url_name:
```
govint: Not this change, the comment is confusing a little, if the format is the latter "/vmfs/volumes/datastore_name/  then this function is returning datastore_name and not the URL.

esx_service/utils/vmdk_utils.py
lipingxue: Done.

esx_service/utils/vmdk_utils.py
lipingxue: Done.

esx_service/utils/vmdk_utils.py
lipingxue: Done.

esx_service/utils/vmdk_utils.py
```diff
@@ -63,44 +63,56 @@
...
 
-    for datastore in ds_objects:
-        dockvols_path, err = vmdk_ops.get_vol_path(datastore.info.name)
-        if err:
-            continue
-        datastores.append((datastore.info.name,
+        for datastore in ds_objects:
+            dockvols_path, err = vmdk_ops.get_vol_path(datastore.info.name)
+            if err:
+                logging.error(" datastore %s is being ignored as the dockvol path can't be created on it")
```
msterin: logging.warning. Looks  like it *can* happen and will happen

Issue #976, opened 2017-02-28
vmdk_ops fails to start after ESX reboot


ashahi1:
Docker cli commands (docker volume create/ls)  does not works after vm reboot if the vm had crashed while the container was still running.
___
Steps:

1. Created a volume
```
root@photon-xVmYMbyTn [ ~ ]# docker volume ls
DRIVER              VOLUME NAME
root@photon-xVmYMbyTn [ ~ ]# docker volume create --driver=vmdk --name=testVol -o size=200mb
testVol
root@photon-xVmYMbyTn [ ~ ]# docker volume ls
DRIVER              VOLUME NAME
vmdk                testVol@sharedVmfs-0
root@photon-xVmYMbyTn [ ~ ]# docker volume inspect testVol@sharedVmfs-0
[
    {
        "Name": "testVol@sharedVmfs-0",
        "Driver": "vmdk",
        "Mountpoint": "/mnt/vmdk/testVol@sharedVmfs-0",
        "Status": {
            "access": "read-write",
            "attach-as": "independent_persistent",
            "capacity": {
                "allocated": "14MB",
                "size": "200MB"
            },
            "clone-from": "None",
            "created": "Tue Feb 28 08:30:56 2017",
            "created by VM": "photon-VM1.3",
            "datastore": "sharedVmfs-0",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "detached"
        },
        "Labels": {},
        "Scope": "global"
    }
]
root@photon-xVmYMbyTn [ ~ ]#
```
2. Ran the container with the volume mounted:
```
root@photon-xVmYMbyTn [ ~ ]# docker volume ls
DRIVER              VOLUME NAME
vmdk                testVol@sharedVmfs-0
root@photon-xVmYMbyTn [ ~ ]# docker run -it --volume-driver=vmdk -v testVol@sharedVmfs-0:/vol1 --name ub ubuntu
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu

d54efb8db41d: Pull complete
f8b845f45a87: Pull complete
e8db7bf7c39f: Pull complete
9654c40e9079: Pull complete
6d9ef359eaaa: Pull complete
Digest: sha256:dd7808d8792c9841d0b460122f1acf0a2dd1f56404f8d1e56298048885e45535
Status: Downloaded newer image for ubuntu:latest
root@e676ea9c8e99:/# ls
[...]

PR #1168, opened 2017-04-15
Cleanup volume status based on VM poweroff

### LINKED ISSUE #369: Auto-cleanup after VM power off (detach volumes)
In some use cases (e.g. force power off/power on) we can get into a situation when a vmdk is still attached to a VM but not used by Docker. This VMDK will be blocked from used in other VMs/ Docker engines . The suggestion is to handle it automatically, and on "power off" event from hostd (if possivle to get one) to check if any of the volumes in voldir are still attached , if they are - detach them 

When docker is not running, detach volumes (if any):
When plugin starts and sees that Docker is not running, we know that no volumes should be attached to the VM. Let's tell that to ESX service so it can detach all docker volumes.
This will do auto-cleanup of stale volumes after VM power on/ off

pshahzeb:
This PR handles the issue of updating status of volumes to detached that belong to a VM which was forcefully powered off.

The approach is as follows:

1. Create a PropertyCollector filter to monitor updates for VM power state
2. Start a thread and wait for updates from host using the property collector.
3. In case of updates where the power state was powerOff, find the attached devices to the VM
4. If the attached devices are docker volumes, set their status to detached in the KV

Testing:
Tested as follows:
1. Creating a volume and attaching it to a container
2. force poweroff the VM
3. The event is captured by vmdkops service and volume status is set to detached. This volume can be attached to another volume.

No side effects: When a VM with no docker volume attached is turned off

Fixes #369


pshahzeb:
Do we need to take lock before updating the status of vmdk?
What do we do if error occurs during updating the status of vmdk? retry?

pdhamdhere:
1. You are monitoring VM power event for every VM. For every powered off VM, we will scan through VM Config to see if there are any (dockvols) disks attached. This may be okay for first version but we should explore alternatives to quickly detect and ignore who are not managed by vDVS. Can we attach custom attribute to VM whenever a volume is attached. On powered off event, we can skip processing if custom attribute is set.
2. I suspect this code may be racing with HA/FDM which is also listening to VM powered off events. Please test with vSphere HA enabled in cluster and kill VMX process.
3. I don't recall how power state changes are handled with vMotion. Please test with vMotion.

esx_service/utils/threadutils.py
```diff
@@ -90,6 +90,12 @@ def start_new_thread(target, args):
     logging.debug("Currently active threads: %s",
                   get_active_threads())
 
+def start_new_daemon(target, args):
+    d = threading.Thread(target=target, args=args)
+    d.daemon = True
+    d.start()
```
govint: Could log what target and args used to start the new thread.

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
 
 #------------------------
 
+def start_vm_changelistener():
+    """
```
govint: How is this service notified to stop and exit itself, if that is needed at all? Say upgrade is done and service is stopped and restarted.

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+    if not device_list:
+        return
+
+    try:
+        for dev in device_list:
+            # if it is a docker volume, construct the vmdk_path and set its status as detached
+            if vmdk_utils.check_docker_volume(dev):
+                datastore, disk_path = dev.backing.fileName.rsplit("]", 1)
+                vmdk_path = os.path.join("/vmfs/volumes/", datastore[1:], disk_path.lstrip())
+                logging.info("Setting detach status for %s", vmdk_path)
```
govint: Suggest making this a function in vmdk_utils which takes the backing file name and returns the "docker volume path".

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+    """
+
+    if not device_list:
+        return
+
+    try:
+        for dev in device_list:
+            # if it is a docker volume, construct the vmdk_path and set its status as detached
+            if vmdk_utils.check_docker_volume(dev):
+                datastore, disk_path = dev.backing.fileName.rsplit("]", 1)
```
govint: Pls. add a comment here showing the format of "dev" and the diskpath and datastore formats. Just so its clear how these are split/joined.

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+
+    # Recurse through the folders
+    visitFolders = TraversalSpec(name='visitFolders', type=vim.Folder,path='childEntity', skip=False)
+    visitFolders.selectSet.extend((SelectionSpec(name='visitFolders'), SelectionSpec(name='dcToVmf'),))
+
+    return SelectionSpec.Array((visitFolders, dcToVmf,))
+
+def set_device_detached(device_list):
+    """
+    For all devices in device_list, if it is a docker volume, set its status to detached in KV
```
govint: As a practice the service code could refer to its managed volumes as DVS volumes - just suggesting vs. calling these docker volumes.

esx_service/vmdk_ops.py
msterin: +1 on that. Also, lI suspct check_docker_volume(dev) is already building the list ...

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+
+    try:
+        for dev in device_list:
+            # if it is a docker volume, construct the vmdk_path and set its status as detached
+            if vmdk_utils.check_docker_volume(dev):
+                datastore, disk_path = dev.backing.fileName.rsplit("]", 1)
+                vmdk_path = os.path.join("/vmfs/volumes/", datastore[1:], disk_path.lstrip())
+                logging.info("Setting detach status for %s", vmdk_path)
+                setStatusDetached(vmdk_path)
+    except Exception as e:
```
msterin: do functions downstairs i n this callchain throw exceptions ?

esx_service/vmdk_ops.py
```diff
@@ -1711,6 +1712,101 @@ def wait_for_tasks(si, tasks):
 
 #------------------------
 
+def start_vm_changelistener():
+    """
+    Listen to power state changes of VMs running on current host
+    """
+    si = get_si()
+    pc = si.content.propertyCollector
+    create_vm_powerstate_filter(pc, si.content.rootFolder)
```
lipingxue: What do we do if it throw a exception  when creating the filter spec in this function?

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+    visitFolders.selectSet.extend((SelectionSpec(name='visitFolders'), SelectionSpec(name='dcToVmf'),))
+
+    return SelectionSpec.Array((visitFolders, dcToVmf,))
+
+def set_device_detached(device_list):
+    """
+    For all devices in device_list, if it is a docker volume, set its status to detached in KV
+    """
+
+    if not device_list:
```
lipingxue: +1 on this

esx_service/vmdk_ops.py
lipingxue: Nit: docker volume-> DVS volume

esx_service/vmdk_ops.py
pshahzeb: Done. Capturing the return value of this function. But I think it would be right to start the main service in spite of any error while starting the VM listener.

esx_service/vmdk_ops.py
pshahzeb: Not really. I don't think try catch block is needed here.

esx_service/vmdk_ops.py
pshahzeb: This won't happen if vm object(moref) is valid. Removing this check and adding a check for moref instead.

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+                    if objectSet.kind == 'modify':
+                        for change in objectSet.changeSet:
+                            if change.name == 'runtime.powerState' and change.val == 'poweredOff':
+                                vm_name = moref.config.name
+                                logging.info("VM poweroff change found for %s", vm_name)
+                                # if the event was powerOff for a VM, set the status of all
+                                # docker volumes attached to the VM to be detached
+                                set_device_detached(moref.config.hardware.device)
+        except Exception as e:
+            logging.error("PropertyChangeListener: error %s", str(e))
```
pshahzeb: Since this is a background operation, we can log an error. Not sure how else we can notify the admin.

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+        try:
+            # process the updates result
+            for filterSet in result.filterSet:
+                for objectSet in filterSet.objectSet:
+                    moref = getattr(objectSet, 'obj', None)
+                    if objectSet.kind == 'modify':
+                        for change in objectSet.changeSet:
+                            if change.name == 'runtime.powerState' and change.val == 'poweredOff':
+                                vm_name = moref.config.name
+                                logging.info("VM poweroff change found for %s", vm_name)
```
pshahzeb: Done

esx_service/vmdk_ops.py
```diff
@@ -1703,6 +1704,101 @@ def wait_for_tasks(si, tasks):
...
+        logging.error("Problem creating PropertyCollector %s", str(e))
+
+def listen_vm_propertychange(pc, empty):
+    logging.info("PropertyChangeListener thread started")
+    version = ''
+    while True:
+        result = pc.WaitForUpdates(version)
+
+        # if the wait timed out, wait again
+        if result is None:
```
pshahzeb: Actually this WaitForUpdates api doesn't time out. We don't need this check.

esx_service/vmdk_ops.py
pshahzeb: Done.

esx_service/vmdk_ops.py
pshahzeb: As of now, the thread is in daemon mode and it stops with the process is restarted.
Had this not been a daemon, the main thread before exiting, waits for this listener to complete( it is waiting for update or updating the vmdk status).

pshahzeb:
@msterin @govint @pdhamdhere @lipingxue 
I have addressed your comments in my latest commit.
Added a unittest which essentially automates steps of testing in PR description.

I have done the following testing:
1. vmdkops service restart: Works fine. No issues
2. VM suspend, resume:  No issue. We don't get the poweroff event here.
3. VM reset: No poweroff event.
4. Vmotion - cross esx : When a VM is moved to another host, it's still live. No poweroff event either.
5. Enabling HA on cluster and VM monitoring: Killing the vmx process. VM shows poweroffed on VC. But we don't get the poweroff event. I am not sure what should be the behavior here.

esx_service/vmdk_ops_test.py
```diff
@@ -64,6 +65,143 @@
...
+            vmdk_ops.wait_for_tasks(si, [task])
+    else:
+        error_info = error_code.generate_error_info(ErrorCode.VM_NOT_FOUND, vm_name)
+        logging.error("Cannot find vm %s", vm_name)
+        return error_info, None
+
+    return None, vm
+
+def destroy_vm_object(si, vm):
+    logging.debug("Trying to destroy VM %s", vm.config.name)
```
msterin: Actually since it's all test , it could be logging.info. what do you think ?

esx_service/vmdk_ops_test.py
```diff
@@ -1450,6 +1429,94 @@ def testPolicyUsageCount(self):
...
+    datastore_name = None
+    datastore_path = None
+
+    def setUp(self):
+        """ Setup run before each test """
+        logging.info("VMListenerTest setUp path =%s", path)
+
+        if not self.datastore_name:
+            datastores = vmdk_utils.get_datastores()
+            datastore = datastores[0]
```
msterin: the logic here seems wrong. 
I'd expect "for each DB, check if dockvol is there or can be created. if not continue. If yes, use it" 

I guess there are some assumptions about test setup here (or copy-paste from other places with assumptions :-)
If that's the case, it's worth at least commenting

msterin:
> Enabling HA on cluster and VM monitoring: Killing the vmx process. VM shows poweroffed on VC. But we don't get the poweroff event. I am not sure what should be the behavior here

What *is* the behavior ? Does VM stay available with docker volume still attached ? 
If yes, then we are good.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,130 @@
...
+                            continue
+
+                        logging.info("VM poweroff change found for %s", moref.config.name)
+
+                        set_device_detached(moref.config.hardware.device)
+        except Exception as e:
+            # Do we need to alert the admin? how?
+            logging.error("VMChangeListener: error %s", str(e))
+
+        version = result.version
```
pshahzeb: To let changenotifier on host know we need further updates. If not done, we keep getting same update changeset again and again.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,130 @@
...
+    visitFolders.selectSet.extend((SelectionSpec(name='visitFolders'), SelectionSpec(name='dcToVmf'),))
+
+    return SelectionSpec.Array((visitFolders, dcToVmf,))
+
+def set_device_detached(device_list):
+    """
+    For all devices in device_list, if it is a DVS volume, set its status to detached in KV
+    """
+
+    for dev in device_list:
```
pshahzeb: Did a param change to invoke actual disk detach. Keeping this same since.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,130 @@
...
+        # if it is a dvs managed volume, set its status as detached
+        vmdk_path = vmdk_utils.find_dvs_volume(dev)
+        if vmdk_path:
+            logging.info("Setting detach status for %s", vmdk_path)
+            vmdk_ops.setStatusDetached(vmdk_path)
```
pshahzeb: Changed to invoke disk_detach.

pshahzeb:
@msterin I have addressed your comments. Regarding the actual disk, I have invoked disk_detach which does actual detach and in turn updates KV as well.

Regarding testing with HA, I think I somehow missed the logs.
In case of killing vmx process, we do get the poweroff event for the VM.

esx_service/vmdk_ops.py
```diff
@@ -1636,7 +1636,14 @@ def main():
 
         kv.init()
         connectLocalSi()
+        err_msg = vm_listener.start_vm_changelistener()
```
pdhamdhere: Hostd (and `si`) may not not be available after ESX reboot. See #976 This means VMListener won't start after ESX reboot.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,148 @@
...
+
+def listen_vm_propertychange(pc):
+    """
+    Waits for updates on powerstate of VMs. If powerstate is poweroff,
+    detach the dvs managed volumes attached to VM
+    """
+    logging.info("VMChangeListener thread started")
+    version = ''
+    while True:
+        result = pc.WaitForUpdates(version)
```
pdhamdhere: 1. Stopping service stops/exits Listerner thread right?
2. Make sure this thread doesn't hog CPU. check esxtop

esx_service/vmdk_ops.py
pshahzeb: Retrying after an interval in such cases.

esx_service/utils/vm_listener.py
pshahzeb: Yes it does stop the listener.  Thread doesn't hog CPU. Confirmed using esxtop

esx_service/utils/threadutils.py
```diff
@@ -84,13 +84,25 @@ def protected(*args, **kwargs):
...
+    if args:
+        new_thread = threading.Thread(target=target, args=args)
+    else:
+        new_thread = threading.Thread(target=target)
+
+    if daemon:
+        new_thread.daemon = True
+
+    new_thread.start()
+    logging.debug("Started new thread - %s with target %s and args %s",
```
pdhamdhere: Should this be info?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+        # Proceed further only after you get si instance
+        logging.warn("VMChangeListener couldn't connect to hostd. Retrying after %s seconds", HOSTD_RECONNECT_INTERVAL)
+        time.sleep(HOSTD_RECONNECT_INTERVAL)
+        si = vmdk_ops.get_si()
+
+    pc = si.content.propertyCollector
+    err_msg = create_vm_powerstate_filter(pc, si.content.rootFolder)
+    if err_msg:
+        return err_msg
+    # Start a separate thread to listen to changes
```
pdhamdhere: Is comment correct? You are not really starting a new thread? "start_vm_changelistener " is thread and listen_vm_propertychange is a function call rt?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+POWERSTATE_POWEROFF = 'poweredOff'
+HOSTD_RECONNECT_INTERVAL = 20
+
+def start_vm_changelistener():
+    """
+    Listen to power state changes of VMs running on current host
+    """
+    si = vmdk_ops.get_si()
+
+    while not si:
```
pdhamdhere: I am always scared of infinite loops. Does this loop break if you stop/kill service?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+        # If hostd is not up yet, sleep for a while and try again
+        # Proceed further only after you get si instance
+        logging.warn("VMChangeListener couldn't connect to hostd. Retrying after %s seconds", HOSTD_RECONNECT_INTERVAL)
+        time.sleep(HOSTD_RECONNECT_INTERVAL)
```
pdhamdhere: Fine for now but I would suggest to file low priority issue to change HOSTD_RECONNECT_INTERVAL to exponential backoff.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+                    for change in objectSet.changeSet:
+                        # if the event was powerOff for a VM, set the status of all
+                        # docker volumes attached to the VM to be detached
+                        if change.name != VM_POWERSTATE or change.val != POWERSTATE_POWEROFF:
+                            continue
+
+                        moref = getattr(objectSet, 'obj', None)
+                        # Do we need to alert the admin? how?
+                        if not moref:
+                            logging.error("Could not retrieve the managed object")
```
govint: Can this be "Could not retrieve the VM managed object."

esx_service/utils/vmdk_utils.py
```diff
@@ -342,13 +342,17 @@ def check_docker_volume(dev):
...
+    # name formatting to remove unwanted characters
+    datastore = datastore[1:]
+    disk_path = disk_path.lstrip()
 
-    if disk_path.lstrip().startswith(vmdk_ops.DOCK_VOLS_DIR):
-        return True
-    return False
+    if disk_path.startswith(vmdk_ops.DOCK_VOLS_DIR):
+        # returning the vmdk path for dvs volume
+        return os.path.join("/vmfs/volumes/", datastore, disk_path)
```
govint: Nit - white space.

esx_service/vmdk_ops.py
```diff
@@ -1636,7 +1636,12 @@ def main():
 
         kv.init()
         connectLocalSi()
+
+        # start the daemon. Do all the task to start the listener through the daemon
+        threadutils.start_new_thread(target=vm_listener.start_vm_changelistener,
+                                 daemon=True)
```
govint: Since we have already connected to the local SI do we still need to have the loops to connect to hostd in the listener. I mean hostd is already available at this point.

esx_service/utils/vmdk_utils.py
```diff
@@ -358,7 +362,7 @@ def check_volumes_mounted(vm_list):
         vm = vmdk_ops.findVmByUuid(vm_id)
         if vm:
             for d in vm.config.hardware.device:
-                if check_docker_volume(d):
+                if find_dvs_volume(d):
```
govint: Since return from find_dvs_volume() is unused does it need to return any data at all?

esx_service/utils/vmdk_utils.py
```diff
@@ -342,13 +342,17 @@ def check_docker_volume(dev):
...
+    datastore, disk_path = dev.backing.fileName.rsplit("]", 1)
     logging.info("backing disk name is %s", disk_path)
+    # name formatting to remove unwanted characters
+    datastore = datastore[1:]
+    disk_path = disk_path.lstrip()
 
-    if disk_path.lstrip().startswith(vmdk_ops.DOCK_VOLS_DIR):
-        return True
-    return False
+    if disk_path.startswith(vmdk_ops.DOCK_VOLS_DIR):
```
govint: Could combine 351 and 349 as in the old code?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+
+    for dev in vm_moref.config.hardware.device:
+        # if it is a dvs managed volume, set its status as detached
+        vmdk_path = vmdk_utils.find_dvs_volume(dev)
+        if vmdk_path:
+            logging.info("Setting detach status for %s", vmdk_path)
+            # disk detach and update the status in KV
+            err_msg = vmdk_ops.disk_detach_int(vmdk_path, vm_moref, dev)
+            if err_msg:
```
govint: Pls. also review the use of handle_stale_attach in the vmdk_ops.py and check if what its doing is relevant anymore.

esx_service/vmdk_ops.py
pshahzeb: connectLocalSi succeeded, we would get si and loop won't be executed. If not, then we need to retry to get si and only then start the listener. When host is restarted, hostd takes time to comeup and we don't get a connection to hostd instantly.

esx_service/utils/vm_listener.py
pshahzeb: Yes. Everything exits with the main process stops.

esx_service/utils/vmdk_utils.py
pshahzeb: Using the trimmed disk_path again to construct vmdk_path on 353

esx_service/utils/vm_listener.py
pshahzeb: Since handle_stale_attach should not find any stale attach,  this check doesn't affect directly.
But handling this through separate issue #1201

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+                            logging.error("Could not retrieve the VM managed object.")
+                            continue
+
+                        logging.info("VM poweroff change found for %s", moref.config.name)
+
+                        set_device_detached(moref)
+        except Exception as e:
+            # Do we need to alert the admin? how?
+            logging.error("VMChangeListener: error %s", str(e))
+
```
govint: Can include a test to restart hostd when this thread is waiting. Does the thread continue to get updates once hostd restarts?

Other than this no comments on the code. Only check behavior if hostd restart when this thread is waiting updates.

govint:
@pshahzeb, please add a comment in the listener code as to the version of the VIM API that you are using.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,154 @@
...
+        # If hostd is not up yet, sleep for a while and try again
+        # Proceed further only after you get si instance
+        logging.warn("VMChangeListener couldn't connect to hostd. Retrying after %s seconds", HOSTD_RECONNECT_INTERVAL)
+        time.sleep(HOSTD_RECONNECT_INTERVAL)
+        si = vmdk_ops.get_si()
+
+    pc = si.content.propertyCollector
+    err_msg = create_vm_powerstate_filter(pc, si.content.rootFolder)
+    if err_msg:
+        return err_msg
```
pdhamdhere: Return to whom? VMDKOps will continue to run without VM-PowerEvent-Listener. Is that okay?

esx_service/utils/vm_listener.py
pdhamdhere: Infinite loop still makes me uncomfortable. If hostd is down then all OPs are going to fail anyway. I dont have alternate suggestions though. Panic after 20 retries could be harsh. Fine leaving it as-is for now.

esx_service/utils/vm_listener.py
pdhamdhere: Did you test this? After Listener has started successfully, `/etc/init.d/hostd restart` and then check if you still get notifications?

pshahzeb: Listener doesn't get notifications after hostd restart. Need to connect to hostd again, get new SI, property collector and start listening again.

esx_service/utils/vm_listener.py
pshahzeb: Yes. Although this error is rare to occur, we can only return with vmdkops running without vm listener. Reconnecting to hostd will not help here.

esx_service/vmdk_ops.py
```diff
@@ -969,12 +969,7 @@ def connectLocalSi(force=False):
                 version=newestVersions.Get('vim'))
         except Exception as e:
             logging.exception("Failed to create the local Service Instance as 'dcui', continuing... : ")
-    elif force:
-        logging.warning("Reconnecting to the local Service Instance")
-        _service_instance = pyVim.connect.Connect(
-            host='localhost',
-            user='dcui',
-            version=newestVersions.Get('vim'))
```
pshahzeb: This call is exactly the same as on L966 but without try catch block. Exception can occur here too. Removing the force flag and resetting _service_instance and calling connectLocalSi for cleaner connect.

pshahzeb:
@govint @pdhamdhere Fixed the hostd restart case.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,205 @@
...
+import log_config
+import vmdk_utils
+import vmdk_ops
+
+from pyVmomi import VmomiSupport, vim, vmodl
+# vim api version used - version11
+
+VM_POWERSTATE = 'runtime.powerState'
+POWERSTATE_POWEROFF = 'poweredOff'
+HOSTD_RECONNECT_INTERVAL = 5 #approx time for hostd to comeup is 10-15 seconds
```
pdhamdhere: I would start with 2secs sleep and increase HOSTD_RECONNECT_ATTEMPT if absolutely needed so that we do not miss any poweroff events.

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,205 @@
...
+
+                        set_device_detached(moref)
+            version = result.version
+        # Capture hostd down exception
+        except http.client.RemoteDisconnected as e:
+            logging.error("VMChangeListener: Hostd connection error %s", str(e))
+            version = ''
+            # Need to get new SI instance, create a new property collector and property filter
+            # for it. Can't use the old one due to stale authentication error.
+            pc, error_msg = get_propertycollector()
```
pdhamdhere: This is fine. Just curious why not exit thread and start a new thread start_vm_changelistener?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,205 @@
...
+        si = vmdk_ops.get_si()
+
+        # exponential backoff for next retry
+        reconnect_interval += reconnect_interval
+
+    # Proceed further only after you get si instance
+    if not si:
+        # could not connect to hostd even after retries
+        # Something is seriously wrong
+        return None, "Hostd Down!"
```
pdhamdhere: "Hostd Down!" => "Unable to connect to hostd. Verify that vmware-hostd is running."

esx_service/utils/threadutils.py
```diff
@@ -84,13 +84,25 @@ def protected(*args, **kwargs):
...
+        new_thread = threading.Thread(target=target, args=args)
+    else:
+        new_thread = threading.Thread(target=target)
+
+    if daemon:
+        new_thread.daemon = True
+
+    new_thread.start()
+    logging.info("Started new thread : %s with target %s and args %s",
+                  new_thread.ident, target, args)
```
govint: Nit - does the thread have a name identifying itself in the logs.

esx_service/utils/vm_listener.py
govint: Agree, codewise for debugging it may be good to exit. Or, can this function return to caller which then runs thru init, setup and wait. In fact the code should be one function that does just that init (get si) - set up (filter setup/prop collector) - wait (listen for events).  If these return then handle the case accordingly in the top level function.

esx_service/utils/threadutils.py
```diff
@@ -84,13 +84,25 @@ def protected(*args, **kwargs):
...
 
-def start_new_thread(target, args):
+def start_new_thread(target, args=None, daemon=False):
     """Start a new thread"""
-    threading.Thread(target=target, args=args).start()
+
+    new_thread = None
+
+    if args:
+        new_thread = threading.Thread(target=target, args=args)
```
msterin: nit: is't L92-95 is the same as simply L93 ? Or threading package has different default for args ?

esx_service/utils/vm_listener.py
```diff
@@ -0,0 +1,207 @@
...
+
+    si = vmdk_ops.get_si()
+
+    reconnect_interval = HOSTD_RECONNECT_INTERVAL
+    for i in range(HOSTD_RECONNECT_ATTEMPT):
+        if si:
+            break
+
+        # If hostd is not up yet, sleep for a while and try again
+        logging.warn("VMChangeListener couldn't connect to hostd.")
```
msterin: nit  - maybe 1 logging.warning is better ? To prevent log pollution :-)

pshahzeb:
Getting this merged. Will do a minor PR for nits (Don't want to commit post LGTMs)

PR #1178, opened 2017-04-17
Rescanning refCounts if first attempt during driver init fails

### LINKED ISSUE #1159: Volume stays attached after abrupt docker plugin kill and container exit
This is seen when the vDVS managed plugin is killed forcefully while the volume is still being used by a container. Reproduced as follows:

1. Create a volume and attach it to a container.
```
root@photon-xVmYMbyTn [ ~ ]# docker volume create -d vsphere:latest --name=testVol1_MP
testVol1_MP
root@photon-xVmYMbyTn [ ~ ]# docker volume ls
vsphere:latest      testVol1_MP@sharedVmfs-0

root@photon-xVmYMbyTn [ ~ ]# docker run -it --volume-driver=vsphere:latest -v testVol1_MP@sharedVmfs-0:/vol1 --name ub ubuntu
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
d54efb8db41d: Pull complete
f8b845f45a87: Pull complete
e8db7bf7c39f: Pull complete
9654c40e9079: Pull complete
6d9ef359eaaa: Pull complete
Digest: sha256:dd7808d8792c9841d0b460122f1acf0a2dd1f56404f8d1e56298048885e45535
Status: Downloaded newer image for ubuntu:latest
root@53fa5061e340:/# df -h
Filesystem                                       Size  Used Avail Use% Mounted on
overlay                                          7.8G  721M  6.7G  10% /
tmpfs                                            2.0G     0  2.0G   0% /dev
tmpfs                                            2.0G     0  2.0G   0% /sys/fs/cgroup
/dev/disk/by-path/pci-0000:03:00.0-scsi-0:0:0:0   93M  1.6M   85M   2% /vol1
/dev/root                                        7.8G  721M  6.7G  10% /etc/hosts
shm                                               64M     0   64M   0% /dev/shm
tmpfs                                            2.0G     0  2.0G   0% /sys/firmware
root@53fa5061e340:/#
root@53fa5061e340:/#
```
2. volume is indicated as attached to the VM
```
[root@sc-rdops-vm16-dhcp-237-252:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume       Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to  Access      Attach-as               Created By  Created Date
-----------  ------------  --------  --------  ----  ----------  ------  -----------  -----------  ----------  ----------------------  ----------  ------------------------
testVol1_MP  sharedVmfs-0  _DEFAULT  100MB     15MB  ext4        N/A     thin         photon       read-write  independent_persistent  photon      Mon Apr 10 19:21:23 2017
```
3. Disable the plugin and kill the container from another terminal
```
root@photon-xVmYMbyTn [ ~ ]# docker plugin disable vsphere:latest -f
vsphere:latest

root@photon-xVmYMbyTn [ ~ ]# docker rm -f 53fa5061e340
53fa5061e340
root@photon-xVmYMbyTn [ ~ ]# docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
```
4. Volume still stays attached.
```
[root@sc-rdops-vm16-dhcp-237-252:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume       Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to  Access      Attach-as               Created By  Created Date
-----------  ------------  --------  --------  ----  ----------  ------  -----------  -----------  ----------  ----------------------  ----------  ------------------------
testVol1_MP  sharedVmfs-0  _DEFAULT  100MB     15MB  ext4        N/A     thin         photon       read-write  independent_persistent  photon      Mon Apr 10 19:21:23 2017
```

While this reproduction is too harsh and destructive, such related and haywire behavior could happen on the docker side leading to not getting an unmount request on esx side.
Note: VM was powered on the time. So just monitoring VM events for cleaning up unattached volumes don't help much as pointed out in #369

pshahzeb:
As per #1159, the only bother issue was when docker itself is restarted and we timeout during refCounting while the init is in process. Need to rescan for refcounting again.

1. If the attempt during first refCounting during driver init fails, create a timer and
redo the refCounting after a delay.

2. Minor nit to replace datastore variable to get rid of a warning.

Fixes #1159

Attaching a volume to a container and letting it run.
Killing docker with -9
Refcounting timeouts:
```
2017-04-17 15:31:16.814522898 -0700 PDT [WARNING] Received signal signal=terminated
2017-04-17 15:31:18.967593466 -0700 PDT [INFO] No config file found. Using defaults.
2017-04-17 15:31:18.967847163 -0700 PDT [WARNING] Failed to load config file /etc/docker-volume-vsphere.conf: open /etc/docker-volume-vsphere.conf: no such file or directory
2017-04-17 15:31:18.967952849 -0700 PDT [INFO] Starting plugin driver=vsphere log_level=info config="/etc/docker-volume-vsphere.conf"
2017-04-17 15:31:18.968065735 -0700 PDT [INFO] Plugin options - port=1019
2017-04-17 15:31:18.968163955 -0700 PDT [INFO] Getting volume data from unix:///var/run/docker.sock
2017-04-17 15:31:20.968893227 -0700 PDT [INFO] Can't connect to unix:///var/run/docker.sock due to (An error occurred trying to connect: context deadline exceeded), skipping discovery
2017-04-17 15:31:20.969059139 -0700 PDT [INFO] Refcounting failed. Scheduling again after 20s seconds
2017-04-17 15:31:20.969078401 -0700 PDT [INFO] Docker VMDK plugin started version="vSphere Volume Driver v0.4" port=1019 mock_esx=false
2017-04-17 15:31:20.969452779 -0700 PDT [INFO] Going into ServeUnix - Listening on Unix socket address="/run/docker/plugins/vsphere.sock"
```
Volume stays attached
```
[root@sc-rdops-vm05-dhcp-138-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume  Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to   Access      Attach-as               Created By    Created Date
------  ------------  --------  --------  ----  ----------  ------  -----------  ------------  ----------  ----------------------  ------------  ------------------------
test1   sharedVmfs-0  _DEFAULT  100MB     15MB  ext4        N/A     thin         ubuntu-VM0.0  read-write  independent_persistent  ubuntu-VM0.0  Mon Apr 17 22:28:36 2017
```

After 20 secods:
Refcounting happens again:
```
2017-04-17 15:31:40.969386829 -0700 PDT [INFO] Getting volume data from unix:///var/run/docker.sock
2017-04-17 15:31:40.985203103 -0700 PDT [INFO] Initiating recovery unmount. name="test1@sharedVmfs-0" refcnt=0 mounted=true dev="/dev/disk/by-path/pci-0000:03:00.0-scsi-0:0:0:0"
2017-04-17 15:31:41.211017706 -0700 PDT [INFO] Discovered 1 volumes in use.
2017-04-17 15:31:41.211100196 -0700 PDT [INFO] Volume name=test1@sharedVmfs-0 count=0 mounted=true device='/dev/disk/by-path/pci-0000:03:00.0-scsi-0:0:0:0'
2017-04-17 15:31:41.211112417 -0700 PDT [INFO] Refcounting reattempt succeded
```

Volume status unmounted:
```
[root@sc-rdops-vm05-dhcp-138-173:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume  Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to  Access      Attach-as               Created By    Created Date
------  ------------  --------  --------  ----  ----------  ------  -----------  -----------  ----------  ----------------------  ------------  ------------------------
test1   sharedVmfs-0  _DEFAULT  100MB     15MB  ext4        N/A     thin         detached     read-write  independent_persistent  ubuntu-VM0.0  Mon Apr 17 22:28:36 2017
```


esx_service/utils/vmdk_utils.py
```diff
@@ -73,10 +73,11 @@ def init_datastoreCache(force=False):
     from local ESX host. force=True will force it to ignore current cache
     and force init
     """
+    global datastores
+
```
govint: The comment on @lipingxue's change was to place this line under the "with....." statement below :-)

pshahzeb: I see. I was seeing a warning because of its usage in logging statement above with statement. So I moved it upwards.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -77,7 +79,23 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
...
+	if refCountStatus == false {
+		timer := time.NewTimer(refCountDelay)
+		log.Infof("Refcounting failed. Scheduling again after %s seconds", refCountDelay)
+		go func() {
+			<-timer.C
+			refCountStatus = d.refCounts.Init(d, mountDir, driverName)
+			if refCountStatus {
+				log.Infof("Refcounting reattempt succeded")
+			} else {
+				log.Infof("Refcounting reattempt failed")
```
govint: This should be a warning log. Preferably have an error status returned vs. a boolean so the user knows exactly why the refcount recovery failed.

There's no recovery if this attempt fails and volumes will be left attached to the VM till the VM is powered off and the other fix comes into play to remove volumes still attached to the VM. Which can be a long time. Do we want to do the recovery a fixed number of times before giving up. Like, docker may have a problem and may not be up for some time and the retry will be useless in that time period. A few fixed retries may be good.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -153,7 +153,7 @@ func newRefCount() *refCount {
 // Init Refcounts. Discover volume usage refcounts from Docker.
 // This functions does not sync with mount/unmount handlers and should be called
 // and completed BEFORE we start accepting Mount/unmount requests.
-func (r RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string) {
+func (r RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string) bool {
```
govint: Should be a specific error so we know exactly why the refcounting failed.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: This can be done. But I am not sure if is really required.
But the first refcount attempt might timeout only when both docker and plugin are being are coming up. In such cases, we retry after sufficient time cushion, taking into consideration the docker backoff time, ensuring it has come up. Can you elaborate on why the docker  might not reply to the api calls the second time?

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: Done

esx_service/utils/vmdk_utils.py
pshahzeb: Reverted. Corrected in #1181

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -77,7 +79,23 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
 		}
 	}
 
-	d.refCounts.Init(d, mountDir, driverName)
+	err := d.refCounts.Init(d, mountDir, driverName)
+	// If refcounting wasn't successfull, schedule one again
+	if err != nil {
```
govint: The rescheduling of the refcount scan seems like a functionality thats owned by the refcounting module. The VMDK driver shouldn't have to care what the refcount module does to handle refcount init failures. It just calls refcnt.Init() and that handles any retry mechanism. In fact that way refcount module has more flexibility on handlng scan retries than if Init() is re-done from here.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +152,33 @@ func newRefCount() *refCount {
...
-// This functions does not sync with mount/unmount handlers and should be called
-// and completed BEFORE we start accepting Mount/unmount requests.
+// tries to calculate refCounts for dvs volumes. If failed, triggers a timer
+// based reattempt to schedule scan after a delay
 func (r RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string) {
+	err := r.calculate(d, mountDir, name)
+	// If refcounting wasn't successfull, schedule one again
+	if err != nil {
+		log.Infof("Refcounting failed: (%v). Scheduling again after %s seconds", err, refCountDelay)
+		r.retryCalculate(d, mountDir, name, refCountRetryAttempts-1)
```
msterin: nit: spaces around -

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +152,33 @@ func newRefCount() *refCount {
...
+// create a timer to calculate refcount after a delay. If failed, retry again
+// until retry attempt limit reached
+func (r RefCountsMap) retryCalculate(d drivers.VolumeDriver, mountDir string, name string, attemptLeft int) {
+	timer := time.NewTimer(refCountDelay)
+	go func() {
+		<-timer.C
+		err := r.calculate(d, mountDir, driverName)
+		if err != nil && attemptLeft > 0 {
+			log.Infof("Refcounting failed: (%v). Scheduling again after %s seconds", err, refCountDelay)
+			r.retryCalculate(d, mountDir, name, attemptLeft-1)
```
msterin: nit: it's better to pass delay to the retryCalculated  instead of using global

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -181,7 +203,7 @@ func (r RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string)
 	err = r.discoverAndSync(c, d)
```
pdhamdhere: discoverAndSync purges existing refCnt map. I don't see any locking. What if mount/unmount requests comes in when we are in retryCalculate->calculate->discoverAndSync?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -92,6 +92,8 @@ const (
 	DockerUSocket           = "unix:///var/run/docker.sock"
 	defaultSleepIntervalSec = 1
 	dockerConnTimeoutSec    = 2
+	refCountDelay           = 20 * time.Second
```
pdhamdhere: Any specific reason it is 20 and not 5 Secs or 10 Secs?

vmdk_plugin/utils/refcount/refcnt.go
msterin: yes, good point. I was thinking "in the worst case, the mount request picks up the earlier data, which is not different from the recalc() simply hitting a few seconds later" but I forgot that the mount does modify the entry.
Lock is needed, per entry or per map. Of course the problem is - if vDVS waits on this lock it cannot reply to Docker. Docker (at least in the past) would also lock internally on the mount table, we got a deadlock and back to square 1..

Maybe we should just fail mount/unmount requests when recalc is in flight ?

pshahzeb: I am a bit skeptical about failing unmount requests while recalc is in flight. The reason being docker ignores failures in unmount. I tried this with a small test.

What we can do is as follows:
While retrying to populate refCount: do it in the reconcile map. The reason being that talking to docker using engine api might be consuming. So we allow parallel mount unmounts during this time.

Once we have reconcile-refmap populated, swap this with original refCount under the refCount mutex. Also sync the mount points  and remove unwanted mounts while maintaing the mutex. After this is done, release the mutex. Since the mount and unmount request too access the refcount, this mutex will hang those requests for a bit.
I understand that using a mutex like this doesn’t principally sound well because unmounting volumes might take a while.

But if we are gonna fail mount/unmount, we might as well hang them for a while. 
CC @govint

pdhamdhere: Seems reasonable and correct approach to me.

govint: Failing the mount requests means containers won't start after the plugin has ack'ed that its done init'ing. Unmounts can't be failed because then we will be missing unmount refcounts. Still needs code to handle refcounts that may be getting restored for the very same containers by the refcount-recalc thread.

This is different from what we discussed offline.  Frankly we need two bitmaps - an active one and one to handle collecting refcounts for containers that started before the plugin restarted.

pdhamdhere: >Frankly we need two bitmaps - an active one and one to handle collecting refcounts

@govint Yes and I believe @pshahzeb is proposing same;
Active one => Original refCount map
Temporary => reconcile-refmap

msterin: (1) is there  indeed a deadlock generated after docker API is initialized() ? 
If we did not check (the code or a test) , a small test is needed (e.g. recount or getMountedVolumesFromDocker() inside of a Create() would do it). @pshahzeb 

(2) In case the deadlock is indeed generated: for 2 maps suggestions (which I like because we ended up doing the same for DS cache in the server side :-)). how are we planning to track changes to the actual bitmap done while AFTER the "new" one slot is filled but BEFORE the maps were swapped ? 

(3) Since we can wait a couple of minutes if we deadlock on start, why not do the same if we deadlock later on ? Similar to ethernet conflict resolution :-) - "back off and retry in a random amount of time (within min/max)" ?

pshahzeb: @msterin I did the small test. Did the refCount recalc as part of Create. There is no deadlock.

msterin: What about during  mount() ?
If there is no deadlock, then it's just a.matter of taking refcnt lock for recalc

pshahzeb: Oh doing it in mount does end up in a deadlock. Seeing it consistently.

pshahzeb:
Approach 1:

We can keep it simple by using a reconcileLock which is taken during the reconciling start and end.
It is also taken at the start of mount and unmount calls. When a deadlock occurs, we timeout releasing the reconcile lock and reschedule it after some time. Then mounts and unmounts proceed. 
This makes the change very simple.

But there can be some undesired unmounts, specifically issued by containers for which we couldn’t initialize the refcounts. We skip unmount when refcount value is 0 or entry is not available for a volume and trigger. But still there are cases when we do undesired unmount.
eg:
Container A attaches v1
Container B attaches v1
plugin restarts -  refcounting failed
Container C attaches v1
Container D attaches v1
Container A exits
Container B exits - this unmount will actually unmount and Container C and D wont be able to access volume

Approach B:
Use the container Ids in refCountMaps and use the plugin start time to separate the containers starter before plugin start and after plugin start
Maintain 2 maps: active map for current mount and unmount. 
reconcile map populated during reconciliation for containers started before the plugin start.
Mount adds container id in list against volume name in active refcount map
Unmount removes ID. If ID is not present in refcountmap for the volume, (this container was started before plugin start) then don’t unmount. This will be handled in reconciliation.

The point when reconcile-map was successfully populated:
acquire refcount mutex lock:
merge the active map and reconcile map
get the mounted volumes on VM
unmount unnecessary volumes
release refcount mutex.

@msterin @govint @pdhamdhere Lets reach a consensus on which is a better approach and worth the probability of issue happening in real life.

pshahzeb:
For approach B: actually there isn't the need for reconcile map and merging it later given that refcount map Incr and Decr happen under a mutex. Just while unmount, if the id is not present in refcountmap, don't unmount, let the reconciliation logic handle it.

pshahzeb:
@govint @pdhamdhere @msterin 
Please review the latest change. It is based on the dual map approach explained above. But with storing mountIDs (ones we get in mount and unmount) and not containerIDs as mentioned above (my bad!).

pshahzeb:
Tested the refcount reattempt scanning by sleeping for a while inside a discoverAndSync and doing a parallel mount from docker. Refcounting breaks and get rescheduled again.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +176,49 @@ func newRefCount() *refCount {
...
-// This functions does not sync with mount/unmount handlers and should be called
-// and completed BEFORE we start accepting Mount/unmount requests.
-func (r RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string) {
+// return if refcounting has been successfull
+func (r *RefCountsMap) GetSuccess() bool {
+	return r.refcntSuccess
+}
+
+// dirty the background refcount process
+func (r *RefCountsMap) Dirty() {
```
shaominchen: Nit: can we use a verb or verb phrase to name a function? For this function, we may rename it to markDirty().

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -434,6 +442,14 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		return volume.Response{Err: ""}
 	}
 
+	// defensive approach to not unmount volumes until refcounting has been successful
+	if d.refCounts.GetSuccess() != true {
+		log.WithFields(
+			log.Fields{"name": r.Name},
+		).Info("Refcounting hasn't been successfull. Skipping unmount request")
```
pdhamdhere: typo: successfull => successful

"Refcounting hasn't been successfull. Skipping unmount request" =>
"Reference count map is not initialized. Unmount is delayed".

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -434,6 +442,14 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
...
+		return volume.Response{Err: "Refcount Error"}
```
pdhamdhere: How does Docker behave for errors on unmount path?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -92,12 +93,21 @@ const (
 	DockerUSocket           = "unix:///var/run/docker.sock"
 	defaultSleepIntervalSec = 1
 	dockerConnTimeoutSec    = 2
+	refCountDelayMin        = 10
```
pdhamdhere: To be consistent with rest of variables, I would add postfix of Sec e.g. refCountDelayMinSec, refCountDelayMaxSec.

Why 10-20Sec. Shouldn't we be more agressive 1-10secs?

vmdk_plugin/utils/refcount/refcnt.go
pdhamdhere: I agree. Change it to SetDirty or MarkDirty?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +176,49 @@ func newRefCount() *refCount {
...
+func (r *RefCountsMap) retryCalculate(d drivers.VolumeDriver, mountDir string, name string, attemptLeft int) {
+	// create a random delay between max and min delay
+	delay := rand.Intn(refCountDelayMax - refCountDelayMin + 1) + refCountDelayMin
+	log.Infof("Scheduling again after %s seconds", delay)
+	timer := time.NewTimer(time.Duration(delay) * time.Second)
+
+	go func() {
+		<-timer.C
+		err := r.calculate(d, mountDir, driverName)
+		if err != nil && attemptLeft > 0 {
```
pdhamdhere: We should exit if we couldn't initialize refCnts in refCountRetryAttempts. If we let plugin continue then disks will never be detached.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -114,8 +124,15 @@ type refCount struct {
...
 // RefCountsMap struct
 type RefCountsMap struct {
+	// does this need to be atomic?
+	// this is just one time state transition
+	refcntSuccess bool
+
 	refMap map[string]*refCount // Map of refCounts
 	mtx    *sync.RWMutex        // Synchronizes RefCountsMap ops
+
+	auxRefMap  map[string]*refCount // do refcount
```
pdhamdhere: Why temporary map is part of global struct?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -114,8 +124,15 @@ type refCount struct {
...
+	curr_state *refcount_state
```
pdhamdhere: Rather than nesting, I would just get rid of refcount_state and add isDirty and dirtyMtx to RefCountsMap structure

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +385,49 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
+		for k, v := range r.auxRefMap {
+			r.refMap[k] = v
+		}
 
-	r.getMountInfo()
-	r.syncMountsWithRefCounters(d)
+		// Check that refcounts and actual mount info from Linux match
+		// If they don't, unmount unneeded stuff, or yell if something is
+		// not mounted but should be (it's error. we should not get there)
+		r.getMountInfo()
```
pdhamdhere: These internally acquire "r.mtx.lock". I dont see any obvious issues but want to make sure you have thought through;
```
r.curr_state.mtx.Lock()
    r.mtx.Lock()
    r.mtx.Unlock()
r.curr_state.mtx. Unlock()
```

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: Error is logged on docker side but unmount proceeds

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: Yes. In this section where state si clean, no mount/unmount can enter since they will wait to make state dirty. So we can proceed to manipulate the actual refmap

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -131,8 +125,8 @@ type RefCountsMap struct {
 	refMap map[string]*refCount // Map of refCounts
 	mtx    *sync.RWMutex        // Synchronizes RefCountsMap ops
 
-	auxRefMap  map[string]*refCount // do refcount
-	curr_state *refcount_state
+	isDirty  bool        // flag to check reconciling has been interrupted
+	dirtyMtx *sync.Mutex // protec the above flag
```
pdhamdhere: typo: protect
how about "Synchronizes access to isDirty flag?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -92,6 +93,9 @@ const (
 	DockerUSocket           = "unix:///var/run/docker.sock"
 	defaultSleepIntervalSec = 1
 	dockerConnTimeoutSec    = 2
+	refCountDelayMin        = 2
```
pdhamdhere: Can you add "Sec" postfix like other variables for e.g. refCountDelayMinSec & refCountDelayMaxSec

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +166,51 @@ func newRefCount() *refCount {
...
+	r.dirtyMtx.Lock()
+	r.isDirty = true
+	r.dirtyMtx.Unlock()
+}
+
+// tries to calculate refCounts for dvs volumes. If failed, triggers a timer
+// based reattempt to schedule scan after a delay
+func (r *RefCountsMap) Init(d drivers.VolumeDriver, mountDir string, name string) {
+	err := r.calculate(d, mountDir, name)
+	// If refcounting wasn't successfull, schedule one again
```
pdhamdhere: Typo: successful

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -577,6 +580,9 @@ func (d *VolumeDriver) Mount(r volume.MountRequest) volume.Response {
 func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Unmounting Volume ")
 
+	// dirty the background refcounting process if any
+	d.refCounts.MarkDirty()
+
 	// if the volume is still used by other containers, just return OK
```
msterin: if it's getting dirty there is no reason to manipulate refcount, is there ? 
I'd expect somethinng like 
```go
 if d.refCounts.MarkDirty() != nil {
   logWithFields(...) # lines 596-600
}

refcnt, err= d.decrRefCount(r.Name)
...

```

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -586,6 +592,14 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		).Error("Refcount error - still trying to unmount...")
 	}
 
+	// defensive approach to not unmount volumes until refcounting has been successful
+	if d.refCounts.GetSuccess() != true {
+		log.WithFields(
+			log.Fields{"name": r.Name},
+		).Info("Reference count map is not initialized. Unmount is delayed")
+		return volume.Response{Err: "Refcount Error"}
```
msterin: Why do we want the unmount 'refcount error' in this case exposed to the docker command ? I think it will only confuse the upstairs. we need to pretend all is good and return success

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -418,6 +423,9 @@ func (d *VolumeDriver) Mount(r volume.MountRequest) volume.Response {
 func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Unmounting Volume ")
 
+	// dirty the background refcounting process if any
+	d.refCounts.MarkDirty()
```
msterin: same comment as for photon driver.  (not part of this PR) but the fact we have copy and paste between driver is not nice

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -114,8 +118,15 @@ type refCount struct {
 
 // RefCountsMap struct
 type RefCountsMap struct {
+	// does this need to be atomic?
+	// this is just one time state transition
+	refcntSuccess bool
```
msterin: just use package sync.atomic

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +166,51 @@ func newRefCount() *refCount {
...
+	timer := time.NewTimer(time.Duration(delay) * time.Second)
+
+	go func() {
+		<-timer.C
+		err := r.calculate(d, mountDir, driverName)
+		if err != nil && attemptLeft > 0 {
+			log.Infof("Refcounting failed: (%v).", err)
+			r.retryCalculate(d, mountDir, name, attemptLeft-1)
+		} else if attemptLeft <= 0 {
+			panic(fmt.Sprintf("Failed to talk to docker to calculate volumes usage. Please restart docker"))
```
msterin: so what happens on panic() in managed plugin ?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +166,51 @@ func newRefCount() *refCount {
...
+		log.Infof("Refcounting failed: (%v).", err)
+		r.retryCalculate(d, mountDir, name, refCountRetryAttempts-1)
+	}
+}
+
+// create a timer to calculate refcount after a delay. If failed, retry again
+// until retry attempt limit reached
+func (r *RefCountsMap) retryCalculate(d drivers.VolumeDriver, mountDir string, name string, attemptLeft int) {
+	// create a random delay between max and min delay
+	delay := rand.Intn(refCountDelayMaxSec-refCountDelayMinSec+1) + refCountDelayMinSec
```
pshahzeb: gofmt we use removes the spaces somehow :|

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +166,51 @@ func newRefCount() *refCount {
...
+	log.Infof("Scheduling again after %s seconds", delay)
+	timer := time.NewTimer(time.Duration(delay) * time.Second)
+
+	go func() {
+		<-timer.C
+		err := r.calculate(d, mountDir, driverName)
+		if err != nil && attemptLeft > 0 {
+			log.Infof("Refcounting failed: (%v).", err)
+			r.retryCalculate(d, mountDir, name, attemptLeft-1)
```
msterin: I am generally suspicious of recursion in  simple operations. 
Why  not something like (pseudo  code)
```
set timer
while (attemptsLeft > 0) {
    <-timer.C
    if calculate()  == nil {
        return nil // all good
   }
   attemptsLeft--
   reset timer
}
 panic("all bad")
```   
it's much easier to see what's going on and less error prone

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: Mark dirty will mark the refcount reattempt dirty. So that the background refcounting breaks if it was in process.
So just mark it dirty and proceed.

lines 596-600 check for a completely separate flag about refcounting in whole has been successful or not.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -277,14 +334,14 @@ func matchNameforVMDK(mount_source string) bool {
...
 
-	r.mtx.Lock() // Lock the RefCountsMap to purge the refcounts
-	for name := range r.refMap {
-		delete(r.refMap, name)
-	}
-	r.mtx.Unlock() // Unlock.
+	r.dirtyMtx.Lock()
+	r.isDirty = false
+	r.dirtyMtx.Unlock()
+
```
msterin: what if Mount() sets dirty right before the lock on L340 , but the actual mount happens after this discovery thread is done talking to Docker? Won't this will reset the Dirty flag and the mount() will be unaccounted for ?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: Yes. We can have a separate issue for to refactor the common code between the drivers

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +375,49 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
 		for _, mount := range containerJSONInfo.Mounts {
 			// check if the mount location belongs to vsphere plugin
 			if matchNameforVMDK(mount.Source) {
-				r.Incr(mount.Name)
-				log.Debugf("  name=%v (driver=%s source=%s) (%v)",
+				// populate temp map
+				rc := auxRefMap[mount.Name]
```
msterin: this looks way too complicated. I thought the idea was 
* set Dirty
* start map discovery, with timeout and retries
* if dirty: 
   * no refcounts on mount
   * no unmount, ,just return success
* when map is properly build
   * unmount hanging stuff at the end
    * clear dirty flag

What is done now does not seem to be any different as far as locks and races, but uses 2 maps for some reason. Can you elaborate why the first map is even needed when Dirty ?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +375,49 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
 		}
 	}
 
-	// Check that refcounts and actual mount info from Linux match
-	// If they don't, unmount unneeded stuff, or yell if something is
-	// not mounted but should be (it's error. we should not get there)
+	// check if the background refcount was dirtied
+	r.dirtyMtx.Lock()
+	if r.isDirty == false {
+		// if not, copy auxRefMap into refMap and unmount unwanted volumes
```
msterin: Why even bother to copy if we can just ignore (and we already do ignore) the map when dirty ?  What am I missing ?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: #1212

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +375,49 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
+
+		// Check that refcounts and actual mount info from Linux match
+		// If they don't, unmount unneeded stuff, or yell if something is
+		// not mounted but should be (it's error. we should not get there)
+		r.getMountInfo()
+		r.syncMountsWithRefCounters(d)
+		// mark reconciling success so that further unmounts can instantly be processed
+		r.refcntSuccess = true
+
+		r.dirtyMtx.Unlock()
```
msterin: this is generally done with `defer r.dirtyMtx.Unlock()`

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +375,49 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
+	} else {
```
msterin: its better to say 
```
if r.isDirty == true {
   return "err msg"
}

<the rest of code is there, with no offset>
return nil
```

msterin:
@pshahzeb - can you add comments on what exactly is the algorithm  ? Either I am misreading it. or there are races.

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: docker logs it as artificial panic and tries to restarts the plugin

pshahzeb:
@msterin 

The algorithm is as follows:

bool to maintain if refcounting is refcounting was clean (clean it in beginning and check it in the end.) any simultaneous mounts and unmounts mark the refcounting dirty.
If dirty, return error and get it scheduled again.
If refcounting was clean, copy the content of temp map (in which we did the background refcounting) into main refcounting. Find out the unnecessary mount points and unmount them if necessary. Mark a bool as the refcounting has been successful.

Meanwhile in unmount, if refcounting hasn’t been successful yet, don't unmount. Skip it.

We need to flush the map before we start the discovery loop to avoid double counting. But this might lead to loss on some accounted mount points, if discovery loop fails. Hence the temp map.

msterin:
> The algorithm is as follows:
> ...

Ah, OK, It's close to what I read. 
Please see my comment about potential race in the begginning of discoverAndSync(). 

> ...
> If refcounting was clean, copy the content of temp map (in which we did the background refcounting) > into main refcounting. Find out the unnecessary mount points and unmount them if necessary. Mark a bool as the refcounting has been successful.

I still do not see why 2 maps are needed - one (and only one) of them is used before the discovery completed (auxMap) and one and only one after (auxMap copied to r). So why not use 1 ?

> We need to flush the map before we start the discovery loop to avoid double counting. 
I still do not understand why do we need to count mounts until discovery was successful

> But this might lead to loss on some accounted mount points, if discovery loop fails. Hence the temp map.
If the discovery fails we don't care what's in the map - we still going to skip unmounts and wait for next discovery , no ?

pshahzeb:
> Ah, OK, It's close to what I read.
Please see my comment about potential race in the begginning of discoverAndSync().

I see it. Still thinking how to handle. Any hints?

>I still do not see why 2 maps are needed - one (and only one) of them is used before the discovery completed (auxMap) and one and only one after (auxMap copied to r). So why not use 1 ?

Before discovery completes we need two maps. One to try out refcounting work and another to cater mounts/unmounts in parallel till we complete the discovery.

msterin:
FWIW I  I'd suggest a slight change to the algorithm which seems to eliminate the race:

* on entry to mount/unmount, take a lock, check GetSuccess() , 
  * If success , drop the lock and finish normal procedure with refcount and single map, no dirty flags
  * if failure,  set local var to skip refcounting and actual unmount, finish [un]mount operation under the same lock , set Dirty before the exit, then release the lock. Note that any ops under this lock in discoverAndSync() should NOT include docker API.

* on entry to discoverAndSync(), 
  * take the lock, clean up isDirty, release the lock (same as now - but there is no gap in Mount/Unmount lock coverage)

* in discoverAndSYnc()  in Docker interrogation loop - check 'lock/IsDirty/unlock' on each loop and break / reschedule if true
* on exit from discoverAndSync():
    * take the lock , if dirty reschedule. If clean, set success and release the lock

pshahzeb:
@msterin @Prashant please have a look.

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -336,6 +336,74 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
+	fstype, exists := status[fsTypeTag]
+	if !exists {
+		fstype = fs.FstypeDefault
+	}
+
+	skipAttach := false
+	// If the volume is already attached to the VM, skip the attach.
+	if state, stateExists := status["State"]; stateExists {
+		if strings.Compare(state.(string), "DETACHED") != 0 {
+			skipAttach = true
```
msterin: that is the wrong way to check mount. You are asking ESX to give you KV metadata file and use that as substitute for mount info. Instead,just check the presense of /mnt/vmdk/<volume> in /proc/mounts (we already have this code for analysing mounted).
BTW - <volume> should probably be <volume@datastore>

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -515,73 +583,45 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
-	// Mount the volume and for now its always read-write.
-	mountpoint, err := d.MountVolume(r.Name, fstype.(string), status["ID"].(string), false, skipAttach)
-	if err != nil {
-		log.WithFields(
-			log.Fields{"name": r.Name, "error": err.Error()},
-		).Error("Failed to mount ")
-		d.decrRefCount(r.Name)
-		return volume.Response{Err: err.Error()}
+		// then unlock the state
+		d.refCounts.StateMtx.Unlock()
```
msterin: please use `defer` to unlock on exit

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -515,73 +583,45 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
 
 // Unmount request from Docker. If mount refcount is drop to 0,
 // Unmount and detach from VM
 func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Unmounting Volume ")
 
-	// dirty the background refcounting process if any
-	d.refCounts.MarkDirty()
+	// lock the state
+	d.refCounts.StateMtx.Lock()
```
msterin:
```go
d.refCounts.StateMtx.Lock()
defer d.refCounts.StateMtx.Unlock()
```

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -515,73 +583,45 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+
+	if d.refCounts.GetSuccess() != true {
+		// if refcounting hasn't been succesful,
+		// no refcounting, no unmount. All unmounts are delayed
+		// until we succesfully populate the refcount map
+		d.refCounts.MarkDirty()
+		d.refCounts.StateMtx.Unlock()
```
msterin: MarkDirty() does not need to lock anymore, so simple var will make do

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: done

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: its volume@datastore if specified in that format while attaching. else if just volume is specified then it is just /mnt/vmdk/volume

pshahzeb: On our plugin side,
docker run -it --volume-driver=vsphere -v test1:/vol1 --name bb busybox
docker run -it --volume-driver=vsphere -v test1@sharedVmfs-0:/vol1 --name bb2 busybox

will result in two mounts
/mnt/vmdk/test1
/mnt/vmdk/test1@sharedVmfs-0

Confirmed this with /proc/mounts.

Do you think we can separately handle this bug (both in photon and vmdk driver)

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: MarkDirty will just set the bool to true. It doesn't take any lock on its own. Just doing it as an exported method

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: #1220

pshahzeb:
@msterin @pdhamdhere Please take a look. For issue about duplicate mount for same volume with refcounting not initialized I have filed #1220. Will do a separate PR for it since it will also require some refactoring.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -349,75 +425,43 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+	// lock the state
+	d.refCounts.StateMtx.Lock()
+	defer d.refCounts.StateMtx.Unlock()
+
+	if d.refCounts.GetSuccess() != true {
+		// if refcounting hasn't been succesful,
+		// no refcounting, no unmount. All unmounts are delayed
+		// until we succesfully populate the refcount map
+		d.refCounts.MarkDirty()
+		return volume.Response{Err: ""}
```
pdhamdhere: I thought we are not going to return error to Docker since this is an plugin internal issue.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -276,21 +337,30 @@ func matchNameforVMDK(mount_source string) bool {
...
+	r.StateMtx.Lock()
+	r.isDirty = false
+	r.StateMtx.Unlock()
 
 	filters := filters.NewArgs()
 	filters.Add("status", "running")
 	filters.Add("status", "paused")
 	filters.Add("status", "restarting")
 
+	if r.checkDirty() {
```
pdhamdhere: Creating filter is lightweight operation. No need to check isDirty here. You can remove this check.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -190,6 +192,80 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+		if refcnt > 1 {
+			log.WithFields(
+				log.Fields{"name": r.Name, "refcount": refcnt},
+			).Info("Already mounted, skipping mount. ")
+			return volume.Response{Mountpoint: getMountPoint(r.Name)}
+		}
+	}
+
+	// This is the first time we are asked to mount the volume, so comply
+	status, err := d.ops.Get(r.Name)
```
pdhamdhere: status should already include "datastore" which can be used as ref counting key and mount dir name. I am okay if you address mount/unmount issue in separate PR

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -349,75 +425,43 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
-		// Fail back to a default version that we can try with.
-		value = fs.FstypeDefault
+	// lock the state
+	d.refCounts.StateMtx.Lock()
+	defer d.refCounts.StateMtx.Unlock()
+	if d.refCounts.GetSuccess() != true {
+		// complete the mount process under the lock to avoid races
+		// dont do refcounting. Just mount
+		reponse := d.processMount(r, false)
+		d.refCounts.MarkDirty()
```
pdhamdhere: Set MarkDirty before processMount so that it may save us some CPU cycles in discoverAndSync

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: Empty string means error is None. That's how we have indicated error to elsewhere to docker.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: Yes will address it separately.

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: done

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +163,58 @@ func newRefCount() *refCount {
...
+// create a timer to calculate refcount after a delay. If failed, retry again
+// until retry attempt limit reached
+func (r *RefCountsMap) retryCalculate(d drivers.VolumeDriver, mountDir string, name string) {
+	// create a random delay between max and min delay
+	attemptLeft := refCountRetryAttempts
+
+	for attemptLeft > 0 {
+		// generate a random delay everytime
+		delay := rand.Intn(refCountDelayMaxSec-refCountDelayMinSec+1) + refCountDelayMinSec
+		log.Infof("Scheduling again after %d seconds", delay)
```
msterin: incremental backoff is better here. 
[1, 2, 4, 8, 16, 16,...]

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -336,6 +336,74 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
 }
 
+// private function that does the job of mounting volume in conjunction with refcounting
+// Called under a lock if refcounting hasn't been successful, else without a lock
+// Don't do refcounting, just the mount if refcounting hasn't been successful.
+func (d *VolumeDriver) processMount(r volume.MountRequest, doRefcnt bool) volume.Response {
+	// If the volume is already mounted , just increase the refcount.
+	//
+	// Note: We are deliberately incrementing refcount first, before trying
+	// to do anything else. If Mount fails, Docker will send Unmount request,
```
pshahzeb: yeah docker doesn't send unmount if mount fails. This comment isn't valid anymore.

msterin: good. Please correct the comment then

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -336,6 +336,74 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
+				log.Fields{"name": r.Name, "refcount": refcnt},
+			).Info("Already mounted, skipping mount. ")
+			return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
+		}
+	}
+
+	// This is the first time we are asked to mount the volume. The disk
+	// ID isn't always cached by the plugin on a host, the disk could have
+	// been created on a different host and getting mounted on another host.
+	// So on the first mount fetch the disk from Photon and add it's ID to
```
pshahzeb: This looks like old comment. Id never gets added to refmap. Can be removed as part of #1212

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: done

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -336,6 +336,74 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
+		log.Debugf("volume name=%s refcnt=%d", r.Name, refcnt)
+		if refcnt > 1 {
+			log.WithFields(
+				log.Fields{"name": r.Name, "refcount": refcnt},
+			).Info("Already mounted, skipping mount. ")
+			return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
+		}
+	}
+
+	// This is the first time we are asked to mount the volume. The disk
```
pshahzeb: done

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -336,6 +336,74 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
+func (d *VolumeDriver) processMount(r volume.MountRequest, doRefcnt bool) volume.Response {
+	// If the volume is already mounted , just increase the refcount.
+	//
+	// Note: We are deliberately incrementing refcount first, before trying
+	// to do anything else. If Mount fails, Docker will send Unmount request,
+	// and we will happily decrement the refcount there, and will fail the unmount
+	// since the volume will have been never mounted.
+	// Note: for new keys, GO maps return zero value, so no need for if_exists.
+
+	if doRefcnt {
```
pshahzeb: done

pshahzeb:
@msterin Addressed your comments.
Fixing the test and avoiding redundant mounts through issue #1220

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +387,34 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
 		for _, mount := range containerJSONInfo.Mounts {
 			// check if the mount location belongs to vsphere plugin
 			if matchNameforVMDK(mount.Source) {
+				// populate temp map
```
pdhamdhere: Remove? There is no temp map?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -87,14 +89,35 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
 	return d
 }
 
+// In following three operations on refcount, if refcount
+// map hasn't been initialized, return 1 to be safe.
```
pdhamdhere: "to be safe." => "to prevent detach and remove"

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -87,14 +89,35 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
...
+
 // Return the number of references for the given volume
-func (d *VolumeDriver) getRefCount(vol string) uint { return d.refCounts.GetCount(vol) }
+func (d *VolumeDriver) getRefCount(vol string) uint {
+	if d.refCounts.GetInitSuccess() != true {
+		return 1
+	}
+
```
pdhamdhere: nit: blank line seem unnecessary

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -87,14 +89,35 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
...
+	return d.refCounts.GetCount(vol)
+}
 
 // Increment the reference count for the given volume
-func (d *VolumeDriver) incrRefCount(vol string) uint { return d.refCounts.Incr(vol) }
+func (d *VolumeDriver) incrRefCount(vol string) uint {
+	if d.refCounts.GetInitSuccess() != true {
+		return 1
+	}
+
```
pdhamdhere: nit: blank line seem unnecessary

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -87,14 +89,35 @@ func NewVolumeDriver(port int, useMockEsx bool, mountDir string, driverName stri
...
+	return d.refCounts.Incr(vol)
+}
 
 // Decrement the reference count for the given volume
-func (d *VolumeDriver) decrRefCount(vol string) (uint, error) { return d.refCounts.Decr(vol) }
+func (d *VolumeDriver) decrRefCount(vol string) (uint, error) {
+	if d.refCounts.GetInitSuccess() != true {
+		return 1, nil
+	}
+
```
pdhamdhere: nit: blank line seem unnecessary

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -116,6 +118,11 @@ type refCount struct {
 type RefCountsMap struct {
 	refMap map[string]*refCount // Map of refCounts
 	mtx    *sync.RWMutex        // Synchronizes RefCountsMap ops
+
+	refcntInitSuccess bool // save refcounting success
+	isDirty           bool // flag to check reconciling has been interrupted
+
```
msterin: nit  - I don't think the empty line is needed as the mutex protects the two boolean and can be grouped together. Format/nit only

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +161,58 @@ func newRefCount() *refCount {
...
+	delay := refCountDelayStartSec
+	for attemptLeft > 0 {
+		// generate a random delay everytime
+		log.Infof("Scheduling again after %d seconds", delay)
+		timer := time.NewTimer(time.Duration(delay) * time.Second)
+
+		<-timer.C
+		err := r.calculate(d, mountDir, driverName)
+		if err != nil {
+			log.Infof("Refcounting failed: (%v).", err)
```
msterin: nit: add attemptsLeft to the log message

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -150,10 +161,58 @@ func newRefCount() *refCount {
...
+			attemptLeft--
+			// exponential backoff
+			delay += delay
+		} else {
+			return // all good
+		}
+	}
+	// couldn't complete refcounting even after retries.
+	panic(fmt.Sprintf("Failed to talk to docker to calculate volumes usage. Please restart docker"))
```
msterin: nit: add a comment about what Docker will do when the plugin panics

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -318,25 +387,34 @@ func (r RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
+				log.Infof("  name=%v (driver=%s source=%s) (%v)",
 					mount.Name, mount.Driver, mount.Source, mount)
 			}
 		}
 	}
 
+	// check if the background refcount was dirtied
+	r.StateMtx.Lock()
+	defer r.StateMtx.Unlock()
+	if r.isDirty == true {
```
msterin: Please use d.CheckDirty() as in all other places. Also allows to drop L399 and 400

pdhamdhere: Don't you have to execute if block with StateMtx acquired to block incoming mount/unmount?

pshahzeb: @pdhamdhere Yes it will be executed under statemtx, unlock is deferred till the end.
@msterin cannot drop L399 and 400 because get mount info and remove unncessary mounts and set refcountInitSuccess under the same lock to avoid races with mount unmount.

esx_service/utils/vmdk_utils.py
```diff
@@ -77,6 +77,7 @@ def init_datastoreCache(force=False):
     with lockManager.get_lock("init_datastoreCache"):
         global datastores
         logging.debug("init_datastoreCache:  %s", datastores)
+
```
shuklanirdesh82: nit: please remove this file from this PR that avoid confusion in future for just new line change

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -105,9 +105,36 @@ func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir
...
+		return 1
+	}
+
+	return d.refCounts.Incr(vol)
+}
+
+// Decrement the reference count for the given volume
+func (d *VolumeDriver) decrRefCount(vol string) (uint, error) {
+	if d.refCounts.GetInitSuccess() != true {
+		return 1, nil
```
shuklanirdesh82: isn't better to return error message instead of `nil`? otherwise we may drop `error` field from return object

pshahzeb: Its not really an error in this case. We just want to hold the refcount so that volume doesn't get deleted till refcounting succeeds. There could be error later on, hence the return type.

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: Yes removing.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: done

PR #1182, opened 2017-04-20
Modified Makefile so plugin build under HGFS

msterin:
When the repo is checked out under VMWare own HGFS on a guest (Linux) , and the Host is WIndows,
the plugin build fails because HGFS in this combination does not support symlinks
and 'docker plugin export` command uses it.

This PR moves the actual plugin files manipulation under /tmp so the build succeeds for anyone who
(like myself) runs linux guest in VMWare Workstation on WIndows , and shares repo dir over HGFS

A side effect of this PR is that ./plugin folder is not polluted with build-only managed  plugin files


plugin/Makefile
```diff
@@ -84,24 +86,24 @@ info:
...
-	docker rm -vf  $(TMP_CONTAINER)
-	docker rmi $(TMP_IMAGE)
+	mkdir -p $(TMP_LOC)/rootfs
+	docker export  $(TMP_CONTAINER)  | tar -x -C $(TMP_LOC)/rootfs
+	cp config.json $(BINARY) $(TMP_LOC)
 	@echo "-- Creating  plugin $(PLUGIN_NAME):$(PLUGIN_TAG) ..."
-	docker plugin create $(PLUGIN_NAME):$(PLUGIN_TAG) .
-
+	docker plugin create $(PLUGIN_NAME):$(PLUGIN_TAG) $(TMP_LOC)
+	@rm -rf $(TMP_LOC)
```
shuklanirdesh82: nit: L89 takes care this .. can it be redundant and removed? Please explain if I am missing something here.

msterin: yes, it's a missed leftover. Will remove. Thanks !

Issue #1189, opened 2017-04-25
Unable to unmount/detach volume after vmgroup membership change


shuklanirdesh82:
volume stays as attached even after detaching from container

Test bed setup:

- 1 ESX (E1) accessing a shared datastore SD1
- 2 VMs (VM1 & VM2): VM1/VM2 are created on SD1 (registered at E1)

Steps to reproduce:

1. VM1: create volume (volume is created under _DEFAULT vmgroup)
`docker volume create -d vsphere vol1`
2. E1: verify volume is created and Admin CLI shows its entry correctly
`vmdkops_admin.py volume ls`
3. VM1/VM2: invoke `docker volume ls` and `docker volume inspect` to verify vol1 properties
4. VM1: attach container and leave it **attached**
`docker run --rm -it -v vol1@sharedVmfs-0:/vol1 --name cnt_abc11 busybox`
5. E1: verify volume is in `attached` state and `docker volume inspect` shows the same
6. E1: create new named vmgroup by adding VM1 ( docker_host where the container is in attached state)
7. E1: `ADMIN CLI volume ls` to make sure the volume is still in attached state
8. VM2: invoke `docker volume inspect vol1` 
9. Make sure `docker volume inspect` or Admin CLI still says volume is attached to VM1
10. VM1: detach/exit container (remove unused containers)
11. VM2: invoke `docker volume inspect vol1` <<< volume is still in **attached** state
```
root@photon-xVmYMbyTn [ ~ ]# docker volume inspect vol1@sharedVmfs-0
[
    {
        "Driver": "vsphere:latest",
        "Labels": null,
        "Mountpoint": "/mnt/vmdk/vol1@sharedVmfs-0",
        "Name": "vol1@sharedVmfs-0",
        "Options": {},
        "Scope": "global",
        "Status": {
            "access": "read-write",
            "attach-as": "independent_persistent",
            "attached to VM": "VM_17_SVMFS",
            "capacity": {
                "allocated": "15MB",
                "size": "100MB"
            },
            "clone-from": "None",
            "created": "Tue Apr 25 06:33:46 2017",
            "created by VM": "VM_17_SVMFS",
            "datastore": "sharedVmfs-0",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "attached"
        }
    }
]
```
12: E1: `vmdkops_admin.py volume ls` still shows it is **attached** state
```
[...]

Issue #1188, opened 2017-04-25
[Multi-ESX] vmgroup is not removable when vms are added from both ESXs (configured as MultiNode)


shuklanirdesh82:
Test bed setup: 
1. 2 ESX (E1 & E2) accessing a shared datastore SD1
2. 2 VMs (VM1 & VM2): 

- VM1 is created on SD1 (registered at E1)
- VM2 is created on SD1 (registered at E2)

Steps to reproduce:
```
1. configure MultiNode state on both E1 & E2
E1: vmdkops_admin.py config init --datastore vsanDatastore
E2: vmdkops_admin.py config init --datastore vsanDatastore

2. Create named vmgroup "T1" and add VM1 & VM2 to T1
E1: vmdkops_admin.py vmgroup create --name=T1 --default-datastore=vsanDatastore
E1: vmdkops_admin.py vmgroup vm add --name=T1 --vm-list=VM1
E2: vmdkops_admin.py vmgroup vm add --name=T1 --vm-list=VM2

3. E1/E2: invoke vmdkops_admin.py vmgroup ls to make sure VM & VM2 are added to T1 vmgroup
E1/E2:vmdkops_admin.py vmgroup ls

4. Invoke vmdkops_admin.py vmgroup rm --name=T1
```

Expectation: Admin CLI should allow to remove `T1`

Current behavior: Admin CLI is not allowing to remove and throwing following error ( vm UUID is not recognized by ESX where vm is not registered)

```
[root@sc2-rdops-vm05-dhcp-160-17:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup rm --name=T1
Cannot complete vmgroup rm. Cannot find vm 564d4e04-f778-487a-5e1c-50e6215af0a5
```
While invoking `vmgroup rm` command from E1 (Admin CLI complains about the UUID of VM2 as it is registered on E2) and on E2 complains about UUID of VM1 

On E1:
```
=== Authorization Config DB: 
DB_SharedLocation: /vmfs/volumes/vsanDatastore/dockvols/vmdkops_config.db
DB_Mode: MultiNode (local symlink pointing to shared DB)
DB_LocalPath: /etc/vmware/vmdkops/auth-db
```

on E2:
```
=== Authorization Config DB: 
DB_LocalPath: /etc/vmware/vmdkops/auth-db
DB_Mode: MultiNode (local symlink pointing to shared DB)
[...]

PR #1190, opened 2017-04-25
Check for mounted volumes when adding VMs to a new vmgroup on create.

govint:
Added check in auth.api.py:_tenant_create() to check for volumes mounted before allowing to move the VMs into the VM group being created.

Test:
```
bin/vmdkops_admin.py vmgroup ls
------------------------------------  --------  -------------------------  -----------------  -------

bin/vmdkops_admin.py vmgroup create --name vmg1 --vm-list 2 --default-datastore /vmfs/volumes/bigone
VMs already have volumes in use, cannot add VMs to vmgroup. VM '2' has volumes mounted.
```


govint:
Fixes #1189

msterin:
Can you elaborate what is the problem ? The bug description is not clear on this (too many symptoms and conditions). It is also not clear why a VM with volumes mounted should be blocked from moving to a group.... I thought unmount() is unrelated to specific tenant location and access right ?

shuklanirdesh82:
> @msterin's comment  https://github.com/vmware/docker-volume-vsphere/pull/1190#issuecomment-297072146

> Can you elaborate what is the problem ? The bug description is not clear on this (too many symptoms and conditions). It is also not clear why a VM with volumes mounted should be blocked from moving to a group....

Sorry Mark for the confusion. I have updated the bug description to reduce the confusion ( #1189 ).

@govint has summarize at https://github.com/vmware/docker-volume-vsphere/issues/1189#issuecomment-296965571

govint:
This PR is only to handle the check for mounted volumes in a VM when a vmgroup is being created. The check is done for vmgroup replace and vmgroup rm commands.

_DEFAULT is a vmgroup in itself so, trying to create a new vmgroup from VMs in _DEFAULT (added by default btw) should make the same check that volumes aren't already attached to those VMs.

The issue with using VM uuid is a separate issue addressed via #1188.

msterin:
I still do not understand what is the issue with placing a VM with an attached volime in a group

shuklanirdesh82:
@govint, I have an offline chat with @msterin and summarizing the discussion here for your reference. 

It would be a troublesome for admin by blocking "vmgroup creation with vms having volumes mounted".

There are two proposed approaches while dealing with VMs having volumes mounted on.

**Approach #1**: Block the request at first attempt and allow Admin to perform the operation (if still wants) by supporting usability feature of `--force` option.

**Approach #2**: Let the request go through and perform necessary cleanup/update at metadata file (KV file) whenever getting the unmount request after the operation (moving vm across vmgroup).

An approach#1 would be preferable for the admin to decide whether moving VMs across vmgroups is desirable at any given point (when volumes are still in use by some of the vms). It would be troublesome for the admin if we stop the operation until VMs are freed up. By supplying `--force` option should unmount all the volumes immediately for the given VM and new mount request will be served on the new datastore that reduces many corner cases too. 

While going with Approach #2 (current behavior), we may run into some corner cases (unknown at this point) to maintain consistency. 

By simply blocking/stopping the request doesn't sound convincing by taking account of usability from an admin point of view.

@msterin, Please update as per necessity if I have misunderstood any part.

govint:
Can I suggest the approach as proposed which matches the behavior in the
documentation. Flexibility  to move a VM is still there  to create a
vmgroup but allowing a force option will need kludge's   in the service
coffee to handle such cases.   Prefer keeping the behavior based on use
case vs. by testing  which is a continuous feature creep.

.

msterin:
I don't think documentation is our north star. Good user experience is. I simply do not understand what an admin is supposed to do if she wants to change a VM membership in a group, and is blocked by "volumes attached" error message.

govint:
This check is in place for these actions on a vmgroup - replace, remove. And the doc says no VM in a vmgroup is allowed to have volumes in datastores not assigned to the VMgroup. User is expected to have stopped using the volumes in the datastore of the old group before assigning.

msterin:
Please do answer my question. I am the admin. I receive this error. What do I do next ?

pshahzeb:
@msterin I think admin can first get ensure the volumes aren't attached (may be stop the containers) and then change the membership? what do you think?

shuklanirdesh82:
> https://github.com/vmware/docker-volume-vsphere/pull/1190#issuecomment-298436822
> I think admin can first get ensure the volumes aren't attached (may be stop the containers) and then change the membership? what do you think?

@pshahzeb Does it show how many volumes are currently attached to which vms? Let's say in the large scale setup 10 VMs are there and some of them are having volumes mounted. There is a requirement to move 10 VM from `_DEFAULT` tenant to "named" tenant and with admin cli, admin can create new named tenant with all 10 vms in one shot. 

what is the expected behavior for `admin_cli vmgroup create` with `--vm-list` set to all 10 vms? It would be nightmare for Admin to detach volumes one by one in addition to that admin *might not* be having access to docker host too. Other problem I am thinking is: when the admin in the process of detaching volumes, it is likely to be new mount request comes up.

msterin:
@pshahzeb - which containers ? How does the admin finds them ? Specific steps/commands please. FWIW I don't think it's trivial or obvious, which is why I think --force option is goodness. But if we decide against it, we need step by step instructions on how to find offending containers.

govint:
It's actually not about which containers. The VM is being moved and it
can't have volumes mounted. The same applies to VM group remove and replace
of vms

esx_service/utils/auth_api.py
```diff
@@ -472,6 +472,12 @@ def _tenant_create(name, default_datastore, description="", vm_list=None, privil
         if error_info:
             return error_info, None
 
+        error_info = vmdk_utils.check_volumes_mounted(vms)
+        if error_info:
+            error_info.msg = "VMs already have volumes in use, cannot add VMs to vmgroup. " + error_info.msg
```
pdhamdhere: `error_info.msg = "Cannot add VM to vmgroup " + error_info.msg`

to be consistent with other error message and error could be "VM not found"

esx_service/utils/auth_api.py
```diff
@@ -692,6 +698,12 @@ def _tenant_vm_add(name, vm_list):
     if error_info:
         return error_info
 
+    error_info = vmdk_utils.check_volumes_mounted(vms)
+    if error_info:
+        error_info.msg = "VMs already have volumes in use, cannot add VMs to vmgroup. " + error_info.msg
```
pdhamdhere: Same thing. 

error_info.msg = "Cannot complete vmgroup add " + error_info.msg

pdhamdhere:
@msterin Agree. Admin experience isn't ideal. However, this is defensive edge case. Generally you would expect Admin to create vmgroup first before handing over control to developers. If needed we can document (KB) this error case.

govint:
User experience is even worse if some containers are running in the VM and we force move the VM to another vmgroup with volumes attached and end up unmounting the volumes and making those containers fail in any way.

The flip side is to allow access automatically in the target VM group for the datastores on which the volumes in use are located. Which is not desireable at all. Thats not the user intention to move a VM and the target group automatically gets access to some datastores. And as VMs get added to a certain group it starts accquiring access to all datastores?

esx_service/utils/auth_api.py
govint: Done.

esx_service/utils/auth_api.py
govint: Done.

msterin:
@pdhamdhere  I agree that user experience being bad in edge case is acceptable, that's why I asked to document step by step what is the manual actions the admin needs to take if she does want to move the VM. IMO the end result of this will be killing containers (if she finds them) which is similar to --force option but more complex and unpredictable.

govint:
@msterin @pdhamdhere I'll make a doc PR to describe how vmgroup migration can be done.

PR #1192, opened 2017-04-25
Cleaned Up VMCI sources to enable shared-lib

akutz:
This patch cleans up the VMCI C-sources to enable the VMCI C client to be utilized as a shared library. Since it used to directly include C files instead of headers and lacked pragma directives, types were being declared multiple times and causing compiler errors when importing the Go package that links to this package.

I also fixed a bug in the `esx_vmdkcmd.go` file after speaking with @pdhamdhere. He noted that it was the intention to provide a sane default for the port instead of leaving it zero. I assigned it a default of 1019 since that is the default if the program is entered through its main file.


pdhamdhere:
CC/ @govint @msterin

msterin:
@shuklanirdesh82 - there is binary junk in test logs again. Can we **please* get it fixed ? https://ci.vmware.run/vmware/docker-volume-vsphere/2324

msterin:
> You could consider doing what I do and defining the C sources along side the Go code so that the Go package is built statically with the C code and Go's own import logic worries about not importing things more than once.

I think there are not many reason to worry about shared libs if we only use it once and link statically, so the rearrangement seems good to me. (FWIW I think we did have C alongside GO code at some point but I moved it out to have more-or-less consistent dir structures with C code. Maybe it was not so wise) 

BTW, the CI is failing in go unit test .... The PR does not seem to impact it but it was passing reliably so worth a look.

akutz:
Hi @msterin,

I know the PR is failing the CI pipeline, and it was caused by me including the new `vmci_client.h` instead of `vmci_client.c`. I realized of course that fails because I'm not linking to an object with the implementations of the header's functions. That's when I realized the crux of the issue, that we either need to make the C code a shared lib or make sure it's statically linked exactly once. I will refactor the sources to achieve the latter as I do with my copies of them.

akutz:
Hi,

I'm going to close this PR so as not to create countless builds while I work out the kinks in the refactor work I'm doing.

PR #1217, opened 2017-05-03
Supporting shared volumes doc [SKIP CI]

govint:
Adding spec for proposal to support shared volumes with the volume plugin. Please review.

### DIFF
--- a/docs/features/shared-volumes-with-NFS.md
+++ b/docs/features/shared-volumes-with-NFS.md
@@ -80,7 +80,6 @@ The configuration items are described below:
 
 1. “RemoteDirs”:  Map of remote dirs.
 1. “<dir-A-label>” – must be unique and admin defined and identifies a specific remote dir or another entry within the list
-1. “addr” – IP or fully qualified host name of server
 1. “args – args to mount remote dir.
 1. “path” – path on server to mount
 1. “type” – FS type of remote dir.
--- a/docs/features/shared-volumes-with-NFS.md
+++ b/docs/features/shared-volumes-with-NFS.md
@@ -1,9 +1,9 @@
-# Supporting shared volumes with vSphere Docker Volumes
+# Supporting NFS shared volumes with vDVS
 
 ## Requirements
-* Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
+* Developer should be able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
 * This allows them to use NFS directories as docker volumes using Docker APIs, Compose etc 
-* Extend the vSphere Docker Volume plugin to support NFS shared volumes and mutliple readers/writers access
+* Extend the vSphere Docker Volume plugin to support NFS shared volumes and multiple readers/writers access
 * Volumes are accessible on multiple docker hosts running on  multiple ESX hosts in a Docker swarm
 * Allow admin to specify what remote dirs are managed by the plugin
 * Support all current docker volume plugin API for shared volumes
@@ -13,8 +13,9 @@
 The design here assumes that the admin is responsible to identify remote directories to export and has configured remote servers to export those. Specifically, the admin needs to have,
 1. Exported remote directories to be used to host docker volumes.
 1. Configured sufficient permissions for the admins on the Docker host to be able to mount and create folders and files on those directories.
-1. The docker host has the necessary packages of the desired version of NFS installed.
-1. The admin can use the same remote directory configuration on multiple docker hosts on separate physical hosts and be able to view and use the same volumes on their containers (the applications must be able to synchronize access to shared volumes).
+1. The docker host has the necessary packages of the desired version of NFS client installed.
+
+The admin can use the same remote directory configuration on multiple docker hosts on separate physical hosts and be able to view and use the same volumes on their containers (the applications must be able to synchronize access to shared volumes).
 
 ### Update plugin configuration
 The plugin needs to be provided the details of the exported diretories and the servers from which to mount them. The details are entered into the plugin configuration file at /etc/docker-volume-vsphere.conf in the format as detailed later.
@@ -112,7 +113,7 @@ Plugin maintains no in-memory state for either volumes or remote dirs. (except t
    - _access perms (ro, rw)_, remote-dir label, type (_file_ only allowed, if not specified defaults to _VMDK_ type)
 
 ## Tenancy
-There is no support for tenancy as supported for VMDK backed volumes. Instead, the admin manages multi-tenancy. User controls which VMs have access to what datastores to support multiple tenants. Remote dirs configuration s available to help the admin determine what datastores are visible per docker host and where volumes get placed.
+There is no support for multi-tenancy for NFS backed volumes, within the plugin. Instead, the admin manages this via the configuration of shared dirs to control which NFS datastores are used to place volumes for which containers or groups of containers(like service stacks for example).
 
 # Authentication
 
--- a/docs/features/shared-volumes-with-NFS.md
+++ b/docs/features/shared-volumes-with-NFS.md
@@ -1,6 +1,5 @@
 # Supporting shared volumes with vSphere Docker Volumes
 
-
 ## Requirements
 * Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
 * This allows them to manage NFS directories as docker volumes using Docker APIs, Compose etc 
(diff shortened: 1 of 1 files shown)


docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+* 
+
+
+## vSphere Docker Volume Plugin Drivers – User experience
+
+  1. Setup
+      a. Set up Plugin Config
+      b. Restart plugin
+  2. Using shared volumes
+      a. docker volume create …
```
govint: Sure, the format isn't changed except to add network FS instances in the config and optionally in the command line. All other commands remain the same. But I'll add examples for each.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+    - vSphere – provide Docker volumes vSAN, VVOL, VMFS, NFS
+    - Photon – vSAN
+* Drivers work with endpoints (currently local-ESX and Photon) to manage volumes
+    - vSphere – works with the local ESX host to provision VMDKs
+    - Photon – works with a Photon Controller endpoint to provision VMDKs
+
+
+# vSphere-cloud driver
+
+* Add a new “vsphere-cloud” (or better name) driver
```
govint: From review of this looks like a new driver may imply less change to existing code but more drivers may be unnecessary. "vsphere" driver is ok to use for all native vsphere storage and other storage backends.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+1. EFS support is enabled as a result of using NFS 4.0/4.1 to access the EFS instance. How authentication is done will
+be addressed in a separate spec.
+2. Docker allows passing secrets to plugins and as such username/passwords can be passed to
+the plugin to handle authentication with EFS.
+
+## Alternative - Support shared volumes as a common module
+* Support shared volumes as a module used by both the vSphere and Photon drivers
+
+### Pros
+* Retains the functionality within the existing drivers - user doesn't need to specify a new driver
```
govint: type=network may be good because the plugin is provisioning a folder vs. file for the volume.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+* Extends the scope of either driver to support more backends as sources for volumes - VMDK(vSphere or Photon), Networked FS/shares, VVOLs (later?)
+
+### Cons
+* Docker volumes with NFS are not related to VMDK based volumes and may confuse the user
```
govint: Ok.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+* Modified time
+* Access time
+* Access permissions
+
+
+# Volume options and Tenancy
+
+## Volume create options
+   - Access perms, volume group, remote-dir (see remote dir config)
+## Tenancy
```
govint: The user is the one responsible for setting up the network shares and handles the items in the assumptions section earlier. No, there is no relation to the vmgroups as that type of tenancy isn't doable for these networked FS (the plugin doesn't interface with ESX at all for networked FS).

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,207 @@
...
+
+* Similar to ESX driver that VMDK driver uses to handle the interface to the ESX (via VMCI)
+* Protocol specific drivers will be defined for the FS types supported – NFS/EFS/CIFS
+   - Note: EFS uses the NFS v4.1 protocol and will be supported in phase-2
+* API similar to the existing vmdkops
+   - Create, Attach, Detach, List, Get, Remove, InitFS (to initialize for a specific network FS instance), etc.
+* To start with one driver that supports NFS 4.0/4.1 will be supported (covers NFS/EFS shares)
+* Support for EFS and CIFS will be handled in phase-2
+
+# Data mover APIs for cloudops – phase 2 of shared volumes
```
govint: Agree.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
+# Supporting shared volumes with vSphere Docker Volumes
+
+## Requirements
+* Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
+* This allows them to manage NFS directories as docker volumes using Docker APIs, Compose etc
```
msterin: nit: allows them to **use** nfs-exposed storage (not manage really)

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
+# Supporting shared volumes with vSphere Docker Volumes
+
+## Requirements
+* Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
+* This allows them to manage NFS directories as docker volumes using Docker APIs, Compose etc 
+* Extend the vSphere Docker Volume plugin to support shared volumes
```
msterin: to support NFS shared volumes and mutliple readers/writers access

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+* Volumes are accessible on multiple docker hosts running on  multiple ESX hosts in a Docker swarm
+* Allow user to specify what remote dirs are managed by the plugin
```
msterin: allows **admin** to specify

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+* Support all current docker volume plugin API for shared volumes
+
+## User experience
+### Set up remote directories for docker volumes
+The design here assumes that the user is responsible to identify remote directories to export and has configured remote servers to export those. Specifically, the user needs to have,
```
msterin: Please use **admin** or **dev** instead of user,  to make use case clearer

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+1. Exported remote directories to be used to host docker volumes.
+1. Configured sufficient permissions for the users on the Docker host to be able to mount and create folders and files on those directories.
```
msterin: that's the assumption, right ? No functionality for this in plugin right ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+
+The plugin configuration defined in /etc/docker-volume-vsphere.conf must be modified to include the details of the remote directories to use to locate volumes. The format describing a remote directory is as shown below,
+
+## /etc/docker-volume-vsphere.conf
+{
+"MaxLogAgeDays": 28,
+"MaxLogSizeMb": 100,
+"LogPath": "/var/log/docker-volume-vsphere.log",
+"LogLevel": "debug",
+_“RemoteDirs”: {_
```
msterin: so is there a CLI or API for this , or the only way to configure is via the conf file ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+	  _"src": "dir-A-label"_
+	  _"default": <0/1>_
+     _},_
+     _},_
+  _}_
+}
+
+The configuration items are described below:
+
+1. “RemoteDirs”:  Map of remote dirs.
```
msterin: how does it map to [mount](https://linux.die.net/man/8/mount)  arguments ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+### Using shared volumes
+Docker volume commands must be used as shown below to use volumes on remote directories. Similar to how local VMDK based volumes are qualified with a datastore name, shared volumes must also be qualified with the label (see remote directory configuration below) that identifies the remote dir. to use to locate and refer a volume.
+1. docker volume create - 
+   1. _docker volume create --driver=vsphere --name=remote-vol -o type=file_ - uses the default entry in the list of remote dirs to locate the volume
+   1. _docker volume create --driver=vsphere --name=remote-vol@remote-dir-label_ - creates the volume in the location identified by the label in the remote dirs list
+1. docker volume list - No change
+1. docker volume inspect - the volume must be qualified by the remote dir label as below,
+   1. _docker volume inspect vol-name@remote-dir-label_
+1. docker run - the volume must be qualified by the remote dir label as below,
+   1. _docker run -it --volume-driver=vsphere -v remote-vol@remote-dir-label:/vol1 --name ub ubuntu_
```
pdhamdhere: Curious, how is plugin going to distinguish between VMDK volume and NFS volume during mount?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+The plugin needs to be restarted to load the updated configuration.
+
+### Using shared volumes
+Docker volume commands must be used as shown below to use volumes on remote directories. Similar to how local VMDK based volumes are qualified with a datastore name, shared volumes must also be qualified with the label (see remote directory configuration below) that identifies the remote dir. to use to locate and refer a volume.
+1. docker volume create - 
+   1. _docker volume create --driver=vsphere --name=remote-vol -o type=file_ - uses the default entry in the list of remote dirs to locate the volume
+   1. _docker volume create --driver=vsphere --name=remote-vol@remote-dir-label_ - creates the volume in the location identified by the label in the remote dirs list
+1. docker volume list - No change
+1. docker volume inspect - the volume must be qualified by the remote dir label as below,
+   1. _docker volume inspect vol-name@remote-dir-label_
```
pdhamdhere: Can you add sample output for inspect

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+1. docker run - the volume must be qualified by the remote dir label as below,
+   1. _docker run -it --volume-driver=vsphere -v remote-vol@remote-dir-label:/vol1 --name ub ubuntu_
+1. docker volume rm - the volume must be qualified by the remote dir label as below,
+   1. _dovker volume rm vol-name@remote-dir-label_
```
pdhamdhere: This will remove directory? What if multiple containers running on different host are using same volume?

pdhamdhere: _dovker => _docker

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+Plugin maintains no in-memory state for either volumes or remote dirs. (except the configuration). Volume met-data listed below are generated when requested via a Get() from Docker,
+1. _Location_ – remote-dir, path within remote-dir
+1. _Type and version_ - volume type (NFS, EFS)
+1. _Capacity used in FS blocks_
+1. _Current usage count-
+1. _Create date/time_
+1. _Modified time_
+1. _Access time_
+1. _Access permissions_
+1. _Use count_
```
pdhamdhere: How do you plan to get Use Count?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+The configuration items are described below:
+
+1. “RemoteDirs”:  Map of remote dirs.
+1. “<dir-A-label>” – must be unique and user defined and identifies a specific remote dir or another entry within the list
+1. “addr” – IP or fully qualified host name of server
+1. “args – args to mount remote dir.
+1. “path” – path on server to mount
+1. “type” – FS type of remote dir.
+1. “vol-path” – a path name that is relative to the mounted path abnove, volumes will be create relative to this path if non-null, if not specified then volumes are located under path above. The _vol-path_ entry allows the user  user to separate volumes for different docker service stacks within a single remote dir.
+1. "src" - indicates that a remote-dir (identified by the assigned label)in the list be used to locate volumes. Combined with a _vol-path_ allows the user to direct placement of volumes on separate folders within a single remote dir. Especially useful when the user wants to separate volumes for different service stacks into separate folders.
```
pdhamdhere: I am not sure if I understand purpose of "src".

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,138 @@
...
+	  _"default": <0/1>_
+     _},_
+     _},_
+  _}_
+}
+
+The configuration items are described below:
+
+1. “RemoteDirs”:  Map of remote dirs.
+1. “<dir-A-label>” – must be unique and user defined and identifies a specific remote dir or another entry within the list
```
pdhamdhere: `docker plugin install` will create directory "path\dir-label\vol-path" rt?

govint:
Thanks for the comments, I'm working on updating the changes and will post the spec today.

docs/features/shared-volumes-with-NFS.md
govint: Done.

docs/features/shared-volumes-with-NFS.md
govint: This is handled in the code with the user having to specify volume@DSLabel. Minus the DSLabel the volume can be NFS if the user also provides a "-o type=nfs" option. Else defaults to the VMDK driver.

docs/features/shared-volumes-with-NFS.md
govint: This has to be handled by the NFS server that the dir. can't be removed if there are open files in the dir. CSI is handling this to ensure that the unmount is sent only if no containers are using the volume. Within the plugin prefer not to keep any meta-data to track this.

docs/features/shared-volumes-with-NFS.md
govint: Yes, the only way is via the config file, I can add this via command line once the changes are submitted. Even authentication for nfs4 isn't included in the first set of the changes.

docs/features/shared-volumes-with-NFS.md
govint: The user provides the volume to mount as <volname>@<label>, the label is used to check if it mapes to an entry in the remote dirs tabel and the args from that are used.

docs/features/shared-volumes-with-NFS.md
govint: No, the first create/mount/remove/list/get will mount the remote dir and if necessary create the dir. path. Its possible user just does docker volume ls with a remote dir spec in the config file.

docs/features/shared-volumes-with-NFS.md
govint: I've updated the spec. I'm pushing it for review again.

docs/features/shared-volumes-with-NFS.md
govint: This is just the local refcount for the volumes mounted to containers from the refcounter.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+1. docker volume create - 
+   1. _docker volume create --driver=vsphere --name=remote-vol -o type=<fstype - nfs/nfs4> - uses the default entry in the list of remote dirs to locate the volume
+   1. _docker volume create --driver=vsphere --name=remote-vol@remote-dir-label_ - creates the volume in the location identified by the label in the remote dirs list
+1. docker volume list - No change
+1. docker volume inspect - the volume must be qualified by the remote dir label as below,
+   1. _docker volume inspect vol-name@remote-dir-label_
+1. docker run - the volume must be qualified by the remote dir label as below,
+   1. _docker run -it --volume-driver=vsphere -v remote-vol@remote-dir-label:/vol1 --name ub ubuntu_
+1. docker volume rm - the volume must be qualified by the remote dir label as below,
+   1. _docker volume rm vol-name@remote-dir-label_
```
pdhamdhere: Git not showing prior comments. So, docker volume rm will issue delete to NFS Server which will delete directory entry. Other containers will continue to run and as soon as all handles are closed, directory is gone? This doesn;t sound right. At very least, it need to be called out in known issues.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+
+The configuration items are described below:
+
+1. “RemoteDirs”:  Map of remote dirs.
+1. “<dir-A-label>” – must be unique and admin defined and identifies a specific remote dir or another entry within the list
+1. “addr” – IP or fully qualified host name of server
+1. “args – args to mount remote dir.
+1. “path” – path on server to mount
+1. “type” – FS type of remote dir.
+1. “vol-path” – a path name that is relative to the mounted path abnove, volumes will be create relative to this path if non-null, if not specified then volumes are located under path above. The _vol-path_ entry allows the admin  admin to separate volumes for different docker service stacks within a single remote dir.
```
pdhamdhere: I am finding combination of RemoteDir, Vol-Path, Src as too confusing. For services, docker prefixes volume names with service name anyway. We either need to simplify or illustrate with examples.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+
+The vSphere plugin driver supports an extendable framework that allows adding support for volumes backed by specific storage types. Presently the driver supports VMDK backed volumes on storage available to the local ESX host. A new module to support NFS backed volumes (on local or NFS mounted directories) is added to the existing framework that supports all of the docker volume functions as listed above.
+
+The new NFS backed volumes will support NFS v3/v4.0 protocols to mount and use remote directories to host docker volumes. Docker volumes will be supported as folders on the mounted directories and exported to containers in the same way that VMDK volumes are made available to containers.
+
+# Plugin configuration for remote directories
+
+The plugin configuration defined in /etc/docker-volume-vsphere.conf must be modified to include the details of the remote directories to use to locate volumes. The format describing a remote directory is as shown below,
+
+## /etc/docker-volume-vsphere.conf
```
shaominchen: Should we put this file in /etc/vmware/vmdkops folder?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+	     _"Src": "dir-A-label"_
+        _},_
+     _},_
+  _}_
+
+The configuration items are described below:
+
+1. “RemoteDirs”:  Map of remote dirs.
+1. “<dir-A-label>” – must be unique and admin defined and identifies a specific remote dir or another entry within the list
+1. “addr” – IP or fully qualified host name of server
```
shaominchen: The item "addr" seems missing in above sample?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+1. _Use count_
+
+
+# Volume options and Tenancy
+
+## Volume create options
+   - _access perms (ro, rw)_, remote-dir label, type (_file_ only allowed, if not specified defaults to _VMDK_ type)
+
+## Tenancy
+There is no support for tenancy as supported for VMDK backed volumes. Instead, the admin manages multi-tenancy. User controls which VMs have access to what datastores to support multiple tenants. Remote dirs configuration s available to help the admin determine what datastores are visible per docker host and where volumes get placed.
```
shaominchen: "User controls which VMs have access to what datastores..."
Do you mean "NFS datastores" here?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+## Requirements
+* Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
+* This allows them to use NFS directories as docker volumes using Docker APIs, Compose etc 
+* Extend the vSphere Docker Volume plugin to support NFS shared volumes and mutliple readers/writers access
+* Volumes are accessible on multiple docker hosts running on  multiple ESX hosts in a Docker swarm
+* Allow admin to specify what remote dirs are managed by the plugin
+* Support all current docker volume plugin API for shared volumes
+
+## User experience
+### Set up remote directories for docker volumes
```
shaominchen: This part is not very clear to me:

1. Exported remote directories...
Are we talking about setting up the remote NFS server here?

2. Configured sufficient permissions for the admins...
How do we configure NFS access permissions? Using docker-volume-vsphere.conf?

3. "The docker host has the necessary packages of the desired version of NFS installed."
Should be "...NFS client installed"?

4. "The admin can use ..."
This item seems inconsistent with above 3 items. Above 3 items are talking about something admin SHOULD do before creating shared volumes. But this item is talking about something admin CAN do with shared volumes feature.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
+# Supporting shared volumes with vSphere Docker Volumes
```
shaominchen: Nit: how about "Supporting NFS shared volumes with vDVS"?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Requirements
+* Developer should able to use externally managed NFS share as “shared docker volume” using “vsphere” volume driver.
```
shaominchen: Nits: 1) should able to => should be able to; 2) what's "NFS share"?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+* This allows them to use NFS directories as docker volumes using Docker APIs, Compose etc 
+* Extend the vSphere Docker Volume plugin to support NFS shared volumes and mutliple readers/writers access
```
shaominchen: Nit: mutliple => multiple

docs/features/shared-volumes-with-NFS.md
govint: Its exactly as how today if a volume is created with a datastore name (or label) then it needs to be specified as such in all the docker commands. As below.

_docker volume create --driver=vsphere --name=TestVol60@bigone -o size=400mb_

_docker volume inspect TestVol60@bigone_
_[
    {
        "Driver": "vsphere:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/TestVol60@bigone",
        "Name": "TestVol60@bigone",
        "Options": {
            "size": "400mb"
        },_
_....._

_docker run -it --volume-driver=vsphere -v Test_Vol60@bigone:/vol1 --name ub1 ubuntu
root@9c842f9aada7:/#
root@9c842f9aada7:/#_

With NFS the proposal is to have a default and gets used on create. But after that on all commands the user needs to specify the volume name as volname@label. This is to ensure that the driver can identify which remote share hosts the volume. The same volume name can be on multiple remote shares, Hence user specifies the label.

Even with VMDK backed volumes I can create ABC@DS1 and later if I specify ABC@DS1 the plugin will pick the correct volume. But if I specify just ABC on the docker command then either the plugin will check the default DS and report _volume not found_ or worse pick the wrong volume.

Regarding accessing and removing a volume on different hosts, prefer the plugin doesn't keep meta-data and the need to maintain it. The best justification is this - the volume plugin isn't cluster aware as yet and will have to rely on Docker doing the _right thing_ when using volumes in a swarm cluster. I'll update the doc.

docs/features/shared-volumes-with-NFS.md
govint: OK, its a flexibility for the user to have simpler definitions of the locations where they want to place their volumes vs. repeating the same remote dir config multiple times.

Remote-dir is the remote folder to mount - like <host addr>:<folder path>
VolPath - suffixes this path to _folder path_ above so user has control where the volumes get placed. The remote folder need not necessarily be used only for docker volumes - its like a tenant-folder in VMDK backed volumes.
Src - allows user to avoid repeating a remote folder definition in the config file. As mentioned in the doc, the remote directories are placed in a table in the config file and an entry can refer another entry by specifying that entry's label as a _Src_. Its only a flexibility to the user.

docs/features/shared-volumes-with-NFS.md
govint: Yes the first three points are to set up the NFS server and export the necessary folders on it. (4) I've moved to a statement below its not a part of the list to set up the NFS server.

docs/features/shared-volumes-with-NFS.md
govint: This is for the client plugin, this is the only config. file thats available to keep this data.

docs/features/shared-volumes-with-NFS.md
govint: Yes, I've removed it. Path is sufficient for the host:/exported/path  to mount.

docs/features/shared-volumes-with-NFS.md
govint: Yes, I'll update the doc.

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+1. Configured sufficient permissions for the admins on the Docker host to be able to mount and create folders and files on those directories.
+1. The docker host has the necessary packages of the desired version of NFS client installed.
+
+The admin can use the same remote directory configuration on multiple docker hosts on separate physical hosts and be able to view and use the same volumes on their containers (the applications must be able to synchronize access to shared volumes).
+
+### Update plugin configuration
+The plugin needs to be provided the details of the exported diretories and the servers from which to mount them. The details are entered into the plugin configuration file at /etc/docker-volume-vsphere.conf in the format as detailed later.
+
+### Restart plugin
+The plugin needs to be restarted to load the updated configuration.
```
msterin: Can we just wait on iNotify ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+### Set up remote directories for docker volumes
+The design here assumes that the admin is responsible to identify remote directories to export and has configured remote servers to export those. Specifically, the admin needs to have,
+1. Exported remote directories to be used to host docker volumes.
+1. Configured sufficient permissions for the admins on the Docker host to be able to mount and create folders and files on those directories.
+1. The docker host has the necessary packages of the desired version of NFS client installed.
+
+The admin can use the same remote directory configuration on multiple docker hosts on separate physical hosts and be able to view and use the same volumes on their containers (the applications must be able to synchronize access to shared volumes).
+
+### Update plugin configuration
+The plugin needs to be provided the details of the exported diretories and the servers from which to mount them. The details are entered into the plugin configuration file at /etc/docker-volume-vsphere.conf in the format as detailed later.
```
msterin: /etc/VMware/... ?
I don't recall where is the existing config file ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Restart plugin
+The plugin needs to be restarted to load the updated configuration.
+
+### Using shared volumes
+Docker volume commands must be used as shown below to use volumes on remote directories. Similar to how local VMDK based volumes are qualified with a datastore name, shared volumes must also be qualified with the label (see remote directory configuration below) that identifies the remote dir. to use to locate and refer a volume.
+1. docker volume create - 
+   1. _docker volume create --driver=vsphere --name=remote-vol -o type=<fstype - nfs/nfs4> - uses the default entry in the list of remote dirs to locate the volume
```
msterin: I think the driver name should be different as it has nothing to do with vsphere really. Maybe a separate driver with separate packaging, shipped separately ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+   1. _docker volume create --driver=vsphere --name=remote-vol@remote-dir-label_ - creates the volume in the location identified by the label in the remote dirs list
+1. docker volume list - No change
+1. docker volume inspect - the volume must be qualified by the remote dir label as below,
+   1. _docker volume inspect vol-name@remote-dir-label_
+1. docker run - the volume must be qualified by the remote dir label as below,
```
msterin: Will it accept :ro and other qualifiers ? Does docker passes this info to the driver ?

docs/features/shared-volumes-with-NFS.md
```diff
@@ -0,0 +1,129 @@
...
+   1. _docker run -it --volume-driver=vsphere -v remote-vol@remote-dir-label:/vol1 --name ub ubuntu_
+1. docker volume rm - the volume must be qualified by the remote dir label as below,
```
msterin: I think as long as volume names (vol@location) are unique we are good, but mixing up vmdk and nfs seems counter intuitive

docs/features/shared-volumes-with-NFS.md
shaominchen: I have the same concern actually. Ideally having just one driver name will make it easier for the user to use our product. However, my concern is it might be difficult to handle all use cases in a consistent way between these two modes, such as volume inspection, multi-tenancy, etc. If we can't handle all those cases consistently, it will be awkward to have the same driver name for two distinct usage modes.

govint:
I'm listing the  pros/cons of using a single driver vs. multiple drivers

Pros of using a single driver:
1. Ease of use and less confusing for user to use appropriate driver type when they can use a single driver name for _all_ docker volumes inside docker hosts on vSphere infra (ESX nodes - anywhere).
1. Encourages to make the design as much as possible to agnostic to the volume type and isolate any differences in configuration alone.
1. Codewise - its simpler to register and run one driver inside the plugin vs. either multiple plugins (why??) or register and run multiple threads inside the plugin as multiple drivers.
1. Multi-tenancy once issues are resolved should be designed in such a way that it can be used for all volume types. Possibly, via ESX service level API to create and use tenants so that even NFS/EFS/CIFS volumes can use the feature as is.
1. One stop for all volume services on Docler/vSphere infra. User doesn't need to pick other plugins.

Cons:
1. A plugin is written for a specific storage type for a specific platform. In that sense vSphere plugin is doing just that and could continue doing that.
1. Related to above point does it even make sense to write a plugin for non-vsphere managed storage? Users should deploy whats already available and make do with other third party plugins for storage not native to vSphere.
1. Code bloat of the plugin in trying to support non-native storage types. Which is non-trivial as the code size grows. Even with common code its not easy maintainig a growing code base.
1. Support one functionality (vSphere storage for docker volumes) well vs. attempting to get a spread of volume types supported. Which at least till now isn't a roadmap item for this plugin.

govint:
Closing per discussion.

PR #1218, opened 2017-05-03
Handling http imports for python 2.x (ESX 6.0.X)

pshahzeb:
This PR is on top of PR #1168 with an additional change to 
handle http imports for python 2.x (ESX 6.0.X). The http module got renamed in python 3.x.
When hostd goes down, capture the http exception and retry to connect to hostd.

Testing:
Tested locally on ESX 6.0.2 based nimbus testbed. test-all target passes


esx_service/utils/vm_listener.py
```diff
@@ -34,6 +33,16 @@
 from pyVmomi import VmomiSupport, vim, vmodl
 # vim api version used - version11
 
+# Need this to capture hostd down exception.
+try:
```
shaominchen: We should have a better way to detect python version, for example: http://sweetme.at/2013/10/21/how-to-detect-python-2-vs-3-in-your-python-script/

pshahzeb: Done.

Issue #1212, opened 2017-05-02
Refactor common code between photon_driver.go and vmdk_driver.go


pshahzeb:
A big amount of code is common in these two files. Need to refactor them and place the common part in a common placeholder.

govint:
The Create() and Mount() are possibly the only functions that are mostly similar in actions between Photon and the VMDK driver. Rest are specific to use the Photon Controller API and not really separable into common code between the two.

shuklanirdesh82:
based on https://github.com/vmware/docker-volume-vsphere/issues/1212#issuecomment-300236648 ... adding `invalid` keyword

PR #1224, opened 2017-05-05
Avoiding redundant mounts by using /proc/mount info

### LINKED ISSUE #1220: Use use full volume name as mountpoint in mount request and check it to avoid redundant mount.
If the refcounting hasn't been successful, we try to mount every time we receive a mount request.
Before mounting, need to check if the volume with a given name is already mounted. If it is skip it.
If not checked and skipped, the second (and further) mount fails with resource busy error.

To check it appropriately, use the full volume name as the mount point. In case of short volume names, get the volume and use the full volume name.
Before mounting, check if the volume is already present in /proc/mounts. If it is, skip the mount. and return the mount point.

This also takes care of a volume getting mounted twice : one with a short name and one with a long name on a VM.

pshahzeb:
1. Convert a short volume name to the long one. If required to a trip to vmdkops service
2. In every mount request, check if the volume is already mounted (info of /proc/mount)
3. refcount test which kills docker and verifies refcounting

Fixes #1220 

Testing:
The refcount test added kills dockerd and verifies refcount (after refcount has been initialized) through logs.

Verified the same manually by forcefully disabling plugin and enabling it.

### DIFF
--- a/misc/scripts/refcnt_test.sh
+++ b/misc/scripts/refcnt_test.sh
@@ -67,10 +67,11 @@ function check_files {
 }
 
 function check_recovery_record {
-    line=`tail -10 /var/log/docker-volume-vsphere.log | $GREP 'Volume name=' | $GREP 'mounted=true'`
-    expected="name=$vname count=$count mounted=true"
+    # restart generates some log.
+    line=`tail -50 /var/log/docker-volume-vsphere.log | $GREP 'Volume name=' | $GREP 'mounted=true'`
+    expected="count=$count mounted=true"
 
-    echo $line | $GREP -q "$expected" ; if [ $? -ne 0 ] ; then
+    echo $line | $GREP "$vname" | $GREP -q "$expected" ; if [ $? -ne 0 ] ; then
        echo Found:  \"$line\"
        echo Expected pattern: \"$expected\"
        return 1
@@ -80,20 +81,20 @@ function check_recovery_record {
 
 function test_crash_recovery {
     timeout=$1
-    echo "Checking recovery for VMDK plugin kill -9"
-    kill -9 `pidof docker-volume-vsphere`
-    until pids=$(pidof docker-volume-vsphere)
+    echo "Checking recovery through docker kill"
+    pkill -9 dockerd
+    until pids=$(pidof dockerd)
     do
-        echo "Waiting for docker-volume-vsphere to restart"
+        echo "Waiting for docker to restart"
         sleep 1
     done
 
     echo "Waiting for plugin init"
-    sleep 3
+    sleep 5
     sync  # give log the time to flush
     wait_for check_recovery_record $timeout
     if [ "$?" -ne 0 ] ; then
-        echo PLUGIN RESTART TEST FAILED. Did not find proper recovery record
+        echo DOCKER RESTART TEST FAILED. Did not find proper recovery record
         exit 1
     fi
 }
@@ -115,8 +116,8 @@ fi
 echo "$(docker volume ls)"
 for i in `seq 1 $count`
 do
-  $DOCKER run -d -v $vname:/v busybox sh -c "touch /v/file$i; sync ; \
-      sleep $timeout"
+  $DOCKER run -d --restart=always -v $vname:/v busybox sh -c "touch /v/file$i; sync ; \
+      while true; do sleep $timeout; done"
 done
 
 echo "Checking the last refcount and mount record"
--- a/misc/scripts/refcnt_test.sh
+++ b/misc/scripts/refcnt_test.sh
@@ -67,7 +67,7 @@ function check_files {
 }
 
 function check_recovery_record {
-    # restart generates some log.
+    # log contains refcounting attempts and after success logs summary.
     line=`tail -50 /var/log/docker-volume-vsphere.log | $GREP 'Volume name=' | $GREP 'mounted=true'`
     expected="count=$count mounted=true"
 
@@ -82,6 +82,7 @@ function check_recovery_record {
 function test_crash_recovery {
     timeout=$1
     echo "Checking recovery through docker kill"
+    # kill docker daemon forcefully
     pkill -9 dockerd
     until pids=$(pidof dockerd)
     do
@@ -116,6 +117,7 @@ fi
 echo "$(docker volume ls)"
 for i in `seq 1 $count`
 do
+  # run containers with restart flag so they restart after docker restart
   $DOCKER run -d --restart=always -v $vname:/v busybox sh -c "touch /v/file$i; sync ; \
       while true; do sleep $timeout; done"
 done
--- a/vmdk_plugin/drivers/photon/photon_driver.go
+++ b/vmdk_plugin/drivers/photon/photon_driver.go
@@ -38,6 +38,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/go-plugins-helpers/volume"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/fs"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/plugin_utils"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/refcount"
 	"github.com/vmware/photon-controller-go-sdk/photon"
 )
@@ -53,6 +54,8 @@ const (
 	capacityGB           = 1
 	fsTypeTag            = "Fs_Type"
 )
+
+var fullVolumeNameMap map[string]string // save full volume names to avoid trip to vmdkops service
 
 // VolumeDriver - Photon volume driver struct
 type VolumeDriver struct {
@@ -78,6 +81,8 @@ func (d *VolumeDriver) verifyTarget() error {
 
 // NewVolumeDriver - creates Driver, creates client for given target
 func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir string) *VolumeDriver {
+
+	fullVolumeNameMap = make(map[string]string)
 
 	d := &VolumeDriver{
 		target:  targetURL,
@@ -362,6 +367,19 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 
 // private function that does the job of mounting volume in conjunction with refcounting
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
+	// get volume metadata
+	status, err := d.GetVolume(r.Name)
+
+	if err != nil {
+		return volume.Response{Err: err.Error()}
+	}
+
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
+		datastore := plugin_utils.GetDatastore(r.Name, status)
+		r.Name = plugin_utils.GetFullVolumeName(r.Name, datastore)
+		fullVolumeNameMap[r.ID] = r.Name
+	}
+
 	// If the volume is already mounted , just increase the refcount.
 	// Note: for new keys, GO maps return zero value, so no need for if_exists.
 	refcnt := d.incrRefCount(r.Name) // save map traversal
@@ -373,12 +391,9 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
 		return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
 	}
 
-	// There can be redundant mounts till refcounts are properly initialized
-	// TODO: #1220
-	status, err := d.GetVolume(r.Name)
-	if err != nil {
-		d.decrRefCount(r.Name)
-		return volume.Response{Err: err.Error()}
+	if plugin_utils.CheckAlreadyMounted(r.Name, d.mountRoot) {
+		log.WithFields(log.Fields{"name": r.Name}).Info("Already mounted, skipping mount. ")
+		return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
 	}
 
 	fstype, exists := status[fsTypeTag]
@@ -616,6 +631,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		// until we succesfully populate the refcount map
 		d.refCounts.MarkDirty()
 		return volume.Response{Err: ""}
+	}
+
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
+		fullVolName := fullVolumeNameMap[r.ID]
+		if fullVolName == "" { // if ID not present in local map, do a get trip
+			status, _ := d.GetVolume(r.Name)
+			datastore := plugin_utils.GetDatastore(r.Name, status)
+			fullVolName = plugin_utils.GetFullVolumeName(r.Name, datastore)
+		}
+		r.Name = fullVolName
 	}
 
 	// if refcount has been succcessful, Normal flow.
--- a/vmdk_plugin/drivers/photon/photon_driver.go
+++ b/vmdk_plugin/drivers/photon/photon_driver.go
@@ -55,16 +55,15 @@ const (
 	fsTypeTag            = "Fs_Type"
 )
 
-var fullVolumeNameMap map[string]string // save full volume names to avoid trip to vmdkops service
-
 // VolumeDriver - Photon volume driver struct
 type VolumeDriver struct {
-	client    *photon.Client
-	hostID    string
-	mountRoot string
-	project   string
-	refCounts *refcount.RefCountsMap
-	target    string
+	client        *photon.Client
+	hostID        string
+	mountRoot     string
+	project       string
+	refCounts     *refcount.RefCountsMap
+	target        string
+	mountIDtoName map[string]string // map of mountID -> full volume name
 }
 
 func (d *VolumeDriver) verifyTarget() error {
@@ -82,8 +81,6 @@ func (d *VolumeDriver) verifyTarget() error {
 // NewVolumeDriver - creates Driver, creates client for given target
 func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir string) *VolumeDriver {
 
-	fullVolumeNameMap = make(map[string]string)
-
 	d := &VolumeDriver{
 		target:  targetURL,
 		project: projectID,
@@ -100,6 +97,7 @@ func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir
 	d.mountRoot = mountDir
 	d.refCounts = refcount.NewRefCountsMap()
 	d.refCounts.Init(d, mountDir, driverName)
+	d.mountIDtoName = make(map[string]string)
 
 	log.WithFields(log.Fields{
 		"version": version,
@@ -368,16 +366,16 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 // private function that does the job of mounting volume in conjunction with refcounting
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
 	// get volume metadata
-	status, err := d.GetVolume(r.Name)
+	volumeMeta, err := d.GetVolume(r.Name)
 
 	if err != nil {
 		return volume.Response{Err: err.Error()}
 	}
 
-	if plugin_utils.IsFullVolumeName(r.Name) != true {
-		datastore := plugin_utils.GetDatastore(r.Name, status)
-		r.Name = plugin_utils.GetFullVolumeName(r.Name, datastore)
-		fullVolumeNameMap[r.ID] = r.Name
+	r.Name, err = plugin_utils.GetFullVolumeName(r.Name, volumeMeta["datastore"].(string), d)
+	if err != nil {
+		log.Errorf("Unable to get full name for volume %s. err:%v", r.Name, err)
+		return volume.Response{Err: err.Error()}
 	}
 
 	// If the volume is already mounted , just increase the refcount.
@@ -391,19 +389,19 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
 		return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
 	}
 
-	if plugin_utils.CheckAlreadyMounted(r.Name, d.mountRoot) {
+	if plugin_utils.AlreadyMounted(r.Name, d.mountRoot) {
 		log.WithFields(log.Fields{"name": r.Name}).Info("Already mounted, skipping mount. ")
 		return volume.Response{Mountpoint: d.getMountPoint(r.Name)}
 	}
 
-	fstype, exists := status[fsTypeTag]
+	fstype, exists := volumeMeta[fsTypeTag]
 	if !exists {
 		fstype = fs.FstypeDefault
 	}
 
 	skipAttach := false
 	// If the volume is already attached to the VM, skip the attach.
-	if state, stateExists := status["State"]; stateExists {
+	if state, stateExists := volumeMeta["State"]; stateExists {
 		if strings.Compare(state.(string), "DETACHED") != 0 {
 			skipAttach = true
 		}
@@ -413,7 +411,7 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
 	}
 
 	// Mount the volume and for now its always read-write.
-	mountpoint, err := d.MountVolume(r.Name, fstype.(string), status["ID"].(string), false, skipAttach)
+	mountpoint, err := d.MountVolume(r.Name, fstype.(string), volumeMeta["ID"].(string), false, skipAttach)
 	if err != nil {
 		log.WithFields(
 			log.Fields{"name": r.Name, "error": err.Error()},
@@ -633,12 +631,14 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		return volume.Response{Err: ""}
 	}
 
-	if plugin_utils.IsFullVolumeName(r.Name) != true {
-		fullVolName := fullVolumeNameMap[r.ID]
-		if fullVolName == "" { // if ID not present in local map, do a get trip
-			status, _ := d.GetVolume(r.Name)
-			datastore := plugin_utils.GetDatastore(r.Name, status)
-			fullVolName = plugin_utils.GetFullVolumeName(r.Name, datastore)
+	if fullVolName, exist := d.mountIDtoName[r.ID]; exist {
+		r.Name = fullVolName
+		delete(d.mountIDtoName, r.ID) //cleanup the map
+	} else {
+		fullVolName, err := plugin_utils.GetFullVolumeName(r.Name, "", d)
+		if err != nil {
+			log.Errorf("Unable to get full name for volume %s. err:%v", r.Name, err)
+			return volume.Response{Err: err.Error()}
 		}
 		r.Name = fullVolName
 	}
--- a/vmdk_plugin/drivers/photon/photon_driver.go
+++ b/vmdk_plugin/drivers/photon/photon_driver.go
@@ -363,20 +363,21 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 	return mountpoint, fs.MountWithID(mountpoint, fstype, id, isReadOnly)
 }
 
-//VolumesInRefMap - get list of volumes from refmap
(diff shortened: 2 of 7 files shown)


pshahzeb:
For the repeating code between photon and vmdk driver: #1212

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -54,6 +55,8 @@ const (
 	fsTypeTag            = "Fs_Type"
 )
 
+var fullVolumeNameMap map[string]string // save full volume names to avoid trip to vmdkops service
```
msterin: it is 'mount id -> fullVOlumeName', right ? Or is is something else  
Please add comment ("mountId->fullVolumeName map" or whatever it is)

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -362,6 +367,19 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 
 // private function that does the job of mounting volume in conjunction with refcounting
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
+	// get volume metadata
+	status, err := d.GetVolume(r.Name)
```
msterin: you need GetVolume on short names only, so the check for IsFullVolumeName should happen before the GetVolume .

 I think the algorithm is supposed to be as follows (for mount , in pseudocode)
```

  if IsFullVolumeName(name) {
         name = r.Name
  } else {
     s.GetVolume () //get volume info 
     name = FormFullVolumeName (r.Name, GetDatastore(result_of_volume_get)
  }
```

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -618,6 +633,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		return volume.Response{Err: ""}
 	}
 
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
+		fullVolName := fullVolumeNameMap[r.ID]
```
msterin: yes, this is the right way of doing things. `mount()` should be similar

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -618,6 +633,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
...
+		if fullVolName == "" { // if ID not present in local map, do a get trip
+			status, _ := d.GetVolume(r.Name)
```
msterin: why no check for GetVolume failure ?

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -618,6 +633,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
...
+			datastore := plugin_utils.GetDatastore(r.Name, status)
+			fullVolName = plugin_utils.GetFullVolumeName(r.Name, datastore)
+		}
+		r.Name = fullVolName
+	}
```
msterin: I suggest to wrap up L639->644 (with slight modification) in `COnstructFullVolumeName (shortName string) string ` function returning full name (and error) . Otherwise cut-n-paste is on the way

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -51,14 +52,18 @@ type VolumeDriver struct {
 	refCounts  *refcount.RefCountsMap
 }
 
-var mountRoot string
+var (
+	mountRoot         string
+	fullVolumeNameMap map[string]string // save full volume names to avoid trip to vmdkops service
```
msterin: for all maps , please comment what do they map (key) and to what (value). E.g. volumeMoundID->fullVolumeName

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -212,6 +217,19 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
 
 // private function that does the job of mounting volume in conjunction with refcounting
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
+	// get volume metadata
+	status, err := d.ops.Get(r.Name)
```
msterin: same comment. Do it for short names only

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -212,6 +217,19 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+
+	if err != nil {
+		return volume.Response{Err: err.Error()}
+	}
+
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
```
msterin: please avoid cut-n-paste and refactor this to a function

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -458,6 +477,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		return volume.Response{Err: ""}
 	}
 
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
+		fullVolName := fullVolumeNameMap[r.ID]
+		if fullVolName == "" { // if ID not present in local map, do a get trip
```
msterin: I think I saw this code a few time already :-). 
Please refactor to avoid cut-n-paste

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+const (
+	linuxMountsFile = "/proc/mounts"
```
msterin: this was already defined somewhere. Did you fully move the definition here ? If not, please do

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// GetMountInfo - return a map of mounted volumes and devices
+func GetMountInfo(mountRoot string) (map[string]string, error) {
+	volumeMap := make(map[string]string)
```
msterin: Please add a comment of what is mapped to what (even if it's an old code). Something like `// maps volume->fullMountPath` or whatever it is doing

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -383,12 +385,20 @@ func (r *RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
 		}
 		log.Debugf("  Mounts for %v", ct.Names)
-
 		for _, mount := range containerJSONInfo.Mounts {
-			// check if the mount location belongs to vsphere plugin
-			if matchNameforVMDK(mount.Source) {
-				r.Incr(mount.Name)
-				log.Infof("  name=%v (driver=%s source=%s) (%v)",
+			// check if the mount location belongs to vmdk plugin
+			if matchNameforVMDK(mount.Source) != true {
```
msterin: I think the comment is better than the function name. It would be nice to name  it  isVMDKMount() or something like that. `matchNameforVMDK` is confusing

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -383,12 +385,20 @@ func (r *RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
+				continue
+			}
+			volname := mount.Name
+			if plugin_utils.IsFullVolumeName(volname) != true {
+				if datastore == "" {
+					volumeMeta, _ := d.GetVolume(volname)
```
msterin: and I certainly saw this code already. Please refactor to avoid cut-n-paste.

vmdk_plugin/drivers/photon/photon_driver.go
pdhamdhere: Too many tiny helper functions. How about;
```
getFullVolumeName(name) {
    if (strings.ContainsAny(name, "@")) {
         return name;
    }
    volMeta, err := d.GetVolume(name)
    if err != nil
         return strings.Join(name, "@", volumeMeta["datastore"])
    else
         ???? return err?
}

File: vmdk_plugin/drivers/photon/photon_driver.go [hunk 14]
```diff
@@ -618,6 +633,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 		return volume.Response{Err: ""}
 	}
 
+	if plugin_utils.IsFullVolumeName(r.Name) != true {
```
pdhamdhere: This check is unnecessary. fullVolumeNameMap[r.ID] if exists will have fullVolName

vmdk_plugin/drivers/photon/photon_driver.go
pdhamdhere: status => volMeta ?

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -618,6 +633,16 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
...
+		fullVolName := fullVolumeNameMap[r.ID]
+		if fullVolName == "" { // if ID not present in local map, do a get trip
+			status, _ := d.GetVolume(r.Name)
+			datastore := plugin_utils.GetDatastore(r.Name, status)
+			fullVolName = plugin_utils.GetFullVolumeName(r.Name, datastore)
+		}
+		r.Name = fullVolName
+	}
+
```
pdhamdhere: Dont you have to delete mountID => volName entry from map?

misc/scripts/refcnt_test.sh
```diff
@@ -80,20 +81,21 @@ function check_recovery_record {
 
 function test_crash_recovery {
     timeout=$1
-    echo "Checking recovery for VMDK plugin kill -9"
-    kill -9 `pidof docker-volume-vsphere`
-    until pids=$(pidof docker-volume-vsphere)
+    echo "Checking recovery through docker kill"
+    # kill docker daemon forcefully
+    pkill -9 dockerd
+    until pids=$(pidof dockerd)
```
govint: How does this impact the refcount module. Unless the plugin is restarted, the refcount module hasn't lost it's refcounts to do a recovery.

misc/scripts/refcnt_test.sh
```diff
@@ -80,20 +81,21 @@ function check_recovery_record {
...
     done
 
     echo "Waiting for plugin init"
-    sleep 3
+    sleep 5
     sync  # give log the time to flush
     wait_for check_recovery_record $timeout
     if [ "$?" -ne 0 ] ; then
-        echo PLUGIN RESTART TEST FAILED. Did not find proper recovery record
+        echo DOCKER RESTART TEST FAILED. Did not find proper recovery record
```
govint: Not sure this change is needed. The test is to see if the plugin restarts then the refcounts are restored. If the plugin restart isn't possible then this test should be made manual to execute with a standalone binary and deprecated from automated test run.

pshahzeb: This test will in turn test plugin restart (docker restart). Plugin restart through docker restart tests the docker api consumption, timeout and handling of parallel mounts/unmounts till refcounting is successful. Not sure why not have this as automated test?

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: In mount, we need GetVolume always, since the volume meta data is used below for fstype and access type.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -383,14 +383,29 @@ func (r *RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
...
-				r.Incr(mount.Name)
-				log.Infof("  name=%v (driver=%s source=%s) (%v)",
-					mount.Name, mount.Driver, mount.Source, mount)
+			// check if the mount location belongs to vmdk plugin
+			if isVMDKMount(mount.Source) != true {
+				continue
+			}
+			volname := mount.Name
+			// gets hit once to retrieve the default datastore. datastoreName is reused henceforth
+			if datastoreName == "" {
```
pshahzeb: this is done once,  but needs to be done inside the loop since we need a volume name to get the datastore.

pshahzeb:
@msterin @pdhamdhere I have addressed your comments in the latest commit. Please take a look.

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -362,6 +365,19 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 
 // private function that does the job of mounting volume in conjunction with refcounting
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
+	// get volume metadata
+	volumeMeta, err := d.GetVolume(r.Name)
```
pdhamdhere: If `plugin_utils.AlreadyMounted(r.Name, d.mountRoot)` then you can skip `d.GetVolume(r.Name)` right?

pshahzeb: saving VolumeMeta as part of retrieving full volume name. If found, avoiding next get trip.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -223,9 +239,17 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
...
-	status, err := d.ops.Get(r.Name)
+	if plugin_utils.AlreadyMounted(r.Name, mountRoot) {
+		log.WithFields(log.Fields{"name": r.Name}).Info("Already mounted, skipping mount. ")
+		return volume.Response{Mountpoint: getMountPoint(r.Name)}
+	}
+
+	// get volume metadata if required
+	if volumeMeta == nil {
+		if volumeMeta, err = d.ops.Get(r.Name); err != nil {
+			return volume.Response{Err: err.Error()}
```
pdhamdhere: Dont you have to decrRefCount?

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,123 @@
...
+			return ""
+		}
+		count++
+		fullname = name
+	}
+	return fullname
+}
+
+// GetFullNameAndMeta - return a qualified volume and metadata(if a get trip was made)
+func GetFullNameAndMeta(name string, datastoreName string, d drivers.VolumeDriver) (string, map[string]interface{}, error) {
```
pdhamdhere: I am not go expert and don't know what is recommended practice here. It definitely seems inconsistent with rest of the code which generally returns "retValue, err" tuple. Should we change it to GetVolumeInfo to return dictionary "volInfo, err" tuple where volInfo.fullName and volInfo.metadata has required details.

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -373,22 +389,26 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
...
+	fstype, exists := volumeMeta[fsTypeTag]
 	if !exists {
 		fstype = fs.FstypeDefault
 	}
 
 	skipAttach := false
 	// If the volume is already attached to the VM, skip the attach.
-	if state, stateExists := status["State"]; stateExists {
+	if state, stateExists := volumeMeta["State"]; stateExists {
 		if strings.Compare(state.(string), "DETACHED") != 0 {
```
pdhamdhere: This is not your change but please file an issue and let's revisit.
Generally we don't make decision based on data in KV file and definitely not inside guest. Disk may be attached to some other VM. If neither refCnt nor mount map suggest disk is attached to this VM then we must issue attach to ESX. If it is already attached, it is a no-op from ESX perspective anyway.

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -360,8 +363,21 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 	return mountpoint, fs.MountWithID(mountpoint, fstype, id, isReadOnly)
 }
 
+//VolumesInRefMap - get list of volumes from refmap
```
msterin: another nit - please mention in the comments that these are full names <volume@datastore> format

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -363,20 +363,20 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
 func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
-	// get volume metadata
-	volumeMeta, err := d.GetVolume(r.Name)
-
-	if err != nil {
-		return volume.Response{Err: err.Error()}
-	}
-
-	r.Name, err = plugin_utils.GetFullVolumeName(r.Name, volumeMeta["datastore"].(string), d)
+	volname, volumeMeta, err := plugin_utils.GetFullNameAndMeta(r.Name, "", d)
```
msterin: nit: the name is accurate but confusing IMO. I'd simply call it GetVolInfo() and added comment 
//returns fullName. volumeMeta map and err if any

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -363,20 +363,20 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
 	if err != nil {
 		log.Errorf("Unable to get full name for volume %s. err:%v", r.Name, err)
 		return volume.Response{Err: err.Error()}
 	}
+	r.Name = volname
+	d.mountIDtoName[r.ID] = volname
```
msterin: nit - I'd use r.Name here, not volname (just like everywhere till the rest of the function)

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -215,18 +220,13 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
-	}
-
-	r.Name, err = plugin_utils.GetFullVolumeName(r.Name, volumeMeta["datastore"].(string), d)
+	volname, volumeMeta, err := plugin_utils.GetFullNameAndMeta(r.Name, "", d)
 	if err != nil {
 		log.Errorf("Unable to get full name for volume %s. err:%v", r.Name, err)
 		return volume.Response{Err: err.Error()}
 	}
+	r.Name = volname
+	d.mountIDtoName[r.ID] = volname
```
msterin: same nit - (use r.Name not volname.,  as everyone till the end of this function)

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -70,27 +70,54 @@ func AlreadyMounted(name string, mountRoot string) bool {
...
-func GetDatastore(name string, d drivers.VolumeDriver) (string, error) {
+func GetDatastore(name string, d drivers.VolumeDriver) (string, map[string]interface{}, error) {
 	volumeMeta, err := d.GetVolume(name)
 	if err != nil {
 		log.Errorf("Unable to get volume metadata %s (err: %v)", name, err)
-		return "", err
+		return "", nil, err
 	}
-	return volumeMeta["datastore"].(string), nil
+	return volumeMeta["datastore"].(string), volumeMeta, nil
```
msterin: nit - in other places volumeMeta is indexed by a named constant, not literal. Why sudden change ?

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -70,27 +70,54 @@ func AlreadyMounted(name string, mountRoot string) bool {
...
 }
 
-// GetFullVolumeName - append datastore to the volume name
-func GetFullVolumeName(name string, datastoreName string, d drivers.VolumeDriver) (string, error) {
+// GetNameFromRefmap - get names from refmap
```
msterin: nit: please add comment. Something like  "finds an entry for volume. Entry can be in vol@ds format, Volums is passed in as a short name, with no '@ds' (btw 0 is this what you intended to do ? )

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -70,27 +70,54 @@ func AlreadyMounted(name string, mountRoot string) bool {
...
+		if refVolName != volName {
+			continue
+		}
+		// if there are collisions, return
+		if count > 0 {
+			return ""
+		}
+		count++
+		fullname = name
+	}
```
msterin: so this one scans the full list regardess. Generally not nice, but I guess these lists are short...

vmdk_plugin/utils/plugin_utils/plugin_utils.go
msterin: I also suggested `GetVolumeInfo`. We can leave it with 3 ret vals but need to accurately document IMO

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -70,27 +70,54 @@ func AlreadyMounted(name string, mountRoot string) bool {
...
+
+// GetFullNameAndMeta - return a qualified volume and metadata(if a get trip was made)
+func GetFullNameAndMeta(name string, datastoreName string, d drivers.VolumeDriver) (string, map[string]interface{}, error) {
 	if strings.ContainsAny(name, "@") {
-		return name, nil
+		return name, nil, nil
 	}
 	if datastoreName != "" {
-		return strings.Join([]string{name, datastoreName}, "@"), nil
+		return strings.Join([]string{name, datastoreName}, "@"), nil, nil
```
msterin: I do not like the joins and splits sprinkled over the code. 
Can it we wrapped in a couple of fucntions (similar to python os.path.join and os.path,basename)  ?

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -264,6 +264,19 @@ func (r *RefCountsMap) GetCount(vol string) uint {
 	return rc.count
 }
 
+//GetVolumeNames - return volume names of entries in refmap
```
msterin: space after //.
Also - full names ? Where do we use a full list of volnames ?

vmdk_plugin/utils/plugin_utils/plugin_utils.go
pshahzeb: Yes. wrapping this in volInfo to contain. It will make code cleaner.

vmdk_plugin/utils/refcount/refcnt.go
pshahzeb: Yes. Just one additional optimization to get full volume name from entries in refmap(if there no multiple entries for given volume name) and avoid a trip to esx

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -267,6 +264,19 @@ func (r *RefCountsMap) GetCount(vol string) uint {
...
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+	var volumeList []string
+
+	for k := range r.refMap {
+		volumeList = append(volumeList, k)
+	}
+
+	return volumeList
+}
```
govint: Possibly add a that creates the volumes with the same name on different datastores.

pshahzeb: Sure. did you mean "list can contain volumes with same name on different datastores?"

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,123 @@
...
+	volumeMeta, err := d.GetVolume(name)
+	if err != nil {
+		log.Errorf("Unable to get volume metadata %s (err: %v)", name, err)
+		return "", nil, err
+	}
+	return volumeMeta["datastore"].(string), volumeMeta, nil
+}
+
+// GetNameFromRefmap - get names from refmap
+func GetNameFromRefmap(volName string, d drivers.VolumeDriver) string {
```
govint: Nit - could make this a little descriptive comment especially how volumes with the same name on different datastores are handled. Plus what happens in that case and this function returns an empty string. Its not wrong to have volumes with the same name on different datastores.

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,123 @@
...
+		return name, nil, nil
+	}
+	if datastoreName != "" {
+		return strings.Join([]string{name, datastoreName}, "@"), nil, nil
+	}
+
+	// find full volume names using refmap if possible
+	if fullVolumeName := GetNameFromRefmap(name, d); fullVolumeName != "" {
+		return fullVolumeName, nil, nil
+	}
```
govint: Unlike what the name suggests "GetFullNameAndMeta()" most of the returns here return NULL for the meta. Is that intended? If only one condition is going to return the metadata for a volume then suggest that this function just return the full name or even merge it with the calling function. 

Prefer not to mix multiple unrelated actions inside a single function. -  doesn't need to return the meta-data for a volume GetVolume() should be used for that.

pshahzeb: Yes renaming it. The idea was to return additional information along with full name so as to avoid trips to esx

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,123 @@
...
+	if _, ok := volumeMap[name]; ok {
+		return true
+	}
+	return false
+}
+
+// GetDatastore - get datastore from volume metadata
+// Note "datastore" key is defined in vmdkops service
+func GetDatastore(name string, d drivers.VolumeDriver) (string, map[string]interface{}, error) {
+	volumeMeta, err := d.GetVolume(name)
```
govint: Function is returning values that are not seemingly related to the name of the function - GetDatastore(). Preferably remove this and let the calling function just call GetVolume() and pick the datastore entry from the return meta-data.

If this function is a must-have then let it return just the datastore name.  I mean we are returning the datastore name and the meta-data which also has the same datastore name in it.

pshahzeb: refactored. This function is no more needed. did the getvolume in the calling function.

vmdk_plugin/utils/refcount/refcnt.go
```diff
@@ -401,22 +403,14 @@ func (r *RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
 			if isVMDKMount(mount.Source) != true {
 				continue
 			}
-			volname := mount.Name
-			// gets hit once to retrieve the default datastore. datastoreName is reused henceforth
-			if datastoreName == "" {
-				datastoreName, _, err = plugin_utils.GetDatastore(volname, d)
```
pshahzeb: this isn't needed anymore. returned datastore as part of volumeInfo. captured here to be reused further (and avoid trip to esx)

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -0,0 +1,147 @@
...
+			return ""
+		}
+		count++
+		fullname = name
+	}
+	return fullname
+}
+
+// GetVolumeInfo - return VolumeInfo which a qualified volume name,
+// datastore name volume metadata if retrieved
```
pdhamdhere: How about changing comment to following?

return VolumeInfo with a qualified volume name.
Optionally returns datastore and volume metadata if retrieved from ESX.
If Volume Metadata is nil then caller can use getVolume()

vmdk_plugin/utils/refcount/refcnt.go
pdhamdhere: // GetVolumeNames - return fully qualified volume names from refMap

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: #1230

PR #1248, opened 2017-05-15
Support shared volumes with NFS [SKIP CI]

govint:
Changes to support NFS volumes. Changes are only compiled at this point and I'll first verify VMDK volume support and then the NFS volumes. Please see #1217 for a description of the proposed support (from user perspective).

Changes are not tested yet and will be done thru this week, along with test cases addition.

Authentication with nfs4 isn't supported yet and will be a separate change once these are in.

Changes are large and I'm also reviewing.

Changes:
1.  New VolumeImpl interface is defined to support the current (vmdk), new (nfs) and future(efs, CIFS, etc) volume types.
2. The existing VMDK driver is now implementing the interface in (1), plus a new network driver thats supporting shared volumes with NFS.
3. Parts of the VMDK driver (Volume interface functions and ref counting) are abstracted into a generic vsphere driver that frontends all volume types on vsphere platform.
4. Refactored changes in plugin_utils, refcount modules to simplify the use of fully qualified names for volumes (as volume@dslabel) for all volume types.
5. The vsphere module now keeps the maps to map mount/unmount IDs to qualified volume names and those names to the backing volume type.
6. At run time on each command, the vsphere driver figures out which backing to use based on the config of remote dirs in the plugin config file and invokes the appropriate (vmdk/nfs) drivers.

Have shared the block diagram of the modules and interfaces in the plugin with the proposed changes.

### DIFF
--- a/plugin/Dockerfile
+++ b/plugin/Dockerfile
@@ -9,7 +9,7 @@
 
 FROM alpine:3.5
 
-RUN apk update ; apk add e2fsprogs xfsprogs
+RUN apk update ; apk add e2fsprogs xfsprogs; apk add nfs-utils
 RUN mkdir -p /mnt/vmdk
 COPY docker-volume-vsphere /usr/bin
 CMD ["/usr/bin/docker-volume-vsphere"]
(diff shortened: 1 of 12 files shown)


pdhamdhere:
Did you folks arrive to conclusion whether to do this as separate plugin?

govint:
Closing per discussion.

Issue #1162, opened 2017-04-12
[Multi-ESX] Need to remove DB  after hit failure in "new_db()" to handle "config init" failure


lipingxue:
When user run "config init" from "NotConfigured" clean state, it will try to call function "new_db()" if the auth db does not exist. 
However, In new_db() function, it can fail in the middle. See the following code snippet: 
```
 def new_db(self):
        """
        Create brand new DB content at self.db_path. Expects clean slate.
        :returns: None for success , a string (with error) for error
        """
        if not self.conn:
            self.__connect()
        err = self.__create_tables()
        if err:
            return err
        err = self.__create_default_tenant()
        if err:
            return err
        err = self.__create_default_privileges()
        if err:
            return err
        return None
```

We need to remove the DB when the new_db() function fails which can ensure that the configuration goes back to a clean state "NotConfigured".

govint:
Simulate error in creating DB,

 bin/vmdkops_admin.py config init --local
Warning: this feature is EXPERIMENTAL
Creating new DB at /etc/vmware/vmdkops/auth-db
Init failed: Vmgroup _DEFAULT does not exist

 bin/vmdkops_admin.py config init --local
Warning: this feature is EXPERIMENTAL
Config DB  is already initialized. Use 'rm --local' to reset

ls -l /etc/vmware/vmdkops/
total 20
-rw-r--r--    1 root     root         15360 May 16 13:33 auth-db
-rw-r--r--    1 root     root          1257 May 16 11:08 log_config.json

govint:
With the fix (but with the simulated error still in place we see the DB is removed if init fails,

bin/vmdkops_admin.py config init --local
Warning: this feature is EXPERIMENTAL
Creating new DB at /etc/vmware/vmdkops/auth-db
Init failed: Vmgroup _DEFAULT does not exist

bin/vmdkops_admin.py config init --local
Warning: this feature is EXPERIMENTAL
Creating new DB at /etc/vmware/vmdkops/auth-db
Init failed: Vmgroup _DEFAULT does not exist

ls -l /etc/vmware/vmdkops/
total 4
-rw-r--r--    1 root     root          1257 May 16 11:08 log_config.json

govint:
Closed via #1259

PR #1259, opened 2017-05-16
Remove DB if DB init fails.

govint:
Verified with error simulation on DB init. Pls. see testing done and test logs in #1162.

### DIFF
--- a/esx_service/utils/auth_data.py
+++ b/esx_service/utils/auth_data.py
@@ -734,13 +734,11 @@ class AuthorizationDataManager:
         return error_code_to_message[ErrorCode.INIT_NEEDED]
 
 
-    def new_db(self):
+    def __init_db(self):
         """
         Create brand new DB content at self.db_path. Expects clean slate.
         :returns: None for success , a string (with error) for error
         """
-        if not self.conn:
-            self.__connect()
         err = self.__create_tables()
         if err:
             return err
@@ -752,6 +750,18 @@ class AuthorizationDataManager:
             return err
         err = self.__create_vm_ds_privileges_for_default_tenant()
         if err:
+            return err
+        return None
+
+    def new_db(self):
+        if not self.conn:
+            # Old DB will be opened or a new DB file will be created.
+            # If the DB exists, __init_db below will fail.
+            # Note that if something is wrong, this will throw so no err checking
+            self.__connect()
+        err = self.__init_db()
+        if err:
+            os.remove(self.db_path)
             return err
         return None
 


msterin:
This clean up has to happen inside new_db() to make it transactional, otherwise every call to new_db() will have a potential to have the same bug

shaominchen:
The current fix works since new_db() is getting invoked by config_init() only. But I agree with Mark it will be safer if we handle this inside new_db(), even though I don't expect new_db() to be invoked by other workflows (we shouldn't be doing that).

govint:
Mark, Sam, can you'll say how to handle this inside of new_db.

shaominchen:
To handle this inside of new_db, we will need to call os.remove(db_path) before each "return err" statement as a rollback mechanism.

Personally I'm fine with current change - as I mentioned above, I don't expect new_db() to be invoked by other workflows - we shouldn't be doing that.

govint:
Sam, Agree, can you approve the change then?

msterin:
> to handle this inside of new_db, we will need to call os.remove(db_path) before each "return err" statement as a rollback mechanism.

the right change would be to separate all `err = _create_xxx' ; if err return err` in a new _init_db method , and in new_db simply say `err  = self.__init_db(); if err {delete db; return err` (in pseudo code)

> Personally I'm fine with current change - as I mentioned above, I don't expect new_db() to be invoked by other workflows - we shouldn't be doing that. 
I don't know what it will invoked from tomorrow, and it should not prevent us from doing it right today.  

> What do you think @msterin ?
I think it's small enough so we should do it right even if we do not have an immediate second usage.

govint:
Got it and thats the correct way to handle the DB init errors. Pls. see udpated changes.

esx_service/utils/auth_data.py
```diff
@@ -734,24 +734,29 @@ def err_config_init_needed(self):
...
-    def new_db(self):
+    def __init_db(self):
         """
         Create brand new DB content at self.db_path. Expects clean slate.
         :returns: None for success , a string (with error) for error
         """
         if not self.conn:
             self.__connect()
         err = self.__create_tables()
+        if not err:
```
shaominchen: Nit: I feel the previous style is simple and straightforward. If we really want to avoid these trivial if-check, we should have a nested structure like this:
    if not err:
        err = xxxx
            if not err:
                err = xxxx
                if not err:
                    err = xxxx
I think this is an overkill. So I'd prefer to keep the previous style.

msterin: please , no nesting.

esx_service/utils/auth_data.py
```diff
@@ -734,24 +734,29 @@ def err_config_init_needed(self):
         return error_code_to_message[ErrorCode.INIT_NEEDED]
 
 
-    def new_db(self):
+    def __init_db(self):
```
msterin: ok, this is what I had in mind. If __connect() fails, there is nothing to delete. 
So 
```python

def __init_db(self):  # this is literally cut-n-paste of old code in new_db()
  """Inits freshly created DB by creating needed tables."""
  err = self.__create_tables()
  if err:
     return err

   err = self.__create_default_tenant()
   if err:
      return err

    err = self.__create_all_ds_privileges_for_default_tenant()
    if err:
      return err

    err = self.__create_vm_ds_privileges_for_default_tenant()
    if err:
       return err


def new_db(self): 
        """
        Create brand new DB content at self.db_path. Expects clean slate.
        :returns: None for success , a string (with error) for error
        """
        if not self.conn:
                # old DB will be opened on a new DB file will be created. 
                # We do not expect the file for already-exists, but if it does , __init_db below will fail.
                # Note that if something is wrong, this will throw so no err checking
                self.__connect()          

       err = __init_db()
       if err:
            os.remove(self.db_path)
            return err
```

govint:
Ok, why is nesting preferred the number of checks doesn't decrease.

msterin:
> Ok, why is nesting preferred the number of checks doesn't decrease.

Agreed - I also prefer to avoid nesting, makes it harder to read and makes deeper decision tree , thus harder to understand

esx_service/utils/auth_data.py
govint: Pushed again with the changes as above.

PR #1260, opened 2017-05-16
Handle stale attach code

govint:
Removed code in handle_stale_attach() to reset KV if the VM is powered off. This code is not relevant any more with the handling of VM events in the service.

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -1155,7 +1155,7 @@ def getStatusAttached(vmdk_path):
 
 def log_attached_volume(vmdk_path, kv_uuid, vol_name):
        '''
-       Log appropriate message for volume thats already attached
+       Log appropriate message for volume thats already attached.
        '''
        cur_vm = findVmByUuid(kv_uuid)
 
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -290,7 +290,7 @@ def cloneVMDK(vm_name, vmdk_path, opts={}, vm_uuid=None, datastore_url=None):
         # Verify if the source volume is in use.
         attached, uuid, attach_as, attached_vm_name = getStatusAttached(src_vmdk_path)
         if attached:
-            log_attached_volume(vmdk_path, uuid)
+            log_attached_volume(vmdk_path, uuid, attached_vm_name)
 
         # Reauthorize with size info of the volume being cloned
         src_vol_info = kv.get_vol_info(src_vmdk_path)
@@ -598,15 +598,11 @@ def removeVMDK(vmdk_path, vol_name=None, vm_name=None, tenant_uuid=None, datasto
 
     # Check the current volume status
     kv_status_attached, kv_uuid, attach_mode, attached_vm_name = getStatusAttached(vmdk_path)
-    # If the volume is attached then don't try removing the volume, the uuid may be
-    # resolved to a VM on the local host or may not if the VM is on another host,
-    # either case attempting to clean the VMDK can leave it in an inconsistent state,
-    # VMDK is present but the KV has been deleted.
     if kv_status_attached:
         if vol_name is None:
             vol_name = vmdk_utils.get_volname_from_vmdk_path(vmdk_path)
         logging.info("*** removeVMDK: %s is in use, volume = %s VM = %s VM-uuid = %s",
-            vmdk_path, vol_name, attached_vm_name, kv_uuid)
+                      vmdk_path, vol_name, attached_vm_name, kv_uuid)
         return err("Failed to remove volume {0}, in use by VM = {1}.".format(vol_name, attached_vm_name))
 
     # Cleaning .vmdk file
@@ -1157,17 +1153,18 @@ def getStatusAttached(vmdk_path):
 
     return attached, uuid, attach_as, vm_name
 
-def log_attached_volume(vmdk_path, kv_uuid):
+def log_attached_volume(vmdk_path, kv_uuid, vol_name):
        '''
        Log appropriate message for volume thats already attached
        '''
        cur_vm = findVmByUuid(kv_uuid)
 
        if cur_vm:
-          msg = "Disk is {0} already attached to VM {1}".format(vmdk_path,
+          msg = "Disk {0} is already attached to VM {1}".format(vmdk_path,
                                                              cur_vm.config.name)
        else:
-          msg = "Failed to find VM (id {0}), disk {1} is already attached".format(kv_uuid,
+          msg = "Failed to find VM {0}({1}), disk {2} is already attached".format(vol_name,
+                                                                                 kv_uuid,
                                                                                  vmdk_path)
        logging.warning(msg)
 
@@ -1247,11 +1244,11 @@ def disk_attach(vmdk_path, vm):
     return error or unit:bus numbers of newly attached disk.
     '''
 
-    kv_status_attached, kv_uuid, attach_mode, _ = getStatusAttached(vmdk_path)
+    kv_status_attached, kv_uuid, attach_mode, attached_vm_name = getStatusAttached(vmdk_path)
     logging.info("Attaching {0} as {1}".format(vmdk_path, attach_mode))
 
     if kv_status_attached:
-       log_attached_volume(vmdk_path, kv_uuid)
+       log_attached_volume(vmdk_path, kv_uuid, attached_vm_name)
 
     # NOTE: vSphere is very picky about unit numbers and controllers of virtual
     # disks. Every controller supports 15 virtual disks, and the unit
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -290,9 +290,7 @@ def cloneVMDK(vm_name, vmdk_path, opts={}, vm_uuid=None, datastore_url=None):
         # Verify if the source volume is in use.
         attached, uuid, attach_as, attached_vm_name = getStatusAttached(src_vmdk_path)
         if attached:
-            if handle_stale_attach(vmdk_path, uuid):
-                return err("Source volume {0} is in use by VM {1} and can't be cloned.".format(src_volume,
-                    attached_vm_name))
+            log_attached_volume(vmdk_path, uuid)
 
         # Reauthorize with size info of the volume being cloned
         src_vol_info = kv.get_vol_info(src_vmdk_path)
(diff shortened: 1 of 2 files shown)


esx_service/cli/vmdkops_admin.py
```diff
@@ -1286,6 +1286,7 @@ def config_init(args):
         auth = auth_data.AuthorizationDataManager(db_path)
         err = auth.new_db()
         if err:
+            os.remove(db_path)
```
shaominchen: This seems the same change as PR #1259?

govint: Yes, I made a branch in the same source tree that had this change from #1259 and it pulled the changes from 1259 as well.

esx_service/vmdk_ops.py
```diff
@@ -1145,24 +1145,7 @@ def handle_stale_attach(vmdk_path, kv_uuid):
...
-                   logging.warning(msg)
-                   return err(msg)
-             else:
-                logging.warning("Failed to find disk %s in powered off VM - %s, resetting volume metadata\n",
-                                vmdk_path, cur_vm.config.name)
-                ret = reset_vol_meta(vmdk_path)
-                if ret:
-                   return ret
-          else:
+          if cur_vm.runtime.powerState != VM_POWERED_OFF:
```
pshahzeb: line 1144 isn't going to return the VM if it is on another host. (in essence if disk is attached to VM on another host) ?

govint: Correct, this code can wrongly reset a volume KV if the volume is accessed across hosts. The ESX service isn't cluster aware and the right way to fix this is to have cluster aware plugins so volumes are used appropriately.

govint:
The check for the volume attached to a live VM is needed and in a cluster
scenario the service doesn't assume the other node is running a version
that handles hostd events

pdhamdhere:
> The check for the volume attached to a live VM is needed 

Why? There are 2 places where `handle_stale_attach` is called and I don't see a reason why that function should exist. In both cases, we check KV status and return error if attached.

In `removeVMDK`,
```
kv_status_attached, kv_uuid, attach_mode = getStatusAttached(vmdk_path)
if kv_status_attached:
         // ret = handle_stale_attach(vmdk_path, kv_uuid)
         logging.info("*** removeVMDK: %s is in use, VM uuid = %s (%s)", vmdk_path, kv_uuid, ret)
         return err("Failed to remove volume {0}, in use by VM uuid = {1}.".format(
                vmdk_path, kv_uuid))
```

In `cloneVMDK`
```
        attached, uuid, attach_as = getStatusAttached(src_vmdk_path)
        if attached:
            // if handle_stale_attach(vmdk_path, uuid):
           return err("Source volume cannot be in use when cloning")
```

>and in a cluster scenario the service doesn't assume the other node is running a version
that handles hostd events

Do you mean some ESX running 0.13 or earlier version and some running 0.14 or later? Firstly, I don't see a reason to optimize code for this corner case. Secondly, I don't understand how keeping `handle_stale_attach` in upcoming release helps.

govint:
If the KV says its attached and we treat it as attached no matter what the actual status of the volume. What happens if the event handler misses an event for some reason? Do we go with what the KV has in it? OTOH the function itself can be removed and instead log based on the status in the KV. Or, leave the function as is  as its a backup for cases where for the event handler has missed the event or doesn't get one and this function resets the KV so the volume is usable.

pdhamdhere:
Leaving the function as-is doesn't help since it only works if VM using volume is on same ESX.

Good point on not failing the request based on KV status. So, we should proceed with remove and clone and return appropriate error?

govint:
Sorry, missed this over the weekend. Yes, I'd say attempt the removal/clone/detach and handle the failure. I'll post the change and with the tests for that.

govint:
Testing, volume is already in use:
$ docker volume inspect test-attached
[
    {
        "Driver": "vsphere:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/test-attached",
        "Name": "test-attached",
        "Options": {},
        "Scope": "global",
        "Status": {
            "access": "read-write",
            "attached to VM": "2",
            "attachedVMDevice": {
                "ControllerPciSlotNumber": "224",
                "Unit": "0"
            },
            "capacity": {
                "allocated": "15MB",
                "size": "100MB"
            },
            "clone-from": "None",
            "created": "Mon May 29 10:06:22 2017",
            "created by VM": "2",
            "datastore": "bigone",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "attached"
        }
    }
]

1. Clone of an in-use volume on the ESX host
docker volume create --driver=vsphere --name=test-attached-clone -o clone-from=test-attached
Error response from daemon: create test-attached-clone: VolumeDriver.Create: Failed to clone volume: Unable to access file [bigone] dockvols/11111111-1111-1111-1111-111111111111/test-attached.vmdk since it is locked

2. Remove a volume in use on the node,
docker volume rm test-attached
Error response from daemon: unable to remove volume: remove test-attached: volume is in use - [ca783cfd2969c425cded3cd07323d936f0d3004fa9d9954250ece119ac50ff08]

3. Attach a volume already attached to the docker host,

$ docker run -it --volume-driver=vsphere -v test-attached:/vol1 --name ub1 ubuntu
root@b260255e3518:/#
root@b260255e3518:/#

govint:
The fix doesn't reset the volume KV just because we don't find the volume on the local node, using the same volume on multiple nodes is disallowed and if a volume is in use one node, its locked and can't be used on another node.

esx_service/vmdk_ops.py
```diff
@@ -601,13 +599,7 @@ def removeVMDK(vmdk_path, vol_name=None, vm_name=None, tenant_uuid=None, datasto
...
     kv_status_attached, kv_uuid, attach_mode, attached_vm_name = getStatusAttached(vmdk_path)
     if kv_status_attached:
-        ret = handle_stale_attach(vmdk_path, kv_uuid)
-        if ret:
-            if vol_name is None:
-                vol_name = vmdk_utils.get_volname_from_vmdk_path(vmdk_path)
-            logging.info("*** removeVMDK: %s is in use, volume = %s VM = %s VM-uuid = %s (%s)",
-                vmdk_path, vol_name, attached_vm_name, kv_uuid, ret)
-            return err("Failed to remove volume {0}, in use by VM = {1}.".format(vol_name, attached_vm_name))
+        log_attached_volume(vmdk_path, kv_uuid)
```
pdhamdhere: Do we really need `log_attached_volume` here and other places?  I can understand if we use it on error path to publish additional information but seems unnecessary on regular code path.

esx_service/vmdk_ops.py
```diff
@@ -1147,44 +1139,19 @@ def getStatusAttached(vmdk_path):
...
-             return err(msg)
+          msg = "Disk {0} already attached to VM {1}".format(vmdk_path,
+                                                             cur_vm.config.name)
        else:
-          logging.warning("Failed to find VM (id %s) attaching the disk %s, resetting volume metadata",
-                          kv_uuid, vmdk_path)
-          ret = reset_vol_meta(vmdk_path)
-          if ret:
-             return ret
+          msg = "Failed to find VM (id {0}), disk {1} is already attached".format(kv_uuid,
```
pdhamdhere: We should publish VM name from KV if available.

esx_service/vmdk_ops.py
govint: Yes, log_attached_volume() is needed because this isn;t expected in the code paths where its called.

esx_service/vmdk_ops.py
govint: Ok.

govint: I checked the code again and while KV does/can give the volume name the log_attached_volume() is actually checking the UUID and if the VM is found and logs the VM name. If the UUID isn't resolved to a VM name then it would be incorrect to print the name from the KV - VM may be renamed (in which case we get the current name or VM is moved to another node and the service is local to the esx host).  The proposed changes are handling printing the VM name as appropriate.

govint:
Swarm test seems to have timed out on a poll and errored out, otherwise build is good.
FAIL: swarm_test.go:91: SwarmTestSuite.TestDockerSwarm

swarm_test.go:141:
    c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
... obtained bool = false
... expected bool = true
... Volume swarm_test_volume_1496839264 is still attached

govint:
CI failed with the issue in #1372

govint:
Depends on #1368 to handle the swarm test error

esx_service/vmdk_ops.py
pshahzeb: Agreed on this reasoning. However, renaming the VM is a rare scenario and i think we can still go ahead with printing th VM name we have in the KV.

govint:
CI is in pogress and once its complete.

tests/e2e/basic_test.go
```diff
@@ -92,6 +93,9 @@ func (s *BasicTestSuite) TestVolumeLifecycle(c *C) {
 		status := verification.VerifyAttachedStatus(s.volName1, host, s.esx)
 		c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volName1))
 
+		out, err = dockercli.DeleteVolume(host, s.volName1)
+		c.Assert(err, Not(IsNil), Commentf(out))
```
shaominchen: Nit: We can use NotNil instead of Not(IsNil)

esx_service/vmdk_ops.py
```diff
@@ -600,14 +598,16 @@ def removeVMDK(vmdk_path, vol_name=None, vm_name=None, tenant_uuid=None, datasto
...
-            if vol_name is None:
-                vol_name = vmdk_utils.get_volname_from_vmdk_path(vmdk_path)
-            logging.info("*** removeVMDK: %s is in use, volume = %s VM = %s VM-uuid = %s (%s)",
-                vmdk_path, vol_name, attached_vm_name, kv_uuid, ret)
-            return err("Failed to remove volume {0}, in use by VM = {1}.".format(vol_name, attached_vm_name))
+        if vol_name is None:
+            vol_name = vmdk_utils.get_volname_from_vmdk_path(vmdk_path)
+        logging.info("*** removeVMDK: %s is in use, volume = %s VM = %s VM-uuid = %s",
+            vmdk_path, vol_name, attached_vm_name, kv_uuid)
+        return err("Failed to remove volume {0}, in use by VM = {1}.".format(vol_name, attached_vm_name))
```
pdhamdhere: AFAICT we are not relying on data in KV for any operation. This would be an exception.

If I understand correctly this is a workaround for DiskLib issue. Rather than adding a workaround here, I would propose to follow solution proposed in original bug. Read KV in memory, delete file, if delete fails, write KV back to disk.

govint: This is not feasible because once the KV is deleted by the issue in disklib a new KV can't be created as the disk can't be opened with the flags needed to create the sidecar. A sidecar can't be created for a disk in use (disk is still attached to some VM). Tried out and figured this out.

govint:
CI failing due to #1371

pdhamdhere:
@govint Please triage CI failure and re-push if needed. This PR has been open for a while.

govint:
@pdhamdhere the CI failed with #1371. Can I merge this change since its not my change thats causing the issue?

pdhamdhere:
> @pdhamdhere the CI failed with #1371. Can I merge this change since its not my change thats causing the issue?

Test that was causing #1371 is fixed. We should get a green CI.

govint:
Green CI.

govint:
Closing.

Issue #1096, opened 2017-03-27
Optimize TestVolumeProperties


pdhamdhere:
[TestVolumeProperties](https://github.com/vmware/docker-volume-vsphere/blob/master/vmdk_plugin/E2E_Tests/e2e_verifyVolmProperties_test.go#L68) test takes more than a minute. 

We run following sequence thrice (for each disk type "thin", "zeroedthick", "eagerzeroedthick");
1.  Create volume (`docker volume create --driver=vsphere ...`
2. Verify size, disk-format type and attached-status with Admin CLI
3. Compare same properties with Docker CLI
4. Start busybox with volume attached
5. Repeat verification from Step 2 
6. Repeat verification from Step 3

Step 4-6 can be independent test with "thin" volume type. No need to repeat these steps for each disk-type.

shuklanirdesh82:
@ashahi1 

As we have gocheck testing framework, let's have a plan to use gocheck instead of go testing so it will be easier to manage testcases.

Please prioritize this task to finish as part of 0.15 milestone.

Thanks.

/CC @tusharnt

shuklanirdesh82:
close via #1310

PR #1266, opened 2017-05-16
Part[2/2] Renaming common util and test files

### LINKED ISSUE #1257: Rename existing go test file names as per naming standard
This is 2nd part of #1243 ... files are moved to /root/tests folder and it is the time to rename the file as per naming conventions discussed in [effective_go](https://golang.org/doc/effective_go.html)

shuklanirdesh82:
fixes #1257 and #1267 

- Files are renamed as per effective_go guide.
- With this PR, removing alias as the package name is used directly ... alias is more appropriate while customizing the name.


tests/e2e/volumepropverification_test.go
```diff
@@ -87,43 +87,43 @@ func TestVolumeProperties(t *testing.T) {
...
-				volmPropertiesAdminCli := verifier.GetVolumePropertiesAdminCli(volName, esx)
+				volmPropertiesAdminCli := verification.GetVolumePropertiesAdminCli(volName, esx)
 				expctdPropsAdmin := []string{"100MB", formatTypes[k], "detached"}
 				if !hasElement(expctdPropsAdmin, volmPropertiesAdminCli) {
 					log.Fatal("Volume properties on ESX fetched using admin cli does not matches with the expected values")
 				}
 
 				if dockerCliCheck {
-					volmPropertiesDkrCli := verifier.GetVolumePropertiesDockerCli(volName, vms[vmIndx])
+					volmPropertiesdockercli := verification.GetVolumePropertiesDockerCli(volName, vms[vmIndx])
```
shaominchen: Nit: volmPropertiesdockercli => volumePropertiesDockerCli

tests/e2e/volumepropverification_test.go
```diff
@@ -87,43 +87,43 @@ func TestVolumeProperties(t *testing.T) {
...
 				if dockerCliCheck {
-					vmNameFrmDockerCli := verifier.GetVmAttachedToVolUsingDockerCli(volName, vms[vmIndx])
+					vmNameFrmDockerCli := verification.GetVmAttachedToVolUsingDockerCli(volName, vms[vmIndx])
 					// TODO: get vm name based on ip and compare ith with the docker cli and admin cli
 					if vmNameFrmDockerCli != vmNameFrmAdminCli {
 						log.Fatalf("Information mis-match - Attached-to-VM field for volume from docker cli is [%s]"+
 							"and attched-to-vm field from admin cli is [%s]", vmNameFrmDockerCli, vmNameFrmAdminCli)
 					}
-					volmPropertiesDkrCli := verifier.GetVolumePropertiesDockerCli(volName, vms[vmIndx])
+					volmPropertiesdockercli := verification.GetVolumePropertiesDockerCli(volName, vms[vmIndx])
```
shaominchen: Ditto.

tests/utils/verification/volumeproperties.go
```diff
@@ -24,22 +24,22 @@ import (
 	sshutil "github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
 )
 
-var ADMIN_CLI_LS = "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls "
-var DOCKER_CLI_INSPC = "docker volume inspect "
+var adminCliLs = "/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls "
```
pshahzeb: these should be consts

tests/e2e/volumepropverification_test.go
```diff
@@ -33,14 +33,14 @@ import (
 )
 
 var (
-	esx              string = os.Getenv("ESX")
+	esx               = os.Getenv("ESX")
 	volSizes                = []string{"100MB"}
 	vms                     = []string{os.Getenv("VM1")}
```
shaominchen: Not the current change, but why do we create a list here if we just need one VM? Same for volSizes. This has caused 2 unnecessary nested for-loop in TestVolumeProperties().

tests/e2e/volumepropverification_test.go
```diff
@@ -69,15 +69,15 @@ Test steps:
...
 		log.Println("running end_to_end tests - current time: ", time.Now())
-		vmIp = vms[vmIndx]
-		log.Println("Running test on VM - ", vmIp)
-		dockerCliCheck = verifier.IsDockerCliCheckNeeded(vms[vmIndx])
+		vmIP = vms[vmIndx]
+		log.Println("Running test on VM - ", vmIP)
+		dockerCliCheck = verification.IsDockerCliCheckNeeded(vms[vmIndx])
 		for i := 0; i < len(volSizes); i++ {
 			for k := 0; k < len(formatTypes); k++ {
 				containerName = "busybox_" + strconv.FormatInt(time.Now().Unix(), 20)
```
shaominchen: Can we extract line 77-126 to a few separate methods to improve the code readability?

tests/e2e/volumepropverification_test.go
shuklanirdesh82: The idea was to run against multiple vms (having it said vm1 is photon and vm2 is ubuntu) ... go testframework is having timeout of 10 mins and there were occurrences this simple test was going over 10 mins causing failure for the PR. So we have turned off running multiple VMs.

Performance improvement is tasked out to leverage running this test against multiple inputs.

for `volSizes`: The test was running against 100MB, 1GB, 1TB due to above issue we trimmed down running against only one input. 

> This has caused 2 unnecessary nested for-loop in TestVolumeProperties().
Kept the for loop to turn on extra combination in the future.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: This is out of the scope of this PR. The idea behind this PR was to rename files and addressing golint warnings. Please share your idea if you still want to include as part of this PR or not.

tests/e2e/volumepropverification_test.go
shuklanirdesh82: done!

tests/e2e/volumepropverification_test.go
shaominchen: Then we should add a TODO comment here to avoid confusion.

tests/e2e/volumepropverification_test.go
shuklanirdesh82: there is already a task to address this https://github.com/vmware/docker-volume-vsphere/issues/1096

tests/utils/verification/volumeproperties.go
pshahzeb: Sure. can do it separately or with other PR as handy :)

shaominchen: I agree with Shahzeb here. Actually we shouldn't rename these commands to camel case, which makes the code harder to read. It's almost a standard convention to define CLI commands in capital letters.

shuklanirdesh82: This is as per golang naming convention to address golint warning :)

```
/tests/utils/verification/volumeproperties.go:27:5: don't use ALL_CAPS in Go names; use CamelCase
../tests/utils/verification/volumeproperties.go:27:5: exported var ADMIN_CLI_LS should have comment
```

shaominchen: Sounds good then.

Issue #1263, opened 2017-05-16
Automate Basic tests


shuklanirdesh82:
Detail steps are mentioned at https://github.com/vmware/docker-volume-vsphere/issues/957#issue-209840924 & https://github.com/vmware/docker-volume-vsphere/issues/957#issuecomment-282431323

shaominchen:
As discussed offline with @shuklanirdesh82, we agree that the test cases described in #957 are quite distinct, so we should move some test scenarios to other test groups. I will file separate issues to track those scenarios. Currently the basic_test group will only include 3 tests: TestVolumeLifecycle, TestBasicVolumeIsolation, and TestVmGroupVolumeIsolation.

shuklanirdesh82:
Thanks @shaominchen for breaking this down and creating separate issues. for the 0.16 there is only one testcase being tracked for this issue. 

Can you please refresh the estimate for `TestVmGroupVolumeIsolation `?

PR #1276, opened 2017-05-19
[Automation Test] Add 2 basic test cases

shaominchen:
1. TestVolumeLifecycle
2. TestVolumeIsolation

### DIFF
--- a/tests/e2e/basic_test.go
+++ b/tests/e2e/basic_test.go
@@ -73,6 +73,10 @@ func (s *BasicTestSuite) TestVolumeCreation(c *C) {
 		out, err = dockercli.RemoveContainer(host, s.containerName)
 		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
 
+		// Verify volume status
+		status = verification.VerifyDetachedStatus(s.volumeName, host, s.esxName)
+		c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
+
 		// Clean up the volume
 		out, err = dockercli.DeleteVolume(host, s.volumeName)
 		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
--- a/tests/e2e/basic_test.go
+++ b/tests/e2e/basic_test.go
@@ -52,9 +52,9 @@ var _ = Suite(&BasicTestSuite{})
 // 1. Create a volume
 // 2. Verify the volume is available
 // 3. Attach the volume
-// 4. Verify volume status
+// 4. Verify volume status is attached
 // 5. Remove the container
-// 6. Verify volume status
+// 6. Verify volume status is detached
 // 7. Remove the volume
 // 8. Verify the volume is unavailable
 func (s *BasicTestSuite) TestVolumeLifecycle(c *C) {
--- a/tests/e2e/basic_test.go
+++ b/tests/e2e/basic_test.go
@@ -37,8 +37,8 @@ type BasicTestSuite struct {
 
 func (s *BasicTestSuite) SetUpTest(c *C) {
 	s.esxName = os.Getenv("ESX")
-	s.volumeName = inputparams.GetVolumeNameWithTimeStamp()
-	s.containerName = inputparams.GetContainerNameWithTimeStamp()
+	s.volumeName = inputparams.GetVolumeNameWithTimeStamp("basic_test")
+	s.containerName = inputparams.GetContainerNameWithTimeStamp("basic_test")
 }
 
 var _ = Suite(&BasicTestSuite{})
--- a/tests/e2e/basic_test.go
+++ b/tests/e2e/basic_test.go
@@ -47,8 +47,18 @@ var _ = Suite(&BasicTestSuite{})
 // VM1 - local VMFS datastore
 // VM2 - shared VMFS datastore
 // VM3 - shared VSAN datastore
-func (s *BasicTestSuite) TestVolumeCreation(c *C) {
-	log.Printf("START: basic-test.TestVolumeCreation")
+//
+// Test steps:
+// 1. Create a volume
+// 2. Verify the volume is available
+// 3. Attach the volume
+// 4. Verify volume status
+// 5. Remove the container
+// 6. Verify volume status
+// 7. Remove the volume
+// 8. Verify the volume is unavailable
+func (s *BasicTestSuite) TestVolumeLifecycle(c *C) {
+	log.Printf("START: basic-test.TestVolumeLifecycle")
 
 	dockerHosts := []string{os.Getenv("VM1"), os.Getenv("VM2"), os.Getenv("VM3")}
 	for _, host := range dockerHosts {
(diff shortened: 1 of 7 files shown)


shaominchen:
Fix issue #1263.

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,110 @@
...
+		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
+
+		// Verify volume status
+		status := verification.VerifyAttachedStatus(s.volumeName, host, s.esxName)
+		c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
+
+		// Clean up the container
+		out, err = dockercli.RemoveContainer(host, s.containerName)
+		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
+
```
ashahi1: Shall we add a check here that volume is detached after we stopped the container just like we are verifying after attaching the volume?

shaominchen: Make sense. I will add this.

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+		// Attach the volume
+		out, err = dockercli.AttachVolume(host, s.volumeName, s.containerName)
+		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
+
+		// Verify volume status
+		status := verification.VerifyAttachedStatus(s.volumeName, host, s.esxName)
+		c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
+
+		// Clean up the container
```
shuklanirdesh82: nit: I would suggest to be "Remove/Delete container"

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+		out, err = dockercli.RemoveContainer(host, s.containerName)
+		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
+
+		// Verify volume status
+		status = verification.VerifyDetachedStatus(s.volumeName, host, s.esxName)
+		c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
+
+		// Clean up the volume
+		out, err = dockercli.DeleteVolume(host, s.volumeName)
```
shuklanirdesh82: You might want to consider moving `DeleteVolume` & `RemoveContainer` into testCleanup() fixture to reuse the code.

tests/utils/dockercli/volumelifecycle.go
```diff
@@ -25,16 +25,15 @@ import (
 	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
 )
 
-// CreateDefaultVolume is going to create vsphere docker volume with
-// defaults.
-func CreateDefaultVolume(ip, name string) ([]byte, error) {
+// CreateVolume is going to create vsphere docker volume with given name.
+func CreateVolume(ip, name string) ([]byte, error) {
```
shuklanirdesh82: how about `CreateVolumeWithDefaults`? inshort .. it would be better to distinguish the volume creation method name. 

Please share your idea on the same.

tests/utils/inputparams/testparams.go
```diff
@@ -42,8 +43,13 @@ func GetVolumeName() string {
...
 // GetVolumeNameWithTimeStamp prepares unique volume name by appending current time-stamp value
-func GetVolumeNameWithTimeStamp(volName string) string {
-	return volName + "_" + strconv.FormatInt(time.Now().Unix(), 10)
+func GetVolumeNameWithTimeStamp() string {
+	return "volume_" + strconv.FormatInt(time.Now().Unix(), 10)
+}
+
+// GetContainerNameWithTimeStamp prepares unique container name by appending current time-stamp value
+func GetContainerNameWithTimeStamp() string {
+	return "busybox_" + strconv.FormatInt(time.Now().Unix(), 10)
```
shuklanirdesh82: I would suggest avoid using `busybox_` ... if would be confusing while creating container of `ubuntu` base image. Let' use generic name like `container_`

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+	accessible := verification.CheckVolumeAccessibility(vm1, s.volumeName)
+	c.Assert(accessible, Equals, true, Commentf("Volume %s is not accessible", s.volumeName))
+
+	// Verify the volume is inaccessible from VM4
+	accessible = verification.CheckVolumeAccessibility(vm2, s.volumeName)
+	//TODO: VM2 inaccessible to this volume is currently not available
+	//c.Assert(accessible, Equals, false, Commentf("Volume %s is accessible", s.volumeName))
+
+	// Clean up the volume
+	out, err = dockercli.DeleteVolume(vm1, s.volumeName)
```
shuklanirdesh82: same as above ... we can reuse the code .. not super urgent but it would be harder later to refactor. 

this what I am thinking .. the moment we create volume next step adds to cleanup list and in the `tearDown` or `testCleanup` iterates over the cleanList and start deletion if the list is not empty.

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+	s.containerName = inputparams.GetContainerNameWithTimeStamp()
+}
+
+var _ = Suite(&BasicTestSuite{})
+
+// Test volume lifecycle management on different datastores:
+// VM1 - local VMFS datastore
+// VM2 - shared VMFS datastore
+// VM3 - shared VSAN datastore
+func (s *BasicTestSuite) TestVolumeCreation(c *C) {
```
shuklanirdesh82: nit: better to add details steps carried from GH issue or confluence page to get a quick sense what the test method is trying to do instead of scanning each line.

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, host := range dockerHosts {
+		//TODO: Remove this check once VM3 is available from the CI testbed
+		if host == "" {
+			continue
+		}
+
+		// Create a volume
+		out, err := dockercli.CreateVolume(host, s.volumeName)
+		c.Assert(err, IsNil, Commentf(misc.FormatOutput(out)))
+
```
ashahi1: It would be nice if we could also verify created volume from - docker volume ls command

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+var _ = Suite(&BasicTestSuite{})
+
+// Test volume lifecycle management on different datastores:
+// VM1 - local VMFS datastore
+// VM2 - shared VMFS datastore
+// VM3 - shared VSAN datastore
+func (s *BasicTestSuite) TestVolumeCreation(c *C) {
+	log.Printf("START: basic-test.TestVolumeCreation")
+
+	dockerHosts := []string{os.Getenv("VM1"), os.Getenv("VM2"), os.Getenv("VM3")}
```
pshahzeb: If these vars are required for entire suite, we can get them once in SetupSuite fixture. Get all three of them and use whatever is required in the tests.

tests/e2e/basic_test.go
```diff
@@ -0,0 +1,114 @@
...
+// Test volume visibility: volume is created on the local datastore attached to VM1:
+// VM1 - accessible to the volume
+// VM2 - inaccessible to the volume
+func (s *BasicTestSuite) TestVolumeAccessibility(c *C) {
+	log.Printf("START: basic-test.TestVolumeAccessibility")
+
+	vm1, vm2 := os.Getenv("VM1"), os.Getenv("VM2")
+
+	// Create a volume from VM1
+	out, err := dockercli.CreateVolume(vm1, s.volumeName)
```
pshahzeb: From these 2 tests, I feel creating volume can be moved to test fixture. If you think this part is not common in view of more tests case to come then it is ok to keep here.

tests/utils/verification/volumeproperties.go
```diff
@@ -96,6 +96,12 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
 	return op
 }
 
+// CheckVolumeAccessibility returns true if the given volume is accessible
+// from the specified VM; false otherwise.
+func CheckVolumeAccessibility(hostName string, volumeName string) bool {
+	return GetVolumeStatusHost(volumeName, hostName) != ""
```
pshahzeb: By accessible, it means the docker volume inspect (to get status of volume) from the given docker host was successful. Is this what you meant?

tests/e2e/basic_test.go
shaominchen: Adding detailed steps will be too tedious, which is also kind of a duplicate of the test code itself. I can add more comments though to make the intention of the test case more clear.

tests/e2e/basic_test.go
shaominchen: OK.

tests/e2e/basic_test.go
shaominchen: RemoveContainer can't be  moved into TearDownTest as the test case needs to verify volume detached status after this. As for DeleteVolume, if all test cases share the same docker hosts (VMs), it will be easier to move this to TearDownTest. So for now I will keep them as is in this test suite. I do agree with you the common cleanup logic should be moved into TearDownTest whenever possible.

tests/e2e/basic_test.go
shaominchen: I agree we should try to reuse the code if possible. On the other hand, keeping code logic symmetric will make it easier to understand. With this considered, if we initialize some resources in SetUpTest(), it will be quite natural (and expected) to clean up the resources in TearDownTest(). For these test cases, we are creating volumes - as part of the testing logic - inside the test case, so I think it's quite acceptable to clean up the resource inside the test case.

tests/utils/dockercli/volumelifecycle.go
shaominchen: I've thought about this. I don't think we actually need many different functions to do the same thing - create a volume with different parameters. Even though Go language does not support function overloading, but we can very easily simulate that using a variadic function - a function that accepts a variable number of parameters. This way, we can call the same API to create different volumes based on needs.

So for now I just rename this, and we can change this to a variadic function later whenever we need it.

tests/utils/inputparams/testparams.go
shaominchen: Sounds fine to me.

tests/e2e/basic_test.go
shaominchen: Make sense. I will also rename this test to testVolumeLifecycle.

tests/e2e/basic_test.go
shaominchen: Different test cases need different VMs: some just need 1 or 2 VMs, some other might need 3 or even more VMs (not available right now). I will consider to move them to SetUp if we have a consistent usage model of these VMs. For now I'd prefer to keep it as is.

tests/e2e/basic_test.go
shaominchen: For these 2 test cases, createVolume is part of the testing logic. It's actually more straightforward to read and understand the testing logic by keeping them in the test case. However, for other testing scenarios, if creating volume is just a prerequisite instead of an important step of the whole testing logic, we can definitely move it to SetUp test fixture.

tests/utils/verification/volumeproperties.go
shaominchen: Yes. We have 2 different way to implement this:
1. Using "docker volume inspect --name" to get the status of given volume name
2. Using "docker volume ls" to get the list of all existing volumes and check if the name is in the list
I used #1 just because it's easier.

tests/e2e/basic_test.go
shuklanirdesh82: test steps would be useful while raising an issue by copy/paste steps to point out where exactly it fails. 

some sort of high level steps are sufficient.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: sorry it sounds confusing to me now .. `CheckVolumeAccessibility`, does this method checks whether the volume is `attached` or not?

in above comment, `docker volume ls` gives you whether volume is visible to the docker host.

can you please help me to understand here?

shaominchen: No, this method does not check volume attach/detach status. It only checks whether the volume is available (or visible/accessible) from the given VM.

tests/e2e/basic_test.go
shaominchen: The in-line comments in the test method are exactly the detailed steps. This is just a preference I think. I'm fine with either way, but let's discuss offline and get to an agreement.

tests/utils/inputparams/testparams.go
```diff
@@ -42,8 +43,13 @@ func GetVolumeName() string {
 }
 
 // GetVolumeNameWithTimeStamp prepares unique volume name by appending current time-stamp value
-func GetVolumeNameWithTimeStamp(volName string) string {
```
shuklanirdesh82: This is kind of useful util while debugging failure to search for testcase customized string .. 

As discussed offline please keep this util.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,27 +96,43 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+
 // VerifyAttachedStatus - verify volume is attached and name of the VM attached
 // is consistent on both docker host and ESX
 func VerifyAttachedStatus(name, hostName, esxName string) bool {
 	log.Printf("Confirming attached status for volume [%s]\n", name)
 
 	vmAttachedHost := GetVMAttachedToVolUsingDockerCli(name, hostName)
 	vmAttachedESX := GetVMAttachedToVolUsingAdminCli(name, esxName)
-	expectedVMName := govc.RetrieveVMNameFromIP(hostName)
+	//expectedVMName := govc.RetrieveVMNameFromIP(hostName)
```
shuklanirdesh82: Please raise an issue to uncomment required check.

pshahzeb: Reopening #1277
Adding followup comment on that issue to uncomment this check.

Issue #1294, opened 2017-05-24
E2E test util verification.ExecCmd may cause the whole e2e test exit


shaominchen:
E2E test util verification.ExecCmd may cause the whole e2e test exit due to log.Fatal() call. Instead of fixing log.Fatal(), we should simply remove this util, because it does nothing, but just checking the error and writing a log. We should use ssh.InvokeCommand directly.

@shuklanirdesh82 @ashahi1 @pshahzeb @govint

govint:
When errors happen in the test run we need to know if the issue was with the command executed (syntax or invalid command options for example) or an issue with the service/plugin. Its best if ssh.InvokeCommand() logs the actual command that's was executed - especially when the command fails (at least).

ssh.InvokeCommandLocally() is checking and logging a fatal log if the command failed.

shuklanirdesh82:
It is a trade-off using `log.fatal()`; are we agreeing to get same kind of failures/cascading failure or stop immediately when one of the test fails to invoke supplied command?

When something goes wrong while invoking command incase of remote host is not available (for example); `log.fatal()` makes sense in that scenario otherwise it needs to be change and use `log.Printf()` and let test decide what to do.

govint:
Actually, fatal level logs aren't needed, the test thats calling InvokeCommand will check the error and do a C.Assert which will stop that test anyway with logs.

shaominchen:
I agree with Govindan here. We shouldn't use fatal log in any tests because it will exit the whole test program, which is definitely not we want. As Govinden suggested, returning the error and let the test cases to decide what to do.

shuklanirdesh82:
Thanks @govint & @shaominchen ... I will take care and make a uniform change.

shuklanirdesh82:
E2E test util verification.ExecCmd has been removed so closing this issue not and opening a separate issue to take care about removing `log.fatalf()` after getting massive refactor from #1096.

PR #1300, opened 2017-05-25
Handling cleanup properly for vmlistener test (#1298)

### LINKED ISSUE #1298: Unable to remove volume even though volume is not attached
Test failed to remove volume even though volume is not attached.
Hit this issue on CI: https://ci.vmware.run/vmware/docker-volume-vsphere/259

Test logs:
```
2017/05/25 19:45:43 START: Test vmdkops service restart + kill vm process
2017/05/25 19:45:43 Creating volume [vmlistener_test_volume_1495741543] on VM [192.168.31.83]
2017/05/25 19:46:01 Attaching volume [vmlistener_test_volume_1495741543] on VM [192.168.31.83]
2017/05/25 19:46:07 Confirming attached status for volume [vmlistener_test_volume_1495741543]
2017/05/25 19:46:11 Restarting vmdkops service on ESX host [192.168.31.62]
2017/05/25 19:46:14 Confirming attached status for volume [vmlistener_test_volume_1495741543]
2017/05/25 19:46:19 Retrieving VM power state for [Ubuntu.14.04]
2017/05/25 19:46:19 VM[Ubuntu.14.04]'s current power state is [poweredOn]
2017/05/25 19:46:19 Retrieving World ID for VM [Ubuntu.14.04] from ESX [192.168.31.62]
2017/05/25 19:46:19 VM's process ID is: 130824
2017/05/25 19:46:19 Killing VM from ESX [192.168.31.62]
2017/05/25 19:46:20 Sleep for 10 seconds
2017/05/25 19:46:30 Retrieving VM power state for [Ubuntu.14.04]
2017/05/25 19:46:30 VM[Ubuntu.14.04]'s current power state is [poweredOff]
2017/05/25 19:46:33 Powering on VM [Ubuntu.14.04]
2017/05/25 19:46:37 Sleep for 35 seconds
2017/05/25 19:47:12 Retrieving VM power state for [Ubuntu.14.04]
2017/05/25 19:47:12 VM[Ubuntu.14.04]'s current power state is [poweredOn]
2017/05/25 19:47:12 Confirming detached status for volume [vmlistener_test_volume_1495741543]
2017/05/25 19:47:19 END: Test vmdkops service restart + kill vm process
2017/05/25 19:47:19 Destroying volume [vmlistener_test_volume_1495741543]

FAIL: vmlistener_test.go:61: VMListenerTestParams.TearDownTest

vmlistener_test.go:65:
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc420164b60), Stderr:[]uint8(nil)} ("exit status 1")
... Error response from daemon: unable to remove volume: remove vmlistener_test_volume_1495741543: volume is in use - [974c2954d9dda822c15c048edf4e320ae6493c2623552b42c9e91729d52f0104]


PANIC: vmlistener_test.go:78: VMListenerTestParams.TestKillVM

... Panic: Fixture has panicked (see related PANIC)
--- FAIL: Test (350.49s)

```

**Test steps:**
```
1. Create a volume
2. Attach it to a container
3. Restart the vmdkops service
4. Verify vmdkops_admin volume ls > verify volume stays as attached
5. Kill the vm
6. Verification: volume status should be detached
7. Power-on the vm
8. Remove the volume  <<<<<<<<<<<<<<<<<<<<<<<<< Failed here.
```

shuklanirdesh82:
fixes #1298

What's in this PR (resolving blocker #1298)

- Retry mechanism for waiting for vm's power state to be changed ( #1301 can be treated as sample)
- Checking whether container is present to delete #1298
- Removing references of verification.ExecCmd() #1294

Not addressed things in this PR
- Major cleanup will take place as part of #1096
- Remaining log.fatal() references (will be removed in the next PR)

Testing done: Yes (locally)

```
=== RUN   Test
2017/05/25 22:20:55 Finding VM name from IP Address [10.20.104.62]
2017/05/25 22:20:56 START: Test vmdkops service restart + kill vm process
2017/05/25 22:20:56 Creating volume [vmlistener_test_volume_1495750856] on VM [10.20.104.62]
2017/05/25 22:21:02 Attaching volume [vmlistener_test_volume_1495750856] on VM [10.20.104.62]
2017/05/25 22:21:04 Confirming attached status for volume [vmlistener_test_volume_1495750856]
2017/05/25 22:21:08 Restarting vmdkops service on ESX host [10.20.105.97]
2017/05/25 22:21:12 Confirming attached status for volume [vmlistener_test_volume_1495750856]
2017/05/25 22:21:17 Retrieving VM power state for [ubuntu]
2017/05/25 22:21:18 VM[ubuntu]'s current power state is [poweredOn]
2017/05/25 22:21:18 Retrieving World ID for VM [ubuntu] from ESX [10.20.105.97]
2017/05/25 22:21:19 VM's process ID is: 3900500
2017/05/25 22:21:19 Killing VM from ESX [10.20.105.97]
2017/05/25 22:21:20 Sleep for 10 seconds
2017/05/25 22:21:30 Retrieving VM power state for [ubuntu]
2017/05/25 22:21:31 VM[ubuntu]'s current power state is [poweredOff]
2017/05/25 22:21:33 Powering on VM [ubuntu]
2017/05/25 22:21:36 Sleep for 45 seconds
2017/05/25 22:22:21 Retrieving VM power state for [ubuntu]
2017/05/25 22:22:22 VM[ubuntu]'s current power state is [poweredOn]
2017/05/25 22:22:22 Confirming detached status for volume [vmlistener_test_volume_1495750856]
2017/05/25 22:22:26 END: Test vmdkops service restart + kill vm process
2017/05/25 22:22:26 Checking container [vmlistener_test_container_1495750856] presence on VM [10.20.104.62]
2017/05/25 22:22:26 Destroying volume [vmlistener_test_volume_1495750856]
OK: 1 passed
--- PASS: Test (93.01s)
```

/CC @ashahi1 @shaominchen


tests/e2e/vmlistener_test.go
```diff
@@ -120,7 +124,7 @@ func (s *VMListenerTestParams) TestKillVM(c *C) {
 
 	// power on vm
 	govc.PowerOnVM(s.dockerHostName)
-	misc.SleepForSec(35)
+	misc.SleepForSec(45)
```
shaominchen: Can we poll VM state instead of blindly waiting for such as long time?

shuklanirdesh82: I have tried the poller and with following sequence needed to rely on the static time out

1. vm is getting powered on within 10 seconds (able to ping the vm but VM is not in fully powered mode)
2. docker daemon takes some time to respond and it is variable
3. even thought docker is available there is no way to find out docker has finished its background processing
4. vDVS plugin refcount thread is running in background and no hook to hang on reliably 

Ideally would like to have poller for step 1 & 4. Step 1 is possible but not 2/3/4.

Let me know what do you think.

tests/e2e/vmlistener_test.go
```diff
@@ -61,6 +61,10 @@ func (s *VMListenerTestParams) SetUpTest(c *C) {
 func (s *VMListenerTestParams) TearDownTest(c *C) {
 	// Note: no need to remove container, it is already removed as
```
ashahi1: Should this be changed since we are checking if container exists?

shuklanirdesh82: yeah, why not? will be updated in the next diff.

tests/e2e/vmlistener_test.go
shaominchen: I don't quite follow. We are just waiting for some time to check the VM status, right? We can introduce a util function to periodically call govc.GetVMPowerState(). If this function returns expected state, return immediately; otherwise, sleep for a very short time and retry. This function should return after a maximum limit. Sample code will be like this:

for {
    state := govc.GetVMPowerState(hostName)
    if state == "expected_state" {
        return true;
    }
    if timePassed > maximumLimit {
        return false;
    }
    sleep(3)
}

pshahzeb: I am not sure if going ahead with tests if VM state is desired one, especially power on is a good idea. there is still a delay where docker itself tries to come up, it itself tries to talk to plugin (btw it does exponential backoff) and then plugin comes up and stabilizes refcount. 
If this isn't clear, we can quickly discuss offline.

shaominchen: We discussed this offline. Now we all agree that we need a general util function to do polling. This util function should accept a function, expected status, maximum time limit as parameters. This way we can avoid blind waiting here and there, which will significantly slow down the whole test suite.

As for now, since issue #1298 is a blocking issue, I'm fine with this quick-and-dirty solution (blind sleep) so that others won't be blocked by this.

shuklanirdesh82: In addition to above comment we definitely need #1294 (not to log.fatal()) as the powerOn status comes back immediately meanwhile guest os has to bootup > NIC has to retrieve IP address until this step "host is not reachable" error hits. 

#1294 is the must need to bypass `log.fatal()`, introduced this static wait time. 

@shaominchen has raised a PR to introduce common poller #1301

shuklanirdesh82: let me see I can squeeze #1294, #1301 as part of this review.

tests/e2e/vmlistener_test.go
```diff
@@ -59,8 +59,10 @@ func (s *VMListenerTestParams) SetUpTest(c *C) {
 }
 
 func (s *VMListenerTestParams) TearDownTest(c *C) {
```
govint: Should be TearDownSuite() for the whole test and TearDownTest for a specific test, in gocheck.

tests/e2e/vmlistener_test.go
```diff
@@ -59,8 +59,10 @@ func (s *VMListenerTestParams) SetUpTest(c *C) {
...
-	// Note: no need to remove container, it is already removed as
-	// vm got killed and `--restart=always` flag is not added
+	if dockercli.IsContainerExist(s.dockerHostIP, s.containerName) {
```
govint: Shouldn't SetupSuite() use the inputparams getEndPoint*() functions to get the dockerHostIP()? Same for the ESX IP.

govint: Instead of removing the container here can we run it with the --rm flag so it will be removed by docker when the container exits.

tests/e2e/volumepropverification_test.go
```diff
@@ -123,7 +123,7 @@ func TestVolumeProperties(t *testing.T) {
 				}
 				log.Println("Finished verifying volume properties like size, disk-format and attached-to-vm fields"+
 					" at vm and esx for volume - ", volName)
-				verification.ExecCmd(vms[vmIndx], "docker stop "+containerName+" ; docker rm "+containerName)
+				ssh.InvokeCommand(vms[vmIndx], "docker stop "+containerName+" ; docker rm "+containerName)
```
govint: Prefer to make this a function to say DeleteRunningContainer() which will stop and then remove the container. And it should be in the dockercli package. Else this model will allow tests to create scripts inside of a call to invokeCommand.

govint: Also please use "dockerHosts" in place of vms. In other tests dockerHosts is being used as the array of endpoints to use. Or standardize on a name - endpoint, vms, dockerHost (preferrable as it has a meaning in these tests).

tests/utils/dockercli/volumelifecycle.go
```diff
@@ -144,3 +144,14 @@ func StopContainer(ip, containerName string) (string, error) {
 	log.Printf("Stopping container [%s] on VM [%s]", containerName, ip)
 	return ssh.InvokeCommand(ip, dockercli.StopContainer+containerName)
 }
+
+// IsContainerExist - return true if container exists otherwise false
+func IsContainerExist(ip, containerName string) bool {
+	log.Printf("Checking container [%s] presence on VM [%s]", containerName, ip)
```
govint: Use gocheck Logf() instead? May be create a logging function in misc.go and use that in all test code.

tests/utils/verification/volumeproperties.go
```diff
@@ -21,32 +21,33 @@ package verification
...
-	pollTimeout = 64
+	pollTimeout = 60
 )
 
 // GetVMAttachedToVolUsingDockerCli returns attached to vm field of volume using docker cli
+// TODO: make this private member after finishing refactoring of volmprop_test.go and remove this TODO
 func GetVMAttachedToVolUsingDockerCli(volName string, hostname string) string {
 	cmd := dockercli.InspectVolume + " --format '{{index .Status \"attached to VM\"}}' " + volName
-	op := ExecCmd(hostname, cmd)
+	op, _ := ssh.InvokeCommand(hostname, cmd)
```
govint: Do we check for error and then the output?

tests/utils/verification/volumeproperties.go
```diff
@@ -21,32 +21,33 @@ package verification
...
 	if op == "" {
 		log.Fatal("Null value is returned by docker cli when looking for attached to vm field for volume. Output: ", op)
```
govint: Pls. convert this to a gocheck Assert()

tests/utils/verification/volumeproperties.go
```diff
@@ -21,32 +21,33 @@ package verification
...
 }
 
 // GetVMAttachedToVolUsingAdminCli returns attached to vm field of volume using admin cli
 func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
 	cmd := admincli.ListVolumes + "-c volume,attached-to 2>/dev/null | grep " + volName
-	op := ExecCmd(hostname, cmd)
+	op, _ := ssh.InvokeCommand(hostname, cmd)
 	volProps := strings.Fields(op)
 	if op == "" {
 		log.Fatal("Null value is returned by admin cli when looking for attached to vm field for volume. Output: ", op)
```
govint: Should be gocheck Assert()

tests/utils/verification/volumeproperties.go
```diff
@@ -21,32 +21,33 @@ package verification
...
 	}
-	return strings.TrimSpace(op)
+	return op
 }
 
 // GetVMAttachedToVolUsingAdminCli returns attached to vm field of volume using admin cli
 func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
 	cmd := admincli.ListVolumes + "-c volume,attached-to 2>/dev/null | grep " + volName
-	op := ExecCmd(hostname, cmd)
+	op, _ := ssh.InvokeCommand(hostname, cmd)
```
govint: Check for error and then output? the command may be deprecated or incorrect syntax.

tests/utils/verification/volumeproperties.go
```diff
@@ -62,7 +63,7 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
 // for volume using Admin cli
 func GetVolumePropertiesAdminCli(volName string, hostname string) string {
 	cmd := admincli.ListVolumes + "-c volume,attached-to,capacity,disk-format 2>/dev/null | grep " + volName
-	op := ExecCmd(hostname, cmd)
+	op, _ := ssh.InvokeCommand(hostname, cmd)
 	if op == "" {
```
govint: Convert to gocheck Assert()

tests/utils/verification/volumeproperties.go
```diff
@@ -77,7 +78,7 @@ func GetVolumePropertiesAdminCli(volName string, hostname string) string {
 // for volume using Docker cli
 func GetVolumePropertiesDockerCli(volName string, hostname string) string {
 	cmd := dockercli.InspectVolume + " --format '{{index .Status.capacity.size}} {{index .Status.diskformat}} {{index .Status \"attached to VM\"}}' " + volName
-	op := ExecCmd(hostname, cmd)
+	op, _ := ssh.InvokeCommand(hostname, cmd)
 	expctedLen := 0
 	if op == "" {
```
govint: All instances of log.Fatal should be gocheck Assert() whcih ensures that only the current test is failed.

tests/utils/verification/volumeproperties.go
```diff
@@ -101,14 +102,17 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
 func CheckVolumeAvailability(hostName string, volumeName string) bool {
 	log.Printf("Checking volume [%s] availability from VM [%s]\n", volumeName, hostName)
```
govint: this should be gocheck Logf() - possibly define a misc.Logf() that uses gocheck Logf to keep all logs uniform,

tests/utils/verification/volumeproperties.go
```diff
@@ -101,14 +102,17 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+		return false
+	}
 	//TODO: add more detailed verification here, e.g. checking volume driver name
 	return strings.Contains(volumes, volumeName)
 }
 
-// GetDockerVolumes returns all docker volumes available from the given host
-func GetDockerVolumes(hostName string) string {
-	return ExecCmd(hostName, dockercli.ListVolumes)
+// getDockerVolumes returns all docker volumes available from the given host
```
govint: Pls. merge with above function, its used nowhere else and its just a call to return InvokeCommand result.

tests/utils/verification/volumeproperties.go
```diff
@@ -132,19 +136,22 @@ func VerifyAttachedStatus(name, hostName, esxName string) bool {
...
-func GetVolumeStatusHost(name, hostName string) string {
+// getVolumeStatusHost - get the volume status on a given host
+func getVolumeStatusHost(name, hostName string) string {
 	cmd := dockercli.InspectVolume + " --format '{{index .Status.status}}' " + name
-	op := ExecCmd(hostName, cmd)
-	return strings.TrimSpace(op)
+	// ignoring the error here, helper is part of polling util
+	// error most likely to be "unable to reach host [ssh:255 error]"
+	// VerifyDetachedStatus takes care of retry mechanism
+	out, _ := ssh.InvokeCommand(hostName, cmd)
```
govint: Or return error and let the caller figure if it wants to continue sleep-retry

shuklanirdesh82:
> https://github.com/vmware/docker-volume-vsphere/pull/1300#pullrequestreview-40504682
> Few comments around using gocheck functions vs. fatal logs and using the input params for all parameter assignments.

Thanks for the review! All the changes that you pointed out were planned to address as part of separate issue. Please refer https://github.com/vmware/docker-volume-vsphere/pull/1300#pullrequestreview-40504682 to see what is being addressed as part of this review and not. We may address all in this PR but the objective of this PR is almost changed and scope is getting bigger.

tests/e2e/volumepropverification_test.go
```diff
@@ -71,7 +71,7 @@ func TestVolumeProperties(t *testing.T) {
 		log.Println("running end_to_end tests - current time: ", time.Now())
 		vmIP = vms[vmIndx]
 		log.Println("Running test on VM - ", vmIP)
-		dockerCliCheck = verification.IsDockerCliCheckNeeded(vms[vmIndx])
+		dockerCliCheck = true
```
shaominchen: If the value is always true, why do we even need this?

tests/e2e/vmlistener_test.go
shaominchen: +1

tests/utils/misc/misc.go
```diff
@@ -21,8 +21,29 @@ import (
...
 
+type vmstatus func(string) string
+
+const (
+	maxAttempt = 60
+)
+
 // SleepForSec sleep for a given number of seconds
 func SleepForSec(sec int) {
 	log.Printf("Sleep for %d seconds", sec)
```
shaominchen: Won't this print too many logs since it's now running in a loop every second?

tests/utils/misc/misc.go
```diff
@@ -21,8 +21,29 @@ import (
 	"time"
 )
 
+type vmstatus func(string) string
```
shaominchen: 1. Can we give this function a better name? "vmstatus" doesn't look like a function. 
2. Is it possible to use interface or pointers to support any type of functions? I'm not sure about this, just want to know if it's possible to extend this to a general solution that accepts any functions.

tests/utils/verification/volumeproperties.go
```diff
@@ -21,32 +21,33 @@ package verification
...
 	"github.com/vmware/docker-volume-vsphere/tests/constants/admincli"
 	"github.com/vmware/docker-volume-vsphere/tests/constants/dockercli"
 	"github.com/vmware/docker-volume-vsphere/tests/constants/properties"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
 	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
 )
 
 const (
-	pollTimeout = 64
+	pollTimeout = 60
```
shaominchen: Nit: why don't we use the same name as above - maxAttempt?

tests/e2e/volumepropverification_test.go
shuklanirdesh82: this is to remove `IsDockerCliCheckNeeded ` quickly; it will be completely removed as part of #1096 ... as pointed out  https://github.com/vmware/docker-volume-vsphere/pull/1300#issue-231476944

tests/e2e/vmlistener_test.go
shuklanirdesh82: > Instead of removing the container here can we run it with the --rm flag so it will be removed by docker when the container exits.

We may miss the scenario when docker hosts comes back .. containers are still exist in `Exited` status even though it was "Up and running" and vDVS plugin is marking its status as `detached` hence I have avoided passing `--rm`.

tests/utils/dockercli/volumelifecycle.go
shuklanirdesh82: Sure .... I would like to extract this comment out as a separate PR to make uniform change.

tests/utils/misc/misc.go
shuklanirdesh82: 1. renamed to `getVMPowerStatus`
2. that's a great idea .. we may explore more to address as part of #1301

tests/utils/misc/misc.go
shuklanirdesh82: It depends on the caller, I have updated existing callers to invoke this not every seconds ... BTW this is indeed a require logging to know during test execution instead of assuming a deadlock or infinite loop.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: done!

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: will be addressed as part of separate PR.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: oh yeah that's correct. Removing this and make it as inline.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: so far it is a private member and there is only one caller .. we can enhance if needed in future.

Issue #1249, opened 2017-05-16
Automate docker swarm P0 testcases


shuklanirdesh82:
Detailed steps are outlined at https://github.com/vmware/docker-volume-vsphere/issues/601#issuecomment-282157695

shaominchen:
Fixed with PR #1309.

Issue #601, opened 2016-09-15
Verify docker service commands with the volume plugin 


govint:
Docker 1.12 and 1.12.1 added up docker service command to create, scale and manage services via docker. We need to verify the volume plugin works seamlessly (especially the --mount option with type=volume) in any of the use cases that docker service is used.

govint:
From last week, ran docker engine in swarm mode on two VMs and created a service that uses a persistent volume with single replica, two replicas and in global mode. The single replica works fine with the volume getting created as part of defining the service via "docker service create".

The two replicas scenario also works fine with the two running on the same node.

Global mode is a bit mixed up where one docker host (VM) gets to mount the volume and the second seems to be using a local docker volume. Ideally it would be correct to have the volume plugin create a snap of the persistent volume and each replica uses a snap of the single volume. The plugin needs to be supporting such an interface (Docker needs to define that). The attach_as attribute should allow a non-persistent mode which should allow creating a snap when a service instance runs.

Will update more on the global mode as I investigate this.

govint:
Besides not being able to share volumes on container instances across VMs we can showcase using the VMDK driver for cases where a service can migrate across hosts and VMs and still be able to access its data, like this one - https://github.com/codedellemc/labs/tree/master/demo-persistence-with-minecraft-docker.
cc @asomaiah

ashahi1:
Created a two seperate swarm setup of three nodes - one master and two worker nodes. 
Docker version: 1.12.1

Setup:
1. All three VMs are on the same host and shared-vmfs datastore.
2. All three VMs are on the same host and vsan datastore.

Performed following steps across both the setups so that we cover  both vsan and vmfs datastores :
--------
Steps for testing docker service commands:
1. Created a thin provisioned volume from master vm
2. Created a service with replica 1
command: docker service create --replicas 1 --name SrvcOnThinVolm --mount type=volume,source=volXXX,target=/vol,volume-driver=vmdk busybox tail -f /dev/null
3. Verified from VC and ESX - volume was successfully attached to the worker-1
4. Scaled the replica to 2
5. Stopped the service on worker-1. Checked in VC and ESX, volume got attached to worker-2 and worker-1 no longer had volume attached.
6. Tried deleting the volume, but failed as volume was still being used.
7. Deleted the service running in swarm. Volume was no longer attached to any node.
8. Removed the volume successfully.
----------

Did the same steps for volumes of different disk-format types: zeroedthick and eagerzeroedthick

[...]

PR #1309, opened 2017-05-30
Add swarm_test to e2e test suite

shaominchen:
Add swarm_test to e2e test suite: fixes issue #1249.

Passed all e2e tests locally.


shuklanirdesh82:
@shaominchen can you please increase default timeout for running our e2e test to 30 mins `-timeout 30m`?

tests/e2e/swarm_test.go
```diff
@@ -0,0 +1,141 @@
...
+// 12. Verify the volume is in attached status
+// 13. Delete the volume - expect fail
+// 14. Remove the service
+// 15. Verify the service is gone
+// 16. Verify the volume is in detached status
+func (s *SwarmTestSuite) TestDockerSwarm(c *C) {
+	log.Printf("START: swarm_test.TestDockerSwarm")
+
+	fullVolumeName := verification.GetFullVolumeName(s.master, s.volumeName)
+	opts := "--replicas 1 --mount type=volume,source=" + fullVolumeName + ",target=/vol,volume-driver=vsphere busybox tail -f /dev/null"
```
shuklanirdesh82: minor: better to reuse const from https://github.com/vmware/docker-volume-vsphere/blob/master/tests/constants/dockercli/cmd.go e.g. `TestContainer, vDVSPluginName ` (need to export `vDVSPluginName`)

tests/e2e/swarm_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	status = verification.VerifyDockerService(s.master, s.serviceName, 2)
+	c.Assert(status, Equals, true, Commentf("Service %s is not running", s.serviceName))
+
+	status, host := verification.VerifyDockerContainer(s.dockerHosts, s.serviceName, 2)
+	c.Assert(status, Equals, true, Commentf("Container %s is not running on any hosts", s.serviceName))
+
+	containerName, err = verification.GetContainerName(host, s.serviceName+".1")
+	c.Assert(err, IsNil, Commentf("Failed to retrieve container name: %s", containerName))
+	out, err = dockercli.StopService(host, containerName)
```
shuklanirdesh82: nit: new line is required to maintain consistency.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
 	return op
 }
 
+// GetContainerName returns full container name based on given short name
+func GetContainerName(hostName, shortName string) (string, error) {
+	cmd := dockercli.ListContainers + "--filter name='" + shortName + "' --format '{{.Names}}'"
```
shuklanirdesh82: I think we should create filter constants for docker volume properties .. e.g. `--format '{{.Names}}` .. it will be easier to make change in one shot to sync with docker framework's change.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+			return status, host
+		}
+		misc.SleepForSec(5)
+	}
+	return false, ""
+}
+
+// allContainersRunning returns true if all replicated containers are up and running; false otherwise.
+func allContainersRunning(dockerHosts []string, serviceName string, replicas int) (bool, string) {
+	//TODO: It looks like all containers will always be spawned on one node. If this assumption
```
shuklanirdesh82: It is better to avoid `TODO` in first revision ... do we still need tis TODO?

tests/utils/verification/volumeproperties.go
```diff
@@ -110,6 +182,22 @@ func CheckVolumeAvailability(hostName string, volumeName string) bool {
...
 
+// GetFullVolumeName returns full volume name from the specified VM; return
+// original short name if any error happens
+func GetFullVolumeName(hostName string, volumeName string) string {
+	log.Printf("Fetching full name for volume [%s] from VM [%s]\n", volumeName, hostName)
+
+	cmd := dockercli.ListVolumes + "--filter name='" + volumeName + "' --format '{{.Name}}'"
+	fullName, err := ssh.InvokeCommand(hostName, cmd)
+	if err != nil {
+		log.Println(err)
```
shuklanirdesh82: this is redundant .. need to be removed.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+		if status {
+			return status, host
+		}
+		misc.SleepForSec(5)
+	}
+	return false, ""
+}
+
+// allContainersRunning returns true if all replicated containers are up and running; false otherwise.
+func allContainersRunning(dockerHosts []string, serviceName string, replicas int) (bool, string) {
```
shuklanirdesh82: `allContainersRunning ` => `areAllContainersRunning ` something more readable

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+	//TODO: It looks like all containers will always be spawned on one node. If this assumption
+	//turns out to be false, then the logic below needs to be changed
+	for _, host := range dockerHosts {
+		out, err := GetAllContainers(host, serviceName)
+		if err != nil {
+			log.Println(err)
```
shuklanirdesh82: this is redundant ... invokCommand routine takes care of printing error message.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+		if !strings.Contains(out, serviceName+"."+strconv.Itoa(i)) {
+			return false
+		}
+	}
+	return true
+}
+
+// VerifyDockerContainer returns true and the host IP if the containers are running on one of
+// the given docker hosts with specified replicas; otherwise returns false and empty string.
+func VerifyDockerContainer(dockerHosts []string, serviceName string, replicas int) (bool, string) {
```
shuklanirdesh82: can you please help me to understand the advantage of returning `host` value here?

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+	log.Printf("Checking running containers for docker service [%s] on docker hosts: %v\n", serviceName, dockerHosts)
+
+	//TODO: Need to implement generic polling logic for better reuse
+	for attempt := 0; attempt < maxAttempt; attempt++ {
```
shuklanirdesh82: I would suggest to move out this code block into another method and put in place at `misc`.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+
+// GetAllContainers returns all running containers on the give host based on the filtering criteria
+func GetAllContainers(hostName, filter string) (string, error) {
+	cmd := dockercli.ListContainers + "--filter name='" + filter + "' --format '{{.Names}}'"
+	return ssh.InvokeCommand(hostName, cmd)
+}
+
+// VerifyDockerService returns true if the given service is running on
+// the specified VM with specified replicas; false otherwise.
+func VerifyDockerService(hostName, serviceName string, replicas int) bool {
```
shuklanirdesh82: as this method is returning `bool` how about renaming to `IsDockerServiceVerified`?

tests/utils/verification/volumeproperties.go
shaominchen: Please refer to swarm_test line 114 - the returned host is needed for the next steps: query the  container name and stop the service *on this specific host*. My first version of this function does not return host actually. But the test failed in CI setup because the container might be spawned in any hosts. So when verifying the running containers, I need to return the host and then run subsequent steps on this specific host.

shuklanirdesh82: got it. Thanks @shaominchen for the explanation.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
 }
 
+// GetContainerName returns full container name based on given short name
+func GetContainerName(hostName, shortName string) (string, error) {
+	cmd := dockercli.ListContainers + "--filter name='" + shortName + "' --format '{{.Names}}'"
+	return ssh.InvokeCommand(hostName, cmd)
+}
+
+// GetAllContainers returns all running containers on the give host based on the filtering criteria
+func GetAllContainers(hostName, filter string) (string, error) {
```
shuklanirdesh82: I think `/volumeproperties.go` is not the better home to hold such utils. I would suggest to create another file or place into existing suitable place.

tests/utils/dockercli/swarm.go
```diff
@@ -0,0 +1,68 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This util provides various helper functions for managing docker swarm cluster.
+
+package dockercli
+
+import (
+	"log"
+
```
pshahzeb: nit: no newline needed

tests/utils/dockercli/swarm.go
```diff
@@ -0,0 +1,68 @@
...
+// CreateService creates a docker service
+func CreateService(ip, name, opts string) (string, error) {
+	log.Printf("Creating a docker service [%s] on VM [%s]\n", name, ip)
+	return ssh.InvokeCommand(ip, dockercli.CreateService+"--name "+name+" "+opts)
+}
+
+// ListService lists the tasks of a docker service
+func ListService(ip, name string) (string, error) {
+	log.Printf("Listing docker services running on VM [%s]\n", ip)
```
pshahzeb: minor: can log the service name as well.

tests/utils/verification/volumeproperties.go
```diff
@@ -97,6 +100,75 @@ func GetVolumePropertiesDockerCli(volName string, hostname string) string {
...
+func allContainersRunning(dockerHosts []string, serviceName string, replicas int) (bool, string) {
+	//TODO: It looks like all containers will always be spawned on one node. If this assumption
+	//turns out to be false, then the logic below needs to be changed
+	for _, host := range dockerHosts {
+		out, err := GetAllContainers(host, serviceName)
+		if err != nil {
+			log.Println(err)
+			continue
+		}
+		if out != "" {
```
pshahzeb: as per go suggestions, we can avoid nesting here. 
```
if out == "" {
    continue
}
....
for loop...
```

tests/e2e/swarm_test.go
```diff
@@ -0,0 +1,141 @@
...
+	volumeName  string
+	serviceName string
+}
+
+func (s *SwarmTestSuite) SetUpTest(c *C) {
+	s.esxName = inputparams.GetEsxIP()
+	s.master = inputparams.GetSwarmManager1()
+	s.worker1 = inputparams.GetSwarmWorker1()
+	s.worker2 = inputparams.GetSwarmWorker2()
+	s.dockerHosts = []string{s.master, s.worker1, s.worker2}
```
pshahzeb: if you think these vars need to be read just once for the entire test suite, you can have them in SetUpSuite instead of SetupTest.

shaominchen: Sure.

tests/e2e/swarm_test.go
shaominchen: Sure.

tests/e2e/swarm_test.go
shaominchen: The intention here is to keep the logic of one test step in one single block.

tests/utils/dockercli/swarm.go
shaominchen: Not sure why gofmt does not take care of this. Fixed.

tests/utils/dockercli/swarm.go
shaominchen: Sure.

tests/utils/verification/volumeproperties.go
shaominchen: I didn't define constant for "--format" because we are using quite some different formats: .Name, .Names, .Status, etc. As for "--filter", it needs a input parameter - defining a constant like "--filter name='" looks weird, as we need attach the matching quote ' separately.

tests/utils/verification/volumeproperties.go
shaominchen: Make sense. Will add a separate dockercontainer.go file.

tests/utils/verification/volumeproperties.go
shaominchen: Renamed it to IsDockerServiceRunning.

tests/utils/verification/volumeproperties.go
shaominchen: We should refactor this (and existing VerifyDetachedStatus) in a separate change.

tests/utils/verification/volumeproperties.go
shaominchen: Sure - renamed it to isContainerRunning.

tests/utils/verification/volumeproperties.go
shaominchen: OK, removed it for now.

tests/utils/verification/volumeproperties.go
shaominchen: Removed.

tests/utils/verification/volumeproperties.go
shaominchen: Fixed.

tests/utils/verification/volumeproperties.go
shaominchen: Removed.

tests/e2e/swarm_test.go
govint: Please see issue #1315, worth adding a test case that verifies vsphere volumes as "external" in the compose file and verify that the volume on a particular datastore is attached to the VM/container.

tests/utils/verification/volumeproperties.go
```diff
@@ -29,10 +29,6 @@ import (
 	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
 )
 
-const (
```
shuklanirdesh82: It is better to have constant to modify the attempt count. I would suggest to add this back; it will be easier to adjust in the future instead of skimming line of code.

tests/utils/verification/dockercontainer.go
```diff
@@ -0,0 +1,95 @@
...
+	"strconv"
+
+	"github.com/vmware/docker-volume-vsphere/tests/constants/dockercli"
+	util "github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
+)
+
+// GetContainerName returns full container name based on given short name
+func GetContainerName(hostName, shortName string) (string, error) {
```
shuklanirdesh82: `GetContainerName ` & `GetAllContainers` should move out from verifier package ... both of them spoils the context of `verifier` .. /util/dockercli would be a better home.

tests/utils/verification/volumeproperties.go
shaominchen: I removed this because there's a conflict on this constant name after splitting this file into two. Also we need to refactor this part anyway (I've already added TODO on both functions). We should have a generic poller function, which accepts a maximum time limit instead of the maximum attempts as the input.

tests/utils/verification/dockercontainer.go
shaominchen: Good idea! Fixed.

tests/e2e/swarm_test.go
shaominchen: I will add a GetSwarmNodes function in inputparams. I don't follow why we need a map here.

tests/e2e/swarm_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+var _ = Suite(&SwarmTestSuite{})
+
+// Test vDVS usage in swarm mode
+//
+// Test steps:
+// 1. Create a docker service with replicas setting to 1
+// 2. Verify the service is up and running with one node
+// 3. Verify one container is spawned
+// 5. Verify the volume is in attached status
```
lipingxue: The --replica option in swarm is not used to scale a stateful application. Instead, it should used to scale a stateless application. So why do we try to add the test with --replica 2 here? Any specific reason? @ashahi1

I think we should just verify the fail-over case here.

tests/e2e/swarm_test.go
shaominchen: First of all, if we need any test to cover new scenarios, that should be done in a separate PR. Moreover, I still don't follow why we need a e2e test for #1315. We can discuss that in the issue itself instead of in this PR.

tests/utils/dockercli/swarm.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
```
shaominchen: As we discussed offline, we should keep each PR small enough so that it’s easier to review and also keep the dependency to the minimum. The current PR is probably a special case as there's only one test case, but it's still pretty big and complicated. Since the review is almost done, I'd prefer to keep the current PR as is instead of wasting time to split it into two.

shuklanirdesh82:
@shaominchen If this PR is targeted for #1249 ... it is better to add ```fixes #1249``` to close the issue automatically with merge.

tests/utils/verification/dockercontainer.go
```diff
@@ -0,0 +1,95 @@
...
+		if status {
+			return status, host
+		}
+		misc.SleepForSec(5)
+	}
+	return false, ""
+}
+
+// isContainerRunning returns true if all replicated containers are up and running; false otherwise.
+func isContainerRunning(dockerHosts []string, serviceName string, replicas int) (bool, string) {
```
lipingxue: Nit：function name change to "IsContainerRunning", which is consistent with other func names

tests/e2e/swarm_test.go
shaominchen: @lipingxue Can we add this comment in the issue #601 itself? We can discuss your comment in the issue and once we have an agreement we can come back to update this test.

tests/utils/verification/dockercontainer.go
shaominchen: This is a convention in Go language: unexported identifiers should start with small letter; exported identifiers should start with capital letter.

tests/e2e/swarm_test.go
lipingxue: OK. I reopened issue #601 to add my comments.

tests/utils/verification/dockercontainer.go
lipingxue: ok

shuklanirdesh82:
@shaominchen  Please revisit the commit history and clean that up correctly.

vmdk_plugin/Makefile
```diff
@@ -287,7 +287,7 @@ CLEANESX_SH   := $(DEPLOY_TOOLS) cleanesx
 deploy-esx: clean-esx
 	$(DEPLOY_ESX_SH) "$(ESX)" "$(VIB_BIN)"
 
-VMS= $(VM1) $(VM2) $(WORKER2)
+VMS= $(VM1) $(VM2)
```
shuklanirdesh82: this is required, isn't it?

shaominchen: Yes, will fix.

PR #1310, opened 2017-05-30
Optimizing volumepropverification_test

ashahi1:
Addresses Issue #1096 

Also converts volumepropverification_test.go to use gocheck testing framework

Ran the test locally. Test output is as follows:
```
2017/06/06 21:35:21 START: volumeproperty_test.TestVolumeProperties
2017/06/06 21:35:21 Creating volume [thin_volume_1496784921] with options [ -o diskformat=thin] on VM [10.192.66.52]
2017/06/06 21:35:23 Creating volume [zeroedthick_volume_1496784923] with options [ -o diskformat=zeroedthick] on VM [10.192.66.52]
2017/06/06 21:35:25 Creating volume [eagerzeroedthick_volume_1496784925] with options [ -o diskformat=eagerzeroedthick] on VM [10.192.66.52]
2017/06/06 21:35:28 Getting size, disk-format and attached-to-vm for volume [thin_volume_1496784921] from ESX [10.192.87.211] using admin cli.
2017/06/06 21:35:30 Getting size, disk-format and attached-to-vm for volume [zeroedthick_volume_1496784923] from ESX [10.192.87.211] using admin cli.
2017/06/06 21:35:31 Getting size, disk-format and attached-to-vm for volume [eagerzeroedthick_volume_1496784925] from ESX [10.192.87.211] using admin cli.
2017/06/06 21:35:33 Getting size, disk-format and attached-to-vm for volume [thin_volume_1496784921] from vm [10.192.66.52] using docker cli
2017/06/06 21:35:33 Getting size, disk-format and attached-to-vm for volume [zeroedthick_volume_1496784923] from vm [10.192.66.52] using docker cli
2017/06/06 21:35:35 Getting size, disk-format and attached-to-vm for volume [eagerzeroedthick_volume_1496784925] from vm [10.192.66.52] using docker cli
2017/06/06 21:35:36 Verify docker and ESX properties of all three volumes are same
2017/06/06 21:35:36 Attaching volume [thin_volume_1496784921] on VM [10.192.66.52]
2017/06/06 21:35:38 Getting size, disk-format and attached-to-vm for volume [thin_volume_1496784921] from ESX [10.192.87.211] using admin cli.
2017/06/06 21:35:39 Getting size, disk-format and attached-to-vm for volume [thin_volume_1496784921] from vm [10.192.66.52] using docker cli
2017/06/06 21:35:40 Verify docker and ESX properties of all three volumes are same
2017/06/06 21:35:40 END: volumeproperty_test.TestVolumeProperties
2017/06/06 21:35:40 Removing all containers on VM [10.192.66.52]
2017/06/06 21:35:43  CLEAN-UP Finished: volumeproperty_test.TestVolumeProperties
```


tests/constants/dockercli/cmd.go
```diff
@@ -76,4 +76,13 @@ const (
 
 	// QueryContainer checks whether container exists or not
 	QueryContainer = docker + "ps -aq --filter name="
+
+	// StopAllContainers refers to stopping all the containers
+	StopAllContainers = docker + "kill $(docker ps -q)"
```
govint: Can this be  "docker ps -aq"?

tests/constants/dockercli/cmd.go
```diff
@@ -76,4 +76,13 @@ const (
...
+
+	// RemoveAllContainers refers to removing all the containers
+	RemoveAllContainers = docker + "rm $(docker ps -a -q)"
+
+	// DeleteAllDockerVolumes refers to removing all the docker volumes
+	DeleteAllDockerVolumes = ListVolumes + "-qf dangling=true | xargs -r docker volume rm"
```
govint: Thanks for adding this.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+	dockerHostIP   string
+	dockerHostName string
+	containerName  string
+	esxIP          string
+}
+
+func (s *VolumePropVerificationSuite) SetUpTest(c *C) {
+	s.dockerHostIP = inputparams.GetDockerHostIP("VM2")
+	s.dockerHostName = govc.RetrieveVMNameFromIP(s.dockerHostIP)
+	s.esxIP = os.Getenv("ESX")
```
govint: Pls. use inputparams.GetEsxIP()

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+	s.volumeName = inputparams.GetVolumeNameWithTimeStamp("volumepropverification_test")
+	s.containerName = inputparams.GetContainerNameWithTimeStamp("volumepropverificationSuite_test")
```
govint: These two can be as constants at the top of this file.

govint: Can check if any of the needed params is not valid and call c.Skip() so that the test is skipped vs. running and spewing errors or Assert()s.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
 }
 
 // method checks if the output from the ssh commands contains the expected values
-func hasElement(volmProps []string, op string) bool {
-	for i := 0; i < len(volmProps); i++ {
-		if !strings.Contains(op, volmProps[i]) {
-			log.Printf("Actual values : %s. Does not contains %s ", op, volmProps[i])
+func hasElement(volumeProps []string, op string) bool {
+	for i := 0; i < len(volumeProps); i++ {
+		if !strings.Contains(op, volumeProps[i]) {
```
govint: Use range to iterate over the array?

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
-				}
-				log.Println("Finished verifying volume properties like size, disk-format and attached-to-vm fields"+
-					" at vm and esx for volume - ", volName)
-				ssh.InvokeCommand(vms[vmIndx], "docker stop "+containerName+" ; docker rm "+containerName)
-			}
+
+func (s *VolumePropVerificationSuite) TestVolumeProperties(c *C) {
+	log.Printf("START: TestVolumeProperties")
+	formatTypes := []string{"thin", "zeroedthick", "eagerzeroedthick"}
+	for i := 0; i < len(formatTypes); i++ {
```
govint: Lines 81 - 90 could be a separate function to "CreateVolumesWithDiskFormat()" and pls. use "range" to iterate over the array.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+			"on ESX fetched using admin cli does not matches with the expected values", s.volumeName))
+
+		volumePropsDockerCli := verification.GetVolumePropertiesDockerCli(s.volumeName, s.dockerHostIP)
+		expectedPropsDockerCli := []string{"100MB", formatTypes[i], "<no value>"}
+		c.Assert(hasElement(expectedPropsDockerCli, volumePropsDockerCli), Equals, true, Commentf("Volume %s properties "+
+			"on docker vm fetched using docker cli does not matches with the expected values", s.volumeName))
+
+		if formatTypes[i] == "thin" {
+			out, err = dockercli.AttachVolume(s.dockerHostIP, s.volumeName, s.containerName)
+			c.Assert(err, IsNil, Commentf("Failed to attach the volume [%s]", s.volumeName))
```
govint: Please print value of "out" so test output shows the cause of failure.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+		s.volumeName = inputparams.GetVolumeNameWithTimeStamp(formatTypes[i])
+		s.containerName = inputparams.GetContainerNameWithTimeStamp("volumepropverification")
+
+		log.Printf("Verifying volume properties like size, disk-format and attached-to-vm fields "+
+			"at vm and esx for volume [%s] ", s.volumeName)
+		out, err := ssh.InvokeCommand(s.dockerHostIP, con.CreateVolume+"--name="+s.volumeName+" -o diskformat="+formatTypes[i])
+		c.Assert(err, IsNil, Commentf(out))
+
+		volumePropsAdminCli := verification.GetVolumePropertiesAdminCli(s.volumeName, s.esxIP)
+		expectedPropsAdminCli := []string{"100MB", formatTypes[i], "detached"}
```
govint: The verification functions themselves seem to be doing comparison and verification. Suggest defining new functions or update existing functions to do the verification needed here.

govint: Please make the expected values as constants with appropriate names so its easy to read and understand the code.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+		c.Assert(hasElement(expectedPropsAdminCli, volumePropsAdminCli), Equals, true, Commentf("Volume %s properties "+
+			"on ESX fetched using admin cli does not matches with the expected values", s.volumeName))
+
+		volumePropsDockerCli := verification.GetVolumePropertiesDockerCli(s.volumeName, s.dockerHostIP)
```
govint: This function is buggy already - please fix it in this PR - its logging a fatal log if inspect fails. We've already decided that fatal logs are disallowed. Instead return a string and error type from this function.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+		expectedPropsDockerCli := []string{"100MB", formatTypes[i], "<no value>"}
```
govint: Pls. use a constant with a meaningful name instead.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+
+		if formatTypes[i] == "thin" {
+			out, err = dockercli.AttachVolume(s.dockerHostIP, s.volumeName, s.containerName)
+			c.Assert(err, IsNil, Commentf("Failed to attach the volume [%s]", s.volumeName))
+
+			// Verifying attached to the vm field for volume
+			vmNameFromAdminCli := verification.GetVMAttachedToVolUsingAdminCli(s.volumeName, s.esxIP)
+			vmNameFromDockerCli := verification.GetVMAttachedToVolUsingDockerCli(s.volumeName, s.dockerHostIP)
+
+			// Get vm name based on ip and compare ith with the docker cli and admin cli
```
govint: comment -  compare ith with

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+			c.Assert(vmNameFromAdminCli, Equals, s.dockerHostName, Commentf("VM name fetched from attached-to-VM field "+
+				"for volume %s is not same as that fetched from its IP.", s.volumeName))
+
+			c.Assert(vmNameFromAdminCli, Equals, vmNameFromDockerCli, Commentf("Information mis-match - "+
+				"Attached-to-VM field for volume from docker cli is [%s] and attched-to-vm field from admin cli is [%s]", vmNameFromDockerCli, vmNameFromAdminCli))
+
+			// Again verifyying volume properties after starting a container
```
govint: Comment - verifyying

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+			volumePropsDockerCli = verification.GetVolumePropertiesDockerCli(s.volumeName, s.dockerHostIP)
+			expectedProprtyDockerCli := []string{"100MB", formatTypes[i]}
```
govint: Use appropriately named constant instead.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+			// Verifying attached to the vm field for volume
+			vmNameFromAdminCli := verification.GetVMAttachedToVolUsingAdminCli(s.volumeName, s.esxIP)
+			vmNameFromDockerCli := verification.GetVMAttachedToVolUsingDockerCli(s.volumeName, s.dockerHostIP)
+
+			// Get vm name based on ip and compare ith with the docker cli and admin cli
+			c.Assert(vmNameFromAdminCli, Equals, s.dockerHostName, Commentf("VM name fetched from attached-to-VM field "+
+				"for volume %s is not same as that fetched from its IP.", s.volumeName))
+
+			c.Assert(vmNameFromAdminCli, Equals, vmNameFromDockerCli, Commentf("Information mis-match - "+
+				"Attached-to-VM field for volume from docker cli is [%s] and attched-to-vm field from admin cli is [%s]", vmNameFromDockerCli, vmNameFromAdminCli))
```
govint: Why shouldn't these two checks on the VM name be done for all disk format types - why is it done only for "thin" disks?

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+		expectedPropsAdminCli := []string{"100MB", formatTypes[i], "detached"}
+		c.Assert(hasElement(expectedPropsAdminCli, volumePropsAdminCli), Equals, true, Commentf("Volume %s properties "+
+			"on ESX fetched using admin cli does not matches with the expected values", s.volumeName))
+
+		volumePropsDockerCli := verification.GetVolumePropertiesDockerCli(s.volumeName, s.dockerHostIP)
+		expectedPropsDockerCli := []string{"100MB", formatTypes[i], "<no value>"}
+		c.Assert(hasElement(expectedPropsDockerCli, volumePropsDockerCli), Equals, true, Commentf("Volume %s properties "+
+			"on docker vm fetched using docker cli does not matches with the expected values", s.volumeName))
+
+		if formatTypes[i] == "thin" {
```
govint: 1. Move this handling for thin disks into a separate function.

2. Add a new function that does "mandatory checks on volume properties"

3. Get the admin CLI and docker CLI volume properties and pass to function in (2) to make the checks.

4. If volume is thin then call another function to verify volume properties specific for thin disks.

Make this function smaller by breaking it up into separate functions

1. One function to create the volumes,
2. Then loop over the volumes and get the admin and docker CLI properties for each volume.
3. Call the property validation function for each volume.

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+				"Attached-to-VM field for volume from docker cli is [%s] and attched-to-vm field from admin cli is [%s]", vmNameFromDockerCli, vmNameFromAdminCli))
+
+			// Again verifyying volume properties after starting a container
+			volumePropsDockerCli = verification.GetVolumePropertiesDockerCli(s.volumeName, s.dockerHostIP)
+			expectedProprtyDockerCli := []string{"100MB", formatTypes[i]}
+			c.Assert(hasElement(expectedProprtyDockerCli, volumePropsDockerCli), Equals, true, Commentf("Volume %s properties "+
+				"on docker vm fetched using docker cli does not matches with the expected values", s.volumeName))
+
+			volumePropsAdminCli = verification.GetVolumePropertiesAdminCli(s.volumeName, s.esxIP)
+			expectedPropsAdminCli = []string{"100MB", formatTypes[i]}
```
govint: Why do we keep getting docker CLI and admin CLI volume properties each time. Get it once for each volume and do the property validations with it.

tests/e2e/volumepropverification_test.go
ashahi1: Test was taking long time to finish and intermittently test was failing on CI because test ran too long. So it was decided to perform certain steps only for "thin" disk format to reduce the overall test duration. Basically wanted to finish CI runs faster.
 
Please refer to Issue #1096

tests/constants/dockercli/cmd.go
shuklanirdesh82: +1

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
 	"log"
 	"os"
 	"strings"
-	"testing"
-	"time"
 
+	con "github.com/vmware/docker-volume-vsphere/tests/constants/dockercli"
```
shuklanirdesh82: minor: con?? ... how about dockerconst

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
-	teardownFunction()
-	os.Exit(retCode)
 }
 
-// clean-up function
-func teardownFunction() {
-	ssh.InvokeCommand(vmIP, dockerVolmRmvCmd)
-	log.Println("-----Clean-up finished - current time: ", time.Now())
+func (s *VolumePropVerificationSuite) TearDownTest(c *C) {
+	out, err := dockercli.StopAllContainers(s.dockerHostIP)
```
shuklanirdesh82: it is better to avoid this step and invoke forceful removal of containers .. unless it is required to have this extra step.

tests/constants/dockercli/cmd.go
```diff
@@ -76,4 +76,13 @@ const (
 
 	// QueryContainer checks whether container exists or not
 	QueryContainer = docker + "ps -aq --filter name="
+
+	// StopAllContainers refers to stopping all the containers
+	StopAllContainers = docker + "kill $(docker ps -q)"
+
+	// RemoveAllContainers refers to removing all the containers
+	RemoveAllContainers = docker + "rm $(docker ps -a -q)"
```
shuklanirdesh82: `docker ps -a -q` => `docker ps -aq` ??

tests/utils/dockercli/volumelifecycle.go
```diff
@@ -155,3 +155,21 @@ func IsContainerExist(ip, containerName string) bool {
...
+}
+
+// RemoveAllContainers - removes all the containers on a particular vm
+func RemoveAllContainers(ip string) (string, error) {
+	log.Printf("Removing all containers on VM [%s]\n", ip)
+	return ssh.InvokeCommand(ip, dockercli.RemoveAllContainers)
+}
+
+// DeleteAllDockerVolumes - removes all the containers on a particular vm
+func DeleteAllDockerVolumes(ip string) (string, error) {
```
shuklanirdesh82: it looks cool but dangerous .. it may delete unwanted volumes as well. one serious problem I can think of running test in parallel. 

Ideal approach should be deleting volumes those are created as part of current running test instance.

tests/constants/dockercli/cmd.go
```diff
@@ -76,4 +76,13 @@ const (
 
 	// QueryContainer checks whether container exists or not
 	QueryContainer = docker + "ps -aq --filter name="
+
+	// StopAllContainers refers to stopping all the containers
```
shaominchen: Nit: Just for consistency, let's remove "refers to" to keep the comments simple.

tests/utils/dockercli/volumelifecycle.go
```diff
@@ -155,3 +155,21 @@ func IsContainerExist(ip, containerName string) bool {
...
+	return ssh.InvokeCommand(ip, dockercli.StopAllContainers)
+}
+
+// RemoveAllContainers - removes all the containers on a particular vm
+func RemoveAllContainers(ip string) (string, error) {
+	log.Printf("Removing all containers on VM [%s]\n", ip)
+	return ssh.InvokeCommand(ip, dockercli.RemoveAllContainers)
+}
+
+// DeleteAllDockerVolumes - removes all the containers on a particular vm
```
shaominchen: The comments need to be updated.

tests/utils/dockercli/volumelifecycle.go
shaominchen: +1. This is useful only if we are sure we are doing the right thing (i.e. no side effects).

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
+	dockerHostName string
+	containerName  string
+	esxIP          string
+}
+
+func (s *VolumePropVerificationSuite) SetUpTest(c *C) {
+	s.dockerHostIP = inputparams.GetDockerHostIP("VM2")
+	s.dockerHostName = govc.RetrieveVMNameFromIP(s.dockerHostIP)
+	s.esxIP = os.Getenv("ESX")
+	s.volumeName = inputparams.GetVolumeNameWithTimeStamp("volumepropverification_test")
```
shaominchen: This name is just too long - can we rename it (and the file name) to volumeproperty_test?

tests/e2e/volumepropverification_test.go
shaominchen: +1

tests/e2e/volumepropverification_test.go
```diff
@@ -24,118 +24,122 @@ import (
...
-				if !hasElement(expctdPropsAdmin, volmPropertiesAdminCli) {
-					log.Fatal("Volume properties on admin cli do not matches with the expected values")
-				}
-				log.Println("Finished verifying volume properties like size, disk-format and attached-to-vm fields"+
-					" at vm and esx for volume - ", volName)
-				ssh.InvokeCommand(vms[vmIndx], "docker stop "+containerName+" ; docker rm "+containerName)
-			}
+
+func (s *VolumePropVerificationSuite) TestVolumeProperties(c *C) {
+	log.Printf("START: TestVolumeProperties")
```
shaominchen: Nit: Just for consistency, let's include the test group name in the log as well.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/govc"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/inputparams"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/verification"
+	. "gopkg.in/check.v1"
+)
+
+const (
+	TestName           = "volumeproperty"
```
shuklanirdesh82: all this should start with lower case unless it is meant for exporting.

tests/utils/verification/volumeproperties.go
```diff
@@ -61,40 +61,26 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
-	op, _ := ssh.InvokeCommand(hostname, cmd)
-	if op == "" {
-		log.Fatal("Null value is returned by admin cli when looking for, size, disk-format and attached to vm. Output: ", op)
-	}
-	if len(strings.Fields(op)) != 4 {
-		log.Fatalf("Output is expected to consist of four elements only - "+
-			"volume name, attached-to-vm status, size and disk-format. Actual output: %s", op)
+	out, err := ssh.InvokeCommand(hostname, cmd)
+	if err != nil {
+		log.Printf("Null value is returned by admin cli when looking for, size, disk-format and attached-to-vm for volume [%s].", volName)
```
shuklanirdesh82: this is not needed as `InvokeCommand ` takes care of this one.

tests/utils/verification/volumeproperties.go
```diff
@@ -61,40 +61,26 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
-		expctedLen = 3
-	}
-	if len(strings.Fields(op)) != expctedLen {
-		log.Fatalf("Docker cli inpect output is expected to consist of three elements only - "+
-			"size, disk-format and attached-to-vm status. Actual output: %s", op)
+func GetVolumePropertiesDockerCli(volName string, hostname string) (string, error) {
+	cmd := dockercli.InspectVolume + " --format ' {{index .Name}} {{index .Status.capacity.size}} {{index .Status.diskformat}} {{index .Status \"attached to VM\"}}' " + volName
+	out, err := ssh.InvokeCommand(hostname, cmd)
+	if err != nil {
+		log.Printf("Null value is returned by docker cli when looking for, size, disk-format and attached-to-vm for volume [%s].", volName)
```
shuklanirdesh82: same as above

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	log.Printf("END: TestVolumeProperties")
+}
+
+// createVolumes - creates volumes of each format type
+func (s *VolumePropVerificationSuite) createVolumes(c *C) {
+	for i, formatType := range s.formatTypes {
+		s.containerName = inputparams.GetContainerNameWithTimeStamp(TestName)
+		s.volumeNames[i] = inputparams.GetVolumeNameWithTimeStamp(TestName)
+		out, err := ssh.InvokeCommand(s.dockerHostIP, dockerconst.CreateVolume+"--name="+s.volumeNames[i]+" -o diskformat="+formatType)
```
shuklanirdesh82: can't we use https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/dockercli/volumelifecycle.go#L37?

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	s.vmAttached = true
+	s.getPropertiesAndVerify(c, s.volumeNames[0])
+
+	log.Printf("END: TestVolumeProperties")
+}
+
+// createVolumes - creates volumes of each format type
+func (s *VolumePropVerificationSuite) createVolumes(c *C) {
+	for i, formatType := range s.formatTypes {
+		s.containerName = inputparams.GetContainerNameWithTimeStamp(TestName)
```
shuklanirdesh82: isn't it keep setting and overriding value and set with the last loop?

ashahi1: I missed it. Will remove it.

tests/e2e/volumeproperty_test.go
ashahi1: I knew that it was in review but wasn't sure if it got checked-in. Will remove it.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+		s.volumeNames[i] = inputparams.GetVolumeNameWithTimeStamp(TestName)
+		out, err := ssh.InvokeCommand(s.dockerHostIP, dockerconst.CreateVolume+"--name="+s.volumeNames[i]+" -o diskformat="+formatType)
+		c.Assert(err, IsNil, Commentf(out))
+	}
+}
+
+// getPropertiesAndVerify - gets properties of a volume from esx and docker host
+func (s *VolumePropVerificationSuite) getPropertiesAndVerify(c *C, volumeName string) {
+	valuesAdminCli, err := verification.GetVolumePropertiesAdminCli(volumeName, s.esxIP)
```
shuklanirdesh82: It is better to invoke admin_cli volume ls once > save the response and perform validation this reduces unnecessary ssh calls for each volumes following.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	log.Printf("Verifying volume properties like size, disk-format and attached-to-vm fields "+
+		"at vm [%s] and esx [%s] for volume - [%s] ", s.dockerHostIP, s.esxIP, volumeName)
+	arrAdminCliValues := strings.Fields(valuesAdmincli)
+	sliceAdminCliValues := arrAdminCliValues[1:len(arrAdminCliValues)]
+	// status := verification.VerifyAttachedStatus(volumeName, s.dockerHostIP, s.esxIP)
+
+	sliceDockerCliValues := s.arrangeDockerCliValues(valuesDockerCli)
+	for i := range sliceDockerCliValues {
+		c.Assert(len(sliceDockerCliValues), Equals, len(sliceAdminCliValues), Commentf("Values mis-match between admin cli and docker cli for volume %s", volumeName))
+		if i == 2 && !s.vmAttached {
```
shuklanirdesh82: what is so special about `i==2`? you may want to put some inline comment to make reviewer's life easier.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *VolumePropVerificationSuite) TestVolumeProperties(c *C) {
+	log.Printf("START: TestVolumeProperties")
+
+	// creating volumes of all disk formats
+	s.createVolumes(c)
+	for _, volumeName := range s.volumeNames {
+		// getting and verifying the admin cli and docker cli values
+		s.getPropertiesAndVerify(c, volumeName)
+		s.dockerVolumeRemoveCmd = s.dockerVolumeRemoveCmd + " " + volumeName
```
govint: This is a global and why is it getting initialized here? if getPropertiesAndVerify() wants to use this global let it use it. Seems odd why this global is set here and not used.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	s.getPropertiesAndVerify(c, s.volumeNames[0])
+
+	log.Printf("END: TestVolumeProperties")
+}
+
+// createVolumes - creates volumes of each format type
+func (s *VolumePropVerificationSuite) createVolumes(c *C) {
+	for i, formatType := range s.formatTypes {
+		s.containerName = inputparams.GetContainerNameWithTimeStamp(TestName)
+		s.volumeNames[i] = inputparams.GetVolumeNameWithTimeStamp(TestName)
```
govint: Watchout, the GetVolumeNameWithTimeStamp() is using time.New().Unix() which returns the number of seconds since some time. Successive calls can return the same name. I'd seen this and use names with a random generated number attached to it. For clarity sake you could have different names - thick, zeroedthick, thin  and attach the timestamp.

tests/e2e/volumeproperty_test.go
govint: Yes, please use dockercli.CreateVolume() API.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+		c.Assert(err, IsNil, Commentf(out))
+	}
+}
+
+// getPropertiesAndVerify - gets properties of a volume from esx and docker host
+func (s *VolumePropVerificationSuite) getPropertiesAndVerify(c *C, volumeName string) {
+	valuesAdminCli, err := verification.GetVolumePropertiesAdminCli(volumeName, s.esxIP)
+	c.Assert(err, IsNil, Commentf(valuesAdminCli))
+
+	valuesDockerCli, err := verification.GetVolumePropertiesDockerCli(volumeName, s.dockerHostIP)
```
govint: Only a suggestion and you can decide how difficult or easy it may be to do this. The verifyProperties() function is may be a bit complex. Is it possible instead for the GetVolumeProperties*() functions to return a map of values so that the verification is easy to do. That would be ideal as it makes verifyProperties() pretty easy.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	for i := range sliceDockerCliValues {
+		c.Assert(len(sliceDockerCliValues), Equals, len(sliceAdminCliValues), Commentf("Values mis-match between admin cli and docker cli for volume %s", volumeName))
+		if i == 2 && !s.vmAttached {
+			c.Assert(sliceDockerCliValues[i], Equals, DiskStatusEmpty, Commentf("Value from dockercli is [%s]", sliceDockerCliValues[i]))
+			c.Assert(sliceAdminCliValues[i], Equals, DiskStatusDetached, Commentf("Value from dockercli is [%s]", i, sliceAdminCliValues[i]))
+		} else {
+			c.Assert(sliceDockerCliValues[i], Equals, sliceAdminCliValues[i], Commentf("Value from admincli is [%s] and that from dockercli is [%s]", sliceAdminCliValues[i], sliceDockerCliValues[i]))
+		}
+	}
+	// Get vm name based on ip and compare ith with the docker cli/admin cli only when vm is attached
```
govint: Nit comment - "ith with"

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if s.vmAttached {
+		c.Assert(sliceDockerCliValues[len(sliceDockerCliValues)-1], Equals, s.dockerHostName, Commentf("VM name fetched from attached-to-VM field "+
+			"for volume %s is not same as that fetched from its IP.", volumeName))
```
govint: This code can be very delicate and can easily fail with these assumed locations of volume properties. Its also the wrong way to compare attributes of a volume based on position of values at specific indices. Lets make the GetVolumeProperties*() functions to return maps vs strings. May be more work but its worth the effort to make the test code here accurate without complexity.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+// Wrote this method because if volume is not attached, attached-to-vm field
+// return as <no value> for docker cli whereas admin cli returns "detached". So length of array increases
+// for docker cli values and this makes comparison difficult.
+// 'status' indicates if volume is attached/detached
+func (s *VolumePropVerificationSuite) arrangeDockerCliValues(valuesDockerCli string) []string {
+	arrDockerCliValues := strings.Fields(valuesDockerCli)
+	sliceDockerCliValues := arrDockerCliValues[1:len(arrDockerCliValues)]
+	if !s.vmAttached {
+		sliceDockerCliValues[len(sliceDockerCliValues)-2] = sliceDockerCliValues[len(sliceDockerCliValues)-2] + sliceDockerCliValues[len(sliceDockerCliValues)-1]
+		sliceDockerCliValues = sliceDockerCliValues[0 : len(sliceDockerCliValues)-1]
```
govint: :-))))), but no Sir, this isn't right at all. Please make those util function return maps and we all can use those to benefit of more tests.

tests/e2e/volumeproperty_test.go
shuklanirdesh82: +1

shaominchen: I've filed issue #1344 for this.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+type VolumePropVerificationSuite struct {
+	volumeNames           [3]string
+	dockerHostIP          string
+	dockerHostName        string
+	containerName         string
+	esxIP                 string
+	formatTypes           []string
+	dockerVolumeRemoveCmd string
```
ashahi1: I am creating a single command to remove all the volumes so that we don't need to ssh for each volume.
This is how final command looks like: docker volume rm vol1 vol2 vol3

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,182 @@
...
+	dockerCliValues := strings.Fields(out)
+	dockerCliMap[volumeName] = [3]string{dockerCliValues[0], dockerCliValues[1], dockerCliValues[2]}
+}
+
+// verifyProperties - verifies admin cli values for a volume are same as the expected values
+// Only for 'thin', after volume is attached, modifying attached-to-vm field with vm name.
+func (s *VolumePropertyTestSuite) verifyProperties(c *C) {
+	for i, volumeName := range s.volumeNames {
+		if s.volumeAttached {
+			tmp := expectedValues[s.volumeNames[0]]
```
govint: If this is the first call; to verifyProperties then expectedValues[] won't have the key (volume names) and will cause a panic of the GO run time.need to do like,
tmp, ok := expectedValues[volumeName]; ok == true {
 _ assign the volume name_
}

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	dockerCliValues := strings.Fields(out)
+	dockerCliMap[volumeName] = [3]string{dockerCliValues[0], dockerCliValues[1], dockerCliValues[2]}
+}
+
+// verifyProperties - verifies admin cli values for a volume are same as the expected values
+// Only for 'thin', after volume is attached, modifying attached-to-vm field with vm name.
+func (s *VolumePropertyTestSuite) verifyProperties(c *C) {
+	for i, volumeName := range s.volumeNames {
+		if s.volumeAttached {
```
govint: This is common for all volumes and hence when thin disk is attached, we will always go thru this path and mark all with the docker host name like as though they are attached, which isn't the case.

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,27 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
+// GetVolumePropertiesAdminCli returns a map of volumes and their corresponding
+// properties - capacity, attached-to-vm and disk-format field using Admin cli
+func GetVolumePropertiesAdminCli(hostName string) map[string][3]string {
+	log.Printf("Getting size, disk-format and attached-to-vm for all volumes from ESX [%s] using admin cli. \n", hostName)
+	cmd := admincli.ListVolumes + "-c volume,capacity,disk-format,attached-to 2>/dev/null | awk -v OFS='\t' '{print $1, $2, $3, $4}' | sed '1,2d'  "
+	out, _ := ssh.InvokeCommand(hostName, cmd)
+	admincliValues := strings.Fields(out)
+	adminCliMap := make(map[string][3]string)
+	for i := 0; i < len(admincliValues); {
+		adminCliMap[admincliValues[i]] = [3]string{admincliValues[i+1], admincliValues[i+2], admincliValues[i+3]}
```
govint: [3] isn't needed []string{......} is ok

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,27 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
+		i = i + 4
 	}
-	return op
+	return adminCliMap
```
govint: Its a global, does it need to be returned?

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+}
+
+// TearDownTest - Removes all the containers and deletes the volumes
+func (s *VolumePropertyTestSuite) TearDownTest(c *C) {
+	out, err := dockercli.RemoveAllContainers(s.dockerHostIP)
+	c.Assert(err, IsNil, Commentf(out))
+	out, err = ssh.InvokeCommand(s.dockerHostIP, s.dockerVolumeRemoveCmd)
+	c.Assert(err, IsNil, Commentf(out))
+}
```
ashahi1: OK, I will add log statements at end of setup and teardown and that automatically prints current time.
e.g.
```
2017/06/05 17:01:15 SETUP Finished: volumeproperty_test.TestVolumeProperties
2017/06/05 17:01:40 CLEAN-UP Finished: volumeproperty_test.TestVolumeProperties
```

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,182 @@
...
+// createVolumes - creates volumes of each format type
+func (s *VolumePropertyTestSuite) createVolumes(c *C) {
+	for i, formatType := range s.formatTypes {
+		s.volumeNames[i] = inputparams.GetVolumeNameWithTimeStamp(formatType)
+		out, err := dockercli.CreateVolumeWithOptions(s.dockerHostIP, s.volumeNames[i], diskFormatOption+formatType)
+		c.Assert(err, IsNil, Commentf(out))
+
+		// creating a single cmd to remove all the created volumes.
+		// Cmd will be used in the test clean-up. e.g. docker volume rm vol1 vol2 vol3
+		s.dockerVolumeRemoveCmd = s.dockerVolumeRemoveCmd + " " + s.volumeNames[i]
```
ashahi1: +1

tests/e2e/volumeproperty_test.go
ashahi1: s.volumeAttached is 'true' only if volume is attached. So this piece of code is only executed if volume is attached else map of expected values is populated. I had tested it locally and it ran fine. So things work fine even for first call.

tests/e2e/volumeproperty_test.go
ashahi1: No, this is not what's happening.  Please see the below output (map of expected values) when thin volume is attached.

```
2017/06/05 20:53:26 Expected values: map[thin_volume_1496695990:[100MB thin ubuntu-VM2.3] zeroedthick_volume_1496695993:[100MB zeroedthick detached] eagerzeroedthick_volume_1496695995:[100MB eagerzeroedthick detached]]
```

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,27 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
-		log.Fatalf("Output is expected to consist of four elements only - "+
-			"volume name, attached-to-vm status, size and disk-format. Actual output: %s", op)
+// GetVolumePropertiesAdminCli returns a map of volumes and their corresponding
+// properties - capacity, attached-to-vm and disk-format field using Admin cli
+func GetVolumePropertiesAdminCli(hostName string) map[string][3]string {
+	log.Printf("Getting size, disk-format and attached-to-vm for all volumes from ESX [%s] using admin cli. \n", hostName)
+	cmd := admincli.ListVolumes + "-c volume,capacity,disk-format,attached-to 2>/dev/null | awk -v OFS='\t' '{print $1, $2, $3, $4}' | sed '1,2d'  "
+	out, _ := ssh.InvokeCommand(hostName, cmd)
+	admincliValues := strings.Fields(out)
+	adminCliMap := make(map[string][3]string)
```
ashahi1: @govint Sorry, I didn't get you. Can you please explain what did you mean in this comment? 
Its global in a test file ```tests/e2e/volumeproperty_test.go``` and not under ```tests/utils/verification/volumeproperties.go```

tests/utils/verification/volumeproperties.go
ashahi1: Its global in a test file tests/e2e/volumeproperty_test.go and not under tests/utils/verification/volumeproperties.go

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,37 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
-	op, _ := ssh.InvokeCommand(hostname, cmd)
-	if op == "" {
-		log.Fatal("Null value is returned by admin cli when looking for, size, disk-format and attached to vm. Output: ", op)
-	}
-	if len(strings.Fields(op)) != 4 {
-		log.Fatalf("Output is expected to consist of four elements only - "+
-			"volume name, attached-to-vm status, size and disk-format. Actual output: %s", op)
+// GetPropertiesForAllVolumesAdminCli returns a map of all the volumes from ESX and their corresponding
+// properties returned - capacity, attached-to-vm and disk-format
+func GetPropertiesForAllVolumesAdminCli(hostName string) map[string][]string {
```
shaominchen: I strongly suggest to add a common util function to return a map of all volume properties in which the key is the property name and the value is the property value. That will be very useful to avoid such complicated function which is only useful for a specific test.

Can we file a issue to track this improvement?

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (s *VolumePropertyTestSuite) SetUpTest(c *C) {
+	s.config = inputparams.GetTestConfig()
+	if s.config == nil {
+		c.Skip("Unable to retrieve test config, skipping volumeproperty_test.TestVolumeProperties.")
+	}
+	s.containerName = inputparams.GetContainerNameWithTimeStamp(testName)
+	s.formatTypes = []string{"thin", "zeroedthick", "eagerzeroedthick"}
+	s.dockerVolumeRemoveCmd = dockerconst.RemoveVolume
+	log.Println(" SETUP Finished: volumeproperty_test.TestVolumeProperties")
```
shaominchen: Nit: remove the space before "SETUP"

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+	out, err := dockercli.RemoveAllContainers(s.config.DockerHosts[1])
+	c.Assert(err, IsNil, Commentf(out))
+	s.dockerVolumeRemoveCmd = s.dockerVolumeRemoveCmd + strings.Join(s.volumeNames, " ")
+	out, err = ssh.InvokeCommand(s.config.DockerHosts[1], s.dockerVolumeRemoveCmd)
+	c.Assert(err, IsNil, Commentf(out))
+	log.Println(" CLEAN-UP Finished: volumeproperty_test.TestVolumeProperties")
+}
+
+var _ = Suite(&VolumePropertyTestSuite{})
+var dockerCliMap = make(map[string][]string)
```
shaominchen: Can we add some comments for these 3 maps?

tests/utils/verification/volumeproperties.go
ashahi1: Filed issue #1365

shuklanirdesh82: The method name is too big .. it should be simplified .. how about his `GetAllVolProperties` and the method doc mentions how are we retrieving.

ashahi1: Ok. I will change it.

tests/e2e/volumeproperty_test.go
ashahi1: Ok. I will add it.

tests/e2e/volumeproperty_test.go
ashahi1: Ok. I will remove it.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: I would prefer to have Map(volumeName,VolPropStruct) where volume name is the key and Struct object is for various properties.

```
struct VolPropStruct {
     capacity
     attached-to-vm
     disk-format
}
```

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/inputparams"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/verification"
+	. "gopkg.in/check.v1"
+)
+
+const (
+	testName           = "volumeproperty"
+	diskStatusDetached = "detached"
```
shuklanirdesh82: better to use https://github.com/vmware/docker-volume-vsphere/blob/master/tests/constants/properties/properties.go#L21

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+	size               = "100MB"
+	diskFormatOption   = " -o diskformat="
+)
+
+type VolumePropertyTestSuite struct {
+	volumeNames           []string
+	containerName         string
+	formatTypes           []string
+	dockerVolumeRemoveCmd string
```
shuklanirdesh82: this should be removed.

tests/e2e/volumeproperty_test.go
shuklanirdesh82: This should be replaced with something similar to https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/misc/misc.go#L53 .. you may introduce new log util.

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+4. Verifies values obtained from admin cli and docker cli.
+5. SSH to the vm and run a container and mount the volume created.
+6. SSH to the vm and esx and verify the attached-to-vm field for volume - both docker cli and admin cli values should be same.
+7. Again verify capacity and disk format using docker cli and admin cli to make sure things are fine after running the container.
+
+NOTE: Do steps 5 ,6 and 7 only for volume of 'thin' disk format
+*/
+
+func (s *VolumePropertyTestSuite) TestVolumeProperties(c *C) {
+	log.Printf("START: volumeproperty_test.TestVolumeProperties")
```
shuklanirdesh82: should be replaced with https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/misc/misc.go#L53

tests/e2e/volumeproperty_test.go
ashahi1: why? and how?

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+	s.volumeAttached = true
+
+	// Get volumes properties for the thin volume from docker and ESX.
+	s.getVolumePropertiesAdminCli(c, s.volumeNames[0])
+	s.getVolumePropertiesDockerCli(c, s.volumeNames[0])
+
+	// Verify if docker and ESX properties of volumes are same and as expected.
+	s.verifyProperties(c)
+
+	log.Printf("END: volumeproperty_test.TestVolumeProperties")
```
shuklanirdesh82: let's use appropriate util from https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/misc/misc.go#L53

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+	for _, formatType := range s.formatTypes {
+		vname := inputparams.GetVolumeNameWithTimeStamp(formatType)
+		out, err := dockercli.CreateVolumeWithOptions(s.config.DockerHosts[1], vname, diskFormatOption+formatType)
+		s.volumeNames = append(s.volumeNames, vname)
+		c.Assert(err, IsNil, Commentf(out))
+	}
+}
+
+// getVolumeProperties - get properties of three volumes from ESX and docker host
+func (s *VolumePropertyTestSuite) getVolumeProperties(c *C) {
```
shuklanirdesh82: as this method is not returning anything .. we should avoid using `getXXX`; it spoils objective of `get` (leads to confusion)

tests/e2e/volumeproperty_test.go
```diff
@@ -0,0 +1,176 @@
...
+		s.getVolumePropertiesAdminCli(c, volumeName)
+	}
+	// get properties for all three volumes from docker host
+	for _, volumeName := range s.volumeNames {
+		s.getVolumePropertiesDockerCli(c, volumeName)
+	}
+}
+
+// getVolumePropertiesAdminCli - get properties of a volume from ESX
+func (s *VolumePropertyTestSuite) getVolumePropertiesAdminCli(c *C, volumeName string) {
```
shuklanirdesh82: it is better to return the map from this and let caller to handle and process instead of setting global variable .. it spoils the code continuation and hard to track who sets it.

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,37 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
+	return adminCliMap
+}
+
+// GetVolumePropertiesAdminCli returns an array of properties of a volume from ESX
+// properties returned - capacity, attached-to-vm and disk-format field
+func GetVolumePropertiesAdminCli(volumeName, hostName string) []string {
+	log.Printf("Getting size, disk-format and attached-to-vm for volume [%s] from ESX [%s] using admin cli. \n", volumeName, hostName)
+	cmd := admincli.ListVolumes + "-c volume,capacity,disk-format,attached-to 2>/dev/null | grep " + volumeName + " | awk -v OFS='\t' '{print $2, $3, $4}' "
+	out, _ := ssh.InvokeCommand(hostName, cmd)
+	admincliValues := strings.Fields(out)
```
shuklanirdesh82: it is better to return from here itself

tests/utils/verification/volumeproperties.go
```diff
@@ -55,42 +55,37 @@ func GetVMAttachedToVolUsingAdminCli(volName string, hostname string) string {
...
-			"volume name, attached-to-vm status, size and disk-format. Actual output: %s", op)
+// GetPropertiesForAllVolumesAdminCli returns a map of all the volumes from ESX and their corresponding
+// properties returned - capacity, attached-to-vm and disk-format
+func GetPropertiesForAllVolumesAdminCli(hostName string) map[string][]string {
+	log.Printf("Getting size, disk-format and attached-to-vm for all volumes from ESX [%s] using admin cli. \n", hostName)
+	cmd := admincli.ListVolumes + "-c volume,capacity,disk-format,attached-to 2>/dev/null | awk -v OFS='\t' '{print $1, $2, $3, $4}' | sed '1,2d'  "
+	out, _ := ssh.InvokeCommand(hostName, cmd)
+	admincliValues := strings.Fields(out)
+	adminCliMap := make(map[string][]string)
+	for i := 0; i < len(admincliValues); {
```
shuklanirdesh82: any reason for not using struct object?

tests/utils/verification/volumeproperties.go
govint: Possibly also remove the AdminCLI and DockerCLI from the function names

tests/e2e/volumeproperty_test.go
ashahi1: Addressed it.

PR #1299, opened 2017-05-25
[Automation] Test cases for volume create, access and vsan 

pshahzeb:
1. Test cases for volume create with valid/invalid volume names
2. Test cases for volume create with valid/invalid volume options
3. Test cases for volume access updates and consistently test read/write operations
4. Test cases for vsan volume

Fixes #1254 

Testing: done locally
test-all target passes
e2-test output
```
=== RUN   Test
2017/05/25 22:48:10 START: basic-test.TestVolumeIsolation
2017/05/25 22:48:10 Creating volume [basic_test_volume_1495752490] on VM [10.192.51.202]
2017/05/25 22:48:13 Checking volume [basic_test_volume_1495752490] availability from VM [10.192.51.202]
2017/05/25 22:48:13 Checking volume [basic_test_volume_1495752490] availability from VM [10.192.51.202]
2017/05/25 22:48:14 Destroying volume [basic_test_volume_1495752490]
2017/05/25 22:48:15 END: basic-test.TestVolumeIsolation
2017/05/25 22:48:15 START: basic-test.TestVolumeLifecycle
2017/05/25 22:48:15 Creating volume [basic_test_volume_1495752495] on VM [10.192.51.202]
2017/05/25 22:48:17 Checking volume [basic_test_volume_1495752495] availability from VM [10.192.51.202]
2017/05/25 22:48:17 Attaching volume [basic_test_volume_1495752495] on VM [10.192.51.202]
2017/05/25 22:48:19 Confirming attached status for volume [basic_test_volume_1495752495]
2017/05/25 22:48:21 Removing container [basic_test_container_1495752495] on VM [10.192.51.202]
2017/05/25 22:48:22 Confirming detached status for volume [basic_test_volume_1495752495]
2017/05/25 22:48:25 Destroying volume [basic_test_volume_1495752495]
2017/05/25 22:48:26 Checking volume [basic_test_volume_1495752495] availability from VM [10.192.51.202]
2017/05/25 22:48:26 Creating volume [basic_test_volume_1495752495] on VM [10.192.51.202]
2017/05/25 22:48:29 Checking volume [basic_test_volume_1495752495] availability from VM [10.192.51.202]
2017/05/25 22:48:29 Attaching volume [basic_test_volume_1495752495] on VM [10.192.51.202]
2017/05/25 22:48:31 Confirming attached status for volume [basic_test_volume_1495752495]
2017/05/25 22:48:33 Removing container [basic_test_container_1495752495] on VM [10.192.51.202]
[...]

PR #1320, opened 2017-06-01
Introducing common e2e test utils

pshahzeb:
This PR is a subset of PR #1299. This only includes changes for common test utils.

1. govc utils for datastores and related constants
2. volume read/write utils
3. volume create with options util
4. admincli util to create a policy and update volume access

Testing: e2e-test target passes locally.


pshahzeb:
@govint I am merging this PR with common utils. Should any need arise to further improve them, we can accommodate it in further PRs.

PR #1352, opened 2017-06-06
Improving error message in plugin state where the refcounting hasn't been successful

### LINKED ISSUE #1348: VMDK plugin returns confusing message
When VMDK plugin is being restarted (e.g. due to VM restart), the background refcnt thread needs some time to initialize. During the initialization process, the plugin might get incoming mount/unmount/remove volume requests. These operations will need to get the ref counts of the volume. In this case , the current logic will return a hard-coded value 1, indicating the refcnt is still in initialization process. This caused very confusing error message being returned to the caller. For example, after verifying the volume status is detached, removing this volume still failed with error message: "Remove failure - volume is still mounted", which is very confusing and misleading.

As we discussed, we should be using a special value, such as -1, to indicate this special status when refcount is still initializing. Using 1 will make it confusing with the normal scenario where the volume is truly being referenced by one container.

We need to fix all 3 functions - vmdk.getRefCount(), vmdk.incrRefCount() and vmdk.decrRefCount() - to return this special value, and fixing the calling APIs accordingly to check this special value and handle the logic appropriately.

This issue has caused vmlistener test case fails intermittently - see issue #1308

pshahzeb:
1. refcount get, incr and decr functions return err instead of 1 in state where refcounting hasn't been successful.
2. updating the callers of above functions to make use of this error

Fixes #1348

### DIFF
--- a/vmdk_plugin/drivers/photon/photon_driver.go
+++ b/vmdk_plugin/drivers/photon/photon_driver.go
@@ -114,7 +114,7 @@ func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir
 
 // Return the number of references for the given volume
 func (d *VolumeDriver) getRefCount(vol string) uint {
-	if d.refCounts.GetInitSuccess() != true {
+	if d.refCounts.IsInitialized() != true {
 		return 1
 	}
 	return d.refCounts.GetCount(vol)
@@ -122,7 +122,7 @@ func (d *VolumeDriver) getRefCount(vol string) uint {
 
 // Increment the reference count for the given volume
 func (d *VolumeDriver) incrRefCount(vol string) uint {
-	if d.refCounts.GetInitSuccess() != true {
+	if d.refCounts.IsInitialized() != true {
 		return 1
 	}
 	return d.refCounts.Incr(vol)
@@ -130,7 +130,7 @@ func (d *VolumeDriver) incrRefCount(vol string) uint {
 
 // Decrement the reference count for the given volume
 func (d *VolumeDriver) decrRefCount(vol string) (uint, error) {
-	if d.refCounts.GetInitSuccess() != true {
+	if d.refCounts.IsInitialized() != true {
 		return 1, nil
 	}
 	return d.refCounts.Decr(vol)
@@ -560,8 +560,10 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Removing volume ")
 
-	if d.refCounts.GetInitSuccess() != true {
-		msg := fmt.Sprintf(plugin_utils.ErrorPluginInit+" Cannot remove volume=%s", r.Name)
+	// Cannot remove volumes till plugin completely initializes (refcounting is complete)
+	// because we don't know if it is being used or not
+	if d.refCounts.IsInitialized() != true {
+		msg := fmt.Sprintf(plugin_utils.PluginInitError+" Cannot remove volume=%s", r.Name)
 		log.Error(msg)
 		return volume.Response{Err: msg}
 	}
@@ -639,7 +641,7 @@ func (d *VolumeDriver) Unmount(r volume.UnmountRequest) volume.Response {
 	d.refCounts.StateMtx.Lock()
 	defer d.refCounts.StateMtx.Unlock()
 
-	if d.refCounts.GetInitSuccess() != true {
+	if d.refCounts.IsInitialized() != true {
 		// if refcounting hasn't been succesful,
 		// no refcounting, no unmount. All unmounts are delayed
 		// until we succesfully populate the refcount map
--- a/vmdk_plugin/drivers/photon/photon_driver.go
+++ b/vmdk_plugin/drivers/photon/photon_driver.go
@@ -560,7 +560,13 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Removing volume ")
 
-	// Docker is supposed to block 'remove' command if the volume is used. Verify.
+	if d.refCounts.GetInitSuccess() != true {
+		msg := fmt.Sprintf(plugin_utils.ErrorPluginInit+" Cannot remove volume=%s", r.Name)
+		log.Error(msg)
+		return volume.Response{Err: msg}
+	}
+
+	// Docker is supposed to block 'remove' command if the volume is used.
 	if d.getRefCount(r.Name) != 0 {
 		msg := fmt.Sprintf("Remove failure - volume is still mounted. "+
 			" volume=%s, refcount=%d", r.Name, d.getRefCount(r.Name))
--- a/vmdk_plugin/drivers/vmdk/vmdk_driver.go
+++ b/vmdk_plugin/drivers/vmdk/vmdk_driver.go
@@ -422,7 +422,13 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Removing volume ")
 
-	// Docker is supposed to block 'remove' command if the volume is used. Verify.
+	if d.refCounts.GetInitSuccess() != true {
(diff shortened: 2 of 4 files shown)


vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -113,25 +113,25 @@ func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir
...
+	return d.refCounts.Incr(vol), nil
 }
 
 // Decrement the reference count for the given volume
 func (d *VolumeDriver) decrRefCount(vol string) (uint, error) {
 	if d.refCounts.GetInitSuccess() != true {
-		return 1, nil
+		return 0, fmt.Errorf(plugin_utils.ErrorPluginInit)
 	}
 	return d.refCounts.Decr(vol)
```
shuklanirdesh82: Don't we need to add `,nil` for the error part here.

pshahzeb: -1 is leading to a lot of change of refcount apis. besides, actual refcount of a volume is stored as uint. using -1 is causing underflow error. So using error instead. Also it makes the logic more clear.

govint: Does it compile if the function doesn't return the number of return values?

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -102,25 +102,25 @@ func (d *VolumeDriver) VolumesInRefMap() []string {
...
+	return d.refCounts.Incr(vol), nil
 }
 
 // Decrement the reference count for the given volume
 func (d *VolumeDriver) decrRefCount(vol string) (uint, error) {
 	if d.refCounts.GetInitSuccess() != true {
-		return 1, nil
+		return 0, fmt.Errorf(plugin_utils.ErrorPluginInit)
 	}
 	return d.refCounts.Decr(vol)
```
govint: return nil for error

vmdk_plugin/drivers/photon/photon_driver.go
shaominchen: +1

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -561,9 +567,16 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Removing volume ")
 
 	// Docker is supposed to block 'remove' command if the volume is used. Verify.
```
shaominchen: Nit: remove "Verify." from the comments.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
shaominchen: +1

pshahzeb: Decr operation  on refcounts does return refcount value and error. that's already a tuple. so no need to specifically return the error.

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: the number of return values are same, Decr returns a tuple.

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -113,25 +113,25 @@ func NewVolumeDriver(targetURL string, projectID string, hostID string, mountDir
 // map hasn't been initialized, return 1 prevent detach and remove.
 
 // Return the number of references for the given volume
-func (d *VolumeDriver) getRefCount(vol string) uint {
+func (d *VolumeDriver) getRefCount(vol string) (uint, error) {
```
pdhamdhere: I would keep the current code AS-IS. Ref count not initialized is not really an error. You can get away with specific error message by modifying "Remove". See comment below.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
```diff
@@ -423,14 +436,21 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
+
+	if err != nil {
+		log.Error(err.Error())
+		return volume.Response{Err: err.Error()}
+	}
+
+	if refcnt != 0 {
 		msg := fmt.Sprintf("Remove failure - volume is still mounted. "+
-			" volume=%s, refcount=%d", r.Name, d.getRefCount(r.Name))
+			" volume=%s, refcount=%d", r.Name, refcnt)
```
pdhamdhere: Here you can check GetInitSuccess and return correct error by keeping the existing code as-is.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
shaominchen: I see. But that means the existing refcount.Incr() and refcount.Decr() are not well designed. A good API design should make the same group of functions/methods consistent. Obviously we are violating this which has caused this confusion. Is it possible to improve these 2 functions to keep them consistent?

pshahzeb:
Will keep existing implemention and modify the Remove handler to call GetInitSuccess and return the message appropriately.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: Sure.

vmdk_plugin/drivers/photon/photon_driver.go
shaominchen: We have actually discussed this option. We can check GetInitSuccess in all the callers, but returning 1 is a bad idea as it's always ambiguous.

vmdk_plugin/drivers/vmdk/vmdk_driver.go
pshahzeb: I think the reason was that incr cannot yield an error : it will be just increment the count.
Decr however needs to indicate error if caller tried to decrement the count below zero.

For this PR, keeping the fix as small as possible and capturing the error in Remove handler only.

vmdk_plugin/drivers/photon/photon_driver.go
pshahzeb: it is ambiguous but keeps the mount/unmount logic consistent. I think there are positive and negative sides to both the options. We can agree on one. 
My preference is to have the small fix for this blocker to distinguish between states for Remove command.
@govint @pdhamdhere @shaominchen

shaominchen: Fine to me. We can have a small fix for now to avoid possible regressions, and we can refactor/improve this part in the future if needed.

vmdk_plugin/utils/plugin_utils/plugin_utils.go
```diff
@@ -32,6 +32,9 @@ const (
 	// index datastore from volume meta
 	// "datastore" key is defined in vmdkops service
 	datastoreKey = "datastore"
+
+	// ErrorPluginInit message to indicate that plugin initialization(refcounting) is not yet complete
+	ErrorPluginInit = "Plugin initialization in progress."
```
shuklanirdesh82: I would suggest to use  `PluginInitError` instead of `ErrorPluginInit`

vmdk_plugin/drivers/photon/photon_driver.go
```diff
@@ -560,7 +560,13 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	log.WithFields(log.Fields{"name": r.Name}).Info("Removing volume ")
 
-	// Docker is supposed to block 'remove' command if the volume is used. Verify.
+	if d.refCounts.GetInitSuccess() != true {
```
govint: Nit may be - could be refcounts.IsInitialized() which can be true or false.

govint: If it remains false then no volume can be unmounted? Its as good as all volumes are attached and remain so? Could add a comment here to briefly explain this.

Please add a test to mount a volume and then on unmount it gets this error, so the volume is now left attached to the VM till the VM is powered off.

Possibly add the volumes (name) to a list and when the refcounting is done initializing, process the list to unmount volumes if needed.

pshahzeb: Sure. IsInitialized sounds better. 👍 
But this is the Remove flow.  The volume can't be removed until refcounting is complete.
I can add a comment about this.

For the unmount, we have the comment in Unmount func.

We are hitting this error through a test while removing the volume when refcount isn't complete yet.

PR #1354, opened 2017-06-06
Introducing test tags to avoid running test multiple times (#1313)

### LINKED ISSUE #1313: Optimize existing deployment and test run scripts
Our deployment scripts performs installation/cleanup for VIB and plugin multiple times while running against CI.

1. there are 2 ESX 6.5 & 6.0 and 8 VMS currently on the CI (need to perform installation in an optimize way)
2. avoid running test multiple time (e.g. plugin test which doesn't care about vsan,vmfs)

shuklanirdesh82:
Fixes #1313 (Part 2)

- avoid running test multiple time (e.g. plugin test which doesn't care about vsan,vmfs)

Adding following tags to the existing tests.

`runonce`: running tests only once doesn't care about underlying storage (vmfs/vsan) e.g. swarm tests, vmlistener test etc

`runalways`: running against all different kind of testbed (various guest OS, ESX version etc.) e.g. volumeproperty_Test, basic_test etc.

Introducing tags to pass `go test` command.


misc/drone-scripts/deploy-and-test-wrapper.sh
```diff
@@ -105,10 +105,10 @@ pluginSanityCheck)
 runtests)
         if [ -e /tmp/$ESX ]
         then
-          TARGET+=" e2e-dkrVolDriver-test test-vm"
+          TARGET+=" e2e-runalways-test test-vm"
         else
           touch /tmp/$ESX
-          TARGET+=" deploy-vm e2e-dkrVolDriver-test testasroot test-esx test-vm"
+          TARGET+=" deploy-vm e2e-runalways-test e2e-runonce-test testasroot test-esx test-vm"
```
shaominchen: Nit: Can we rename them to: e2e-test-runonce and e2e-test-runalways?

tests/e2e/basic_test.go
```diff
@@ -15,6 +15,8 @@
 // This test suite includes test cases to verify basic functionality
 // in most common configurations
 
+// +build runalways
```
shaominchen: One question: can we apply this tag on specific test cases?

tests/e2e/vmlistener_test.go
```diff
@@ -16,6 +16,8 @@
 // In case of vm/esx host restart vDVS cleans up stale attachment details
 // the volume.
 
+// +build unstable
```
shaominchen: One question: can we apply this tag on specific test cases? We do need this flexibility as most of the time we may just need to disable one test case instead of one test suite.

tests/e2e/swarm_test.go
```diff
@@ -112,7 +114,7 @@ func (s *SwarmTestSuite) TestDockerSwarm(c *C) {
 	status, host := verification.IsDockerContainerRunning(s.swarmNodes, s.serviceName, 2)
 	c.Assert(status, Equals, true, Commentf("Container %s is not running on any hosts", s.serviceName))
 
-	containerName, err = dockercli.GetContainerName(host, s.serviceName+".1")
+	containerName, err := dockercli.GetContainerName(host, s.serviceName+".1")
```
shaominchen: Just curious - why didn't we run into any issues due to this?

ashahi1:
@shuklanirdesh82 Going forward, is it mandatory that all tests need to have at least one tag?

shaominchen:
I actually have similar question as Anup asked above. If we don't add any tags, what's the default behavior?

tests/e2e/swarm_test.go
ashahi1: @shaominchen I did run into this issue. 

CI: https://ci.vmware.run/vmware/docker-volume-vsphere/362

```
# github.com/vmware/docker-volume-vsphere/tests/e2e
../tests/e2e/swarm_test.go:115: undefined: containerName
../tests/e2e/swarm_test.go:116: undefined: containerName
../tests/e2e/swarm_test.go:117: undefined: containerName
FAIL	github.com/vmware/docker-volume-vsphere/tests/e2e [build failed]
make[1]: *** [e2e-dkrVolDriver-test] Error 2
make: *** [e2e-dkrVolDriver-test] Error 2
```

shaominchen: Yes, I know this is a syntax error, but just curious why it's still passing locally and also in most CI runs. It seems very weird.

misc/drone-scripts/deploy-and-test-wrapper.sh
shuklanirdesh82: sure!

tests/e2e/swarm_test.go
shuklanirdesh82: Yeah .. it failed for me only on CI I got curious too... `:=` was required so did not bother much.

shuklanirdesh82:
> Going forward, is it mandatory that all tests need to have at least one tag?

Yeah @ashahi1! Thanks for bringing this up ... we need to add such tags  otherwise it falls into default category i.e. run always.

shuklanirdesh82:
> a comment will help explain what the runonce/always is all about in the code. 

Sure @govint  I will add that information in the Makefile ... 

> Or in the TBD test-writers-bible.

I will raise separate issue to work on go-to guide for adding testing details like VIC or Docker team is having `Testing for vDVS` something like section.

tests/e2e/basic_test.go
shuklanirdesh82: Unfortunately not.

tests/e2e/swarm_test.go
shaominchen: No, containerName is a local variable, which is generated based on s.serviceName.

shaominchen:
"we need to add such tags otherwise it falls into default category i.e. run always."
If that's the case, then I think we can remove the "runalways" tag to keep it simple.

tests/e2e/vmlistener_test.go
shuklanirdesh82: There is not any straight forward way that I know of. Selective run is possible through `go test -run TestNameRegexp` .. we need to come up with the `TestNameRegexp` such that we can exclude certain tests. 

There are other alternative that go test community user recommends like writing interceptor that filter out tests.

shuklanirdesh82:
> If that's the case, then I think we can remove the "runalways" tag to keep it simple.

How can we avoid running `unstable` or `runonce` .. there is no way to exclude tags that I know of.

tests/e2e/vmlistener_test.go
shaominchen: Fine. Before we figure out how to do this, we can use some workaround - for example, add a "disable" prefix to the test case name to skip one single unstable test case.

shuklanirdesh82:
Note: with the latest commit `gofmt` auto corrects formatting for `tests/e2e/vmgroups_test.go`

Issue #1301, opened 2017-05-26
[E2E] Need a general poller util to avoid blind sleeping in the tests


shaominchen:
In some e2e tests, we have to wait until some time-consuming operations (such as VM power on/off, Kill Docker Container, etc.) to complete. Currently we are just blindly sleeping for some fixed time, and then verify the expected result. This is not only inefficient, but also unstable.

It's inefficient because we have to set a time limit big enough to make sure the operation is completed, and even if the operation completes in a shorter time, we are still waiting the fixed time, which is just a waste.

It's unstable because in some cases (depending on machine, OS, or any other conditions) the operations may not complete in the fixed time, then the test case will still fail.

To solve this issue, we need to introduce a general util function to do polling. This util function should accept a function, expected result, maximum time limit as parameters. This way we can avoid blind waiting here and there, which will significantly slow down the whole test suite. The basic polling logic will be like this:

for {
    result := time_consuming_func()
    if result == "expected_result" {
        return true;
    }

    if timePassed > timeLimit {
        return false;
    }
    sleep(short_interval)
}

shuklanirdesh82:
@shaominchen  I am parking this with you .. let's see we need it as part of 0.16 or we may defer it for  v1GA.

shaominchen:
This will improve the total running time of e2e test, but not a must fix for Kubecon.

Issue #1291, opened 2017-05-24
vmdk_opsd does not properly restart on signal 10 (and instead throws an exception) 


ashahi1:
Seeing ```Function not implemented error``` in the vmdkops log after I run the command to Initialize the Config DB to SingleNode.

Command: /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py config init --local

/var/log/vmware/vmdk_ops.log snippet:
```
05/24/17 18:50:42 1386157 [MainThread] [ERROR  ] [Errno 38] Function not implemented: '/usr/lib/vmware/vmdkops/bin/vmdk_ops.py'
Traceback (most recent call last):
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 1706, in main
    handleVmciRequests(port)
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 1637, in handleVmciRequests
    c = lib.vmci_get_one_op(g_vmci_listening_socket, byref(cartel), txt, c_int(bsize))
  File "/usr/lib/vmware/vmdkops/bin/vmdk_ops.py", line 1530, in signal_handler_restart
    os.execve(__file__, sys.argv, os.environ)

```

Logs: 
[vmdk_ops.txt](https://github.com/vmware/docker-volume-vsphere/files/1026768/vmdk_ops.txt)

govint:
Happens in the signal handler when the service is restarted. Saw this a few times last week but then its not consistent. Are we trying to restart the script when the signal is received?

msterin:
To clarify what is going on: 

* for some reason ` os.execve` in `signal_handler_restart` throws the "function not implemented" exception (see above). **This causes the whole service to exit, and it is restarted by init.d watchdog , up to 10 times (per config in vmdk-opsd). After the 10th time the service stays down** 

* this exposes a few obvious glitches, but unfortunately it is not clear what is going on:
  * os.execve() works fine on ESX (we double-checked it), but fails when called in the context of C waiting for VMCI connection. Not sure if it's an issue or a coincidence
  * we close vmci socket in `vmci_release_listening_socket` but do not clear up global `g_vmci_listening_socket`
  * os.execve params seem to be inaccurate (see suggestion below)

---

Note: Fixing the above items **does not** fix the issue, but we need to fix it still to remove dirt. 

Here is how execve params should look like, probably: 
```python
   executable = os.path.realpath(sys.executable)
[...]

PR #1368, opened 2017-06-07
Add a basic test: TestVmGroupVolumeIsolation

shaominchen:
1. Added a new test case: TestVmGroupVolumeIsolation
2. Refactored existing test cases to use new common utilities
3. Fixed a cleanup issue in vmgroups_test
4. Updated common utils LogTestStart & LogTestEnd and all related test cases.

Fixes issue #1263, #1372 and #1394 .


tests/e2e/basic_test.go
```diff
@@ -20,34 +20,54 @@
...
+	}
+
+	s.esx = s.config.EsxHost
+	s.vm1 = s.config.DockerHosts[0]
+	s.vm2 = s.config.DockerHosts[1]
+	s.vm1Name = s.config.DockerHostNames[0]
+	s.vm2Name = s.config.DockerHostNames[1]
+	s.volName1 = inputparams.GetUniqueVolumeName(testName)
+	s.volName2 = inputparams.GetUniqueVolumeName(testName)
+	s.containerName = inputparams.GetContainerNameWithTimeStamp(testName)
```
govint: I was thinking maybe just put this into a dummy_test.go file and let these esx, ... variables be globals there. All test modules in the e2e package would be able to refer those variables without each module doing the above. Any repetition of code anywhere could be avoided.

tests/e2e/basic_test.go
```diff
@@ -59,70 +79,123 @@ var _ = Suite(&BasicTestSuite{})
...
+	out, err = dockercli.DeleteVolume(s.vm2, s.volName1)
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = dockercli.DeleteVolume(s.vm1, s.volName2)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Remove Config DB
+	admincli.ConfigRemove(s.esx)
+
+	misc.LogTestEnd(testName, "TestVmGroupVolumeIsolation")
```
govint: This test seems very similar if not exact to vmgroups_test.go:TestVmGroupVolumeAccessAcrossVmGroups(). If neeeding to enhance then the existing test can be enhanced vs. adding a test here. Plus its best to keep all vmgroup + volume related tests in one place.

tests/e2e/vmgroups_test.go
```diff
@@ -111,6 +111,9 @@ func (vg *VmGroupTest) TearDownSuite(c *C) {
 	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
 	log.Printf(out)
 
+	// Remove Config DB
+	adminutils.ConfigRemove(vg.config.EsxHost)
+
```
govint: This must not be done, its creating a fresh and clean plate for the next test to run and we may potentially wipe out a scenario that may lead to tests failing and bugs getting exposed. In fact, for all the e2e tests create the config DB once and let the tests run thru without removng the config DB (except for those tests that test the config DB itself).

tests/utils/misc/misc.go
```diff
@@ -48,18 +48,13 @@ func WaitForExpectedState(fn getVMPowerStatus, vmName, expectedState string) boo
 	log.Printf("Timed out to poll status\n")
 	return false
 }
-// LogTestStart - Print a start log with given test name and
-// current time stamp
+
+// LogTestStart - Print a start log with given test group and test case name
 func LogTestStart(testGroup string, testName string) {
-	log.Printf("START:%s %s %s", testGroup, testName, curTime())
+	log.Printf("START: %s.%s", testGroup, testName)
```
govint: Please retain timestamp, when debugging we know exactly when a test started and when it finished and hence correlate other logs (daemon, plugin). It also lets us know which tests are taking time. Hence it was defined this way.

tests/utils/misc/misc.go
```diff
@@ -48,18 +48,13 @@ func WaitForExpectedState(fn getVMPowerStatus, vmName, expectedState string) boo
...
-// LogTestEnd - Print a stop log with given test name and
-// current time stamp
+// LogTestEnd - Print a stop log with given test group and test case name
 func LogTestEnd(testGroup string, testName string) {
-	log.Printf("END:%s %s %s", testGroup, testName, curTime())
-}
-
-func curTime() string {
-	return time.Now().Format(time.UnixDate)
+	log.Printf("END: %s.%s", testGroup, testName)
```
govint: Same comment for the timestamp.

tests/e2e/basic_test.go
lipingxue: +1 on this

tests/e2e/basic_test.go
```diff
@@ -20,34 +20,54 @@
...
+	s.vm2Name = s.config.DockerHostNames[1]
+	s.volName1 = inputparams.GetUniqueVolumeName(testName)
+	s.volName2 = inputparams.GetUniqueVolumeName(testName)
+	s.containerName = inputparams.GetContainerNameWithTimeStamp(testName)
 }
 
 var _ = Suite(&BasicTestSuite{})
 
 // Test volume lifecycle management on different datastores:
 // VM1 - local VMFS datastore
```
lipingxue: // VM1 - local VMFS datastore -> VM1 - created on local VMFS datastore

tests/e2e/basic_test.go
shaominchen: Personally I don't really have much concern about this. We have already introduced common TestConfig, which I feel is sufficient. Here I introduced these local variables just for convenience because I personally prefer short names locally. Otherwise I would just use s.config.XXX.

tests/e2e/basic_test.go
shaominchen: I completely agree with you - currently the test cases we defined on the Dashboard have a lot of overlap and might be duplicated. I will take a closer look at the test case you pointed out, and decide what we should do to avoid the duplication.

tests/e2e/vmgroups_test.go
lipingxue: If we decide to do this, we must make sure each test clean up itself properly. For example, if a test create a VM group, it must remove the VM group when test end (not only in successful case, but also in failure case).  I don't think current test code did this.

tests/utils/verification/volumeproperties.go
```diff
@@ -173,7 +173,7 @@ func VerifyDetachedStatus(name, hostName, esxName string) bool {
 	log.Printf("Confirming detached status for volume [%s]\n", name)
 
 	//TODO: Need to implement generic polling logic for better reuse
-	for attempt := 0; attempt < 30; attempt++ {
+	for attempt := 0; attempt < 60; attempt++ {
```
lipingxue: define a const for this instead of using hard coded "60".

vmdk_plugin/Makefile
```diff
@@ -334,6 +334,7 @@ checkremote:
 # expects binaries to be deployed ot the VM and ESX (see deploy-all target)
 test-vm: checkremote deploy-vm-test
 	$(log_target)
+	-$(SSH) root@$(ESX) '/etc/init.d/vmdk-opsd start'
```
lipingxue: why we need this here?

tests/e2e/vmgroups_test.go
shaominchen: I don't follow this. What's the default customer scenario we are expecting? I may be wrong , but I think we are expecting a typical customer environment will be a clean vDVS installation without initializing local or shared DB. So most of e2e tests should be run in the default setup, i.e. NotConfigured mode.

Multi-tenancy features are experimental. So only VmGroup related tests should take care of ConfigInit and ConfigRemove.

tests/utils/misc/misc.go
shaominchen: I don't follow this. As I've already explained in the PR description, go log itself will print out timestamp for each log message, for example:

2017/06/05 02:46:12 START: restart_test.TestPluginKill
2017/06/05 02:46:12 Attaching volume [restart_test_volume_1496630768] on VM[10.192.232.148]
2017/06/05 02:46:14 Confirming attached status for volume [restart_test_volume_1496630768]
2017/06/05 02:46:16 Killing vDVS plugin on VM [10.192.232.148]
2017/06/05 02:46:18 Sleep for 2 seconds

Why do we want to append a redundant timestamp?

tests/utils/misc/misc.go
shaominchen: Same as above,.

tests/utils/verification/volumeproperties.go
shaominchen: I've already added TODO here. We need a general poller util to avoid blind sleeping in the tests - see issue #1301.

vmdk_plugin/Makefile
shaominchen: Due to issue #1291, we need to restart the service because it has crashed due to "Config Remove".

tests/e2e/basic_test.go
shuklanirdesh82: > on the Dashboard have a lot of overlap and might be duplicated.

We have broke down tests into smaller steps so you might see the overlap for some steps with others. As discussed in the past we have to avoid such duplication during the automation by enhancing the existing one.

+1 on @govint's comment.

tests/utils/verification/volumeproperties.go
shuklanirdesh82: Isn't it better to control const variable to adjust `attempt` until we come to #1301? I would prefer to have const so we don't need to dive into code to adjust it and do some adjustment by controlling const variable.

tests/e2e/basic_test.go
```diff
@@ -20,34 +20,54 @@
...
 // VM2 - shared VMFS datastore
-// VM3 - shared VSAN datastore
+// VM3 - shared VSAN datastore (TODO: currently not available)
```
shuklanirdesh82: what is the need for VM3 here? if I understand correctly ... VM3 should be registered on the same ESX where VM1 & VM2 are there then it is easily configured on the CI as well as local testbed.

tests/e2e/basic_test.go
shaominchen: Will clarify this.

tests/e2e/basic_test.go
shaominchen: This test is to verify volume creation on different datastores. VM1 is created on local VMFS datastore, VM2 is created on shared VMFS datastore, and VM3 is created on shared VSAN datastore. I will update these comments to avoid confusion.

tests/utils/verification/volumeproperties.go
shaominchen: Fine. Since you both prefer adding consts now, I will add it.

tests/e2e/basic_test.go
shaominchen: @govint I took a further look into vmgroups_test.TestVmGroupVolumeAccessAcrossVmGroups(). This test is very similar to basic_test.TestVmGroupVolumeIsolation(). However, I think it makes sense to keep both because the former is trying to verify the volume isolation between two user-defined vmgroups, while the latter is trying to verify the volume isolation between _DEFAULT and user-defined vmgroup.

I agree we should keep all vmgroup related tests in one single place. I tried to move my test case into vmgroups_test.go. However, this needs too much refactoring to the existing logic in vmgroups_test.go, because the test assumption is different. In vmgroups_test.go, all tests assume that VM1&VM2 belong a user-defined vmgroup (we shouldn't call it "default" in the comments, which will cause confusion with _DEFAULT vmgroup). The current test case doesn't have this assumption. To avoid significant change to the existing vmgroups_test.go, I will keep this test in basic_test.go for now.

shaominchen:
@govint @lipingxue @shuklanirdesh82 Please review the latest updates.

vmdk_plugin/Makefile
shuklanirdesh82: Please remove this file and rebase with master to grab fix for #1291

shaominchen:
Removed the workaround of restarting vmdk-opsd. Please take a look @govint @shuklanirdesh82

tests/e2e/vmgroups_test.go
govint: @lipingxue current test code removes all vmgroups in TearDownSuite() which is _always_ called irrespective of pass/fail of the test suite.

A test must remove all artifacts that are created by it. But why remove the config DB? 
After removing all test artifacts is the DB sane, how do we know that it is?

If we remove the config DB (which btw isn't a test artifact) then most likely we are removing bugs as well. If the removal of test artifacts leaves the DB is a corrupt/inconsistent state - whether local or clustered - then shouldn't that be discovered in the testing?

We can't run without the config DB - customers are trying it out and hence it must be in all our tests.

tests/e2e/basic_test.go
```diff
@@ -59,70 +77,124 @@ var _ = Suite(&BasicTestSuite{})
...
-	out, err = dockercli.DeleteVolume(vm1, s.volumeName)
+	out, err = dockercli.CreateVolume(s.vm1, s.volName2)
 	c.Assert(err, IsNil, Commentf(out))
 
-	log.Printf("END: basic-test.TestVolumeIsolation")
+	accessible = verification.CheckVolumeAvailability(s.vm1, s.volName2)
+	c.Assert(accessible, Equals, true, Commentf("Volume %s is not available on [%s]", s.volName2, s.vm1))
+
+	accessible = verification.CheckVolumeAvailability(s.vm2, s.volName2)
+	c.Assert(accessible, Equals, false, Commentf("Volume %s is still available on [%s]", s.volName2, s.vm2))
```
shuklanirdesh82: nit: `Volume %s is still available on` => `Volume %s is available on` or `Volume %s should not be available on`

tests/e2e/vmgroups_test.go
shaominchen: @govint Yesterday we (sorry you were not included) discussed this and all agreed that most of our e2e tests should be running in the "default" mode. The "default" mode here means "No Config DB", i.e. a fresh installation without running "Config Init". For now multi-tenancy feature is experimental, so (at least for now) we are expecting most of our customers will be using vDVS in the "default" mode. This assumption will remain true until multi-tenancy is fully/correctly implemented and becomes an official product feature.

Coming back to e2e tests, we are expecting most of the test groups will be run in the "default" mode without DB configured. Only vmgroup related tests will need to run "config init", and thus it is expected these tests should be responsible for cleaning up the DB (i.e. run "config rm") so that other e2e tests will still be running in a clean "default" mode.

Hope this helps. Let me know if you still have question about this.

tests/e2e/basic_test.go
shaominchen: The phrase "... should not be available on" looks fine to me. I searched our exiting test code - it looks like we are all using "is still available" or "is still attached/detached" - please refer to restart_test, vmlistener_test, basic_test, swarm_test, etc. So I'd prefer to keep it consistent.

govint:
@shaominchen, does the swarm test time delays added need to be explained in the docs? Can a customer face the same issue as whats seen with these tests?

PR #1369, opened 2017-06-07
Converting UUID to tenant name in log messages

### LINKED ISSUE #955: Use the tenant name in the logs instead of tenant IDs
The tenant UUIDs are long sequences of digits and make for difficult reading in the logs and error messages as the user needs to lookup the tenant mapping from UUID to tenant name to figure which tenant is in use for volume ops (when debugging).

When a tennat dir. (uuid is dir name) is made then also create a symlink to that uuid-dir which is the name of the tenant. Similar to ESX datastores that use a readable name for every datastore.

shivanshu21:
Fixes #955 

Fixing out of bounds issue

More places printing uuid changed

All non swarm CI tests passing
UUID not printing in log messages


esx_service/utils/vm_listener.py
```diff
@@ -212,9 +211,13 @@ def set_device_detached(vm_moref):
         # if it is a dvs managed volume, set its status as detached
         vmdk_path = vmdk_utils.find_dvs_volume(dev)
         if vmdk_path:
-            logging.info("Setting detach status for %s", vmdk_path)
+            readable_vmdk_path = vmdk_ops.gets_readable_path(vmdk_path)
```
msterin: I think checking for conversion errors adds tons of code and loses the original value, without a real reason for it. 

It would be easier if gets_readable_path(vmdk_path) (btw, why `gets` and not `get`?) would be just returning the original path in case of an error. Then the lines 214->218 would be simply
```python
 logging.info("Setting detach status for %s", vmdk_ops.get_readable_path(vmdk_path))

File: esx_service/utils/vmdk_utils.py [hunk 2]
```diff
@@ -346,7 +346,11 @@ def find_dvs_volume(dev):
     # "[<datastore name>] <parent-directory>/tenant/<vmdk-descriptor-name>"
     # Trim the datastore name and keep disk path.
     datastore_name, disk_path = dev.backing.fileName.rsplit("]", 1)
-    logging.info("backing disk name is %s", disk_path)
+    readable_vmdk_path = vmdk_ops.gets_readable_path(disk_path)
+    if readable_vmdk_path == '':
```
msterin: same comment as above

msterin: ditto

esx_service/utils/vsan_info.py
```diff
@@ -69,11 +69,16 @@ def set_policy(vmdk_path, policy_string):
     Returns None on success else return the output error. This can be
     displayed at the client
     """
+    readable_vmdk_path = vmdk_ops.gets_readable_path(vmdk_path)
+    if readable_vmdk_path == '':
+        logging.warning("Failed to fetch readable vmdk path")
```
msterin: and again

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
 
     return (s, o)
 
+
+# Translates the vmdk_path into a readable path by
+# removing the UUID string and using tenant_name
+def gets_readable_path(vmdk_path):
```
msterin: nit : missing docstring

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+    if searchObj is None:
+        return vmdk_path
+
+    # iterate over the directory and figure out
+    # which symlink's data matches the t_uuid
+    dvpath = ''
+    for i in range(1, idx + 1):
+        dvpath += '/'
+        dvpath += words[i]
+    dvcontents = os.listdir(dvpath)
```
msterin: This  means bunch of disk IO (well, maybe it will hit the buffer cache, but still) just to print a single error message.
This is not really acceptable, especially since there is more (like symlink reading etc.

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
 
     return (s, o)
 
+
+# Translates the vmdk_path into a readable path by
+# removing the UUID string and using tenant_name
+def gets_readable_path(vmdk_path):
+    idx = 0
+    t_name = ''
```
msterin: what is t_name ? Is it tenant_name ? Then it should be named so. The aqme about t_uuid. There is no reason to save 5 characters in the name and make it harder to read

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+                        t_uuid,
+                        re.M|re.I)
+    if searchObj is None:
+        return vmdk_path
+
+    # iterate over the directory and figure out
+    # which symlink's data matches the t_uuid
+    dvpath = ''
+    for i in range(1, idx + 1):
+        dvpath += '/'
```
msterin: os.path package is your friend. Please use it for getting path components. If you really want to do it yourself, I think split("/") is an easier approach

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+        dvpath += words[i]
+    dvcontents = os.listdir(dvpath)
+    for k in dvcontents:
+        if os.path.islink(dvpath + '/' + k):
+            if t_uuid in os.readlink(dvpath + '/' + k):
+                t_name = k
+
+    if t_name == '':
+        logging.warning("Failed to find tenant name in vmdk_path")
```
msterin: this should be logging.debug at best. Or no message at all. If anything goes wrong, simply return original string - no damage done

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+        return ''
+    dvpath += '/'
+    dvpath += t_name
+    for i in range (idx + 2, len(words)):
```
msterin: Please do not manipulate *nix paths this way, use os.path

esx_service/vmdk_ops.py
```diff
@@ -244,13 +301,18 @@ def createVMDK(vmdk_path, vm_name, vol_name,
...
-    logging.info("*** cloneVMDK: %s opts = %s vm_uuid=%s datastore_url=%s", vmdk_path, opts, vm_uuid, datastore_url)
-
     # Get source volume path for cloning
     error_info, tenant_uuid, tenant_name = auth.get_tenant(vm_uuid)
     if error_info:
         return err(error_info)
 
+    readable_vmdk_path = gets_readable_path(vmdk_path)
+    if readable_vmdk_path == '':
+        logging.warning("Failed to fetch readable vmdk path")
```
msterin: same comment, no need to check for return value, just return the passed value on error and use it.

esx_service/vmdk_ops.py
```diff
@@ -244,13 +301,18 @@ def createVMDK(vmdk_path, vm_name, vol_name,
...
 
 def cloneVMDK(vm_name, vmdk_path, opts={}, vm_uuid=None, datastore_url=None):
-    logging.info("*** cloneVMDK: %s opts = %s vm_uuid=%s datastore_url=%s", vmdk_path, opts, vm_uuid, datastore_url)
-
     # Get source volume path for cloning
     error_info, tenant_uuid, tenant_name = auth.get_tenant(vm_uuid)
     if error_info:
         return err(error_info)
 
+    readable_vmdk_path = gets_readable_path(vmdk_path)
```
msterin: why did you move the loggign after get_tenant ? It is the "function entered" log, it should be before other code

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
                opts={}, vm_uuid=None, tenant_uuid=None, datastore_url=None):
-    logging.info("*** createVMDK: %s opts = %s vm_name=%s tenant_uuid=%s datastore_url=%s",
-                  vmdk_path, opts, vm_name, tenant_uuid, datastore_url)
+
+    e_info, t_uuid, tenant_name = auth.get_tenant(vm_uuid)
+    if e_info:
+        return err(e_info)
+
+    readable_vmdk_path = gets_readable_path(vmdk_path)
+    if readable_vmdk_path == '':
```
msterin: same comment,  please drop the return value check and return original one on error in get_readable_path

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+
 # returns error, or None for OK
 # opts is  dictionary of {option: value}.
 # for now we care about size and (maybe) policy
 def createVMDK(vmdk_path, vm_name, vol_name,
                opts={}, vm_uuid=None, tenant_uuid=None, datastore_url=None):
-    logging.info("*** createVMDK: %s opts = %s vm_name=%s tenant_uuid=%s datastore_url=%s",
-                  vmdk_path, opts, vm_name, tenant_uuid, datastore_url)
+
+    e_info, t_uuid, tenant_name = auth.get_tenant(vm_uuid)
```
msterin: why dd you add get_tenant here ?   It's a fairly expensive op (goes to DB) , so doing it for logging info only is not worth it

esx_service/utils/vmdk_utils.py
```diff
@@ -346,7 +346,11 @@ def find_dvs_volume(dev):
...
+        logging.warning("Failed to fetch readable vmdk path")
+
+    logging.info("backing disk name is %s", readable_vmdk_path)
```
pshahzeb: At this point, if readable_vmdk_path is empty, we will print warning on 351 but empty vmdk_path. This isn't worth. Lets print the whatever vmdk_path we have if we failed to convert it to readable one

esx_service/utils/vm_listener.py
pshahzeb: +1

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+            break
+        idx += 1
+
+    # The format is not always the same for vmdk_path
+    # If you suspect this is not the known format, return
+    # the original string
+    if (len(words) <= idx + 1):
+        return vmdk_path
+    t_uuid = words[idx + 1]
+    searchObj = re.search(r'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}',
```
pshahzeb: minor: can add a small comment here.

esx_service/vmdk_ops.py
```diff
@@ -159,17 +159,74 @@ def RunCommand(cmd):
...
+
+    e_info, t_uuid, tenant_name = auth.get_tenant(vm_uuid)
+    if e_info:
+        return err(e_info)
+
+    readable_vmdk_path = gets_readable_path(vmdk_path)
+    if readable_vmdk_path == '':
+        logging.warning("Failed to fetch readable vmdk path")
+
+    logging.info("*** createVMDK: %s opts = %s vm_name=%s tenant_name=%s datastore_url=%s",
```
pshahzeb: this log should be the first thing in this func.

shivanshu21:
Changed approach after discussion with Mark. Closing this request.

PR #1387, opened 2017-06-11
Print tenant name in logs instead of tenant ID

### LINKED ISSUE #955: Use the tenant name in the logs instead of tenant IDs

shivanshu21:
Description
=========

Changed approach after discussion with Mark. Closing previous pull request.
This fixes #955 

If the request comes with a tenant_name already, I return a path with symlink name instead of tenant UUID. The OS will handle resolving symlinks. 

Testing done:
==========
Manual tests:
1. Create VMDK
2. Remove VMDK

Ran multiple runs of CI automation.
After multiple ESX server operations, logs do not show tenant UUID in vmdk_path anymore.
Here is a CI test I scheduled on build server: https://ci.vmware.run/vmware/docker-volume-vsphere/458

Log Excerpts
==========
06/09/17 20:57:14 5055174 [photon-VM0.2-sharedVmfs-0._DEFAULT.option_volume_3132227180552437724] [INFO   ] *** createVMDK: /vmfs/volumes/sharedVmfs-0/dockvols/_DEFAULT/option_volume_3132227180552437724.vmdk opts = {'fstype': 'ext4', 'size': '10gb'} vm_name=photon-VM0.2 tenant_uuid=11111111-1111-1111-1111-111111111111 datastore_url=/vmfs/volumes/591cfe9a-12f9884b-340d-020037ced216

06/09/17 20:57:27 5055174 [photon-VM0.2-sharedVmfs-0._DEFAULT.option_volume_8185258463521160967] [INFO   ] *** removeVMDK: /vmfs/volumes/sharedVmfs-0/dockvols/_DEFAULT/option_volume_8185258463521160967.vmdk

06/09/17 20:57:18 5055174 [photon-VM0.2-sharedVmfs-0._DEFAULT.option_volume_182702735967048418] [INFO   ] *** attachVMDK: /vmfs/volumes/sharedVmfs-0/dockvols/_DEFAULT/option_volume_182702735967048418.vmdk to photon-VM0.2 VM uuid = 564d5cac-bc2b-bfde-45ca-14f6f8437476

### DIFF
--- a/esx_service/vmdk_ops.py
+++ b/esx_service/vmdk_ops.py
@@ -690,14 +690,18 @@ def detachVMDK(vmdk_path, vm_uuid):
 
 # Check existence (and creates if needed) the path for docker volume VMDKs
 def get_vol_path(datastore, tenant_name=None):
-    # If the command is NOT running under a tenant, the folder for Docker
+    # If tenant_name is set to None, the folder for Docker
     # volumes is created on <datastore>/DOCK_VOLS_DIR
-    # If the command is running under a tenant, the folder for Dock volume
+    # If tenant_name is set, the folder for Dock volume
     # is created on <datastore>/DOCK_VOLS_DIR/tenant_uuid
     # a symlink <datastore>/DOCK_VOLS_DIR/tenant_name will be created to point to
     # path <datastore>/DOCK_VOLS_DIR/tenant_uuid
+    # If the dock volume folder already exists,
+    # the path returned contains tenant name not UUID.
+    # This is to make logs more readable. OS will resolve this path
+    # as a symlink with tenant_name will already be present.
 
-    path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
+    readable_path = path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
 
     if tenant_name:
         error_info, tenant = auth_api.get_tenant_from_db(tenant_name)
@@ -705,11 +709,17 @@ def get_vol_path(datastore, tenant_name=None):
             logging.error("get_vol_path: failed to find tenant info for tenant %s", tenant_name)
             path = dock_vol_path
         path = os.path.join(dock_vol_path, tenant.id)
+        readable_path = os.path.join(dock_vol_path, tenant_name)
 
     if os.path.isdir(path):
-        # If the path exists then return it as is.
-        logging.debug("Found %s, returning", path)
-        return path, None
+        # If the readable_path exists then return, else return path with no symlinks
+        if os.path.exists(readable_path):
+            logging.debug("Found %s, returning", readable_path)
+            return readable_path, None
+        else:
+            logging.warning("Internal: Tenant name symlink not found for path %s", readable_path)
+            logging.debug("Found %s, returning", path)
+            return path, None
 
     if not os.path.isdir(dock_vol_path):
         # The osfs tools are usable for DOCK_VOLS_DIR on all datastores.
@@ -738,7 +748,7 @@ def get_vol_path(datastore, tenant_name=None):
             logging.info("Symlink %s is created to point to path %s", symlink_path, path)
 
     logging.info("Created %s", path)
-    return path, None
+    return readable_path, None
 
 def parse_vol_name(full_vol_name):
     """


esx_service/vmdk_ops.py
```diff
@@ -696,20 +696,24 @@ def get_vol_path(datastore, tenant_name=None):
     # is created on <datastore>/DOCK_VOLS_DIR/tenant_uuid
     # a symlink <datastore>/DOCK_VOLS_DIR/tenant_name will be created to point to
     # path <datastore>/DOCK_VOLS_DIR/tenant_uuid
+    # If the command is under a tenant then the path returned contains tenant
+    # name not UUID. This is to make logs more readable. OS will resolve this path
+    # as a symlink with tenant_name is already present.
 
-    path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
+    rpath = path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
```
msterin: what does "r" stands for in "rpath" ?  It would be better to use full names, whatever it is .... - easier to read

esx_service/vmdk_ops.py
```diff
@@ -696,20 +696,24 @@ def get_vol_path(datastore, tenant_name=None):
     # is created on <datastore>/DOCK_VOLS_DIR/tenant_uuid
     # a symlink <datastore>/DOCK_VOLS_DIR/tenant_name will be created to point to
     # path <datastore>/DOCK_VOLS_DIR/tenant_uuid
+    # If the command is under a tenant then the path returned contains tenant
```
msterin: I am not sure what "command is under a tenant" mean. Can you elaborate ?

esx_service/vmdk_ops.py
```diff
@@ -696,20 +696,24 @@ def get_vol_path(datastore, tenant_name=None):
...
 
     if tenant_name:
         error_info, tenant = auth_api.get_tenant_from_db(tenant_name)
         if error_info:
             logging.error("get_vol_path: failed to find tenant info for tenant %s", tenant_name)
             path = dock_vol_path
         path = os.path.join(dock_vol_path, tenant.id)
+        rpath = os.path.join(dock_vol_path, tenant_name)
 
     if os.path.isdir(path):
```
msterin: should not it be if os.path.ispath(rpath) ?

esx_service/vmdk_ops.py
```diff
@@ -696,20 +696,24 @@ def get_vol_path(datastore, tenant_name=None):
...
         # If the path exists then return it as is.
-        logging.debug("Found %s, returning", path)
-        return path, None
+        logging.debug("Found %s, returning", rpath)
```
msterin: You also should check (out of healty paranoia) if os.path.exists(os.path.realpath(rpath)) and  thow or print a warning if it does not

esx_service/vmdk_ops.py
```diff
@@ -696,20 +696,24 @@ def get_vol_path(datastore, tenant_name=None):
...
-    path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
+    rpath = path = dock_vol_path = os.path.join("/vmfs/volumes", datastore, DOCK_VOLS_DIR)
 
     if tenant_name:
         error_info, tenant = auth_api.get_tenant_from_db(tenant_name)
         if error_info:
             logging.error("get_vol_path: failed to find tenant info for tenant %s", tenant_name)
             path = dock_vol_path
         path = os.path.join(dock_vol_path, tenant.id)
+        rpath = os.path.join(dock_vol_path, tenant_name)
```
luomiao: why rpath? (in the sense, what's "r" stands for?

And it looks like rpath is in fact the same thing as the symlink_path on Line 739. Can you just use symlink_path here?

esx_service/vmdk_ops.py
shivanshu21: This is just an extension of the already present comment at the start of this function mentioned below:

"# If the command is NOT running under a tenant, the folder for Docker
 # volumes is created on <datastore>/DOCK_VOLS_DIR"

By this the original commenter means that tenant_name is not none in the args.

msterin: then let's change the comment (including the original one) so it is clear

esx_service/vmdk_ops.py
```diff
@@ -690,26 +690,36 @@ def detachVMDK(vmdk_path, vm_uuid):
...
             logging.error("get_vol_path: failed to find tenant info for tenant %s", tenant_name)
             path = dock_vol_path
         path = os.path.join(dock_vol_path, tenant.id)
+        readable_path = os.path.join(dock_vol_path, tenant_name)
 
-    if os.path.isdir(path):
-        # If the path exists then return it as is.
-        logging.debug("Found %s, returning", path)
-        return path, None
+    if os.path.isdir(readable_path) or os.path.isdir(path):
```
luomiao: I just feel the new logic here is different from old ones...
Before this fix, if path is not a dir, it will go ahead to the following steps to create it or run necessary things.
But the new logic is, as long as the readable_path is there (even path is not a dir), it will never go to the rest of the code.
Is it possible that, the original path is getting deleted but the symlink path is still existing. Then the original path will never be created again?
I am not sure if this is a possible situation. Just raise the concern.
So I think it's better to keep this check as it was "if os.path.isdir(path):", but add your following checks.

luomiao:
And sorry I just noticed here: [line 751](https://github.com/vmware/docker-volume-vsphere/pull/1387/files/9f7d2a255729cab3d4da81d4fb7d183b67de7e30#diff-2658c43b7daffd5e1ec9419a511da1d6R751)
I guess you may want to add the return of readable_path here too (with the checks).

And btw for [line 734](https://github.com/vmware/docker-volume-vsphere/pull/1387/files/9f7d2a255729cab3d4da81d4fb7d183b67de7e30#diff-2658c43b7daffd5e1ec9419a511da1d6R734), it looks like the second half "not os.path.isdir(path)" check is not necessary.

shivanshu21:
@luomiao We can leave line 734 for abundance of caution. The original writer might have put it for some corner case. Fixed line 751.

luomiao:
LGTM but need Mark's final call :)

esx_service/vmdk_ops.py
```diff
@@ -690,26 +690,36 @@ def detachVMDK(vmdk_path, vm_uuid):
...
     if os.path.isdir(path):
-        # If the path exists then return it as is.
-        logging.debug("Found %s, returning", path)
-        return path, None
+        # If the readable_path exists then return, else return path with no symlinks
+        if os.path.exists(readable_path):
+            logging.debug("Found %s, returning", readable_path)
+            return readable_path, None
+        else:
+            logging.debug("Tenant name symlink not found for path %s", readable_path)
```
msterin: this should be logging.warning("internal: Tenant name symlink not found for path %s. Using UUID path instead")  - it is not a normal situation

pshahzeb:
Merging this. Comments have been addressed by @shivanshu21 .

Issue #1371, opened 2017-06-08
Metadata load error from volume KV during access update for the volume.


pshahzeb:
While updating the volume access from read-write (default) to read-only, through automated volume access test, the update fails.
The access is updated using admin cli.

The exact log for the failure are

```
06/08/17 00:42:37 73825 [MainThread] [ERROR  ] Failed to decode meta-data for /vmfs/volumes/datastore1/dockvols/11111111-1111-1111-1111-111111111111/vol_access_volume_1496882546.vmdk
Traceback (most recent call last):
  File "/usr/lib/vmware/vmdkops/Python/kvESX.py", line 280, in load
    return json.loads(kv_str)
  File "/build/mts/release/bora-4564106/bora/build/esx/release/vmvisor/sys-boot/lib64/python3.5/json/__init__.py", line 319, in loads
  File "/build/mts/release/bora-4564106/bora/build/esx/release/vmvisor/sys-boot/lib64/python3.5/json/decoder.py", line 339, in decode
  File "/build/mts/release/bora-4564106/bora/build/esx/release/vmvisor/sys-boot/lib64/python3.5/json/decoder.py", line 357, in raw_decode
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

pshahzeb:
The datastore1 type is local.
Can you please confirm @shuklanirdesh82

shuklanirdesh82:
Yeah it is local VMFS data store.

shuklanirdesh82:
https://ci.vmware.run/vmware/docker-volume-vsphere/411

changing readOnly volume to writable volume also fails intermittently.

```
2017/06/08 06:31:32 START:vol_access TestAccessUpdate_R_RW Thu Jun  8 06:31:32 UTC 2017
2017/06/08 06:31:32 Creating volume [vol_access_volume_1496903492@datastore1] with options [ -o access=read-only] on VM [192.168.31.81]
2017/06/08 06:31:37 Writing message_by_host1 to file test.txt on volume [vol_access_volume_1496903492@datastore1] from VM[192.168.31.81]
2017/06/08 06:31:38 Failed to invoke command [docker run  -v vol_access_volume_1496903492@datastore1:/vol1 --name vol_access_container_1496903497 busybox  /bin/sh -c 'echo "message_by_host1" > /vol1/test.txt']: exit status 1
[...]

PR #1388, opened 2017-06-11
Add log to dump kv_str

lipingxue:
This PR includes:
1. add log to dump kv_str to help reproduce #1371 
2. set "policyList" to empty in SetupTest for VsanTestSuite, which is the left over from my previous merged PR.

### DIFF
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -277,6 +277,8 @@ def load(volpath):
                 return None
 
     try:
+        # Adding this log for DEBUG
+        logging.warning("kv_str from meta file is %s", kv_str)
         return json.loads(kv_str)
     except ValueError:
         logging.exception("Failed to decode meta-data for %s", volpath)
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -278,7 +278,7 @@ def load(volpath):
 
     try:
         # Adding this log for DEBUG
-        logging.warning("kv_str from meta file is %s", kv_str)
+        logging.warning("kv_str from meta file is %s ", kv_str)
         return json.loads(kv_str)
     except ValueError:
         logging.exception("Failed to decode meta-data for %s", volpath)
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -277,10 +277,10 @@ def load(volpath):
                 return None
 
     try:
-        # Adding this log for DEBUG
-        logging.warning("kv_str from meta file is %s ", kv_str)
         return json.loads(kv_str)
     except ValueError:
+        # Adding this log for DEBUG
+        logging.warning("kv_str from meta file is %s ", kv_str)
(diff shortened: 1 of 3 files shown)


tests/e2e/volume_access_test.go
```diff
@@ -15,7 +15,7 @@
 // The goal of this test suite is to verify read/write consistency on volumes
 // in accordance with the access updates on the volume
 
-// +build runonce
+// +build runalways
```
govint: this is the default and hence need not be specified.

esx_service/utils/kvESX.py
```diff
@@ -277,6 +277,8 @@ def load(volpath):
                 return None
 
     try:
+        # Adding this log for DEBUG
+        logging.warning("kv_str from meta file is %s ", kv_str)
```
govint: Should print the json formatted text instead?

tests/e2e/volume_access_test.go
lipingxue: Specified explicitly is more clear, and I found "basic_test.go" also specify "runalways".

esx_service/utils/kvESX.py
lipingxue: I moved the log so that it only print when the "kv_str" is not a json file.

tests/e2e/vsan_test.go
```diff
@@ -57,6 +57,10 @@ func (s *VsanTestSuite) SetUpSuite(c *C) {
 	}
 }
 
+func (s *VsanTestSuite) SetUpTest(c *C) {
+	s.policyList = []string{}
```
pshahzeb: Minor: As this is cleaning the list, this step should go to teardown.
You can empty this slice as s.policyList = s.policyList[:0]

lipingxue: I put it in the setup is to make sure each test is running from a clean state. I think either way works. So I will keep it as it is.

tests/e2e/volume_access_test.go
shaominchen: Actually we do need to add "runalways" explicitly - @shuklanirdesh82 can you please explain the difference between default (no tag) and "runalways"?

shuklanirdesh82: There is a limitation of `go test -tags <tag_name>` command. There is no way to exclude any tag, as per `-tags` tests are picked up with passed value and run. Any test files not labeled with tag falls into default category and end up running multiple times hence tag is required.

Let me explain with following example. let's say we have 3 files a_test.go, b_test.go, c_test.go.
```
a_test.go => // +build runalways
b_test.go => // +build runonce
c_test.go => no label 
```

We have two targets `runalways` and `runonce` (https://github.com/vmware/docker-volume-vsphere/blob/master/vmdk_plugin/Makefile#L357).

targets `runalways`: includes and ends up running `a_test.go` & `c_test.go`
targets `runonce`: includes and ends up running `b_test.go` & `c_test.go`

`c_test.go` falls into default and ends up running multiple times hence the tag explicitly needed.

PR #1390, opened 2017-06-12
Fix tenant_add_vm

govint:
Minor error in return from _tenant_vm_add when checking for attached volumes to the VM being added to the vmgroup.

The code is using two functions to do the same work in auth_api.py - _tenant_vm_add() and _tenant_vm_replace() both are about the same in functionality - the former is moving the VM from default to a named tenant and the latter from a named tenant to another named tenant. These need to be refactored into a single function so code isn't duplicated.

### DIFF
--- a/esx_service/utils/auth_api.py
+++ b/esx_service/utils/auth_api.py
@@ -695,7 +695,7 @@ def _tenant_vm_add(name, vm_list):
     if error_info:
         error_info.msg = "Cannot add VM to vmgroup " + error_info.msg
         logging.error(error_info.msg)
-        return error_info, None
+        return error_info
 
     error_info, auth_mgr = get_auth_mgr_object()
 


govint:
CI failed with below error, 
FAIL: volumecreate_test.go:177: VolumeCreateTestSuite.TestValidOptions

volumecreate_test.go:82:
    // positive test case
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc420363140), Stderr:[]uint8(nil)} ("exit status 1")
... Error response from daemon: create option_volume_n9ocblmgyv6t: Post http://%!F(MISSING)run%!F(MISSING)docker%!F(MISSING)plugins%!F(MISSING)334cbf101410e47e1360d8793088a0e9b8fe25e263fe87cbe190729a8020203c%!F(MISSING)vsphere.sock/VolumeDriver.Create: http: ContentLength=69 with Body length 0

volumecreate_test.go:82:
    // positive test case
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc42045d360), Stderr:[]uint8(nil)} ("exit status 1")
... Error response from daemon: create option_volume_1x211hubfnfyi: Post http://%!F(MISSING)run%!F(MISSING)docker%!F(MISSING)plugins%!F(MISSING)334cbf101410e47e1360d8793088a0e9b8fe25e263fe87cbe190729a8020203c%!F(MISSING)vsphere.sock/VolumeDriver.Create: http: ContentLength=80 with Body length 0

govint:
Thanks for the review @lipingxue @shaominchen

PR #1392, opened 2017-06-12
Addition

nnf97:
Installing VIB on ESXi 5.5 generates the error: VIB VMWare_bootbank_esx-vmdkops-service_0.14.0577889-0.0.1 requires esx-version >= 6.0.0

However, this version requirement isnt documented. Just like he Docker version is called out, it would be beneficial to call out that the ESXi version should be 6.0 or higher.


pdhamdhere:
Good catch. Looks like [README](https://github.com/vmware/docker-volume-vsphere/blob/master/README.md) calls out ESX versions but documentation doesn't. 

@shuklanirdesh82 Should ESX and Docker version be part of [prerequisite.md](https://github.com/vmware/docker-volume-vsphere/edit/gh-pages/jekyll-docs//prerequisite.md)

shuklanirdesh82:
> Should ESX and Docker version be part of prerequisite.md

Yeah, I will add such information. Thanks @nnf97 for catching!

nnf97:
Right. I saw that on the readme. However, it phrases it as 'tested on', which to my mind means, it is possible that this thing will work on other versions, but no one had the time or resources to actually test it on the other versions.

The error message clearly says it should be ESXi 6.0 as the  lowest version.

I will update the readme as well to indicate that 6.0 and up is a 'requirement', if you don't mind.

On Jun 12, 2017, at 5:28 PM, Prashant Dhamdhere <notifications@github.com<mailto:notifications@github.com>> wrote:


Good catch. Looks like README<https://github.com/vmware/docker-volume-vsphere/blob/master/README.md> calls out ESX versions but documentation doesn't.

@shuklanirdesh82<https://github.com/shuklanirdesh82> Should ESX and Docker version be part of prerequisite.md<https://github.com/vmware/docker-volume-vsphere/edit/gh-pages/jekyll-docs//prerequisite.md>

pdhamdhere:
@nnf97 Correct. Anything below ESXi 6.0 is not supported. 

>I will update the readme as well to indicate that 6.0 and up is a 'requirement', if you don't mind.

Sure thing and thanks in advance,

PR #1393, opened 2017-06-12
Revert "Fix e2e utils for generating unique numbers"

shuklanirdesh82:
Reverts vmware/docker-volume-vsphere#1385


shaominchen:
Why are we reverting this? Any regression have been found? Please add some explanation in the PR description.

shuklanirdesh82:
master was failing with the last merge continuously so kicking off the test run with revert change.

Haven't seen such failures in past 5-6 runs so closing this as invalid.

PR #1404, opened 2017-06-13
Silencing e2e test ssh util

### LINKED ISSUE #1389: Remove failure logs thats cluttering the e2e test logs.
The below logs make it look like the e2e tests have errors, these need to be removed till we have log levels for the tests:

2017/06/11 18:33:34 Failed to invoke command [docker volume  create --driver=vsphere  --name= 8F2qNfHK5a84jjJkwzDkh9h2fhfUVuS9jZ8uVbhV3vC5AWX39IVUWSP2NcHciWvqZTa2N95RxRTZHWUsaD6HEdz0ThbXfQ6pYSQ3n]: exit status 1

2017/06/11 18:33:35 Failed to invoke command [docker volume  create --driver=vsphere  --name= Volume_volume_4swm71tvdge@invalidDatastore]: exit status 1

2017/06/11 18:33:35 Failed to invoke command [docker volume  create --driver=vsphere  --name= Volume-000000]: exit status 1

pshahzeb:
ssh util to execute a command remotely yells a print  "Failed to invoke command" in case the command failed. But the command is expected to fail in some negative test cases. Such prints become misleading.
removing the print, since all tests handle error prints in test assertion comments (assert Commentf).

Testing: Done locally. e2e-test passes.
Fixes #1389

### DIFF
--- a/tests/utils/ssh/ssh.go
+++ b/tests/utils/ssh/ssh.go
@@ -39,9 +39,6 @@ func InvokeCommand(ip, cmd string) (string, error) {
 	sshIdentity := []string{sshKeyOpt[0], sshKeyOpt[1], "-q", "-kTax", "-o StrictHostKeyChecking=no"}
 
 	out, err := exec.Command("/usr/bin/ssh", append(sshIdentity, "root@"+ip, cmd)...).CombinedOutput()
-	if err != nil {
-		log.Printf("Failed to invoke command [%s]: %v", cmd, err)
-	}
 	return strings.TrimSpace(string(out[:])), err
 }
 


pshahzeb:
L50 check does a log.fatal if check succeeds. Since InvokeCommandLocally doesn't return error, not sure if we must silence it as of now.

shaominchen:
@shuklanirdesh82 @govint I remember Nirdesh had some comments before about removing the error logs in the test cases (or verification utils) because the ssh util itself has already logged the errors. Now we are removing this log from ssh util, so we will be missing some important logs in those test cases, which will make debugging more difficult. For example, without this log, I can't debug this issue: https://github.com/vmware/docker-volume-vsphere/issues/1405

I understand that printing error messages in (successful) test output is a little bit annoying, but personally I'm not really bothered with those messages. If everybody feel it's quite annoying, I'm fine to silent it, but then please make sure to write sufficient debugging logs in your test cases.

PR #1410, opened 2017-06-13
Adding more logs when reading/writing meta file

lipingxue:
This PR added more logs when reading/writing meta file to triage #1371 . Please review it.

### DIFF
--- a/esx_service/utils/kvESX.py
+++ b/esx_service/utils/kvESX.py
@@ -28,6 +28,7 @@ import time
 import threadutils
 import vmdk_utils
 import os
+import traceback
 
 # Python version 3.5.1
 PYTHON64_VERSION = 50659824
@@ -259,9 +260,6 @@ def load(volpath):
     logging.info("Load dictionary from sidecar for vol path %s", volpath)
     meta_file = lib.DiskLib_SidecarMakeFileName(volpath.encode(),
                                                 DVOL_KEY.encode())
-    if os.stat(meta_file).st_size == 0:
-        logging.warning("load entry: meta file %s is empty", meta_file)
-
     retry_count = 0
     vol_name = vmdk_utils.get_volname_from_vmdk_path(volpath)
     while True:
@@ -279,6 +277,11 @@ def load(volpath):
             else:
                 logging.exception("load:Failed to access %s", meta_file)
                 return None
+        except:
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
+            logging.exception('load: Unexpected exception:'.join('!! ' + line for line in lines))
+            return None
 
     if os.stat(meta_file).st_size == 0:
         logging.warning("load exit: meta file %s is empty", meta_file)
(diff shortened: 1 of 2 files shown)


esx_service/utils/kvESX.py
```diff
@@ -256,32 +257,42 @@ def load(volpath):
     """
     Load and return dictionary from the sidecar
     """
+    logging.info("Load dictionary from sidecar for vol path %s", volpath)
```
msterin: it feels like a debug, not info. Can you elaborate why do we want it on every log all the time ?

esx_service/utils/kvESX.py
```diff
@@ -256,32 +257,42 @@ def load(volpath):
...
+                logging.exception("load:Failed to access %s", meta_file)
                 return None
 
     try:
         return json.loads(kv_str)
     except ValueError:
         # Adding this log for DEBUG
-        logging.warning("kv_str from meta file is %s ", kv_str)
-        logging.exception("Failed to decode meta-data for %s", volpath)
+        logging.warning("load:Dump kv_str from meta file")
```
msterin: Do we really need 3 lines in log for that ? Can it be compressed into 1 ?

esx_service/utils/kvESX.py
lipingxue: Mark, the reason I put it as info is the failure cannot hit every time, that is why I want to merge it to master so that every CI run will able to log the enough information that I need. I have more chance to reproduce this issue with needed log.

msterin: so this is for a bug hunting ? Then there should be an issue to roll it back when the bug is found. 

I do not think we should print that much information which really has no value outside of catching this bug.

esx_service/utils/kvESX.py
```diff
@@ -298,7 +310,15 @@ def save(volpath, kv_dict):
     vol_name = vmdk_utils.get_volname_from_vmdk_path(volpath)
     while True:
         try:
+            if os.stat(meta_file).st_size > 0:
```
pshahzeb: I think this block of logging is redundant.



Agreed that this logs are temp and would be removed once we hit the root cause of the issue about the volume metadata  being unavailable(either corrupted or empty); do you mind wrapping this lines into a function and call it twice. The only reason being this change is being merged into master.

esx_service/vmdk_ops.py
```diff
@@ -1474,7 +1474,7 @@ def set_vol_opts(name, tenant_name, options):
 
     vmdk_path = vmdk_utils.get_vmdk_path(path, vol_name)
 
-    logging.debug("set_vol_opts: path=%s vmdk_path=%s", path, vmdk_path)
+    logging.info("set_vol_opts: path=%s vmdk_path=%s", path, vmdk_path)
```
pshahzeb: Essentially  this log (and the related ones) are meant to be debug itself.

Do you mind registering the issue to revert this change ?(either due to test issue or product issue)

pshahzeb:
Contingent to CI pass and no merge conflicts (not expecting any though)

shuklanirdesh82:
CI failure: https://ci.vmware.run/vmware/docker-volume-vsphere/548

@lipingxue  tests from test-esx are failing.

esx_service/utils/kvESX.py
lipingxue: #1419 has been filed to track removing the log added in this PR after the root cause of #1371 is found.

PR #1414, opened 2017-06-14
Remove retries in delete volume.

govint:
Removed the retries in DeleteVolume(), callers need to retry if the use case is expecting delays.

### DIFF
--- a/tests/utils/dockercli/volumelifecycle.go
+++ b/tests/utils/dockercli/volumelifecycle.go
@@ -90,21 +90,8 @@ func ReadFromVolume(ip, volName, containerName, fileName string) (string, error)
 
 // DeleteVolume helper deletes the created volume as per passed volume name.
 func DeleteVolume(ip, name string) (string, error) {
-	log.Printf("Destroying volume [%s]\n", name)
-	var out string
-	var err error
-
-	for attempt := 0; attempt < maxRemoveVolAttempt; attempt++ {
-		out, err = ssh.InvokeCommand(ip, dockercli.RemoveVolume+name)
-		if err != nil && strings.Contains(out, pluginInitError) {
-			misc.SleepForSec(waitTime)
-			log.Printf("Volume cannot be deleted yet as plugin initialization still in progress. Retrying...")
-			continue
-		} else {
-			break
-		}
-	}
-	return out, err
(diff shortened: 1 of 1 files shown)


tests/utils/dockercli/volumelifecycle.go
```diff
@@ -90,21 +90,8 @@ func ReadFromVolume(ip, volName, containerName, fileName string) (string, error)
...
-		if err != nil && strings.Contains(out, pluginInitError) {
-			misc.SleepForSec(waitTime)
-			log.Printf("Volume cannot be deleted yet as plugin initialization still in progress. Retrying...")
-			continue
-		} else {
-			break
-		}
-	}
-	return out, err
+	log.Printf("Deleting volume [%s]\n", name)
```
pshahzeb: I think this retrial was added due to a design a design decision that volume cannot be deleted in phase where plugin initialization is in progress.

//CC @ashahi1  @shuklanirdesh82 


Since there isn't a specific reason in PR description as to why this change is needed.. can you point to the necessary reason?

If I am not wrong, did you imply test case should handle volume deletion(or prior step to it) depending on test scenario and retry again. 

As far as I know, doing this change will intermittently fail VM listener test case.

govint: Yes, callers of this function should be doing the retry, depending on whether a retry must happen or not. If deletion fails then most likely its going to fail multiple times in the retries. Saw this in the CI when debugging the missing KV file error.

pshahzeb: In that case, I think we should modify the signature of the util to specify if retry should be made or not.
what do you think?

pshahzeb:
//CC @shaominchen @shuklanirdesh82 @ashahi1

tests/utils/dockercli/volumelifecycle.go
govint: May be that will prevent the existing code from having to change to accommodate retries. My only concern is when looking at logs and the test code we see multiple delete attempts and thats not in the test code. The util code should typically not have to do anymore than the basic op requested.

govint: Let me take this PR out for now, discuss offline and then figure if needed. Too many CI failures to keep looking at.

PR #1423, opened 2017-06-15
Revert "Making volumecreate and volumeaccess test as unstable to improve CI stability"

ashahi1:
Reverts vmware/docker-volume-vsphere#1420


shaominchen:
@ashahi1 Please make sure at least passing e2e tests locally. I think the issue with #1420 is that the ErrorVolumeCreate const defined in volumecreate_test.go is being used by vsan_test.go.

PR #1425, opened 2017-06-15
Making volume create and access tests unstable

ashahi1:
Minor fix to paste test error const in common place.

Tested locally.
```
target e2e-test-runalways :
OK: 6 passed
--- PASS: Test (207.20s)
PASS

target e2e-test-runonce:
--- PASS: Test (555.88s)
PASS
OK: 16 passed
```


shuklanirdesh82:
> Tested locally.

Good but not informative, it is good to add some short of test run output. How was the change tested?

PR #1427, opened 2017-06-15
[Trivial] fixing a bug to grab photon vm syslog incase of test failure

shuklanirdesh82:
`dump_log.sh` fails to collect syslog incase of test failure on photon vm.

This PR is going to fix the above problem and make sure not generating empty syslog while truncating the logs.

/CC @pdhamdhere @govint


shuklanirdesh82:
> https://github.com/vmware/docker-volume-vsphere/pull/1427#pullrequestreview-44201438

Thanks @msterin! I will sync up with you offline to more idea on the above comment.

shuklanirdesh82:
> https://github.com/vmware/docker-volume-vsphere/pull/1427#pullrequestreview-44198104

It was a regression at TOT. Photon VMs are not having syslog (we need to capture using `journalctl`) and following line (log truncation) is creating an empty syslog files.

`$SSH $USER@$1 "echo > $SYS_LOGFILE"`

Ideal way could be to find out VM type (photon vs ubuntu) and invoke the command to capture syslog (not sure whether it is more time consuming to find out OS type)

pdhamdhere:
Got it. LGTM.

Issue #1331, opened 2017-06-01
Automate Misc Test case: Restart docker service


lipingxue:
Detail in issue #1010

shuklanirdesh82:
fixed by #1428

PR #1287, opened 2017-05-23
[Automation] Add test for volume detach verification after plugin kill

pshahzeb:
1. Add test for volume detach verification after plugin kill
2. Moved common part of volume create to test fixture
3. Introduced a couple of utils

Testing: Done locally. test-all target passes
```
2017/05/23 16:54:56 START: restart_test.TestPluginKill
2017/05/23 16:54:56 Attaching volume [restart_test_volume_1495558494] on VM[10.192.33.29]
2017/05/23 16:54:57 Confirming attached status for volume [restart_test_volume_1495558494]
2017/05/23 16:55:00 Killing vDVS plugin on VM [10.192.33.29]
2017/05/23 16:55:03 Stopping container [restart_test_container_1495558494] on VM [10.192.33.29]
2017/05/23 16:55:14 Starting container [restart_test_container_1495558494] on VM [10.192.33.29]
2017/05/23 16:55:16 Stopping container [restart_test_container_1495558494] on VM [10.192.33.29]
2017/05/23 16:55:28 Confirming detached status for volume [restart_test_volume_1495558494]
2017/05/23 16:55:30 END: restart_test.TestPluginKill
```

Issue #1252, opened 2017-05-16
[3 tests] Automate refcount P0 testcases


shuklanirdesh82:
- [ ] Testcase 1: Test stale mounts 
Steps:
```
    a. Start container without restart flag
    b. kill docker and wait for docker to start
    c. Check if disk is detached
```

- [ ] Testcase 2: Kill Plugin
```
Kill Plugin
   a. Start container with restart flag
   b. kill plugin
   c. stop container
   d. start container (should succeed)
   e. stop container (should succeed & volume should be detached.
```

- [ ] Testcase 3: kill docker host vm
```
kill docker host vm

1. create volume
2. mount more than one container with restart policy `--restart=always` (let's mount 5 containers)
3. perform verification through docker volume inspect and vmdkops_admin volume ls to check for the attached status
4. kill docker host vm
5. start docker host vm
6. stop all containers
7. verifies the volume is in detached state and containers are not in running mode
8. remove volume and make sure volume rm works fine

```

Note: New change for 0.14
 
reference count rescan
With --restart=always flag containers are restarted even after  docker host restart.
Verify tests 1 to 4 with and without this flag. If this flag is used, containers are restarted and disk will be attached again. If using this flag, explicitly stop the containers and then verify if disk has been detached.
If this flag is not used, then containers don’t automatically restart and volume remains detached. This is docker behaviour

pshahzeb:
Test 3 doesn't hit refcount logic. It is because of step 4, which will trigger VM listener event and disk would be detached after step 4 itself. This gets tested through #1256 .

pshahzeb:
Automated through #1265 and #1287

Issue #1010, opened 2017-03-08
[test automation] Miscellaneous driver tests


ashahi1:
Test -1: Test to delete the policy which is applied to a volume

```
- Create a vsan policy P1
- Do vmdkops_admin policy ls to verify policy fields.
- Create a volume using the policy P1.
- Again do vmdkops_admin policy ls to check policy fields
- Try deleting the policy - deletion should fail as is volume is still using the policy

```

Test - 2: Test to update a vsan policy which doesn't exist
```
- Try updating a vsan policy which doesn't exist

```

Test - 3: Verifying ```docker volume ls and admin cli ls``` after restarting the hostd
  ```
- Create a volume
- Do docker volume ls and admin cli volume ls
- Restart hostd from esx
- Again verify output of docker volume ls and admin cli ls
```

Test - 4: Restart docker service
```
- Create a volume - vol1
- Do docker volume ls and admin cli volume ls
- Run the container with the volume mounted - vol1
- Restart docker service (systemctl docker restart)
- Run another container and mount the same volume - vol1
- Verify volume should get successfully mounted
```

Test - 5: Create a volume using vsan policy on non-vsan datastore

```
- Create a valid vsan policy P1
- Do vmdkops_admin policy ls to verify policy fields.
- Create a volume using the policy P1 on a non-vsan datastore (vm should be on vmfs datastore).
- Volume creation operation should fail.
```

**Comment by shuklanirdesh82** [comment]
@ashahi1 

> Test - 4: Verify 'docker volume ls' works after migration of vm

As discussed offline we need to add some verification after migrating VM (```Migrate the vm to different esx```)

**Comment by ashahi1** [comment]
Test -1 : Test to delete the policy which is applied to a volume
---

- Created a vsan policy

```
[root@sc-rdops-vm18-dhcp-32-67:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy create --name some-policy --content '(("proportionalCapacity" i50)''("hostFailuresToTolerate" i0))'
Successfully created policy: some-policy
[root@sc-rdops-vm18-dhcp-32-67:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py policy ls
Policy Name  Policy Content                                               Active
-----------  -----------------------------------------------------------  ------
some-policy  (("proportionalCapacity" i50)("hostFailuresToTolerate" i0))  Unused

[root@sc-rdops-vm18-dhcp-32-67:~]

```

[...]

PR #1428, opened 2017-06-15
Added tests for restarting docker service.

govint:
Enhanced restart tests to verify volume is available after docker restart with containers running.

### DIFF
--- a/tests/constants/dockercli/cmd.go
+++ b/tests/constants/dockercli/cmd.go
@@ -37,8 +37,11 @@ const (
 	// KillDocker kill docker
 	KillDocker = "pkill -9 dockerd "
 
-	// StartDocker - manually start docker
-	StartDocker = "systemctl start docker"
+	// RestartDockerWithSystemd - restart docker with systemctl
+	RestartDockerWithSystemd = "systemctl restart docker"
+
+	// RestartDockerService - restart docker service
+	RestartDockerService = "service docker restart"
 
 	// VDVSPluginName name of vDVS plugin
 	VDVSPluginName = "vsphere "
--- a/tests/e2e/restart_test.go
+++ b/tests/e2e/restart_test.go
@@ -1,4 +1,4 @@
-// Copyright 2017 VMware, Inc. All Rights Reserved.
+// Copyright 2017 VMware, Inc. All Rights Reserved. 
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,67 +20,100 @@
 package e2e
 
 import (
-	"log"
-	"os"
-
 	. "gopkg.in/check.v1"
 
 	"github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
 	"github.com/vmware/docker-volume-vsphere/tests/utils/inputparams"
 	"github.com/vmware/docker-volume-vsphere/tests/utils/verification"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
 )
 
-type PluginSuite struct {
+type RestartTestData struct {
+	config        *inputparams.TestConfig
 	volumeName    string
-	hostIP        string
 	containerName string
-	esxIP         string
 }
 
-func (s *PluginSuite) SetUpTest(c *C) {
-	s.hostIP = os.Getenv("VM2")
-	s.esxIP = os.Getenv("ESX")
+func (s *RestartTestData) SetUpSuite(c *C) {
+	s.config = inputparams.GetTestConfig()
+	if s.config == nil {
+		c.Skip("Unable to retrieve test config, skipping docker/plugin restart tests")
+	}
+
 	s.volumeName = inputparams.GetUniqueVolumeName("restart_test")
 	s.containerName = inputparams.GetUniqueContainerName("restart_test")
 
+	// ensure there are no remanants from an earlier run
+	out, err := dockercli.RemoveContainer(s.config.DockerHosts[1], s.containerName)
+
 	// Create a volume
-	out, err := dockercli.CreateVolume(s.hostIP, s.volumeName)
+	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName)
 	c.Assert(err, IsNil, Commentf(out))
 }
 
-func (s *PluginSuite) TearDownTest(c *C) {
-	out, err := dockercli.RemoveContainer(s.hostIP, s.containerName)
-	c.Assert(err, IsNil, Commentf(out))
-	out, err = dockercli.DeleteVolume(s.hostIP, s.volumeName)
+func (s *RestartTestData) TearDownSuite(c *C) {
+	out, err := dockercli.DeleteVolume(s.config.DockerHosts[1], s.volumeName)
 	c.Assert(err, IsNil, Commentf(out))
 }
 
-var _ = Suite(&PluginSuite{})
+var _ = Suite(&RestartTestData{})
 
-// Test Stale mount
-// 1. Attach a volume to a container
+// TestVolumeDetached - verifies a volume is detached when the docker
+// daemon is restarted.
+// 1. Attach a volume to a container on a host
 // 2. Verify attached status
 // 3. Restart docker
 // 4. Verify detached status. Volume should be detached (within the timeout)
-func (s *PluginSuite) TestVolumeStaleMount(c *C) {
-	log.Printf("START: restart_test.TestVolumeStaleMount")
+// 5. Verify a container can be started to use the same volume on another host
+// 6. Restart the docker daemon on the other host
+// 7. Verify detached status for the volume
+// 8. Verify a container can be started to use the same volume on the original host
+func (s *RestartTestData) TestVolumeDetached(c *C) {
+	misc.LogTestStart(c.TestName())
 
-	out, err := dockercli.AttachVolume(s.hostIP, s.volumeName, s.containerName)
+	// 1. Attach a volume to a container on a host
+	out, err := dockercli.AttachVolume(s.config.DockerHosts[1], s.volumeName, s.containerName)
 	c.Assert(err, IsNil, Commentf(out))
 
-	status := verification.VerifyAttachedStatus(s.volumeName, s.hostIP, s.esxIP)
+	// 2. Verify attached status
+	status := verification.VerifyAttachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
 
-	out, err = dockercli.KillDocker(s.hostIP)
+	// 3. Restart docker
+	out, err = dockercli.RestartDocker(s.config.DockerHosts[1])
 	c.Assert(err, IsNil, Commentf(out))
 
-	status = verification.VerifyDetachedStatus(s.volumeName, s.hostIP, s.esxIP)
+	// 4. Verify detached status. Volume should be detached (within the timeout)
+	status = verification.VerifyDetachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
 
-	log.Printf("END: restart_test.TestVolumeStaleMount")
+	out, err = dockercli.RemoveContainer(s.config.DockerHosts[1], s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// 5. Verify a container can be started to use the same volume on another host
+	out, err = dockercli.AttachVolume(s.config.DockerHosts[0], s.volumeName, s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// 6. Restart the docker daemon on the other host
+	out, err = dockercli.RestartDocker(s.config.DockerHosts[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Clean up the container on host0
+	out, err = dockercli.RemoveContainer(s.config.DockerHosts[0], s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// 7. Verify detached status for the volume
+	status = verification.VerifyDetachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
+
+	// 8. Verify a container can be started to use the same volume on the original host
+	out, err = dockercli.ExecContainer(s.config.DockerHosts[1], s.volumeName, s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	misc.LogTestEnd(c.TestName())
 }
 
-// Test vDVS plugin restart
+// TestPluginKill - Test vDVS plugin kill
 // 1. Attach a volume with restart=always flag
 // 2. Verify volume attached status
 // 3. Kill vDVS plugin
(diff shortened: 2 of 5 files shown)


govint:
Fixes #1331

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,130 @@
...
-	esxIP         string
 }
 
-func (s *PluginSuite) SetUpTest(c *C) {
-	s.hostIP = os.Getenv("VM2")
-	s.esxIP = os.Getenv("ESX")
+func (s *RestartTestData) SetUpSuite(c *C) {
+	s.config = inputparams.GetTestConfig()
+	if s.config == nil {
+		c.Skip("Unable to retrieve test config, skipping vmgroup tests")
```
pshahzeb: i think you meant restart test.

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,130 @@
...
+	}
+
 	s.volumeName = inputparams.GetUniqueVolumeName("restart_test")
 	s.containerName = inputparams.GetUniqueContainerName("restart_test")
 
+	// ensure there are no remanants from an earlier run
+	out, err := dockercli.RemoveContainer(s.config.DockerHosts[1], s.containerName)
```
pshahzeb: I still feel this should have been part of teardown since it's a cleanup step.

tests/e2e/restart_test.go
shuklanirdesh82: nit: skipping restartingDocker tests?

tests/e2e/restart_test.go
shuklanirdesh82: does it mean L42 doesn't give unique container name?

govint: Thats what was in the original code, but the stop container doesn't remove the container so creating a new one fails with "there is a container with the same name already" and test fails. Since kill is being done for  the docker and plugin the containers get stopped not removed.

govint: Thats done once for the suite vs. per test.

tests/e2e/restart_test.go
govint: Yup, cut paste from vmgroup

govint:
Since this is but a step in an existing test (renamed) I've not added a new test.

shuklanirdesh82:
Thanks @pshahzeb !

@govint I have confusion about the steps mentioned in the PR and the steps described in the issue #1331 which redirects to #1010.

Can you please clarify about the steps?

govint:
@shuklanirdesh82 these are the steps from #1010 ,

- Create a volume - vol1
- Do docker volume ls and admin cli volume ls
- Run the container with the volume mounted - vol1
- Restart docker service (systemctl docker restart)
- Run another container and mount the same volume - vol1
- Verify volume should get successfully mounted

the existing test did everything except (5) above.

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
+// daemon is stopped.
+// 1. Attach a volume to a container on a host
 // 2. Verify attached status
 // 3. Restart docker
 // 4. Verify detached status. Volume should be detached (within the timeout)
-func (s *PluginSuite) TestVolumeStaleMount(c *C) {
-	log.Printf("START: restart_test.TestVolumeStaleMount")
-
-	out, err := dockercli.AttachVolume(s.hostIP, s.volumeName, s.containerName)
+// 5. Verify a container can be started to use the same volume on another host
```
shuklanirdesh82: I am confused now .. by taking reference of https://github.com/vmware/docker-volume-vsphere/pull/1428#issuecomment-309191133 .. step#5-#8 doesn't match to #1010

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
+	// ensure there are no remanants from an earlier run
+	out, err := dockercli.RemoveContainer(s.config.DockerHosts[1], s.containerName)
+
 	// Create a volume
-	out, err := dockercli.CreateVolume(s.hostIP, s.volumeName)
+	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName)
 	c.Assert(err, IsNil, Commentf(out))
 }
 
-func (s *PluginSuite) TearDownTest(c *C) {
```
shuklanirdesh82: can you please share the reason behind using TearDownSuite and not TearDownTest?

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
 	c.Assert(err, IsNil, Commentf(out))
 
-	status := verification.VerifyAttachedStatus(s.volumeName, s.hostIP, s.esxIP)
+	// 2. Verify attached status
+	status := verification.VerifyAttachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
 
-	out, err = dockercli.KillDocker(s.hostIP)
+	// 3. Restart docker
+	out, err = dockercli.KillDocker(s.config.DockerHosts[1])
```
shuklanirdesh82: #3 says `Restart docker` .. any reason for invoking `dockercli.KillDocker` ?

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
+	out, err = dockercli.RemoveContainer(s.config.DockerHosts[1], s.containerName)
 	c.Assert(err, IsNil, Commentf(out))
 
-	status = verification.VerifyDetachedStatus(s.volumeName, s.hostIP, s.esxIP)
+	// 5. Verify a container can be started to use the same volume on another host
+	out, err = dockercli.AttachVolume(s.config.DockerHosts[0], s.volumeName, s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// 6. Restart the docker daemon on the other host
+	out, err = dockercli.KillDocker(s.config.DockerHosts[0])
```
shuklanirdesh82: if the test asks for `docker restart` then we should use `systemctl restart docker` approach instead of `dockercli.KillDocker` which kills docker using `pidof docker`.

tests/e2e/restart_test.go
govint: The test verifies that the volume is usable.

Not sure these tests were reviewed earlier but as I'm writing these tests, I'm also reviewing to see what makes the test correct and complete. For example, verifying detached state (original code) isn't right, the KV may say its detached but the volume may be stuck in a VM.

The correct verification is "can the volume be used again" which is by running a container and can it be used across docker host VMs. If a volume is stuck in a VM, then attach will work for it. So the correct test would be:

1. Run a container,
2. Restart docker
3. Assert the volume is detached (its not a sure test of correctness of volume state).
4. Attach volume to a different docker host VM - this ensures that the volume is truly detached.
5. Restart the docker on the VM in (4)
6. Verify again that on the first docker host VM we are able to re-attach the volume  - bonus verification that the original docker host is able to attach the VM.

tests/e2e/restart_test.go
govint: this is the original code, was it reviewed before submission?

tests/e2e/restart_test.go
govint: It should be using service or systemctl as the case may be. I'll send that change out as a separate PR.

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
 	c.Assert(err, IsNil, Commentf(out))
 
-	status := verification.VerifyAttachedStatus(s.volumeName, s.hostIP, s.esxIP)
+	// 2. Verify volume attached status
+	status := verification.VerifyAttachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
 
-	out, err = dockercli.KillVDVSPlugin(s.hostIP)
+	// 3. Kill vDVS plugin
+	out, err = dockercli.KillVDVSPlugin(s.config.DockerHosts[1])
```
govint: Is this kill Ok, when instead the plugin should be "disabled" via docker plugin disable?

tests/e2e/restart_test.go
govint: This is how the code was originally and wanted to allow docker create the volume when the container runs. Seems like AttachVolume() and AttachVolumeWithRestart() do not specify the volume driver name and hence while one test will use the vsphere volume the second one will use  local volume. Good. I'll fix.

tests/e2e/restart_test.go
shuklanirdesh82: I think @pshahzeb may shed more idea on this step.

/CC @pshahzeb

pshahzeb: dockercli.KillDocker essentially kills docker. Docker daemon is expected to restart automatically. In case if it doesn't comeup automatically, systemctl is used  to bring docker up.
We can rename this util to RestartDocker if that avoids confusion.

tests/e2e/restart_test.go
pshahzeb: disable is graceful. We needed ungraceful kill.

tests/e2e/restart_test.go
pshahzeb: Yes it was reviewedas part of #1265 and #1287. The StaleMount test was originally written just to verify following scenario
as part of #1252
```
    a. Start container without restart flag
    b. kill docker and wait for docker to start
    c. Check if disk is detached
```

> Create a volume - vol1
Do docker volume ls and admin cli volume ls
Run the container with the volume mounted - vol1
Restart docker service (systemctl docker restart)
Run another container and mount the same volume - vol1
Verify volume should get successfully mounted

> the existing test did everything except (5) above.

I see the additional steps (good and bonus to have); but those come across as a different test scenario. I am fine with with either coming with an altogether separate test or modifying the existing one when the scenario warrants for a substantial overlap.

tests/constants/dockercli/cmd.go
```diff
@@ -37,8 +37,11 @@ const (
 	// KillDocker kill docker
 	KillDocker = "pkill -9 dockerd "
 
-	// StartDocker - manually start docker
-	StartDocker = "systemctl start docker"
+	// StartDockerWithSystemd - start docker with systemctl
```
shaominchen: Please update the comments accordingly.

tests/constants/dockercli/cmd.go
```diff
@@ -37,8 +37,11 @@ const (
...
+	RestartDockerWithSystemd = "systemctl restart docker"
+
+	// StopDockerWithSystemd - stop docker with systemctl
```
shaominchen: Same as above.

tests/e2e/restart_test.go
```diff
@@ -1,6 +1,4 @@
-// Copyright 2017 VMware, Inc. All Rights Reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Copyright 2017 VMware, Inc. All Rights Reserved.  // // Licensed under the Apache License, Version 2.0 (the "License");
```
shaominchen: Why changing the format?

tests/e2e/restart_test.go
shaominchen: I think it's better to get unique volume/container names for each test. We have done so in other test groups. This will reduce potential conflicts between different test cases.

shaominchen: @govint I don't follow either - the original code was removing (not stopping) container in teardown, why was that not working? I'd suggest to do cleanup in tearDown, and also use a new unique volume/container name per test. This way, we can avoid potential conflicts due to dirty data left over, and also it's easier to figure out which tests have left over dirty data.

tests/e2e/restart_test.go
govint: How does killing docker make a mount stale? Killing docker scenario seems unclear exactly what thats testing. If these are restart tests then restarting the docker/plugin is the correct way.

tests/e2e/restart_test.go
govint: Happens that docker will never allow a plugin to be disabled if there are volumes using it - even volumes that are just created and unused docker doesn't allow disabling the plugin.

tests/e2e/restart_test.go
govint: Mis-edited, will correct.

govint:
The setup and teardown are done in the setup suite and teardown suite functions. I don't see a reason to create a volume for each test. The test is cleaning up all items it creates. vmgroup tests are also doing this.

I think we don't have a specific preferred way to run the tests - cleanup after each test inside a group or after the test completes and don't see why one is preferred.

Again, cleaning up after each test inside a group also implies bugs can get covered up. For example, test updates access mode on a volume, then resets it, but ESX service has a bug that doesn't change the access mode. Test cleans up and next test works ok?

It may be individual style and thats ok, but its not a preferred approach to scrub the created entities off between tests and miss issues.

govint:
Why would we need a unique container for each test? Removing a container needs to work, else there is a bug in docker or in the test code. Getting a new container name for each test means we have left a container running in the system.

Removing the container and volumes may look like they succeed but they may not have because of bugs - and doing it in teardowntest or teardownsuite doesn't matter then.

tests/e2e/restart_test.go
pshahzeb: Correct.  Hence the kill signal to plugin for have forceful and ungraceful kill.

shaominchen:
@govint 
Yes, at first we were also reusing the same volume name and container name for the whole test group (i.e. a single x.go test file). However, later after running into some issues, we discussed and all agreed that we should not reuse the same volume/container for different test cases. The main reason is:
    
Each test case should be independent of another. In other words, one test case failure should not fail the entire test group.

It's not possible to guarantee that all test cases will always pass 100% in any case. In some cases one or two test cases might fail due to any reason: product issue, testbed issue, or test case issue, etc. If that happens, we want to make sure other test cases in the same test group should not be affected.

To answer your questions:
1. "Again, cleaning up after each test inside a group also implies bugs can get covered up..."
I don't follow this. Again, each test should be independent of another. If one test case fails, other test cases should continue to run independently without being affected. In the example as mentioned, if "ESX service has a bug that doesn't change the access mode", then the current test case should fail with appropriate assertion, instead of leaving over dirty data to fail the next test case.

2. "Getting a new container name for each test means we have left a container running in the system."
I don't follow here - the TearDownTest() should take care of cleaning the containers, why will the running container being left over? Even if the running containers are left over, the next test case won't be affected as it will create a different container. In the end after all tests are completed, we can figure out which test group/test case has left over dirty data, and we should definitely fix that.

tests/constants/dockercli/cmd.go
shuklanirdesh82: +1

tests/constants/dockercli/cmd.go
shuklanirdesh82: +1

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
 	c.Assert(err, IsNil, Commentf(out))
 }
 
-var _ = Suite(&PluginSuite{})
+var _ = Suite(&RestartTestData{})
 
-// Test Stale mount
-// 1. Attach a volume to a container
+// TestVolumeDetached - verifies a volume is detached when the docker
+// daemon is stopped.
```
shuklanirdesh82: nit:  `daemon is stopped.` => ` daemon is restarted.`

tests/e2e/restart_test.go
```diff
@@ -20,97 +18,141 @@
...
-	log.Printf("END: restart_test.TestVolumeStaleMount")
+	// 8. Verify a container can be started to use the same volume on the original host
+	out, err = dockercli.ExecContainer(s.config.DockerHosts[1], s.volumeName, s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	misc.LogTestEnd(c.TestName())
 }
 
-// Test vDVS plugin restart
+// TestPluginKill - Test vDVS plugin restart
```
shuklanirdesh82: nit: `test vDVS plugin restart` => `test vDVS plugin killed`

tests/utils/dockercli/docker_mgmt.go
```diff
@@ -0,0 +1,94 @@
...
+	"log"
+	"strings"
+
+	"github.com/vmware/docker-volume-vsphere/tests/constants/dockercli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
+)
+
+const (
+	dockerRestartCmd = "pidof systemd && " + dockercli.RestartDockerWithSystemd + " || " + dockercli.RestartDockerService
```
pshahzeb: Minor: a comment is a good to have here.

govint:
Its not possible to say which combination of actions lead to a bug and hence impossible to write tests to run specific scenarios.

Issue #1379, opened 2017-06-08
Automate fail-over test for docker host created on shared vmfs


shaominchen:
please refer to #957 Scenario 5.

shuklanirdesh82:
@shaominchen There should be some confusion about this issue's title .. please revisit and correct it. It says "Automate VSAN policy test - fail-over for docker host created on shared vmfs" I guess VSAN policy for shared vmfs either one should be correct .. do you mean `docker host created on VSAN`?

PR #1431, opened 2017-06-16
Add a failover e2e test

shaominchen:
Added a new failover e2e test and also refactored existing test cases.

Fixes issue #1379 

Testing Done:

=> Running target e2e-test-runalways Wed Jun 7 18:49:10 UTC 2017
--- PASS: Test (194.08s)
PASS
OK: 6 passed

=> Running target e2e-test-runonce Wed Jun 7 18:52:27 UTC 2017
--- PASS: Test (540.44s)
PASS
OK: 18 passed


tests/e2e/vmlistener_test.go
```diff
@@ -104,34 +116,85 @@ func (s *VMListenerTestParams) TestServiceRestart(c *C) {
...
 
-	// kill VM
-	isVMKilled := esxcli.KillVM(s.esxIP, processID)
-	c.Assert(isVMKilled, Equals, true, Commentf("Unable to kill VM %s ...", s.dockerHostName))
+// Test failover by killing a VM and then mount the volume to a different VM
+// 1. Setup: Create a volume
+// 2. Setup: Attach it to a container
+// 3. Kill the VM
+// 4. Verification: volume status should be detached
+// 5. Attach it to a different container
```
pshahzeb: I think it should be attach to a container on different VM.
Thats what it looks like from line 157

tests/e2e/vmlistener_test.go
```diff
@@ -104,34 +116,85 @@ func (s *VMListenerTestParams) TestServiceRestart(c *C) {
...
+	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
+
+	// Remove the container
+	out, err = dockercli.RemoveContainer(s.vm2, s.containerName)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Power on VM
+	govc.PowerOnVM(s.vm1Name)
+	isStatusChanged := misc.WaitForExpectedState(govc.GetVMPowerState, s.vm1Name, properties.PowerOnState)
+	c.Assert(isStatusChanged, Equals, true, Commentf("VM [%s] should be powered on state", s.vm1Name))
```
pshahzeb: This isn't part of original test steps but I think  it would be a good idea to add a check here (after L172) to ensure that status of the volume in detached. The reason being that container from VM2 is removed and VM1 has come up too.

tests/e2e/vmlistener_test.go
shaominchen: Yes, will update the comments.

tests/e2e/vmlistener_test.go
shaominchen: Will add.

shaominchen: With this change, I think the existing TestKillVM can be removed. What do you think?

pshahzeb: Agree that there is an overlap. I had a discussion with @shuklanirdesh82 in such cases. We should have it for now since we do not want to miss any bug in any of the possible permutation.

shaominchen: I don't think it's a permutation in the current case. Anyway, I will keep it for now.

PR #1433, opened 2017-06-16
Automate basic vsan test cases.

### LINKED ISSUE #1377: Automate VSAN policy test - basic
Please refer to #957 Scenario 3.

lipingxue:
Fixed #1377 

The fix is implemented by extending the existing vsan test.


tests/e2e/vsan_test.go
```diff
@@ -63,6 +64,10 @@ func (s *VsanTestSuite) SetUpTest(c *C) {
 }
 
 func (s *VsanTestSuite) TearDownTest(c *C) {
+	if len(s.containerList) > 0 {
+		out, err := dockercli.RemoveAllContainers(s.config.DockerHosts[0])
```
shuklanirdesh82: This is scary and stop us moving to run test in parallel. It is better to remove the container created by this test.

tests/e2e/vsan_test.go
```diff
@@ -77,12 +82,24 @@ func (s *VsanTestSuite) TearDownTest(c *C) {
 
 var _ = Suite(&VsanTestSuite{})
 
+func (s *VsanTestSuite) newCName() string {
```
shuklanirdesh82: method doc or improved method name would be useful to know about the intention behind this helper.

tests/e2e/vsan_test.go
```diff
@@ -77,12 +82,24 @@ func (s *VsanTestSuite) TearDownTest(c *C) {
...
+	s.containerList = append(s.containerList, cname)
+	return cname
+}
+
 // Steps:
 // 1. Create a valid vsan policy
 // 2. Volume creation with valid policy should pass
 // 3. Valid volume should be accessible
+// 4. Start a container to mount the volume, and write message to the volume
+// 5. Start another container to mount the volume, and read message from volume and verify the
```
shaominchen: I'm confused here: I don't see the test logic related to this step - start *another* container to mount the volume. Can you please clarify?

lipingxue: I mean function ```dockercli.WriteToVolume``` internally will create a container and mount the volume.

shaominchen: I still don't follow: the test logic shows that we are starting a container to mount the volume and write data to the volume, which is step 4. But where is the logic of "**start another container to mount the volume**" as described in step 5?

I guess step 5 should be just: "Read message from volume and verify the message..."

shaominchen: As discussed offline, we are not *attaching* the volume to a different VM here, but just to *mounting* it to a different container *in the same VM*. That makes sense. Sorry for the confusion.

tests/e2e/vsan_test.go
shaominchen: Looks like we are already using this in volumeproperty_test.go.

govint: Running test groups in parallel most likely will simply not work, given that all of these are operating off the default vmgroup and some are doing things like removing access privileges for the default vmgroup.

tests/e2e/vsan_test.go
lipingxue: Done

tests/e2e/vsan_test.go
lipingxue: OK. Will keep it as it is.

shuklanirdesh82: Running test parallel is an action item for the 0.16 sprint to achieve "Overall time taken by each test run will be ~30minutes." ... we should such `dockercli.RemoveAllContainers` that blocks us moving test run in parallel. 

I don't want to repeat again ... test container is a test item and its test responsibility to remove.

> Running test groups in parallel most likely will simply not work,

Which ever the test is not possible they will be identified and we may put into serial run bucket. 

> Looks like we are already using this in volumeproperty_test.go.

It was a miss from the review and by taking account of above example we should not use this approach. I discussed with @ashahi1  to remove usage of `dockercli.RemoveAllContainers` and remove only the containers created by test. 

/CC @ashahi1

shuklanirdesh82: @lipingxue By looking at your test code I don't see a need for cleaning up containers. Let's remove it.

tests/e2e/vsan_test.go
```diff
@@ -77,12 +82,25 @@ func (s *VsanTestSuite) TearDownTest(c *C) {
 
 var _ = Suite(&VsanTestSuite{})
 
+// This function is used to return a unique container name used for "vsan_test"
+func (s *VsanTestSuite) newCName() string {
```
shuklanirdesh82: this method can be removed and replaced by inline invocation.

lipingxue: Code is more clean with this helper function and I will keep it as it is.

tests/e2e/vsan_test.go
ashahi1: I will be removing ```dockercli.RemoveAllContainers``` function and will only remove the containers that test created.

tests/e2e/vsan_test.go
shuklanirdesh82: There is not much left in the method other than one line `inputparams.GetUniqueContainerName("vsan_test")` after syncing up with master code. Containers will be removed automatically hence no required of L88 `s.containerList = append(s.containerList, cname)` ... This method left out to be as inline function.

tests/e2e/vsan_test.go
```diff
@@ -41,9 +41,10 @@ import (
 type VsanTestSuite struct {
 	config *inputparams.TestConfig
 
-	vsanDSName string
-	volumeName string
-	policyList []string
+	vsanDSName    string
+	volumeName    string
+	policyList    []string
+	containerList []string
```
shuklanirdesh82: minor: this should be removed.

Issue #1449, opened 2017-06-21
Clone volume fails with "VM None does not belong to any vmgroup"


vxav:
When I try to clone a volume I always get an error "VM None does not belong to any vmgroup".
Of course my photon host is in a vmgroup.
I can make sure of it because if I create a docker volume with the vsphere driver on Photon1 (on ESX1), Photon0 (on ESX0) picks it up instantly.
Tried with a volume attached and with a detached one.

**Photon output**

`root@photon1 [ ~ ]# docker volume create --name own_Clone --driver=vsphere -o clone-from=own_own
Error response from daemon: create own_Clone: VolumeDriver.Create: VM None does not belong to any vmgroup
`

**/var/log/vmware/vmdk_ops.log:**

`06/21/17 07:36:19 68514 [MainThread] [INFO   ] Started new thread : 879477143296 with target <function execRequestThread at 0xccc2c17510> and args (11, 69292, b'{"cmd":"get","details":{"Name":"ClonedVol"},"version":"2"}')
06/21/17 07:36:19 68514 [Thread-71] [INFO   ] Checking DB mode for /etc/vmware/vmdkops/auth-db...
06/21/17 07:36:19 68514 [Thread-71] [INFO   ] Config DB found. maj_ver=1 min_ver=2 mode MultiNode
06/21/17 07:36:19 68514 [Thread-71] [INFO   ] cmd=get opts={} vmgroup=MYVMGROUP datastore_url=/vmfs/volumes/582b1324-e49d9a0e-9bc3-000c29c1501d is allowed to execute
06/21/17 07:36:19 68514 [Photon0-SHARED-Datastore.MYVMGROUP.ClonedVol] [INFO   ] executeRequest 'get' completed with ret={'Error': 'Volume ClonedVol not found (file: /vmfs/volumes/SHARED-Datastore/dockvols/0f2d53d7-091e-4b04-a6b0-1d2a43c454eb/ClonedVol.vmdk)'}
06/21/17 07:36:19 68514 [MainThread] [INFO   ] Started new thread : 879477143296 with target <function execRequestThread at 0xccc2c17510> and args (13, 69292, b'{"cmd":"create","details":{"Name":"ClonedVol","Opts":{"clone-from":"TestVol1"}},"version":"2"}')
06/21/17 07:36:19 68514 [Thread-72] [INFO   ] Checking DB mode for /etc/vmware/vmdkops/auth-db...
[...]

PR #1451, opened 2017-06-21
Pass vm_uuid to createVMDK

govint:
Pass the vm_uuid to createVMDK() and hence to cloneVMDK(), passing None causes clone op to fail if the VM isn't in the default vmgroup.

Added a test to the vmgroup test group to verify that a VM in a non-default vmgroup can create a volume and then clone it.


pdhamdhere:
Fixes #1449

Issue #1450, opened 2017-06-21
Need to add more verification steps for basic_test


shuklanirdesh82:
CI run: https://ci.vmware.run/vmware/docker-volume-vsphere/648

```
=== RUN   Test
2017/06/21 08:09:31 START: BasicTestSuite.TestBasicVolumeIsolation
2017/06/21 08:09:31 Creating volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] on VM [192.168.31.81]
2017/06/21 08:09:42 Checking volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] availability from VM [192.168.31.81]
2017/06/21 08:09:42 Checking volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] availability from VM [192.168.31.85]
2017/06/21 08:09:43 Destroying volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte]
2017/06/21 08:09:45 END: BasicTestSuite.TestBasicVolumeIsolation
2017/06/21 08:09:45 START: BasicTestSuite.TestVmGroupVolumeIsolation
2017/06/21 08:09:45 Initializing the SingleNode Config DB on esx [192.168.31.62] 
2017/06/21 08:09:47 Creating volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] on VM [192.168.31.81]
2017/06/21 08:09:55 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [192.168.31.81]
2017/06/21 08:09:56 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [192.168.31.85]
2017/06/21 08:09:56 Creating a vmgroup [T1] on esx [192.168.31.62]
2017/06/21 08:09:59 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [192.168.31.81]
2017/06/21 08:09:59 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [192.168.31.85]
[...]

Issue #1411, opened 2017-06-14
Admin CLI should return proper exit code when requested operation fails


shuklanirdesh82:
While running test locally basic_test fails with following error message.

```
2017/06/14 03:16:25 Finding VM name from IP Address [10.20.104.40]
2017/06/14 03:16:26 Finding VM name from IP Address [10.20.104.40]
2017/06/14 03:16:27 Finding Datastores available on ESX
2017/06/14 03:16:28 START: BasicTestSuite.TestBasicVolumeIsolation
2017/06/14 03:16:28 Creating volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] on VM [10.20.104.40]
2017/06/14 03:16:35 Checking volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] availability from VM [10.20.104.40]
2017/06/14 03:16:36 Checking volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte] availability from VM [10.20.104.40]
2017/06/14 03:16:37 Destroying volume [BasicTestSuite.TestBasicVolumeIsolation_volume_16ddfwutxgjte]
2017/06/14 03:16:39 END: BasicTestSuite.TestBasicVolumeIsolation
2017/06/14 03:16:39 START: BasicTestSuite.TestVmGroupVolumeIsolation
2017/06/14 03:16:39 Initializing the SingleNode Config DB on esx [10.20.105.97] 
2017/06/14 03:16:41 Creating volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] on VM [10.20.104.40]
2017/06/14 03:16:42 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [10.20.104.40]
2017/06/14 03:16:43 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [10.20.104.40]
2017/06/14 03:16:44 Creating a vmgroup [T1] on esx [10.20.105.97]
2017/06/14 03:16:46 Checking volume [BasicTestSuite.TestVmGroupVolumeIsolation_volume_uo7w8oxnzncz] availability from VM [10.20.104.40]

----------------------------------------------------------------------
FAIL: basic_test.go:155: BasicTestSuite.TestVmGroupVolumeIsolation

[...]

PR #1454, opened 2017-06-21
Add more verification steps for basic_test

shaominchen:
Fixes issue #1450


tests/e2e/basic_test.go
```diff
@@ -156,9 +156,10 @@ func (s *BasicTestSuite) TestVmGroupVolumeIsolation(c *C) {
 	misc.LogTestStart(c.TestName())
 
 	// Initialize Config DB
-	admincli.ConfigInit(s.esx)
+	out, err := admincli.ConfigInit(s.esx)
+	c.Assert(err, IsNil, Commentf(out))
```
shuklanirdesh82: I was thinking to extend verification further by invoking adminCLI config status and verify its output to find out configuration mode as changed or not that helps to reach out root cause easily later we will need such verification as well. 

IMO, we should little bit verification here to validate L#159 activity.

tests/e2e/basic_test.go
```diff
@@ -193,8 +194,13 @@ func (s *BasicTestSuite) TestVmGroupVolumeIsolation(c *C) {
 	out, err = dockercli.DeleteVolume(s.vm1, s.volName2)
 	c.Assert(err, IsNil, Commentf(out))
 
+	// Clean up the vm group
+	out, err = admincli.DeleteVMgroup(s.esx, vmgroup)
+	c.Assert(err, IsNil, Commentf(out))
+
 	// Remove Config DB
-	admincli.ConfigRemove(s.esx)
+	out, err = admincli.ConfigRemove(s.esx)
```
shuklanirdesh82: same as above ... AdminCLI config status > verify mode is `Notconfigured`

tests/e2e/basic_test.go
shaominchen: We should have a separate test case for that. I will file a issue for this.

tests/e2e/basic_test.go
```diff
@@ -193,8 +194,13 @@ func (s *BasicTestSuite) TestVmGroupVolumeIsolation(c *C) {
 	out, err = dockercli.DeleteVolume(s.vm1, s.volName2)
 	c.Assert(err, IsNil, Commentf(out))
 
+	// Clean up the vm group
+	out, err = admincli.DeleteVMgroup(s.esx, vmgroup)
+	c.Assert(err, IsNil, Commentf(out))
+
```
ashahi1: I thought admincli does not returns error so this check does not work. Shouldn't we check if vmgroup was actually deleted to be sure that admincli.DeleteVMgroup() actually deleted the vmgroup?

shaominchen: Let's not mix different issues together. We already have an issue for that: https://github.com/vmware/docker-volume-vsphere/issues/1411

tests/e2e/basic_test.go
ashahi1: +1. err check is not doing anything in case of admin cli. We should add this check to verify mode is ```Not configured```.

pshahzeb:
Verification is good.
Agree that admincli doesn't return proper exit code and that fix should be handled in a separate PR.
However, without that fix, these assertions are anyway not helping us, since the err is not going to be NotNil in case of errors. 
We can merge this change but the aim is not fully resolved. 
What do you suggest?

tests/e2e/basic_test.go
shaominchen: Filed issue https://github.com/vmware/docker-volume-vsphere/issues/1456

shuklanirdesh82: I am totally agree for the separate testcase.

what is the harm of doing extra check here? It is a programatic step and helps to find out the root cause immediately instead of passing to next step and fail eventually with out proper cause.

It reduces the cost of extra debugging.

/CC @msterin

tests/e2e/basic_test.go
shaominchen: "err check is not doing anything in case of admin cli" - this is true, but this is not the test case issue, this is a product issue, and we already have a tracking issue for that: #1411 

"We should add this check to verify mode is Not configured." - We need an entire test group to verify different state changes: #1456

shaominchen:
Since everybody agree that it makes debugging tricky due to issue #1411, we should fix that issue right away, instead of wasting time to tweak ALL (not just this test case) test cases to bypass that issue. If @govint has no bandwidth to work on that right now, can @pshahzeb pick up that issue?

shaominchen:
As discussed offline, we came to agreement on this:
1. Due to issue #1411, the assertion statement in the test logic cannot really guarantee the operation is completed as expected. So we need to fix this issue as high priority. @govint 
2. Before issue #1411 is addressed, we can add more verification steps to make the debugging easier - in addition to verify the ssh exit code (which don't work right now), we should also verify the real business data to make sure the operation is completed as expected.

To address 2, we will improve existing util functions to verify the data instead of adding these extra logic to all the test cases. I will file an issue for this.

PR #1473, opened 2017-06-23
Adds Microsoft/go-winio vendor package

venilnoronha:
This commit adds the `Microsoft/go-winio` vendor package, which will be used for implementing the `npipe` based Docker plugin on windows.

I'm skipping the test here as there is no direct code change, and this addition was already validated in PR #1447.


venilnoronha:
@msterin that's right.

Issue #1469, opened 2017-06-23
User cannot delete volumes if its vmgroup gets deleted


ashahi1:
Steps:
1. Create a vmgroup and add a vm to it 
2. Create a volume from the vmgroups's vm 
3. Remove the vm
4. Delete the vmgroup
5. Try deleting the volume, operation fails

Volumes should get deleted irrespective of vmgroup being present or not.

-----------------

Test steps and their output are as follows:

1. Create a vmgroup and add a vm to it 
```
[root@sc-rdops-vm05-dhcp-129-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name vmgroup1 --default-datastore=_VM_DS --vm-list=ubuntu-VM2.2
vmgroup 'vmgroup1' is created. Do not forget to run 'vmgroup vm add' to add vm to vmgroup.
[root@sc-rdops-vm05-dhcp-129-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
Uuid                                  Name      Description                Default_datastore  VM_list
------------------------------------  --------  -------------------------  -----------------  ------------
11111111-1111-1111-1111-111111111111  _DEFAULT  This is a default vmgroup  _VM_DS
41bb861a-7181-4505-bacb-eda799b6c782  vmgroup1                             _VM_DS             ubuntu-VM2.2
```

2. Create a volume from the vmgroups's vm 
```
root@sc-rdops-vm02-dhcp-52-237:~# docker volume create -d vsphere:latest --name TestVol1
TestVol1
root@sc-rdops-vm02-dhcp-52-237:~# docker volume ls
DRIVER              VOLUME NAME
vsphere:latest      TestVol1@sharedVmfs-0


[root@sc-rdops-vm05-dhcp-129-159:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume    Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to  Access      Attach-as               Created By    Created Date
--------  ------------  --------  --------  ----  ----------  ------  -----------  -----------  ----------  ----------------------  ------------  ------------------------
TestVol1  sharedVmfs-0  vmgroup1  100MB     13MB  ext4        N/A     thin         detached     read-write  independent_persistent  ubuntu-VM2.2  Fri Jun 23 01:45:22 2017
```

3. Remove the vm
```
[...]

PR #1474, opened 2017-06-23
Automation test for removing user created vmgroup.

### LINKED ISSUE #1341: [E2E] Automate Tenancy Test case - Verify volume creation
- [x] Verify volume creation from new tenant - P0
- [ ] Verify volume creation after deleting user created tenant - P0

Refer to issue #899

lipingxue:
Fixed #1341 

This PR includes 
1. the automation test for removing user created vmgroup with "--remove-volumes" option
2. code refactoring for function ```VerifyDetachedStatus``` to meet the need for automating this test

Test Done:
Only swarm related tests failed since I do not have a swarm setup
```
=> Running target test-e2e-runalways Tue Jun 27 22:57:46 UTC 2017
GO15VENDOREXPERIMENT=1 go test -v -timeout 30m -tags runalways github.com/vmware/docker-volume-vsphere/tests/e2e

PASS
OK: 6 passed
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	124.073s


=> Running target test-e2e-runonce Tue Jun 27 22:59:53 UTC 2017

GO15VENDOREXPERIMENT=1 go test -v -timeout 30m -tags runonce github.com/vmware/docker-volume-vsphere/tests/e2e

-- FAIL: Test (481.53s)
FAIL
OOPS: 24 passed, 2 FAILED, 1 MISSED
exit status 1
FAIL	github.com/vmware/docker-volume-vsphere/tests/e2e	481.539s


```

### DIFF
--- a/tests/constants/admincli/cmd.go
+++ b/tests/constants/admincli/cmd.go
@@ -95,9 +95,6 @@ const (
 	// VMlist VM of the vmgroup
 	VMlist = " --vm-list="
 
-	// AddDatastoreAccess adds datastore to vmgroup
-	AddDatastoreAccess = vmdkopsAdmin + "vmgroup access add --name="
-
 	// GetDBMode get current DB config
 	GetDBMode = vmdkopsAdmin + "config status"
 )
--- a/tests/constants/admincli/cmd.go
+++ b/tests/constants/admincli/cmd.go
@@ -49,6 +49,9 @@ const (
 	// where --name will be name of the vmgroup
 	ReplaceVMFromVMgroup = vmdkopsAdmin + "vmgroup vm replace --name="
 
+	// AddAccessForVMgroup create access perms on vmgroup
+	AddAccessForVMgroup = vmdkopsAdmin + "vmgroup access add --name "
+
 	// SetAccessForVMgroup modify access perms on vmgroup
 	SetAccessForVMgroup = vmdkopsAdmin + "vmgroup access set --name "
 
--- a/tests/e2e/swarm_test.go
+++ b/tests/e2e/swarm_test.go
@@ -65,7 +65,7 @@ func (s *SwarmTestSuite) SetUpTest(c *C) {
 	out, err := dockercli.CreateVolume(s.master, s.volumeName)
 	c.Assert(err, IsNil, Commentf(out))
 
-	status := verification.VerifyDetachedStatus(s.volumeName, s.dataStores[0], s.master, s.esxName)
+	status := verification.VerifyDetachedStatus(s.volumeName, s.master, s.esxName)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", s.volumeName))
 }
 
@@ -247,7 +247,7 @@ func (s *SwarmTestSuite) TestFailoverAcrossReplicas(c *C) {
 	out, err = dockercli.ListService(s.master, s.serviceName)
 	c.Assert(err, NotNil, Commentf("Expected error does not happen"))
 
-	status = verification.VerifyDetachedStatus(s.volumeName, s.dataStores[0], s.master, s.esxName)
+	status = verification.VerifyDetachedStatus(s.volumeName, s.master, s.esxName)
 	c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
 
 	misc.LogTestEnd(c.TestName())
--- a/tests/e2e/swarm_test.go
+++ b/tests/e2e/swarm_test.go
@@ -35,6 +35,7 @@ import (
 
 type SwarmTestSuite struct {
 	esxName     string
+	dataStores  []string
 	master      string
 	worker1     string
 	worker2     string
@@ -45,6 +46,7 @@ type SwarmTestSuite struct {
 
 func (s *SwarmTestSuite) SetUpSuite(c *C) {
 	s.esxName = inputparams.GetEsxIP()
+	s.dataStores = inputparams.GetDatastores()
 	s.master = inputparams.GetSwarmManager1()
 	s.worker1 = inputparams.GetSwarmWorker1()
 	s.worker2 = inputparams.GetSwarmWorker2()
@@ -63,7 +65,7 @@ func (s *SwarmTestSuite) SetUpTest(c *C) {
 	out, err := dockercli.CreateVolume(s.master, s.volumeName)
 	c.Assert(err, IsNil, Commentf(out))
 
-	status := verification.VerifyDetachedStatus(s.volumeName, s.master, s.esxName)
+	status := verification.VerifyDetachedStatus(s.volumeName, s.dataStores[0], s.master, s.esxName)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", s.volumeName))
 }
 
@@ -245,7 +247,7 @@ func (s *SwarmTestSuite) TestFailoverAcrossReplicas(c *C) {
 	out, err = dockercli.ListService(s.master, s.serviceName)
 	c.Assert(err, NotNil, Commentf("Expected error does not happen"))
 
-	status = verification.VerifyDetachedStatus(s.volumeName, s.master, s.esxName)
+	status = verification.VerifyDetachedStatus(s.volumeName, s.dataStores[0], s.master, s.esxName)
 	c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
 
 	misc.LogTestEnd(c.TestName())
--- a/tests/e2e/swarm_test.go
+++ b/tests/e2e/swarm_test.go
@@ -35,7 +35,6 @@ import (
 
 type SwarmTestSuite struct {
 	esxName     string
-	dataStores  []string
 	master      string
 	worker1     string
 	worker2     string
@@ -46,7 +45,6 @@ type SwarmTestSuite struct {
 
 func (s *SwarmTestSuite) SetUpSuite(c *C) {
 	s.esxName = inputparams.GetEsxIP()
-	s.dataStores = inputparams.GetDatastores()
 	s.master = inputparams.GetSwarmManager1()
 	s.worker1 = inputparams.GetSwarmWorker1()
 	s.worker2 = inputparams.GetSwarmWorker2()
--- a/tests/e2e/vmgroups_test.go
+++ b/tests/e2e/vmgroups_test.go
@@ -459,7 +459,7 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Status should be detached
-	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.DockerHosts[0], vg.config.EsxHost)
+	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.Datastores[0], vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
 
 	// Verify volume can be mounted and unmounted for the second volume
@@ -467,7 +467,7 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Status should be detached
-	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
+	status = verification.VerifyDetachedStatus(vg.volName2, vg.config.Datastores[1],
 		vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
 
@@ -528,7 +528,7 @@ func (vg *VmGroupTest) TestVmgroupRemove(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Status should be detached
-	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.DockerHosts[0], vg.config.EsxHost)
+	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.Datastores[0], vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
 
 	// Verify volume can be mounted and unmounted for the second volume
@@ -536,7 +536,7 @@ func (vg *VmGroupTest) TestVmgroupRemove(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Status should be detached
-	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
+	status = verification.VerifyDetachedStatus(vg.volName2, vg.config.Datastores[1],
 		vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
 
--- a/tests/e2e/vmgroups_test.go
+++ b/tests/e2e/vmgroups_test.go
@@ -454,29 +454,27 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	// need to use full name for volume name
 	vg.createVolumes(c, vg.volName2+"@"+vg.config.Datastores[1])
 
-	// Verify volume can be mounted and unmounted for the first volume
+	// Run a container and then remove it using the first volume
 	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName1, vg.testContainer)
 	c.Assert(err, IsNil, Commentf(out))
 
-	// Status should be detached
+	// Status should be detached after removing the container
 	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.Datastores[0], vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
 
-	// Verify volume can be mounted and unmounted for the second volume
+	// Run a container and then remove it using the first volume
 	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName2+"@"+vg.config.Datastores[1], vg.testContainer)
 	c.Assert(err, IsNil, Commentf(out))
 
-	// Status should be detached
+	// Status should be detached after removing the container
 	status = verification.VerifyDetachedStatus(vg.volName2, vg.config.Datastores[1],
 		vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
 
 	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
-	vmList := []string{vg.config.DockerHostNames[0], vg.config.DockerHostNames[1]}
-	for _, vm := range vmList {
-		out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
-		c.Assert(err, IsNil, Commentf(out))
-	}
+	vmList := vg.config.DockerHostNames[0] + "," + vg.config.DockerHostNames[1]
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vmList)
+	c.Assert(err, IsNil, Commentf(out))
 
 	out, err = adminutils.DeleteVMgroup(vg.config.EsxHost, vgTestVMgroup1, true)
 	c.Assert(err, IsNil, Commentf(out))
@@ -496,10 +494,9 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
 	c.Assert(err, IsNil, Commentf(out))
 
-	for _, vm := range vmList {
-		out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
-		c.Assert(err, IsNil, Commentf(out))
-	}
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vmList)
+	c.Assert(err, IsNil, Commentf(out))
+
 	misc.LogTestEnd(c.TestName())
 }
 
--- a/tests/e2e/vmgroups_test.go
+++ b/tests/e2e/vmgroups_test.go
@@ -47,7 +47,6 @@ type VmGroupTest struct {
 	volName1      string
 	volName2      string
 	volName3      string
-	volName4      string
 }
 
 var _ = Suite(&VmGroupTest{})
@@ -126,7 +125,6 @@ func (vg *VmGroupTest) vmgroupGetVolName(c *C) {
 	vg.volName1 = inputparams.GetUniqueVolumeName(c.TestName())
 	vg.volName2 = inputparams.GetUniqueVolumeName(c.TestName())
 	vg.volName3 = inputparams.GetUniqueVolumeName(c.TestName())
-	vg.volName4 = inputparams.GetUniqueVolumeName(c.TestName())
 }
 
 // Tests to validate behavior with the __DEFAULT_ vmgroup.
@@ -452,14 +450,6 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	out, err := adminutils.AddCreateAccessForVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.Datastores[1])
 	c.Assert(err, IsNil, Commentf(out))
 
-	cmd := adminconst.GetAccessForVMgroup + vgTestVMgroup1
-	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
-	if err == nil {
-		log.Printf(out)
-	}
-
-	misc.SleepForSec(2)
-
 	// Create another volume on second datastore
 	// need to use full name for volume name
 	vg.createVolumes(c, vg.volName2+"@"+vg.config.Datastores[1])
@@ -467,9 +457,6 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	// Verify volume can be mounted and unmounted for the first volume
 	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName1, vg.testContainer)
 	c.Assert(err, IsNil, Commentf(out))
-
-	// Docker may not have completed the detach yet with the host.
-	misc.SleepForSec(2)
 
 	// Status should be detached
 	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.DockerHosts[0], vg.config.EsxHost)
@@ -479,46 +466,40 @@ func (vg *VmGroupTest) TestVmgroupRemoveWithRemoveVol(c *C) {
 	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName2+"@"+vg.config.Datastores[1], vg.testContainer)
 	c.Assert(err, IsNil, Commentf(out))
 
-	// Docker may not have completed the detach yet with the host.
-	misc.SleepForSec(2)
-
 	// Status should be detached
 	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
 		vg.config.DockerHosts[0], vg.config.EsxHost)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
 
 	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
-	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
-	c.Assert(err, IsNil, Commentf(out))
+	vmList := []string{vg.config.DockerHostNames[0], vg.config.DockerHostNames[1]}
+	for _, vm := range vmList {
+		out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
+		c.Assert(err, IsNil, Commentf(out))
+	}
 
-	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
-	c.Assert(err, IsNil, Commentf(out))
-
-	log.Printf("Removing test vmgroup %s", vgTestVMgroup1)
 	out, err = adminutils.DeleteVMgroup(vg.config.EsxHost, vgTestVMgroup1, true)
 	c.Assert(err, IsNil, Commentf(out))
 
 	// vmgroup has been removed and all volumes belong to it have been removed
 	// "err.Error()" will be filled with "exit status 1"
-	out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, vg.volName1)
-	log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, vg.volName1)
-	c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", vg.volName1))
-
-	out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, vg.volName2)
-	log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, vg.volName2)
-	c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", vg.volName2))
+	volumeList := []string{vg.volName1, vg.volName2}
+	for _, volume := range volumeList {
+		out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, volume)
+		log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, volume)
+		c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", volume))
+	}
 
 	// Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
-	cmd = adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+	cmd := adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
 	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
 	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
 	c.Assert(err, IsNil, Commentf(out))
 
-	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
-	c.Assert(err, IsNil, Commentf(out))
-
-	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
-	c.Assert(err, IsNil, Commentf(out))
+	for _, vm := range vmList {
+		out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
+		c.Assert(err, IsNil, Commentf(out))
+	}
 	misc.LogTestEnd(c.TestName())
 }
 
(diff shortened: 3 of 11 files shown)


tests/constants/admincli/cmd.go
```diff
@@ -49,6 +49,9 @@ const (
 	// where --name will be name of the vmgroup
 	ReplaceVMFromVMgroup = vmdkopsAdmin + "vmgroup vm replace --name="
 
+	// AddAccessForVMgroup create access perms on vmgroup
+	AddAccessForVMgroup = vmdkopsAdmin + "vmgroup access add --name "
```
govint: #1381 is also modifying this code pls. check if there is any overlap.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+
+	// Verify volume can be mounted and unmounted for the second volume
+	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName2+"@"+vg.config.Datastores[1], vg.testContainer)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Docker may not have completed the detach yet with the host.
+	misc.SleepForSec(2)
+
+	// Status should be detached
+	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
```
govint: Can't the same VerifyDetachedStatus() be used here? Why have a separate function for the non-default DS, the DS name is being passed anyway.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
+	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
+	c.Assert(err, IsNil, Commentf(out))
+	misc.LogTestEnd(c.TestName())
```
govint: Why do we need to recreate and add the VMs back? is it for maintaining the test config?

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+}
+
+// test to verify volume is removed after user created vmgroup is removed without "--remove-volume" option
+// 1. create a user created vmgroup "vmgroup_test1" and add VMs to this vmgroup
+// 2. create two volumes for this vmgroup, one is on default datastore, and the other is on non-default datastore
+// 3. verify mount/unmount for this volume
+// 4. after unmount, verify the volume is detached for both volumes
+// 5. remove VMs from this vmgroup, and remove the vmgroup with "--remove-volume" option
```
govint: with or without the option.

tests/e2e/vmgroups_test.go
```diff
@@ -46,6 +47,7 @@ type VmGroupTest struct {
 	volName1      string
 	volName2      string
 	volName3      string
+	volName4      string
```
pshahzeb: This isn't because of your change but this should have been a slice instead. You might wanna handle it if it is trivial.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+		vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
+
+	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
+	c.Assert(err, IsNil, Commentf(out))
+
```
ashahi1: If possible, please combine them as to me it seems that we are removing VMs from the same vmgroup.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	misc.LogTestStart(c.TestName())
+
+	// Create a volume in vgTestVMgroup1 on default_datastore
+	vg.createVolumes(c, vg.volName1)
+
+	// Add create privilege for the second datastore
+	out, err := adminutils.AddCreateAccessForVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.Datastores[1])
+	c.Assert(err, IsNil, Commentf(out))
+
+	cmd := adminconst.GetAccessForVMgroup + vgTestVMgroup1
```
pshahzeb: I think this is a verification step after access has been added. Might wanna wrap it into a util.  If we always do this after adding the access, you might wanna add these steps to AddCreateAccessForVMgroup util so that the test steps itself are clear.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
+	if err == nil {
+		log.Printf(out)
+	}
+
+	misc.SleepForSec(2)
```
pshahzeb: Why this sleep? Does it take time for access to be added? if so please comment.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
+
+	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
+	c.Assert(err, IsNil, Commentf(out))
+
+	log.Printf("Removing test vmgroup %s", vgTestVMgroup1)
```
ashahi1: Minor: We do not really need this statement as DeleteVMgroup method itself has log message.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	// Status should be detached
+	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
+
+	// Verify volume can be mounted and unmounted for the second volume
+	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName2+"@"+vg.config.Datastores[1], vg.testContainer)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Docker may not have completed the detach yet with the host.
+	misc.SleepForSec(2)
```
pshahzeb: ExecContainer fires a container with --rm flag. So this detach should happen instantly. I am not sure if a sleep is needed here. Are we hitting a race without this sleep?

tests/e2e/vmgroups_test.go
pshahzeb: +1

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	// Status should be detached
+	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
+		vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
+
+	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
```
pshahzeb: Minor: wrap this removal and assert in a loop.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+
+	// vmgroup has been removed and all volumes belong to it have been removed
+	// "err.Error()" will be filled with "exit status 1"
+	out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, vg.volName1)
+	log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, vg.volName1)
+	c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", vg.volName1))
+
+	out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, vg.volName2)
+	log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, vg.volName2)
+	c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", vg.volName2))
```
pshahzeb: minor: these 3 repeating lines can be wrapped after you the volume name as slice.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +434,175 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	// Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
+	cmd = adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
+	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[0])
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.DockerHostNames[1])
```
pshahzeb: Should be wrapped in a loop

tests/utils/verification/volumeproperties.go
```diff
@@ -152,6 +155,29 @@ func VerifyDetachedStatus(name, hostName, esxName string) bool {
 	return false
 }
 
+// VerifyDetachedStatusOnNonDefaultDS - For volume on non-default DS,
+// check if the status gets detached within the timeout
+func VerifyDetachedStatusOnNonDefaultDS(name, datastore, hostName, esxName string) bool {
```
pshahzeb: Do we need this since we already have a util to verify detached status?

tests/e2e/vmgroups_test.go
lipingxue: I introduce this because most of the existing tests only need to verify volumes created on default datastore and already call function ```VerifyDetachedStatus(```.  By introducing this new function, we can keep those part unchanged. Those two util functions  can be used by caller in different cases.

tests/e2e/vmgroups_test.go
lipingxue: Yes. Just for maintaining the test config.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Here we do remove VMs from same vmgroup, and it is what we need to do here.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: This is not a verification step, remove this.

tests/e2e/vmgroups_test.go
lipingxue: I copied from existing code, and verified that this is not needed. So I remove the sleep.

tests/e2e/vmgroups_test.go
lipingxue: Done.

lipingxue:
@govint @ashahi1 @pshahzeb I have addressed your comments,  please take a look.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,143 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	// Status should be detached
+	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
+
+	// Verify volume can be mounted and unmounted for the second volume
+	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName2+"@"+vg.config.Datastores[1], vg.testContainer)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Status should be detached
+	status = verification.VerifyDetachedStatusOnNonDefaultDS(vg.volName2, vg.config.Datastores[1],
```
govint: Same function as above can be used for the verification. Verification is best done by exec'ing a container on another docker host VM, which will be successful only when the volume is really detached from the VM.

tests/utils/verification/volumeproperties.go
```diff
@@ -152,6 +155,29 @@ func VerifyDetachedStatus(name, hostName, esxName string) bool {
...
+	log.Printf("Confirming detached status for volume [%s] on datastore[%s]\n", name, datastore)
+
+	//TODO: Need to implement generic polling logic for better reuse
+	const maxAttempt = 60
+	for attempt := 0; attempt < maxAttempt; attempt++ {
+		misc.SleepForSec(2)
+		status := getVolumeStatusHost(name+"@"+datastore, hostName)
```
pshahzeb: The only difference between this util and VerifyDetachedStatus is the datastore arg which is used to construct a fully qualified name to get volume status from docker host.
If it is worth, please put the common part of these utils into a func which can be called from these utils.

I still think we can use VerifyDetachedStatus, unless we are worried about same volume names on different datastores.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,143 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	c.Assert(err, IsNil, Commentf(out))
+
+	volumeList := []string{vg.volName1, vg.volName2}
+	for _, volume := range volumeList {
+		// volume does not belong to any vmgroup, so "out" is expected to be "N/A"
+		out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, volume)
+		log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, volume)
+		c.Assert(out, Equals, "N/A", Commentf("volume %s should not belong to any vmgroup", volume))
+	}
+
```
lipingxue: Those volumes are no orphan volumes, and user have no way to access/cleanup those orphan volumes. So we cannot cleanup them through adminCLI or dockerCLI.
As we dicussed in https://github.com/vmware/docker-volume-vsphere/issues/1469, we should NOT allow user to remove a vmgroup if volumes still in the vmgroup. So testing the scenario that "remove vmgroup without removing volumes" is not needed if we decided to not support remove vmgroup without removing volumes. 

I plan to remove this test from the PR.

lipingxue:
@govint @pshahzeb I have addressed your comments and disable the test that removing vmgroup without "--remove-volumes" option.

tests/constants/admincli/cmd.go
shuklanirdesh82: +1

tests/utils/verification/volumeproperties.go
```diff
@@ -130,15 +133,37 @@ func getVolumeStatusHost(name, hostName string) string {
...
+// VerifyDetachedStatus- check if the status gets detached within the timeout
+func VerifyDetachedStatus(name, datastore, hostName, esxName string) bool {
+	log.Printf("Confirming detached status for volume [%s] on datastore[%s]\n", name, datastore)
 
 	//TODO: Need to implement generic polling logic for better reuse
 	const maxAttempt = 60
 	for attempt := 0; attempt < maxAttempt; attempt++ {
 		misc.SleepForSec(2)
-		status := getVolumeStatusHost(name, hostName)
+		status := getVolumeStatusHost(name+"@"+datastore, hostName)
```
shuklanirdesh82: can you please shade some information why do we need this change?

tests/utils/verification/volumeproperties.go
```diff
@@ -130,15 +133,37 @@ func getVolumeStatusHost(name, hostName string) string {
 	return out
 }
 
-// VerifyDetachedStatus - check if the status gets detached within the timeout
-func VerifyDetachedStatus(name, hostName, esxName string) bool {
-	log.Printf("Confirming detached status for volume [%s]\n", name)
+// // VerifyDetachedStatus - check if the status gets detached within the timeout
```
shuklanirdesh82: I guess this is dead code and can be cleaned up.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	// Add create privilege for the second datastore
+	out, err := adminutils.AddCreateAccessForVMgroup(vg.config.EsxHost, vgTestVMgroup1, vg.config.Datastores[1])
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Create another volume on second datastore
+	// need to use full name for volume name
+	vg.createVolumes(c, vg.volName2+"@"+vg.config.Datastores[1])
+
+	// Verify volume can be mounted and unmounted for the first volume
```
shuklanirdesh82: As per the inline comment, please add verification step when volume was attached.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	out, err = dockercli.ExecContainer(vg.config.DockerHosts[0], vg.volName1, vg.testContainer)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Status should be detached
+	status := verification.VerifyDetachedStatus(vg.volName1, vg.config.Datastores[0], vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName1))
+
+	// Verify volume can be mounted and unmounted for the second volume
```
shuklanirdesh82: same as above

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+
+	// Status should be detached
+	status = verification.VerifyDetachedStatus(vg.volName2, vg.config.Datastores[1],
+		vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
+
+	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
+	vmList := []string{vg.config.DockerHostNames[0], vg.config.DockerHostNames[1]}
+	for _, vm := range vmList {
+		out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
```
shuklanirdesh82: only one call is sufficient to remove both vms .. `admin_cli vmgroup vm remove --name --vm-list ` .. am I missing anything here?

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+		c.Assert(err, IsNil, Commentf(out))
+	}
+
+	out, err = adminutils.DeleteVMgroup(vg.config.EsxHost, vgTestVMgroup1, true)
+	c.Assert(err, IsNil, Commentf(out))
+
+	// vmgroup has been removed and all volumes belong to it have been removed
+	// "err.Error()" will be filled with "exit status 1"
+	volumeList := []string{vg.volName1, vg.volName2}
+	for _, volume := range volumeList {
```
shuklanirdesh82: could be replaced with one remote call invocation ..`admin_cli volume ls` > verifies volumes created as part of this test method should not be present.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+		out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, volume)
+		log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, volume)
+		c.Assert(err.Error(), Equals, "exit status 1", Commentf("volume %s should be removed", volume))
+	}
+
+	// Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
```
shuklanirdesh82: sorry we should avoid this approach .. this leads to confusion any of the step fails (L494-502) instead test should create its own vmgroup and deletes it

tests/e2e/vmgroups_test.go
lipingxue: We should use util function instead of call "admin_cli volume ls" directly.

tests/e2e/vmgroups_test.go
lipingxue: Why? I checked the existing tests in "vmgroups_test.go", and I followed the approach as other tests. Why it will lead to confusion? Comments already say what is the purpose of the code from L494-502.

tests/utils/verification/volumeproperties.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Update the comments.

tests/utils/verification/volumeproperties.go
lipingxue: Function ```getVolumeStatusHost``` calls dockercli to get the volume status. It actually run ``` docker volume inspect``` command, so for volume which is not created on the default_datastore, we need to pass in full name of the volume.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: As we discussed, https://github.com/vmware/docker-volume-vsphere/issues/1496 is filed to track this.

tests/e2e/vmgroups_test.go
lipingxue: As we discussed, https://github.com/vmware/docker-volume-vsphere/issues/1497 is filed to track this.

tests/constants/admincli/cmd.go
shuklanirdesh82: this needs to be removed.

tests/utils/verification/volumeproperties.go
```diff
@@ -130,15 +133,15 @@ func getVolumeStatusHost(name, hostName string) string {
 	return out
 }
 
-// VerifyDetachedStatus - check if the status gets detached within the timeout
-func VerifyDetachedStatus(name, hostName, esxName string) bool {
-	log.Printf("Confirming detached status for volume [%s]\n", name)
+// VerifyDetachedStatus- check if the status gets detached within the timeout
```
shuklanirdesh82: minor: golint is complaining about this method document .. 

Reference: https://ci.vmware.run/vmware/docker-volume-vsphere/763

`../tests/utils/verification/volumeproperties.go:136:1: comment on exported function VerifyDetachedStatus should be of the form "VerifyDetachedStatus ..."`

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,141 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	c.Assert(err, IsNil, Commentf(out))
+
+	// Status should be detached after removing the container
+	status = verification.VerifyDetachedStatus(vg.volName2, vg.config.Datastores[1],
+		vg.config.DockerHosts[0], vg.config.EsxHost)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", vg.volName2+"@"+vg.config.Datastores[1]))
+
+	// remove VM1 and VM2 from vgTestVMgroup1 and then remove the vmgroup
+	vmList := vg.config.DockerHostNames[0] + "," + vg.config.DockerHostNames[1]
+	out, err = adminutils.RemoveVMFromVMgroup(vg.config.EsxHost, vgTestVMgroup1, vmList)
```
govint: Suggestion, can instead pass the vg.config.DockerHostNames to RemoveVMFromVMgroup() and let it do the creation of the vmlist to pass to the admin cli command. Basically, the test functions are simple and shouldn't have to create special args for the util functions.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,141 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+	}
+
+	// Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
+	cmd := adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
+	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
+	c.Assert(err, IsNil, Commentf(out))
+
+	out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vmList)
+	c.Assert(err, IsNil, Commentf(out))
```
govint: This is ok, just place it in a separate function RestoreVmgroups() so this test function doesn't look like its cluttered with this restoration code.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,141 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+// 	Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
+// 	cmd := adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+// 	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
+// 	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
+// 	c.Assert(err, IsNil, Commentf(out))
+
+// 	for _, vm := range vmList {
+// 		out, err = adminutils.AddVMToVMgroup(vg.config.EsxHost, vgTestVMgroup1, vm)
+// 		c.Assert(err, IsNil, Commentf(out))
+// 	}
```
govint: Makes sense to have a separate function to keep this code to restore the configuration.

tests/utils/verification/volumeproperties.go
```diff
@@ -161,3 +164,25 @@ func GetAssociatedPolicyName(hostname string, volName string) (string, error) {
...
+	}
+	volProps := strings.Fields(op)
+	if op == "" {
+		msg := fmt.Sprintf("Null value is returned by admin cli when looking for vmgroup field for volume %s", volName)
+		log.Printf(msg)
+		return op, errors.New(msg)
+	}
+	if len(volProps) != 2 {
+		log.Printf("Admin cli output is expected to consist of two elements only - "+
+			"volume name and vmgrop. Actual output %s ", op)
```
govint: what happens if the same volume name is on two separate datastores, then we will get two lines (I had this scenario and verification/volumeproperties.go:GetVMAttachedToVolUsingAdminCli() failed).

I'll raise an issue for docker volume inspect to return the vmgroup for a volume so we can use it directly.

I've added a function to return a map of volume properties in PR #1500 and that can be modified to return the vmgroup for a volume.

The above approach is prone to errors in different scenarios and have a lot of this code,  as much as possible should be using docker to get properties filtered out for our use.

tests/utils/verification/volumeproperties.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Done.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,144 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
...
+// 	for _, volume := range volumeList {
+// 		volume does not belong to any vmgroup, so "out" is expected to be "N/A"
+// 		out, err = verification.GetVMGroupForVolume(vg.config.EsxHost, volume)
+// 		log.Println("GetVMGroupForVolume return out[%s] err[%s] for volume %s", out, err, volume)
+// 		c.Assert(out, Equals, "N/A", Commentf("volume %s should not belong to any vmgroup", volume))
+// 	}
+
+// 	Recreate the test VM group1 and add VM1, VM2 back to the vmgroup
+// 	cmd := adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+// 	log.Printf("Creating test vmgroup %s", vgTestVMgroup1)
```
lipingxue: Done. Move the restore logic to function ```RestoreVmgroup```

tests/e2e/vmgroups_test.go
lipingxue: Done. Same as above.

tests/utils/verification/volumeproperties.go
lipingxue: The volume name used in test is unique since it use random generator. So  it is not very possible that we hit the scenario that we have same volume name with two different datastores.

For a orphan volume, ``` docker volume inspect ``` may not able to get the vmgroup associated with the volume.

lipingxue:
I think more and decide that we should add function "VerifyDetachedStatusNonDefaultDS" back.
This function is used to check whether volume is detached if the volume is created on the non default datastore. Original function "VerifyDetachedStatus"  should be used to check whether volume created on the default datastore is detached or not. 

We should not try to combine function "VerifyDetachedStatusNonDefaultDS" and "VerifyDetachedStatus" into one.

Here are the reasons:
1.  From user point of view, for volume created on the default datastore, user should use short name instead of full vol name. Full vol  should be used if the volume is not in the default datastore. So it is make sense to have separate functions to handle these two different scenarios. If the language support function overloading, we should have overload function for these two scenarios. Since go does not support function overloading, we should have two functions with different names here.

2.  For automation tests which need to check whether a volume is detached or not, most of the existing tests create the volume on the default datastore and the default_datastore is vm_datastore for most cases.  It does not make sense to find out the vm_datastore and then pass in to function "VerifyDetachedStatus" . In order to do that, we must write another util function to get the vm_store for a VM, which is not needed.  And then most existing tests need to be refactored.  I don't see much value to do this. 
  
@shuklanirdesh82  @govint 
I decide to add function "VerifyDetachedStatusNonDefaultDS" back, and please review it.

tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -31,16 +31,14 @@ func CreateVMgroup(ip, name, vmName, dsName string) (string, error) {
...
-
-// DeleteVMgroupAndItsVolumes method deletes a vmgroup and removes its volumes
-func DeleteVMgroupAndItsVolumes(ip, name string) (string, error) {
-	log.Printf("Deleting a vmgroup and its corresponding volumes [%s] on esx [%s]\n", name, ip)
-	return ssh.InvokeCommand(ip, admincli.RemoveVMgroup+name+admincli.RemoveVolumes)
+// DeleteVMgroup method deletes a vmgroup and removes its volumes as well if "delete_vol" is set
+func DeleteVMgroup(ip, name string, delete_vol bool) (string, error) {
+	log.Printf("Deleting a vmgroup [%s] on esx [%s] with delete_vol[%d]\n", name, ip, delete_vol)
+	if delete_vol {
+		return ssh.InvokeCommand(ip, admincli.RemoveVMgroup+name+admincli.RemoveVolumes)
```
shuklanirdesh82: minor: actually `command` should have been prepared in `if ...else .. ` block that improve readability and `ssh.InvokeCommand(` would be placed at once out of `if` block.

tests/constants/admincli/cmd.go
shuklanirdesh82: L#99 is doing the same thing .. Please remove this one.

tests/utils/verification/volumeproperties.go
```diff
@@ -158,6 +161,29 @@ func VerifyDetachedStatus(name, hostName, esxName string) bool {
...
 
+// VerifyDetachedStatusNonDefaultDS checks if the status gets detached within the timeout for volume on non
+// default datastore
+func VerifyDetachedStatusNonDefaultDS(name, datastore, hostName, esxName string) bool {
+	log.Printf("Confirming detached status for volume [%s] on datastore[%s]\n", name, datastore)
+
+	//TODO: Need to implement generic polling logic for better reuse
+	const maxAttempt = 60
+	for attempt := 0; attempt < maxAttempt; attempt++ {
+		misc.SleepForSec(2)
```
shuklanirdesh82: this needs to be removed; it may hide if there is any race condition.

govint:
@lipingxue, no the VerifyDetachedState() function is sufficient for any scenario - whether default or non-default datastore. The user has the choice to specify the datastore whether its a default or not. The function is only figuring out whether a _volume_ is in detached or attached state. Why should that depend on the datastore?

In fact, the restart tests explicitly test for use case where user can attach a volume via short and long name on the default and non-default datastores.

lipingxue:
@govint I don't get your point what do you mean  "user has the choice to specify the datastore whether its a default or not".  Which parameter that then function need to take so that "user has the choice to specify the datastore whether its a default or not"? The function calls both dockerCLI and adminCLI to figure out the volume is detached or not. For dockerCLI, short name can only be used if the volume is created in the default datastore. Otherwise, user need to specify the full vol name.

shuklanirdesh82:
CI is failing: https://ci.vmware.run/vmware/docker-volume-vsphere/842

```
2017/06/30 15:40:00 START: vgBasicSuite.TestDSAccessPrivilegeForUserVG
2017/06/30 15:40:00 Creating create access for vmgroup vg_basictest, datastore vsanDatastore on esx [192.168.31.62] 
2017/06/30 15:40:03 Getting vmgroup's [vg_basictest] access rights to datastore vsanDatastore ,  on esx [192.168.31.62] 
2017/06/30 15:40:05 Access rights to datastore vsanDatastore for vmgroup - vg_basictest are: [vsanDatastore True] . 

FAIL: vmgroupmisc_test.go:157: vgBasicSuite.TestDSAccessPrivilegeForUserVG

vmgroupmisc_test.go:165:
    c.Assert(isDatastoreAccessible, Equals, false, Commentf("Unexpected Behavior: Datastore %s is accessible for vmgroup %s .", s.config.Datastores[1], vmGroupName))
... obtained bool = true
... expected bool = false
... Unexpected Behavior: Datastore vsanDatastore is accessible for vmgroup vg_basictest .
```

Let's continue the investigation in the next sprint.

tests/e2e/vmgroups_test.go
```diff
@@ -431,3 +432,138 @@ func (vg *VmGroupTest) TestVmGroupVolumeClone(c *C) {
 
 	misc.LogTestEnd(c.TestName())
 }
+
+// Restore VMgroup
+func (vg *VmGroupTest) RestoreVmgroup(c *C) {
```
shaominchen: Nit: This function should start with small letter since we don't really want to export it.

tests/utils/verification/volumeproperties.go
shaominchen: Agree with @pshahzeb. I think we can do the same thing as what we are doing in VerifyAttachedStatus - please refer to issue #1506 and the fix: https://github.com/vmware/docker-volume-vsphere/pull/1492/commits/54ea0f1120db6ab683b83f4d8f6e033cfd9273a6

tests/e2e/swarm_test.go
```diff
@@ -35,6 +35,7 @@ import (
 
 type SwarmTestSuite struct {
 	esxName     string
+	dataStores  []string
```
lipingxue: Done.

tests/e2e/vmgroups_test.go
lipingxue: Done.

lipingxue:
@shaominchen  I have addressed your comments, please take a look. Thanks!

Issue #1315, opened 2017-06-01
compose yml with datastore selection


nbaillie:
im ok with the creation of volumes like the bellow:
docker volume create --driver=vsphere --name=MyTestVol@VMWARE-DOCKER -o size=1gb

But how can i add the @VMWARE-DOCKER component to the yml for stack deployment?

i have tried Like.....
volumes:
    db:
       driver: vsphere
       driver_opts:
         size: 120Gb
         datastore: VMWARE-DOCKER

and.....
volumes:
    db@VMWARE-DOCKER:
       driver: vsphere
       driver_opts:
         size: 120Gb
         datastore: VMWARE-DOCKER

and some other variations.

pdhamdhere:
@nbaillie Thanks for reporting issue. `@datastore` is optional. By default volumes are created on same datastore as VM. All VMs on same datastore can use volume without specifying `@datastore` qualifier.

Not able to specify `@datastore` in compose file is a BUG. Thanks for catch.

lipingxue:
@nbaillie The following is a workaround:
1. Create the volume 
```
docker volume create --driver=vsphere --name=MyTestVol@VMWARE-DOCKER -o size=1gb


```
2. use the volume as "external volume" in compose file:
```
volumes:
db:
   external:
     name: "MyTestVol@VMWARE-DOCKER" 
```

May I know why you need to specify the datastore name explicitly? By default volumes are created on the same datastore as VM.  Thanks!

nbaillie:
Thanks for the workaround, and I'm happy to help catch this.

We have the VMs Primary Volume on a vVol datastore and want to have the Docker volumes not on vVol as our storage layer has better support and features for non-vVol datastore additionally we are looking at using different performance tiers of storage for certain volumes, Logs,db,archives
 etc.

ckotte:
this is also necessary if your worker nodes are distributed over several datastores.. If you don't specify the datastore and the container fails over to a worker that is stored in another datastore, the container can't have access to its persistent data..

pdhamdhere:
Tracking issue with [docker/compose #4886](https://github.com/docker/compose/issues/4886)

lipingxue:
[docker/compose #4886](https://github.com/docker/compose/issues/4886)  was closed. Filed a new issue with [moby/moby #33529](https://github.com/moby/moby/issues/33529).

madjam002:
[...]

PR #1475, opened 2017-06-23
[CI SKIP] Known issue doc.liping

lipingxue:
Add issue #1315 as a known issue in the README.md.


govint:
Do we still have RPM users?

Issue #1324, opened 2017-06-01
Automate Docker Swarm interop test


shaominchen:
Details refer to - Issue #927.

shaominchen:
Fixed with PR #1492.

Issue #1490, opened 2017-06-27
swarm_test.TestDockerSwarm failed as volume is not attached


ashahi1:
CI: https://ci.vmware.run/vmware/docker-volume-vsphere/739

Sam, as swarm test failed, can you please take a look at it?

I was trying to fix issue https://github.com/vmware/docker-volume-vsphere/issues/1284#issuecomment-311200955 and this test failed.

Test logs:
```
2017/06/27 03:11:08 START: swarm_test.TestDockerSwarm
2017/06/27 03:11:08 Fetching full name for volume [swarm_test_volume_184158] from VM [192.168.31.81]
2017/06/27 03:11:09 Full volume name: [swarm_test_volume_184158@datastore1]
2017/06/27 03:11:09 Creating a docker service [swarm_test_service_239730] on VM [192.168.31.81]
2017/06/27 03:11:10 Checking docker service [swarm_test_service_239730] status on VM [192.168.31.81]
2017/06/27 03:11:10 Listing docker service [swarm_test_service_239730] running on VM [192.168.31.81]
2017/06/27 03:11:10 Checking running containers for docker service [swarm_test_service_239730] on docker hosts: [192.168.31.81 192.168.31.85 192.168.31.83]
2017/06/27 03:11:11 Sleep for 5 seconds
2017/06/27 03:11:23 Containers running on docker host [192.168.31.83]: swarm_test_service_239730.1.lww2tro7aw22uh57s0pq4bbq6
2017/06/27 03:11:23 Confirming attached status for volume [swarm_test_volume_184158]
2017/06/27 03:11:27 Expected Attached VM name is [photon.vmfs]
2017/06/27 03:11:27 Attached VM name from Docker CLI is [Ubuntu.14.04]
2017/06/27 03:11:27 Attached VM name from Admin CLI is [Ubuntu.14.04]

----------------------------------------------------------------------
FAIL: swarm_test.go:91: SwarmTestSuite.TestDockerSwarm

swarm_test.go:106:
    c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
... obtained bool = false
... expected bool = true
... Volume swarm_test_volume_184158 is not attached

2017/06/27 03:11:27 Destroying volume [swarm_test_volume_184158]

----------------------------------------------------------------------
FAIL: swarm_test.go:65: SwarmTestSuite.TearDownSuite

swarm_test.go:68:
    c.Assert(err, IsNil, Commentf(out))
[...]

PR #1492, opened 2017-06-27
Add swarm test - TestFailoverAcrossSwarmNodes

shaominchen:
Fixes issue #1324, #1490  and #1506.


tests/e2e/swarm_test.go
```diff
@@ -47,30 +50,133 @@ func (s *SwarmTestSuite) SetUpSuite(c *C) {
...
+	misc.LogTestStart(c.TestName())
+
+	const (
+		testData = "Hello World!"
+		testFile = "hello.txt"
+		volPath  = "/vol"
+	)
+
+	fullVolumeName := verification.GetFullVolumeName(s.master, s.volumeName)
+	opts := "--mount type=volume,source=" + fullVolumeName + ",target=" + volPath + ",volume-driver=" + constant.VDVSPluginName + "--constraint node.role==worker --restart-condition on-failure" + constant.TestContainer
```
lipingxue: Add some comments about the service to be created.  "Service will only be scheduled in the worker node and it will restart on failure".

tests/e2e/swarm_test.go
```diff
@@ -47,30 +50,133 @@ func (s *SwarmTestSuite) SetUpSuite(c *C) {
...
+	c.Assert(status, Equals, true, Commentf("Container %s is not running", s.serviceName))
+
+	status = verification.VerifyAttachedStatus(s.volumeName, s.master, s.esxName)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
+
+	containerName, err = dockercli.GetContainerName(host, s.serviceName)
+	c.Assert(err, IsNil, Commentf("Failed to retrieve container name: %s", containerName))
+
+	// TODO: this verification does not pass during auto run - need further investigation
+	// out, err = dockercli.ReadFromContainer(host, containerName, volPath, testFile)
```
shuklanirdesh82: are you planning to uncomment before merging?

tests/utils/dockercli/containers.go
```diff
@@ -86,3 +86,24 @@ func GetVolumeProperties(volumeName, hostName string) (string, error) {
 	return ssh.InvokeCommand(hostName, cmd)
 }
 
+// WriteToContainer write data to a file in the running container
+func WriteToContainer(ip, containerName, volPath, fileName, data string) (string, error) {
```
shuklanirdesh82: we do have the util in place https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/dockercli/volumelifecycle.go#L98

Any reason for introducing this one?

tests/utils/dockercli/containers.go
```diff
@@ -86,3 +86,24 @@ func GetVolumeProperties(volumeName, hostName string) (string, error) {
...
+
+	writeCmd := " /bin/sh -c 'echo \"" + data + "\" > " + volPath + "/" + fileName + "'"
+	fullCmd := dockercli.RunCmdInContainer + containerName + writeCmd
+
+	log.Println(fullCmd)
+	return ssh.InvokeCommand(ip, fullCmd)
+}
+
+// ReadFromContainer read content from the file in the running container
+func ReadFromContainer(ip, containerName, volPath, fileName string) (string, error) {
```
shuklanirdesh82: same as above we have the util in place .. on the other hand .. ReadFromVolume makes more sense. We should remove this and use the existing one.

tests/utils/esx/govc.go
```diff
@@ -54,11 +55,19 @@ func GetVMPowerState(vmName string) string {
 // PowerOnVM util powers on the VM
 // govc vm.power -on=true photon
 func PowerOnVM(vmName string) string {
-	// log.Printf("Powering on VM [%s]\n", vmName)
+	log.Printf("Powering on VM [%s]\n", vmName)
```
shuklanirdesh82: have a look at L#20 for the reason being commented out this line.

tests/e2e/swarm_test.go
shaominchen: Sure.

tests/e2e/swarm_test.go
shaominchen: I plan to address this separately. I will enable this whenever I figure out the root cause.

tests/utils/dockercli/containers.go
shaominchen: We can't use that util here. That util writes data to a volume by running a **new** container. But this util writes data to an **existing** container scheduled by swarm cluster.

tests/utils/dockercli/containers.go
shaominchen: Same as above.

tests/utils/esx/govc.go
shaominchen: Actually I don't really follow the comments at L#20. I think those one line log messages are pretty useful. Anyway, I will comment it out just for local consistency for now. Later we should re-enable it after introducing logging levels.

tests/e2e/swarm_test.go
shuklanirdesh82: sounds good! Please add the issue# here or in the code for the future reference.

tests/utils/dockercli/containers.go
shuklanirdesh82: I see. It is better to call this out at the method documentation to avoid such confusion in the future. 

nit: `write data to a file in the running container` is not sufficient in my opinion ... something like `this util writes data from an existing running container  as per passed container name` in short some more elaborative.

tests/e2e/swarm_test.go
shaominchen: Filed issue #1505

tests/utils/dockercli/containers.go
shaominchen: I will update both - something like this:

// WriteToVolume write data to a given file on given volume by starting a new container
vs.
// WriteToContainer write data to a file in the existing running container

Issue #1518, opened 2017-06-29
Supply a fix to wait until vm recovers fully 


shuklanirdesh82:
Need to fix common util to wait until vm recovers from restart or shutdown abruptly.

```
2017/06/29 04:54:16 END: SwarmTestSuite.TestFailoverAcrossSwarmNodes
2017/06/29 04:54:16 Destroying volume [swarm_test_volume_544544]
2017/06/29 04:54:18 Initializing the SingleNode Config DB on esx [192.168.31.62] 
2017/06/29 04:54:23 Vmgroup vmgroup_test1 does not exist
2017/06/29 04:54:23 Creating test vmgroup vmgroup_test1
2017/06/29 04:54:25 Adding VM [photon.vmfs] to a vmgroup [vmgroup_test1] on esx [192.168.31.62] 
2017/06/29 04:54:28 Adding VM [govc: no such VM] to a vmgroup [vmgroup_test1] on esx [192.168.31.62]...

the node hasn't came back fully ... I believe SwarmTestSuite.TestFailoverAcrossSwarmNodes .. checked in yesterday and causing this failure .. it seems the vm is not fully recovered has the IP is not coming back correctly
```

ashahi1:
Closing this as I have provided the fix in PR #1519

PR #1519, opened 2017-06-29
Check if vDVS is running after vm power-on

ashahi1:
Fixes issue #1518 

After vm is powered-on, checking if vDVS is running.

```
=> Running target test-e2e-runalways
OK: 6 passed
--- PASS: Test (227.47s)
PASS

=> Running target test-e2e-runonce
OK: 26 passed
--- PASS: Test (984.70s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	984.700s
```


shaominchen:
@ashahi1 I think we have agreed that the fix is to improve the util function WaitForExpectedState. Otherwise, the same issue will happen for other test cases where we need to kill or restart VMs.

ashahi1:
@shaominchen https://github.com/vmware/docker-volume-vsphere/issues/1521

PR #1523, opened 2017-06-29
    Add support of new plugin/driver "vsphere-shared".

luomiao:
To build the shared plugin: make build-shared
Only docker-based build is supported for now.
Default config file path: /etc/docker-volume-shared.conf
Default log file path: /var/log/docker-volume-shared.log


shivanshu21:
Suggestion: Would it make sense to call this driver "kahuna" to avoid confusion with vmdk / vsphere / vsphere-shared? Using homophones can be confusing.

luomiao:
I don't think so...
Kahuna is an internal project codename.
It should not show up in any open-source code :)
Although it may make sense to change from vsphere-shared to vsphere-global.

shuklanirdesh82:
@luomiao can you please post some reference from the log impacting your change?

It would be good to have "Testing Done" section, I guess request is to merge to vDVS master so providing comment as per our code contributing guide line.

Thanks!

msterin:
Can you elaborate why this is needed ?

luomiao:
@msterin 
We need this to enable the global plugin.
With this in, Shivanshu and I will be able to parallel working on different functions and make separate PRs.
And according to the guideline, we need to cut the milestones into fine-grained check-ins :)

Or do you prefer we first working on another repo and then submitting PRs in group?

msterin:
What do you mean "enable" ? I thought it would be a different plugin ?

luomiao:
Yes this is a different plugin.
I might used the wrong word.
We need this to add the global plugin.
This is like the current vmdk/photon plugin.
User can set the config file with
"Driver":"vsphere/photon/vsphere-global" to choose which plugin to install.

govint:
That would make it a new driver in the existing plugin, is the idea to have a new plugin all together? Which means it can be separately installed along with the vsphere plugin.

luomiao:
Sorry about the confusion.
I was under the impression that driver = plugin.
So let me clarify.
There are two options then:

**1. Use the same plugin but add a new driver type.**
**What users need to do:**
```
docker plugin install --grant-all-permissions --alias vsphere vmware/docker-volume-vsphere:latest
docker plugin install --grant-all-permissions --alias vsphere-global vmware/docker-volume-vsphere:latest --config /etc/docker-volume-vsphere-global.conf
```
Content of docker-volume-vsphere-global.conf should include:
```
"Driver": "vsphere-global"
```
**What we need to do:**
Add new driver under vmdk_plugin/drivers. It shares the same main.go as existing drivers.
Keep the same binary and one Dockerfile as before.

**2. Use a different plugin**
**What users need to do:**
```
docker plugin install --grant-all-permissions --alias vsphere vmware/docker-volume-vsphere:latest
docker plugin install --grant-all-permissions --alias vsphere-global vmware/docker-volume-vsphere-global:latest 
```
Users don't need to provide a separate config files.
**What we need to do:**
Create another global_plugin. Generate a different binary and docker image.

govint:
Option 2 was what came up as possibly preferred in the review of the NFS support some time back.

luomiao:
OK!
Let me update the PR then.

msterin:
@luomiao - should this one be closed ?

Issue #1463, opened 2017-06-22
[E2E] test-e2e should allow running test against testbed where only one datastore is present.


shuklanirdesh82:
Test execution should fail gracefully instead of `invalid memory address or nil pointer dereference`

```
=> Running target test-e2e-runalways Thu Jun 22 17:41:22 UTC 2017

GO15VENDOREXPERIMENT=1 go test -v -timeout 30m -tags runalways github.com/vmware/docker-volume-vsphere/tests/e2e 
=== RUN   Test

----------------------------------------------------------------------
PANIC: volumeproperty_test.go:78: VolumePropertyTestSuite.TearDownSuite

... Panic: runtime error: invalid memory address or nil pointer dereference (PC=0x4582EC)

/usr/local/go/src/runtime/panic.go:458
  in gopanic
/usr/local/go/src/runtime/panic.go:62
  in panicmem
/usr/local/go/src/runtime/sigpanic_unix.go:24
  in sigpanic
volumeproperty_test.go:79
  in VolumePropertyTestSuite.TearDownSuite
/usr/local/go/src/reflect/value.go:302
  in Value.Call
/usr/local/go/src/runtime/asm_amd64.s:2086
  in goexit
--- FAIL: Test (5.72s)
FAIL
OOPS: 0 passed, 1 skipped, 1 FIXTURE-PANICKED
exit status 1
FAIL	github.com/vmware/docker-volume-vsphere/tests/e2e	5.728s
make: *** [test-e2e-runalways] Error 1
make: Leaving directory `/go/src/github.com/vmware/docker-volume-vsphere/vmdk_plugin'
make[1]: *** [test-e2e] Error 2
make: *** [test-e2e] Error 2
```

shuklanirdesh82:
minor improvement is required at inputparams.go
1. need to fix condition `len(config.Datastores) < 1 {` (test-e2e should run against 1datastsore setup and let the test to decide whether stop execution incase of insufficient datastore.

2. log the error message about insufficient datastore

```
if config.DockerHostNames[0] == "" || config.DockerHostNames[1] == "" ||
		len(config.Datastores) < 1 {
		return nil
	}
```

ashahi1:
Fixed as part of #1530

Issue #1398, opened 2017-06-12
[E2E] automated e2e tests should fail gracefully when test bed setup is not ideal or some environment vars are not set


shuklanirdesh82:
With the swarm test automation and govc integration, there are bunch of env variables need to be set inorder to execute tests locally. In an absence of ideal testbed or env. vars, tests are failing with out giving proper error message. It creates lot of confusion to the user and ends up spending more time. 

This issue is to streamline the failure and provides meaningful information to user when run into test setup issues.

- remove unnecessary environment variables from contributing.md too. 

/CC @msterin

shaominchen:
Here's my suggestion (copied from #1517 )

1. Verify the required environment variables are properly initialized before running e2e test
2. Clean up the Config DB if it already exists before running each test group
3. Maybe more...

shuklanirdesh82:
@ashahi1  is already having partial fix so bringing back to his plate

ashahi1:
Fixed as part of PR #1530

Issue #1515, opened 2017-06-29
[Test]Simplify the testbed for basic precheckin test


shaominchen:
Currently the precheckin tests (i.e. "test-all" target) requires a setup with at least 3 nodes deployed in swarm cluster mode. This is too heavy for a basic precheckin test. For internal developers, we can use nimbus to deploy a testbed with one single command. But for external developers, it will be too complicated to set up the environment simply for running the precheckin test.

On the other hand, we also want to make sure our CI can capture most common issues by running all important e2e test cases. Also we should still provide option to run any complicated test cases locally with some optional parameters.

To achieve these different goals, here's our plan:

1. Developers should be able to run test-all on a single VM. This is achieved by moving all extensive test cases (which require complicated setup) to one or more special labels
2. Developers should be able to run complicated test case by specifying some optional parameters. The assumption is that the environment should be ready for running those tests
3. CI should continue to run all extensive test cases as we have already planned

@msterin @shuklanirdesh82

shuklanirdesh82:
@ashahi1, @shaominchen & I discussed following approach while invoking `make test-all` locally.

1. Basic testbed (1 ESX + 1 VM) : run tests eligible for such configuration and print out nice message about test run and suggesting to use recommended testbed.

2. Recommended testbed (1 ESX + 2 VMs): makes sure required parameters (env. variables) are set correctly and run tests otherwise stop the execution immediately before running any test.

3. Special need like Swarm testbed (as of now): Such specific test should assert out if the test bed need doesn't satisfy without blocking other test run. Such test setup should provide sufficient error message while asserting out (should not use `log.fatal()`)

govint:
@shuklanirdesh82 - is this issue about identifying tests that are good to run with 1 VM and 1 ESX? Make a separate target for those to be run?

There are two issues here::

1. Using 1VM + 1ESX vs. 1ESX + 3VMs - I'm not sure why this is needed, a developer who can create 1VM + 1ESX can create two more VMs, set up a swarm and run their tests. And most likely those set ups will live for a long enough to be reused.

2. Identifying a "basic test set" - The issue with identifying a set of tests as "basic" is that soon they aren't enough to validate the code being checked in. Dev. is submitting code that the pre-checkin or basic tests aren't testing.

We can instead identify a set for regression that must be run before posting to CI perhaps. That set can exercise core workflows.

Even then, reducing the testbed to 1VM + 1ESX seems too constraining and I'm not sure of the value.

[...]

PR #1530, opened 2017-06-30
Fix to fail e2e tests gracefully if setup is not ideal

ashahi1:
Fixes issue #1463 and #1398

1. Added a fix to ensure E2E tests will run even with one datastore - #1463
2. Added the fix to verify the minimum required environment variables are properly initialized before running e2e test else fail the test run.
3. Made changes to inputparams.GetTestConfig() so we have only one instance of config and not each E2E test creating a new instance.

Tested locally: All tests passed other than swarm tests as there was no swarm configured
```
=> Running target test-e2e-runalways
OK: 6 passed
--- PASS: Test (120.40s)
PASS

=> Running target test-e2e-runonce
--- FAIL: Test (601.62s)
FAIL
OOPS: 23 passed, 3 skipped, 1 FAILED, 2 MISSED
exit status 1
FAIL	github.com/vmware/docker-volume-vsphere/tests/e2e	601.626s
make: *** [test-e2e-runonce] Error 1
```

### DIFF
--- a/tests/e2e/swarm_test.go
+++ b/tests/e2e/swarm_test.go
@@ -46,11 +46,11 @@ type SwarmTestSuite struct {
 func (s *SwarmTestSuite) SetUpSuite(c *C) {
 	s.esxName = inputparams.GetEsxIP()
 	s.master = inputparams.GetSwarmManager1()
-	c.Assert(s.master, Not(IsNil), Commentf("Master node is nil"))
+	c.Assert(s.master, Not(IsNil), Commentf("Master node is not set."))
 	s.worker1 = inputparams.GetSwarmWorker1()
-	c.Assert(s.worker1, Not(IsNil), Commentf("Worker1 node is nil"))
+	c.Assert(s.worker1, Not(IsNil), Commentf("Worker1 node is not set."))
 	s.worker2 = inputparams.GetSwarmWorker2()
-	c.Assert(s.worker2, Not(IsNil), Commentf("Worker2 node is nil"))
+	c.Assert(s.worker2, Not(IsNil), Commentf("Worker2 node is not set."))
 	s.swarmNodes = inputparams.GetSwarmNodes()
 
 	// Verify if swarm cluster is already initialized
--- a/tests/e2e/swarm_test.go
+++ b/tests/e2e/swarm_test.go
@@ -46,8 +46,11 @@ type SwarmTestSuite struct {
 func (s *SwarmTestSuite) SetUpSuite(c *C) {
 	s.esxName = inputparams.GetEsxIP()
 	s.master = inputparams.GetSwarmManager1()
+	c.Assert(s.master, Not(IsNil), Commentf("Master node is nil"))
 	s.worker1 = inputparams.GetSwarmWorker1()
+	c.Assert(s.worker1, Not(IsNil), Commentf("Worker1 node is nil"))
 	s.worker2 = inputparams.GetSwarmWorker2()
+	c.Assert(s.worker2, Not(IsNil), Commentf("Worker2 node is nil"))
 	s.swarmNodes = inputparams.GetSwarmNodes()
 
 	// Verify if swarm cluster is already initialized
--- a/tests/utils/inputparams/testparams.go
+++ b/tests/utils/inputparams/testparams.go
@@ -42,7 +42,7 @@ var (
 	endPoint1  string
 	endPoint2  string
 	volumeName string
-	mu         sync.Mutex
+	once       sync.Once
 	config     *TestConfig
 )
 
@@ -102,27 +102,18 @@ func GetEndPoint2() string {
 // GetSwarmManager1 returns swarm manager node IP from the configured swarm cluster
 func GetSwarmManager1() string {
 	manager1 := os.Getenv("MANAGER1")
-	if manager1 == "" {
-		log.Printf("Manager node not found. Need this to run swarm test.")
-	}
 	return manager1
 }
 
 // GetSwarmWorker1 returns 1st swarm worker node IP from the configured swarm cluster
 func GetSwarmWorker1() string {
 	worker1 := os.Getenv("WORKER1")
-	if worker1 == "" {
-		log.Printf("Worker1 node not found. Need this to run swarm test.")
-	}
 	return worker1
 }
 
 // GetSwarmWorker2 returns 2nd swarm worker node IP from the configured swarm cluster
 func GetSwarmWorker2() string {
 	worker2 := os.Getenv("WORKER2")
-	if worker2 == "" {
-		log.Printf("Worker2 node not found. Need this to run swarm test.")
-	}
 	return worker2
 }
 
@@ -149,7 +140,7 @@ func getInstance() *TestConfig {
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM1"))
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM2"))
 	if config.DockerHosts[0] == "" && config.DockerHosts[1] == "" {
-		log.Fatal("No docker hosts found. Atleast one host is needed to run tests.")
+		log.Fatal("No docker hosts found. At least one host is needed to run tests.")
 	}
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[0]))
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[1]))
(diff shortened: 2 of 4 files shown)


tests/e2e/swarm_test.go
```diff
@@ -46,8 +46,11 @@ type SwarmTestSuite struct {
 func (s *SwarmTestSuite) SetUpSuite(c *C) {
 	s.esxName = inputparams.GetEsxIP()
 	s.master = inputparams.GetSwarmManager1()
+	c.Assert(s.master, Not(IsNil), Commentf("Master node is nil"))
```
shuklanirdesh82: nit: Please improve the error message like `Master node is not set`

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,68 @@ func GetEndPoint2() string {
...
 // GetTestConfig - returns the configuration of IPs for the
 // ESX host, docker hosts and the datastores on the host
-func GetTestConfig() *TestConfig {
-	var config *TestConfig
-
+func getInstance() *TestConfig {
+	noVMName := "no such VM"
 	config = new(TestConfig)
 	config.EsxHost = os.Getenv("ESX")
+	if config.EsxHost == "" {
```
shuklanirdesh82: as discussed offline .. should be replaced by `GetEsxIP()`

tests/e2e/swarm_test.go
lipingxue: +1

tests/e2e/swarm_test.go
```diff
@@ -46,8 +46,11 @@ type SwarmTestSuite struct {
...
 	s.worker1 = inputparams.GetSwarmWorker1()
+	c.Assert(s.worker1, Not(IsNil), Commentf("Worker1 node is nil"))
```
lipingxue: Same as above. "Worker1 node is not set".

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,68 @@ func GetEndPoint2() string {
...
+		log.Fatal("ESX host not found. Stopping the test run.")
+	}
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM1"))
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM2"))
+	if config.DockerHosts[0] == "" && config.DockerHosts[1] == "" {
+		log.Fatal("No docker hosts found. Atleast one host is needed to run tests.")
+	}
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[0]))
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[1]))
+	if config.DockerHostNames[0] == noVMName && config.DockerHostNames[1] == noVMName {
```
lipingxue: So do we allow the test to run if only one VM is set?

ashahi1: Yes, we will allow it. It will be taken care as part of PR #1515

tests/e2e/swarm_test.go
ashahi1: Done.

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,65 @@ func GetEndPoint2() string {
...
+		log.Fatalf("No names found for docker hosts - %s , %s ", config.DockerHosts[0], config.DockerHosts[1])
+	}
 	config.Datastores = esx.GetDatastoreList()
-
-	if config.DockerHostNames[0] == "" || config.DockerHostNames[1] == "" ||
-		len(config.Datastores) <= 1 {
-		return nil
+	if len(config.Datastores) < 1 {
+		log.Fatalf("No datastores found. Atleast one datastore is needed to run tests.")
```
shaominchen: Nit: Atleast => At least

tests/utils/inputparams/testparams.go
shaominchen: If we allow one VM setup, we should just check one host name here.

tests/utils/inputparams/testparams.go
```diff
@@ -153,3 +171,15 @@ func GetRandomNumber() string {
 	bytes := min + rand.Intn(max)
 	return strconv.Itoa(int(bytes))
 }
+
+// GetTestConfig - Creates one instance of TestConfig
+func GetTestConfig() *TestConfig {
+	if config == nil {
```
shaominchen: An Idiomatic Singleton Approach in Go is using sync.Once:

```go
import (
    "sync"
)

type singleton struct {
}

var instance *singleton
var once sync.Once

func GetInstance() *singleton {
    once.Do(func() {
        instance = &singleton{}
    })
    return instance
}
```

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,65 @@ func GetEndPoint2() string {
 
 // GetSwarmManager1 returns swarm manager node IP from the configured swarm cluster
 func GetSwarmManager1() string {
-	return os.Getenv("MANAGER1")
+	manager1 := os.Getenv("MANAGER1")
+	if manager1 == "" {
+		log.Printf("Manager node not found. Need this to run swarm test.")
```
ashahi1: @shaominchen I have removed the log.Printf from here and keeping the assert in the swarm test. I do not think we should change it to log.Fatal as in that case, whole test run will stop. And keeping assert in swarm test will only stop the swarm test and not the whole test run.

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,65 @@ func GetEndPoint2() string {
...
-
+func getInstance() *TestConfig {
+	noVMName := "no such VM"
 	config = new(TestConfig)
-	config.EsxHost = os.Getenv("ESX")
+	config.EsxHost = GetEsxIP()
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM1"))
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM2"))
+	if config.DockerHosts[0] == "" && config.DockerHosts[1] == "" {
+		log.Fatal("No docker hosts found. Atleast one host is needed to run tests.")
```
ashahi1: Done

tests/utils/inputparams/testparams.go
ashahi1: Done

tests/utils/inputparams/testparams.go
shaominchen: That's not the case. After addressing #1515, swarm test and other complicated tests will be moved out of the basic precheckin test suite. So it doesn't matter if we use log.Fatal() or not here. When swarm test and other complicated tests are REQUIRED to run - for example, in CI - we still want to fail the whole test if these environment variables are not available.

tests/utils/inputparams/testparams.go
ashahi1: @shaominchen It was one of the approaches mentioned here: http://marcio.io/2015/07/singleton-pattern-in-go/
But looks like the preferred approach is to use sync.once. 
Will make this change.

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,56 @@ func GetEndPoint2() string {
 
 // GetSwarmManager1 returns swarm manager node IP from the configured swarm cluster
 func GetSwarmManager1() string {
-	return os.Getenv("MANAGER1")
+	manager1 := os.Getenv("MANAGER1")
```
shaominchen: let's revert this change - no need to introduce a local variable here. Same for below.

tests/utils/inputparams/testparams.go
```diff
@@ -96,52 +101,56 @@ func GetEndPoint2() string {
...
-	config.EsxHost = os.Getenv("ESX")
+	config.EsxHost = GetEsxIP()
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM1"))
 	config.DockerHosts = append(config.DockerHosts, os.Getenv("VM2"))
+	if config.DockerHosts[0] == "" && config.DockerHosts[1] == "" {
+		log.Fatal("No docker hosts found. At least one host is needed to run tests.")
+	}
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[0]))
 	config.DockerHostNames = append(config.DockerHostNames, esx.RetrieveVMNameFromIP(config.DockerHosts[1]))
+	if config.DockerHostNames[0] == noVMName && config.DockerHostNames[1] == noVMName {
```
shaominchen: The AND logic here should be OR, since we want to check both. Please fix the logs message below appropriately.

tests/utils/inputparams/testparams.go
ashahi1: Done

tests/utils/inputparams/testparams.go
ashahi1: Done

PR #1534, opened 2017-07-03
Add support of new global plugin compilation.

luomiao:
Plugin name: global
Driver name: global
Default config file path: /etc/docker-volume-global.conf
Default log file path: /var/log/docker-volume-global.log
To build the new global plugin: make build-global
Only docker-based build is supported for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,11 +42,16 @@ build-all: dockerized-build-ui
 	$(MAKE) --directory=vmdk_plugin $@
 	$(MAKE) --directory=plugin all
 
+build-global: dockerized-build-ui
+	$(MAKE) --directory=global_plugin build-all
+	$(MAKE) --directory=global_plugin_docker all
+
 # clean inside docker run to avoid sudo make clean
 # dev builds are inside docker which creates folders
 # as root.
 clean: dockerized-clean-ui
 	$(MAKE) --directory=vmdk_plugin $@
+	$(MAKE) --directory=global_plugin $@
 
 # Non dockerized build, used by CI
 build:
@@ -54,6 +59,7 @@ ifeq ($(INCLUDE_UI), true)
 	$(MAKE) --directory=ui $@
 endif
 	$(MAKE) --directory=vmdk_plugin $@
+	$(MAKE) --directory=global_plugin $@
 
 # Forward to UI inside docker run
 dockerized-build-ui:
--- a/global_plugin/Makefile
+++ b/global_plugin/Makefile
@@ -1,0 +1,140 @@
+# Copyright 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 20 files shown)


Makefile
```diff
@@ -42,18 +42,24 @@ build-all: dockerized-build-ui
 	$(MAKE) --directory=vmdk_plugin $@
 	$(MAKE) --directory=plugin all
 
+build-global: dockerized-build-ui
```
msterin: "global" is very confusing. "Shared" is way better. Yes, I know all is shared on one VM, but it's easy to explain in comments

global_plugin/Makefile
```diff
@@ -0,0 +1,140 @@
+# Copyright 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
msterin: This seems to be cut n paste of plugin makefiles and I needs rework, more below

global_plugin/Makefile
```diff
@@ -0,0 +1,140 @@
...
+# all binaries for VMs - plugin and tests
+#VM_BINS = $(PLUGIN_BIN) $(BIN)/$(PLUGNAME).test
+VM_BINS = $(PLUGIN_BIN)
+
+# plugin name, for go build
+PLUGIN := github.com/$(GOPATH_ORG)/$(GOPATH_PLUGNAME)
+
+GO := GO15VENDOREXPERIMENT=1 go
+
+# All sources. We rebuild if anything changes here
```
msterin: Plugin should use go modules a thus go build should take care if dependencies .

global_plugin/drivers/driver.go
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package drivers
+
+// VolumeDriver interface used by the refcountedVolume module to handle
+// recovery mounts/unmounts.
+type VolumeDriver interface {
```
msterin: Why it has to be re defined ? Can't we reuse the existing ?

luomiao:
That's why I preferred option one :)

I did a little bit touch of current existing code such as moving the log_formatter into utils, in order to reuse the current code.
To reuse the windows related code across different plugins, we may need further re-org of current code structure.

global_plugin/drivers/driver.go
luomiao: You're right. This one is not needed.

govint:
@luomiao Can't we use a single code base and build two separate plugins? It can share a lot of the code, they don't need to be complete separate code bases. Parts of the code will become shared packages, especially the E2E tests have to be reusable

luomiao:
@govint Yep this is what we are planning to do now.
In fact I already updated the other PR:
https://github.com/vmware/docker-volume-vsphere/pull/1523
But somehow it's not triggering the auto-testings.
I can submit another PR if needed.
Can you take a look at the updated PR 1523? Thanks!

govint:
@luomiao can you post the changes with the common code separated out so we know which is the code and the new plugin. Also, better to first post a write up of the code level separation of the new plugin and check that in first so we have a reference to understand the changes.

govint:
May be also disable CI till the code is in the shape as discussed and then re-enable it.

PR #1539, opened 2017-07-05
Add support of new plugin/driver "vsphere-shared".

luomiao:
**Layout after the PR:**
vmdk_plugin now contains two plugins: original_plugin and shared_plugin.
Both of the two plugins have their own main.go in order to support building separate binaries.
A new shared driver is also added into vmdk_plugin/drivers, which shares the same VolumeDriver interface as vmdk/photon drivers.

A few utility functions are separated out from original plugin and moved under vmdk_plugin/utils since they are used by both original and shared plugins:
**log_formatter**: for VmwareFormatter struct and logInit functions
**plugin_server**: for PluginServer struct and SockPluginServer for Linux and NpipePluginServer for windows.

Both original_plugin and shared_plugin share the same Makefile.
The shared_plugin only supports docker-based build for now.
To build original_plugin: follow the current build steps.
To build shared_plugin: `make build-shared`

Default config file path for shared_plugin: /etc/docker-volume-shared.conf
Default log file path for shared_plugin: /var/log/docker-volume-shared.log

The Dockerfiles of docker plugin are also separated into two sub-directories: plugin/original and plugin/shared, while sharing the same Makefile.
To build original plugin: `make all`
To build shared plugin: `make shared-all`
Original plugin will be pushed into docker hub repo: `$(DOCKER_HUB_REPO)/docker-volume-vsphere`
Shared plugin will be pushed into docker hub repo: `$(DOCKER_HUB_REPO)/docker-volume-shared`

**Installation step**:
Original plugin: the same as before
Shared plugin: `docker plugin install --grant-all-permissions --alias shared [DOCKER_HUB_REPO]/docker-volume-shared:[TAG-EXTRATAG]`


plugin/Makefile
```diff
@@ -60,10 +60,12 @@ PLUGIN_TAG := $(VERSION_TAG)$(EXTRA_TAG)
 
 # plugin name - used as a base for full plugin name and container for extracting rootfs
 PLUGIN_NAME=$(DOCKER_HUB_REPO)/docker-volume-vsphere
+SHARED_PLUGIN_NAME=$(DOCKER_HUB_REPO)/docker-volume-shared
```
venilnoronha: Should it be `docker-volume-vsphere-shared` here and at other places?

plugin/shared/config.json
```diff
@@ -0,0 +1,65 @@
...
+		"AllowAllDevices": true,
+		"Capabilities": ["CAP_SYS_ADMIN"],
+		"Devices": null
+
+	},
+	"Env": [ {
+      "name": "VDVS_LOG_LEVEL",
+      "description": "Log level - info, warn, error or debug",
+      "value": "info",
+      "Settable": [ "value"]
```
venilnoronha: nit: fix indentation.

plugin/shared/config.json
```diff
@@ -0,0 +1,65 @@
...
+	},
+	{
+      "name": "VDVS_TEST_PROTOCOL_VERSION",
+      "description": "Client protocol version used in test",
+      "value": "",
+      "Settable": [ "value"]
```
venilnoronha: nit: fix indentation.

plugin/shared/config.json
```diff
@@ -0,0 +1,65 @@
...
+	},
+	"Interface" : {
+		"Types": ["docker.volumedriver/1.0"],
+		"Socket": "shared.sock"
+	},
+	"Linux": {
+		"AllowAllDevices": true,
+		"Capabilities": ["CAP_SYS_ADMIN"],
+		"Devices": null
+
```
venilnoronha: nit: remove empty line.

vmdk_plugin/drivers/shared/shared_driver.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return d
+}
+
+// In following three operations on refcount, if refcount
+// map hasn't been initialized, return 1 to prevent detach and remove.
+
+// Return the number of references for the given volume
+func (d *VolumeDriver) getRefCount(vol string) uint {
+	if d.refCounts.IsInitialized() != true {
```
venilnoronha: nit: `!= true` isn't needed here and at other places.

vmdk_plugin/utils/log_formatter/log_formatter.go
```diff
@@ -17,12 +17,15 @@
...
 
-package main
+package log_formatter
 
 import (
 	"bytes"
 	"fmt"
 	log "github.com/Sirupsen/logrus"
+	"github.com/natefinch/lumberjack"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config"
```
venilnoronha: nit: better to keep vendor imports separate from native ones.

vmdk_plugin/shared_plugin/main.go
```diff
@@ -0,0 +1,111 @@
...
+	// config file or the default.
+	if *driverName == "" {
+		if err == nil && c.Driver != "" {
+			*driverName = c.Driver
+		} else {
+			*driverName = sharedDriver
+		}
+	}
+
+	// The windows plugin only supports the vsphere driver.
```
venilnoronha: We could remove this block entirely as we only support a single driver.

vmdk_plugin/shared_plugin/constants.go
```diff
@@ -0,0 +1,31 @@
...
+	// Docker driver types.
+	volumeDriver = "VolumeDriver"
+
+	// Docker plugin handshake endpoint.
+	// Also see https://docs.docker.com/engine/extend/plugin_api/#handshake-api
+	pluginActivatePath = "/Plugin.Activate"
+
+	// Docker volume plugin endpoints.
+	// Also see https://docs.docker.com/engine/extend/plugins_volume/#volume-plugin-protocol
+	volumeDriverCreatePath = "/VolumeDriver.Create"
```
venilnoronha: We wouldn't need `pluginActivatePath` and `volumeDriverCreatePath ` here as they're only used by the `plugin_server` module.

vmdk_plugin/utils/config/config_windows.go
```diff
@@ -24,4 +24,8 @@ var (
 	DefaultConfigPath = filepath.Join(os.Getenv("PROGRAMDATA"), "docker-volume-vsphere", "docker-volume-vsphere.conf")
 	// DefaultLogPath is the default location of the log (trace) file.
 	DefaultLogPath = filepath.Join(os.Getenv("LOCALAPPDATA"), "docker-volume-vsphere", "logs", "docker-volume-vsphere.log")
+	// DefaultSharedPluginConfigPath is the default location of configuration file for shared plugin
+	DefaultSharedPluginConfigPath = filepath.Join(os.Getenv("PROGRAMDATA"), "docker-volume-shared", "docker-volume-shared.conf")
+	// DefaultSharedPluginLogPath is the default location of log (trace) file for shared plugin
+	DefaultSharedPluginLogPath = filepath.Join(os.Getenv("LOCALAPPDATA"), "docker-volume-shared", "logs", "docker-volume-shared.log")
```
venilnoronha: Are we planning to support the shared plugin on Windows too? If it isn't planned currently then we could remove `DefaultSharedPluginConfigPath` and `DefaultSharedPluginConfigPath`.

vmdk_plugin/utils/log_formatter/log_formatter.go
```diff
@@ -83,3 +86,51 @@ func (f *VmwareFormatter) appendKeyValue(b *bytes.Buffer, key string, value inte
...
+			config.SetDefaults(&c)
+		} else {
+			panic(fmt.Sprintf("Failed to load config file %s: %v",
+				*configFile, err))
+		}
+	}
+
+	path := c.LogPath
+	if path == "" {
+		path = config.DefaultLogPath
```
venilnoronha: `DefaultSharedPluginLogPath` isn't used?

vmdk_plugin/utils/log_formatter/log_formatter.go
luomiao: I guess this is the style that is used by vDVS so far :)
But to double-check, what do you mean by keeping vendor imports separate?

vmdk_plugin/drivers/shared/shared_driver.go
luomiao: It's needed to prevent operations to the volumes when they are still in initialization stage.

vmdk_plugin/utils/log_formatter/log_formatter.go
luomiao: Thanks for catching this.
I lost this change when refactoring this LogInit function.
will update it.

vmdk_plugin/shared_plugin/constants.go
luomiao: Do you think it's ok to move this const into plugin_server_windows.go?

vmdk_plugin/utils/config/config_windows.go
luomiao: Let me remove them for now.

vmdk_plugin/drivers/shared/shared_driver.go
venilnoronha: I meant `if d.refCounts.IsInitialized() {` would be enough. Sorry for the confusion.

vmdk_plugin/utils/log_formatter/log_formatter.go
venilnoronha: I meant that we could add a line break before vendor imports. :)

vmdk_plugin/shared_plugin/constants.go
venilnoronha: I had put it in the `main_windows.go` file earlier, but since these weren't platform specific constants we decided to put them in `constants.go`. Here, we could have a  `vmdk_plugin/utils/plugin_server/constants.go` file and move `pluginActivatePath` and `volumeDriverCreatePath` there.

plugin/Makefile
msterin: Well, to  be super accurate it needs to be docker-volume-vsphere-shared but it's a mouthful. 
maybe simply vsphere-shared ? "docker-volume-" seems redundand

plugin/shared/config.json
msterin: I suspect it's github tab handing, not the actual indentation issues

vmdk_plugin/utils/log_formatter/log_formatter.go
msterin: Whatever gofmt is happy with :-)

vmdk_plugin/utils/config/config_windows.go
msterin: If all goes well we will look into supporting  it on Windows in v2, though it will use native SMB I'd guess.

plugin/shared/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Dockerile for packaging https://github.com/vmware/docker-volume-vsphere as
+# Docker managed plugin.
```
msterin: it looks identical to non-shared, why do we need a copy ?

plugin/shared/config.json
```diff
@@ -0,0 +1,65 @@
+{
+	"Description": "VMWare vSphere Shared Docker Volume plugin",
```
msterin: I think it's better to use 1 json and just sed the difference before building plugin image.
Having 2 copies is error prone

vmdk_plugin/Makefile
```diff
@@ -95,10 +98,18 @@ VMDKOPS_TEST_MODULE := vmdkops
 VMDKOPS_MODULE_SRC = $(VMDKOPS_MODULE)/*.go $(VMCI_SRC)
 
 # All sources. We rebuild if anything changes here
-SRC = main.go log_formatter.go utils/refcount/refcnt.go \
+SRC = original_plugin/main.go original_plugin/constants.go \
```
msterin: "original" is not semantically clear name. We can call it "block_device" or something along these lines, but not "original"

vmdk_plugin/Makefile
```diff
@@ -95,10 +98,18 @@ VMDKOPS_TEST_MODULE := vmdkops
...
+	utils/refcount/refcnt.go utils/log_formatter/log_formatter.go \
+	utils/plugin_server/plugin_server.go \
 	utils/fs/fs.go utils/config/config.go utils/plugin_utils/plugin_utils.go\
 	drivers/photon/photon_driver.go drivers/vmdk/vmdk_driver.go
 
+SHARED_PLUGIN_SRC = shared_plugin/main.go shared_plugin/constants.go \
```
msterin: can we use $(subst) to build SHARED_PLUGIN_SRC instead of copy and paste ? https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

vmdk_plugin/drivers/shared/shared_driver.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/docker/go-plugins-helpers/volume"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/plugin_utils"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/refcount"
+)
+
+const (
+	version = "vSphere Volume Shared Driver v0.1"
+)
+
+// VolumeDriver - vsphere shared volume driver struct
```
msterin: this does not look shared- or block- specific.

vmdk_plugin/drivers/shared/shared_driver.go
```diff
@@ -0,0 +1,209 @@
...
+		return volume.Response{Err: ""}
+	}
+
+	log.Errorf("VolumeDriver Unmount to be implemented")
+	return volume.Response{Err: ""}
+}
+
+// Capabilities - Report plugin scope to Docker
+func (d *VolumeDriver) Capabilities(r volume.Request) volume.Response {
+	return volume.Response{Capabilities: volume.Capability{Scope: "global"}}
```
msterin: SO I guess the text above will change for actual shared- driver and is cut-n-paste just do the split, right ?

vmdk_plugin/original_plugin/main.go
```diff
@@ -27,64 +27,13 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/go-plugins-helpers/volume"
-	"github.com/natefinch/lumberjack"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/drivers/photon"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/drivers/vmdk"
 	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/log_formatter"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/plugin_server"
```
msterin: Please do not use "original_", that will become confusing really fast

vmdk_plugin/shared_plugin/main.go
```diff
@@ -0,0 +1,111 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/go-plugins-helpers/volume"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/drivers/shared"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/config"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/log_formatter"
+	"github.com/vmware/docker-volume-vsphere/vmdk_plugin/utils/plugin_server"
+)
+
+// main for docker-volume-vsphere
+// Parses flags, initializes and mounts refcounters and finally initializes the server.
```
msterin: looks like the majority of the code is redundant in both drivers and should be separated in main_utils.go :-)

vmdk_plugin/utils/log_formatter/log_formatter.go
msterin: Is this code shared between 2 drivers?

Makefile
```diff
@@ -42,6 +42,10 @@ build-all: dockerized-build-ui
 	$(MAKE) --directory=vmdk_plugin $@
 	$(MAKE) --directory=plugin all
 
+build-shared: dockerized-build-ui
```
luomiao: lol we really need a short clear name!

plugin/shared/Dockerfile
luomiao: The binary is different.
I will update the docker plugin folder to merge duplicated files.

vmdk_plugin/utils/log_formatter/log_formatter.go
luomiao: Yes. if you mean drivers -> plugins.

vmdk_plugin/drivers/shared/shared_driver.go
luomiao: We will have more shared-specific things here later.
It's the same as the VolumeDriver struct used for photon-specific and block-specific driver.

govint:
@luomiao, is it possible to rename original_plugin as vmdk_plugin and retain the new one as "shared_plugin"? The top level dir can be just plugins/ maybe. Is there a command that will build all the plugins? Can _make_ build all plugins while _make vmdk_ or _make shared_ build the specific plugins?

vmdk_plugin/drivers/shared/shared_driver.go
luomiao: Yes

luomiao:
@govint There is already another plugin folder which holds the dockerfile for building docker plugins... We may need to think about another name for the top plugin code.
About the command to build all plugins: we can introduce this after the new shared plugin is complete to some degree. Now it's still in progress so I think it's better do not build all plugins together.

vmdk_plugin/utils/plugin_server/plugin_server_linux.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package plugin_server
```
govint: Why would the linux, windows, shared plugin servers get built into each plugin? Pls. separate into separate packages for vmdk, windows, shared. And instantiate the appropriate one from main().

vmdk_plugin/shared_plugin/main.go
```diff
@@ -0,0 +1,111 @@
...
+		log.Warning("Unknown driver or invalid/missing driver options, exiting - ", *driverName)
+		os.Exit(1)
+	}
+
+	if reflect.ValueOf(driver).IsNil() == true {
+		log.Warning("Error in driver initialization exiting - ", *driverName)
+		os.Exit(1)
+	}
+
+	server := plugin_server.NewPluginServer(*driverName, &driver)
```
govint: So, this will instantiate a vmdk or shared or windows plugin server? It should instantiate based on the selection in plugin_server.go.

plugin/Makefile
```diff
@@ -74,15 +76,21 @@ TMP_LOC := /tmp/plugin
 
 # default target
 all: info plugin push
+shared-all: shared-info shared-plugin shared-push
```
luomiao: I still suggest postpone this to a little bit later stage.
These changes will have to change the CI and testings too.
We don't want to disturb the testing code while the new shared plugin is still under development.
So, please think this new plugin as an experimental feature, which won't be a default choice for now :)

plugin/Makefile
```diff
@@ -91,18 +99,42 @@ clean:
...
+	rm -f $(SHARED_BINARY)
+	-docker plugin rm $(SHARED_PLUGIN_NAME):$(PLUGIN_TAG) -f
+	-docker rm -vf $(TMP_CONTAINER)
+	-docker rmi $(TMP_IMAGE)
+
 plugin: clean
 	@echo "== building Docker image, unpacking to ./rootfs and creating plugin..."
-	cp $(BIN_LOC)/$(BINARY) .
-	docker build -q -t $(TMP_IMAGE) .
+	cp $(BIN_LOC)/$(BINARY) ./original
```
luomiao: This will be updated soon. I have already merged the dockerfiles.

vmdk_plugin/drivers/shared/shared_driver.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	// Docker is supposed to block 'remove' command if the volume is used.
+	if d.getRefCount(r.Name) != 0 {
+		msg := fmt.Sprintf("Remove failure - volume is still mounted. "+
+			" volume=%s, refcount=%d", r.Name, d.getRefCount(r.Name))
+		log.Error(msg)
+		return volume.Response{Err: msg}
+	}
+
```
luomiao: This will be changed when we filled in the function.

vmdk_plugin/utils/plugin_server/plugin_server_linux.go
luomiao: Sorry I don't quite understand what you mean by separate into separate packages for vmdk, windows, shared? Linux/windows plugin_server are just platform specific compilation choice for plugin servers. It's getting compiled according to the platform where you build the code. It will be automatically instantiated. I guess you may understand this part of code wrong... Please ping me offline so we can discuss more about it. @venilnoronha

venilnoronha: There would be only 2 types of `plugin_server`s i.e. `SockPluginServer` for linux and `NpipePluginServer` for windows. We should be able to service `vmdk`, `photon` and `shared` drivers using these `plugin_server`s by providing the corresponding driver instance to `NewPluginServer(..)`. The `plugin_server`'s job is just to expose the driver APIs over unix sock / windows npipe.

govint: Sorry, wanted to rephrase the comment and missed. Basically, is it possible to ensure that only the relevant code is even compiled into the plugins. Like if all plugin servers are part of the plugin_server package does all of the code get included into the binary? Can that be avoided?

venilnoronha: From [go build tags](https://golang.org/pkg/go/build/) documentation:
```
Naming a file dns_windows.go will cause it to be included only when building the package for Windows; similarly, math_386.s will be included only when building the package for 32-bit x86.
```
I'm not sure if that answers your question.

ashahi1:
vsphere-shared name can be confusing for end-user. User can have questions like - do I need to use shared plugin to use shared datastores ? e.g. sharedVmfs

luomiao:
Replaced by:
https://github.com/vmware/docker-volume-vsphere/pull/1555
https://github.com/vmware/docker-volume-vsphere/pull/1567
https://github.com/vmware/docker-volume-vsphere/pull/1574

PR #1549, opened 2017-07-06
[Trivial] changed plugin staging place

msterin:
making sure managed plugin stage does not clash with /tmp/plugin used during CI build 
@shuklanirdesh82


msterin:
not needed, the actual tmp folder for build is /tmp/docker-volume-vsphere, so there is no clash

Issue #1378, opened 2017-06-08
Automate fail-over test for docker host created on vsan


shaominchen:
please refer to #957 Scenario 4.

shuklanirdesh82:
isn't it duplicate to #1377? If agree close this one by adding label `duplicate`, please explain otherwise.

shaominchen:
@shuklanirdesh82 This is for #957 Scenario 4, while #1377 is for #957 Scenario 3. They are different, right?

ashahi1:
Yes, scenario 3 and scenario 4 mentioned in #957 are different. Nirdesh can correct me If I am wrong.

shuklanirdesh82:
> This is for #957 Scenario 4, while #1377 is for #957 Scenario 3. They are different, right?

after modifying the description of this issue no more confusion. Earlier issue description was pointing to scenario 3.

PR #1550, opened 2017-07-06
Add failover test for volume created on VSAN

shaominchen:
Fixes issue #1378 

Testing Done:

OK: 31 passed
--- PASS: Test (922.68s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	922.686s


tests/e2e/vmlistener_test.go
```diff
@@ -178,11 +227,101 @@ func (s *VMListenerTestParams) TestFailoverAcrossVM(c *C) {
 	esxutil.PowerOnVM(s.vm1Name)
 	isStatusChanged := esxutil.WaitForExpectedState(esxutil.GetVMPowerState, s.vm1Name, properties.PowerOnState)
 	c.Assert(isStatusChanged, Equals, true, Commentf("VM [%s] should be powered on state", s.vm1Name))
+	c.Assert(verification.IsVDVSIsRunning(s.vm1), Equals, true, Commentf("vDVS is not running on [%s]", s.vm1Name))
 
 	// Status should be still detached
 	status = verification.VerifyDetachedStatus(s.volumeName, s.vm1, s.esx)
 	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", s.volumeName))
 
+	// Remove the volume
```
pshahzeb: Same comment to remove the container as above.

tests/e2e/vmlistener_test.go
```diff
@@ -178,11 +227,101 @@ func (s *VMListenerTestParams) TestFailoverAcrossVM(c *C) {
...
+	// Power on VM1 which has been killed
+	esxutil.PowerOnVM(s.vm1Name)
+	isStatusChanged := esxutil.WaitForExpectedState(esxutil.GetVMPowerState, s.vm1Name, properties.PowerOnState)
+	c.Assert(isStatusChanged, Equals, true, Commentf("VM [%s] should be powered on state", s.vm1Name))
+	c.Assert(verification.IsVDVSIsRunning(s.vm1), Equals, true, Commentf("vDVS is not running on [%s]", s.vm1Name))
+
+	// Status should be still detached
+	status = verification.VerifyDetachedStatus(s.volumeName, s.vm1, s.esx)
+	c.Assert(status, Equals, true, Commentf("Volume %s is not detached", s.volumeName))
+
```
pshahzeb: If the container on the first VM exists, remove it.

tests/e2e/vmlistener_test.go
```diff
@@ -127,29 +156,49 @@ func (s *VMListenerTestParams) TestBasicFailover(c *C) {
...
 	isStatusChanged := esxutil.WaitForExpectedState(esxutil.GetVMPowerState, s.vm1Name, properties.PowerOnState)
 	c.Assert(isStatusChanged, Equals, true, Commentf("VM [%s] should be powered on state", s.vm1Name))
+	c.Assert(verification.IsVDVSIsRunning(s.vm1), Equals, true, Commentf("vDVS is not running on [%s]", s.vm1Name))
 
 	// Status should be detached
-	status := verification.VerifyDetachedStatus(s.volumeName, s.vm1, s.esx)
+	status = verification.VerifyDetachedStatus(s.volumeName, s.vm1, s.esx)
 	c.Assert(status, Equals, true, Commentf("Volume %s is still attached", s.volumeName))
 
+	// Remove the volume
```
shaominchen: Yes, I have considered the same. However, this is not a must, right? Even if the container exists after VM restarts, the volume should still be in detached status, correct? So removing volume should not run into issue here.

I think both are acceptable. I keep it in TearDown just to avoid a little bit code duplication. Comments?

tests/e2e/vmlistener_test.go
```diff
@@ -56,20 +58,12 @@ func (s *VMListenerTestParams) SetUpSuite(c *C) {
...
-	c.Assert(err, IsNil, Commentf(out))
-
-	out, err = dockercli.AttachVolume(s.vm1, s.volumeName, s.containerName)
-	c.Assert(err, IsNil, Commentf(out))
-
-	status := verification.VerifyAttachedStatus(s.volumeName, s.vm1, s.esx)
-	c.Assert(status, Equals, true, Commentf("Volume %s is not attached", s.volumeName))
 }
 
 func (s *VMListenerTestParams) TearDownTest(c *C) {
```
shuklanirdesh82: this cleanup method is inconsistent .. I have seen the test is cleaning up created container ..

tests/e2e/vmlistener_test.go
```diff
@@ -80,41 +74,77 @@ func (s *VMListenerTestParams) TearDownTest(c *C) {
...
+// 3. Verification: volume is in attached status
+// 4. Restart the vmdkops service
+// 5. Verification: volume stays as attached
+// 6. Remove the container
+// 7. Remove the volume
 func (s *VMListenerTestParams) TestServiceRestart(c *C) {
 	misc.LogTestStart(c.TestName())
 
+	// Create the volume
+	out, err := dockercli.CreateVolume(s.vm1, s.volumeName)
```
shuklanirdesh82: @shaominchen now this test adds some inconsistency what we have decided earlier .. as each test is creating volume why not use setupTest fixture. 

For adding new test requires these amount of refactor, could be avoided by creating a separate file for vsan failure test

tests/e2e/vmlistener_test.go
shaominchen: Ah, I forgot to remove this TearDownTest. I will take care of this in my follow-up PR.

tests/e2e/vmlistener_test.go
shaominchen: I don't think so. Only truly common logic should be handled by SetUp/TearDown. For some test groups, such as vsan_test, basic_test, swarm_test, it's not the case. The reason I move the logic to test itself is because it's not common for all tests in this group.

I don't think it's a good idea to create separate test groups based on the common logic in SetUp/TearDown. Instead, test group should be created based on different test scenarios.

We can sync up offline if you have further comments.

ashahi1: What do you guys think of having separate files? Just a suggestion.:)

PR #1552, opened 2017-07-07
Test util to verify detached status and poll the detached status of a volume

### LINKED ISSUE #1528: Test should not wait unnecessary  while verifying DetachedStatus
current util is waiting for 2 seconds then invokes `adminCli volume ls` to check for "attached/detached" status. It is scary and might be hiding some race condition if there is any. we should get rid of blind wait time for each and every tests. Only special test step should use the poller when required. 

This issue to adjust the util by calling *VerifyDetachedStatus* with out wait time i.e. immediately after performing an operation.

https://github.com/vmware/docker-volume-vsphere/blob/0caae1cd2a0f5b53b3b5a1c5c3a972a53757e4a5/tests/utils/verification/volumeproperties.go#L146

interesting code snip
```
// VerifyDetachedStatus - check if the status gets detached within the timeout
func VerifyDetachedStatus(name, hostName, esxName string) bool {
	log.Printf("Confirming detached status for volume [%s]\n", name)

	const maxAttempt = 60
	for attempt := 0; attempt < maxAttempt; attempt++ {
		misc.SleepForSec(2)
		status := getVolumeStatusHost(name, hostName)
		if status != properties.DetachedStatus {
			continue
		}
```

We should add such poller with wait time whenever is required with selective test. 

/CC @msterin

pshahzeb:
This PR does two things.
1. Removes unnecessary wait in VerifyDetachedStatus. It instantly checks if the status is "detached". This is used in most of the places except destructive test scenarios i.e. restarts and failovers.

2. PollDetachedStatus retries to make check if volume status is detached. This is used in scenarios of restarts and failovers where plugin takes a bit of time to stabilize.

Testing:
test-e2e passes locally.
```
2017/07/07 00:10:51 END: VsanTestSuite.TestValidPolicy
2017/07/07 00:10:51 Destroying volume [vsanVol_volume_400787@vsanDatastore]
2017/07/07 00:10:52 Removing policy [validPolicy] on esx [10.192.248.54]
--- PASS: Test (926.32s)
PASS
OK: 30 passed
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	926.330s
```

Fixes #1528

### DIFF
--- a/tests/e2e/restart_test.go
+++ b/tests/e2e/restart_test.go
@@ -20,21 +20,22 @@
 package e2e
 
 import (
-	. "gopkg.in/check.v1"
-	"strings"
 	"log"
+	"strings"
 
-	"github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
-	"github.com/vmware/docker-volume-vsphere/tests/utils/inputparams"
-	"github.com/vmware/docker-volume-vsphere/tests/utils/verification"
-	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
+	. "gopkg.in/check.v1"
+
 	adminconst "github.com/vmware/docker-volume-vsphere/tests/constants/admincli"
 	adminutil "github.com/vmware/docker-volume-vsphere/tests/utils/admincli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/dockercli"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/inputparams"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
+	"github.com/vmware/docker-volume-vsphere/tests/utils/verification"
 )
 
 type RestartTestData struct {
-	config        *inputparams.TestConfig
-	volumeName    string
+	config            *inputparams.TestConfig
+	volumeName        string
 	containerNameList []string
 }
 
@@ -51,8 +52,8 @@ func (s *RestartTestData) SetUpSuite(c *C) {
 
 	s.volumeName = inputparams.GetUniqueVolumeName("restart_test")
 	s.containerNameList = []string{inputparams.GetUniqueContainerName("restart_test"),
-				inputparams.GetUniqueContainerName("restart_test"),
-				inputparams.GetUniqueContainerName("restart_test")}
+		inputparams.GetUniqueContainerName("restart_test"),
+		inputparams.GetUniqueContainerName("restart_test")}
 
 	// Create a volume
 	out, err := dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName)
@@ -61,7 +62,7 @@ func (s *RestartTestData) SetUpSuite(c *C) {
 	// Get volume status
 	status, err := dockercli.GetVolumeStatus(s.config.DockerHosts[1], s.volumeName)
 	c.Assert(err, IsNil, Commentf("Failed to fetch status for volume %s", s.volumeName))
-	
+
 	ds2 = s.config.Datastores[0]
 	if strings.Compare(status["datastore"], s.config.Datastores[0]) == 0 {
 		ds2 = s.config.Datastores[1]
@@ -72,7 +73,7 @@ func (s *RestartTestData) SetUpSuite(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Create a volume on the second datastore
-	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName + "@" + ds2)
+	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName+"@"+ds2)
 	c.Assert(err, IsNil, Commentf(out))
 
 	log.Printf("Restart tests setup complete")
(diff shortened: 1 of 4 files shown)


tests/e2e/restart_test.go
```diff
@@ -84,7 +84,7 @@ func (s *RestartTestData) TestVolumeDetached(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// 4. Verify detached status. Volume should be detached (within the timeout)
-	status = verification.VerifyDetachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
+	status = verification.PollDetachedStatus(s.volumeName, s.config.DockerHosts[1], s.config.EsxHost)
```
govint: If the function is using a different method of verifying status lets retain the same name vs. changes to multiple files for a name change.

tests/utils/verification/volumeproperties.go
```diff
@@ -143,24 +143,34 @@ func getVolumeStatusHost(name, hostName string) string {
 func VerifyDetachedStatus(name, hostName, esxName string) bool {
```
govint: Let the new functionality be the preferred way and lets replace whats there without having to change test code.

tests/utils/verification/volumeproperties.go
```diff
@@ -143,24 +143,34 @@ func getVolumeStatusHost(name, hostName string) string {
...
-		// Use full name to check volume status on docker host
-		status := getVolumeStatusHost(fullName, hostName)
-		if status != properties.DetachedStatus {
-			continue
-		}
-		// Use short name to check volume status on ESX
-		// this api returnes "detached" in when volume is detached
-		status = GetVMAttachedToVolUsingAdminCli(name, esxName)
-		if status == properties.DetachedStatus {
+		if VerifyDetachedStatus(name, hostName, esxName) {
```
govint: Let this be unexported - pollDetachedStatus(), all test code can use the Verify...() function as is.

tests/e2e/restart_test.go
shaominchen: I think the intention here is: VerifyDetachedStatus will be used in scenarios where we want to check the status immediately, while PollDetachedStatus will be used in scenarios where we want to poll the status, e.g. VM restart. So these two names are clear and make sense to me.

pshahzeb: Yes. That's the intention. Since there is a clear distinction between to case, (to retry and not to retry), hence the two utils. Some of the test codes is bound to change.

govint: The test code is getting really complex with different functions with the same purpose doing things differently. Test code development and reviews becomes that much more tricky with having to use a particular function (for the same purpose) in different scenarios. In this case all that a test code needs to figure is if a volume is detached, correct? For that a function which checks the current state and returns if its detached or wait  if the state must become detached. A single API that takes a waitTillVolumeDetached parameter is sufficient to embed both requirements if that's needed (still don't see why a function just wants the current state which need not be correct if its coming from the KV). Its simpler to have a single function used in all the test code.

pshahzeb: @govint 
> A new function is there in utils/docker/containers.go:GetVolumeStatus() which will return a map of the volume properties including the volume state. Use that to get the instantaneous state of a volume. Why keep adding more functions.

I agree adding more redundant utils is useless. But here we need to verify detached status from both docker host side and esx side. That's exactly VerifyDetachedStatus is doing.

> A single API that takes a waitTillVolumeDetached parameter

Existing VerifyDetachedStatus always waits till volume gets detached (within timeout). Adding an extra parameter to decide whether to wait or not would change the func signature and still induce test code change (but we will achieve the goal of single util to do single work). And the if/else part would always hit through this util.

Since most of the cases we need to instantly determine the detached status and few of them need to poll (restart and failover scenarios), I introduced PollDetachedStatus.
We can do also it through the one API way and have extra parameter and if block inside, but I think two separate utils is cleaner. Let me know what you think about this.

govint: Actually, the admin CLi and docker cli ways of getting the volume status is the same - from the KV. So there's no difference in getting the status from the ESX side. Just doing a docker inspect is sufficient.  Which is why I was pointing to the newly added function. An instantaneous result is something that can change and doesn't require needing multiple checks (which are based off the same source anyway).

pshahzeb: > Actually, the admin CLi and docker cli ways of getting the volume status is the same - from the KV. So there's no difference in getting the status from the ESX side. Just doing a docker inspect is sufficient.

Correct and logical. However, we had run into some issues in the past where there were inconsistent status results on docker host and admin cli.
CC @shuklanirdesh82

pshahzeb: Not exactly this #934 but we can run into issues that demonstrate similar symptoms. Hence the check from both esx and docker host

shuklanirdesh82: > https://github.com/vmware/docker-volume-vsphere/pull/1552#discussion_r126505314

@govint / @pshahzeb correct and valid point. We came across many instances during our initial phase with vDVS where `docker inspect` and esxcli was giving different issue hence the extensive verification. I

govint: We've never seen it in any of the tests till date on CI at least. If anything thats a bug or the thread synch errors we have seen in the past. How can they give different results when the source is exactly the same. If anything those bugs have to be fixed. No user is going to do double verification when all they would need is docker volume inspect and it better be correct.

govint: #934 isn't a support scenario at all with two ESX sharing the same datastores.

Issue #1551, opened 2017-07-06
Refactor and rename of vDVS code


luomiao:
In order to add the new shared plugin, the current layout will need some changes in order to better utilize common code and eliminate confusions.

A brief plan of the new layout:
```
docker-volume-vsphere
    - (Not mentioned directories will stay the same as before)
    - plugin -> managed_plugin (RENAME)
            this folder hold the Makefile and dockerfile template for building the final docker plugins.
    - vmdk_plugin -> client_plugin (RENAME)
            - vmdk_plugin (NEW)
                   hold existing plugin’s entry main function
            - shared_plugin (NEW)
                   hold the new shared plugin’s entry main function
            - drivers
                    - vmdk
                    - photon
                    - shared (NEW)
            - utils
                    - config
                    - fs
                    - plugin_utils
                    - refcount
                    - log_formatter (NEW)
                    - plugin_server (NEW)
 
```

pshahzeb:
>            - vmdk_plugin (**NEW**) _hold existing plugin’s entry main function_

Since we are renaming this, how about calling it vsphere instead of vmdk

luomiao:
@msterin Your call please! vmdk or vsphere plugin?

govint:
managed_plugin could be plugin_config (thats all it has)

client_plugin could be managed_plugin

vmdk_plugin could be vmdk (its a plugin as we know :-))

shared_plugin could be shared (needs a different name as it looks like its a shared component with the other drivers) or _network_

The names for drivers are a mix of volume type and platform type - vmdk and shared are volume types and photon is a platform which gives vmdks. Could rename so they are all platform perhaps.

luomiao:
To make progress, let's use shared_plugin for now and we can update it when we have a better name!

luomiao:
managed_plugin could be plugin_config: I am ok with this one.
client_plugin could be managed_plugin: so since we have esx_service, this client_plugin means it's the codes that are running on docker clients. So I still think client_plugin is a better name.

msterin:
Client plugin is a generic code, has nothing to do with managed. Current ./plugin is managed plugin and thus is in rename list. Please.keep client_plugin

shuklanirdesh82:
addressed through #1555

PR #1555, opened 2017-07-07
Rename top folders of vDVS

luomiao:
vmdk_plugin -> client_plugin
plugin -> plugin_dockerbuild

This is the first PR for Issue: https://github.com/vmware/docker-volume-vsphere/issues/1551
This PR is only for renaming purpose.
Refactoring of common utility functions will be another separate PR.


shaominchen:
From the comments in #1551, one suggestion is to rename "plugin" to "plugin_config". That makes sense to me. Can we get finalized on the decision before making the change?

luomiao:
@shaominchen 
In my opinion I prefer to name it to plugin_dockerbuild :)
It does include config.json but Dockerfile is equally important inside...
So as Mark has suggested, it's the real managed plugin is built here.
So I think it's better to still use managed_plugin (or plugin_dockerbuild :) )

luomiao:
@msterin agreed and updated :)

luomiao:
Update:
vmdk_plugin -> client_plugin
plugin -> plugin_dockerbuild

Commonvars.mk
```diff
@@ -61,4 +61,4 @@ MANAGED_PLUGIN_NAME := "vsphere:latest"
 
 # build places binaries here:
 BIN := ../build
-MANAGED_PLUGIN_LOC := ../plugin
\ No newline at end of file
+MANAGED_PLUGIN_LOC := ../plugin_dockerbuild
```
shuklanirdesh82: https://github.com/vmware/docker-volume-vsphere/blob/master/misc/scripts/deploy-tools.sh

files needs to be changed too.

luomiao: Done. Can you please approve?

Issue #1521, opened 2017-06-29
[E2E] Make WaitForExpectedState util to also check if vDVS is running


ashahi1:
Make WaitForExpectedState more robust to also check if vDVS is running if expected power state for vm is power-on.

https://github.com/vmware/docker-volume-vsphere/blob/master/tests/utils/esx/esxcli.go#L66

PR #1478, opened 2017-06-24
Adding vmgroup tests

ashahi1:
Added two tests and their related utils:
- TestVGNameForOrphanedVolumes
- TestDSAccessPrivilegeForUserVG

Original PR is having 3 LGTM, so this new PR is ready to merge as CI passed as well

These tests have already been reviewed in PR #1381  
Had to create separate PR as one of the test related to deleting default vmgroup fails.  

Ran make-testall locally and all tests pass.

```
=> Running target test-e2e-runalways
OK: 6 passed
--- PASS: Test (211.76s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	211.763s

=> Running target test-e2e-runonce 
PASS
OK: 24 passed
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	938.302s

```

PR #1563, opened 2017-07-07
Add a new util function to check if VM is fully back online

shaominchen:
Fix issue #1521.

Testing Done:
OK: 31 passed
--- PASS: Test (1069.05s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	1069.055s


tests/utils/esx/esxcli.go
```diff
@@ -77,3 +78,39 @@ func WaitForExpectedState(fn getVMPowerStatus, vmName, expectedState string) boo
...
+		return false
+	}
+
+	return true
+}
+
+// isVDVSRunning checks if vDVS is running. This util can be useful
+// in scenarios where vm is powered-on and user wants to find out if
+// vm is fully up to be able to run docker volume commands.
+func isVDVSRunning(ip string) bool {
```
ashahi1: @shaominchen This is not an esxcli util. Why was it moved to this file?

shaominchen: After refactoring, this util is not being used by any test case directly. It's being called by IsVMBackOnline function only. That's why it's an unexported function here.

ashahi1: No, that shouldn't be a criteria for moving functions around. We have separate files because it's better to keep similar utilities together. 
Also, if someone has to only verify if vDVS is running, then we will have to call IsVMBackOnline() ?
And later on we can add more utils to vDVSUtil.go file

shaominchen: That depends on how you look at the dependency. VM is running on ESX and we have utils in esx package to check VM status. vDVS is also running on ESX, so it's fair enough to have esx utils to check vDVS status as well. Of course, if we need to add more utils specific to vDVS, then I agree with you - it makes perfect sense to extract them out in a separate file. But I didn't see any requirements right now.

shuklanirdesh82: @shaominchen I agree with @ashahi1 here. We should not put `isVDVSRunning ` into esxcli.go  that leads to confusion. 

It's good that you have renamed the method but `vDVSUtil.go` is the ideal home; Please move that back.

tests/utils/esx/esxcli.go
```diff
@@ -77,3 +78,39 @@ func WaitForExpectedState(fn getVMPowerStatus, vmName, expectedState string) boo
 	log.Printf("Timed out to poll status\n")
 	return false
 }
+
+// IsVMBackOnline checks if VM is fully back online after being restarted
+func IsVMBackOnline(vmIP, vmName string) bool {
```
shuklanirdesh82: minor: `IsVMBackOnline` is good but not ideal when `isVDVSRunning=false` .. in that case vm comes back online and vDVS did not. The name does not serve the whole purpose.

tests/utils/esx/esxcli.go
shuklanirdesh82: > vDVS is also running on ESX, so it's fair enough to have esx utils to check vDVS status as well.

There should be some confusion, `isVDVSRunning` is not checking the service running on ESX instead it is checking on VM. 

On the other thing just noticed, this method is not exported that requires another refactoring when required. Where as original proposal is having in exported and the code was reviewed at #1381 and later broke it down checked through #1478 (see https://github.com/vmware/docker-volume-vsphere/pull/1478#issuecomment-310925916)

shaominchen: You are correct that we are checking the vDVS **plugin** (not the driver) which is running on VM. As for why this method is not exported, I have already explained the reason above. Original proposal was valid because at that time this util was invoked by multiple test cases directly, which is not valid anymore.

Anyway, since you both have the same comments, I will extract this util to a separate file.

tests/utils/esx/esxcli.go
shaominchen: I will rename it to IsVDVSRunningAfterVMRestart. Let me know if you have better suggestions.

shaominchen:
@shuklanirdesh82 @ashahi1 The review comments have already been addressed. Can you please take a look?

tests/utils/esx/esxcli.go
```diff
@@ -77,3 +78,39 @@ func WaitForExpectedState(fn getVMPowerStatus, vmName, expectedState string) boo
...
+	isStatusOnline := WaitForExpectedState(GetVMPowerState, vmName, properties.PowerOnState)
+	if !isStatusOnline {
+		log.Printf("VM %s is not in PoweredOn state", vmName)
+		return false
+	}
+
+	isVDVSRunning := isVDVSRunning(vmIP)
+	if !isVDVSRunning {
+		log.Printf("vDVS is not running on VM [%s]. Please check if the previous VM IP [%s] has changed.", vmName, vmIP)
```
shaominchen: This message is just helpful for debugging - we have seen this happening a few times, and it's difficult to debug if we don't realize that VM IP might have changed after restart.

tests/utils/esx/vdvs.go
```diff
@@ -15,7 +15,7 @@
 // This util provides various helper methods that can be used by different tests to
 // fetch information related to vSphere docker-volume-service.
 
-package verification
+package esx
```
govint: This should be utis/docker, its checking for a docker managed plugin.

Should also be checking that docker is running

shuklanirdesh82: @govint  It was placed earlier at `utils/verification` package ... is it a good idea to move back to `utils/verification` package?

PR #1567, opened 2017-07-10
Refactoring utility functions

luomiao:
Create and move new utility functions into utils folder:
main/log_formatter.go -> utils/log_formatter/log_formatter.go
main/main.go PluginServer main -> utils/plugin_server/plugin_server.go StartServer
main/main.go logInit  main -> utils/config/config.go LogInit InitConfig

The new utility functions will be shared between different plugins.

### DIFF
--- a/client_plugin/utils/config/config.go
+++ b/client_plugin/utils/config/config.go
@@ -19,10 +19,27 @@ package config
 
 import (
 	"encoding/json"
+	"flag"
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/natefinch/lumberjack"
+	"github.com/vmware/docker-volume-vsphere/client_plugin/utils/log_formatter"
 	"io/ioutil"
+	"os"
+	"runtime"
 )
 
 const (
+	// PhotonDriver is the driver to handle single attach vmdk-based volumes in Photon Platform
+	PhotonDriver = "photon"
+	// VMDKDriver is the driver to handle single attach vmdk-based in vSphere/vCenter 6.0+ (deprecated)
+	VMDKDriver = "vmdk"
+	// VSphereDriver is the driver to handle single attach vmdk-based in vSphere/vCenter 6.0+
+	VSphereDriver = "vsphere"
+
+	// DefaultPort is the default ESX service port.
+	DefaultPort = 1019
+
 	// Local constants
 	defaultMaxLogSizeMb  = 100
 	defaultMaxLogAgeDays = 28
@@ -41,6 +58,14 @@ type Config struct {
 	Host          string `json:",omitempty"`
 }
 
+// LogInfo stores parameters for setting up logs
+type LogInfo struct {
+	LogLevel       *string
+	LogFile        *string
+	DefaultLogFile string
+	ConfigFile     *string
+}
+
 // Load the configuration from a file and return a Config.
 func Load(path string) (Config, error) {
 	jsonBlob, err := ioutil.ReadFile(path)
@@ -51,15 +76,12 @@ func Load(path string) (Config, error) {
 	if err := json.Unmarshal(jsonBlob, &config); err != nil {
 		return Config{}, err
 	}
-	SetDefaults(&config)
+	setDefaults(&config)
 	return config, nil
 }
 
-// SetDefaults for any config setting that is at its `bottom`
-func SetDefaults(config *Config) {
-	if config.LogPath == "" {
-		config.LogPath = DefaultLogPath
-	}
+// setDefaults for any config setting that is at its `bottom`
+func setDefaults(config *Config) {
 	if config.MaxLogSizeMb == 0 {
 		config.MaxLogSizeMb = defaultMaxLogSizeMb
 	}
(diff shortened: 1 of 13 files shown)


client_plugin/utils/plugin_server/plugin_server_linux.go
```diff
@@ -26,8 +26,8 @@ import (
 )
 
 const (
-	mountRoot     = "/mnt/vmdk"           // VMDK and photon volumes are mounted here
-	pluginSockDir = "/run/docker/plugins" // Unix sock for the plugin is maintained here
+	// Unix sock for the plugin is maintained here
+	pluginSockDir = "/run/docker/plugins"
```
venilnoronha: nit: since we only have a single `const`, we could have it without the `()` block.

client_plugin/vmdk_plugin/main.go
```diff
@@ -17,185 +17,44 @@ package main
...
-	useMockEsx := flag.Bool("mock_esx", false, "Mock the ESX service")
-
-	flag.Parse()
-
-	logInit(logLevel, nil, configFile)
-
-	// Load the configuration if one was provided.
-	c, err := config.Load(*configFile)
+	cfg, err := config.InitConfig(config.DefaultConfigPath, config.DefaultLogPath,
+		config.VSphereDriver, config.VSphereDriver)
```
venilnoronha: Can we NOT pass any params here? `InitConfig(..)` should be able to pick these up easily, since they're also from within the `config` module.

We could also define `DefaultDriver` under `config_linux.go` and `config_windows.go` to provide a platform specific default driver name.

client_plugin/vmdk_plugin/sanity_test.go
```diff
@@ -75,7 +76,7 @@ func init() {
 	flag.IntVar(&parallelVolumes, "parallel_volumes", 3, "Volumes per docker daemon for create/delete concurrent tests")
 	flag.IntVar(&parallelClones, "parallel_clones", 2, "Volumes per docker daemon for clone concurrent tests")
 	flag.Parse()
-	usingConfigFileDefaults := logInit(logLevel, logFile, configFile)
+	usingConfigFileDefaults := config.LogInit(logLevel, logFile, config.DefaultLogPath, configFile)
```
venilnoronha: Similar to `InitConfig(..)`, we could leave out the `config.DefaultLogPath` here.

client_plugin/vmdk_plugin/sanity_test.go
```diff
@@ -86,7 +87,7 @@ func init() {
 		"conf_file":                *configFile,
 		"using_conf_file_defaults": usingConfigFileDefaults,
 	}).Info("VMDK plugin tests started ")
-	log.SetFormatter(new(VmwareFormatter))
+	log.SetFormatter(new(log_formatter.VmwareFormatter))
```
luomiao: Nice catch of this one. This looks like a duplicated call even before this PR. I will fix it by this chance.

client_plugin/vmdk_plugin/main.go
luomiao: We need to pass parameters when we add the new shared plugin.
For now we only have the vmdk plugin so it looks unnecessary.
But this is required when we have the new shared plugin in the following PRs.

It will become the following for vmdk plugin:
`
cfg, err := config.InitConfig(config.DefaultVMDKPluginConfigPath, config.DefaultVMDKPluginLogPath, config.VSphereDriver, config.VSphereDriver)`

And for shared plugin:
`
cfg, err := config.InitConfig(config.DefaultSharedPluginConfigPath, config.DefaultSharedPluginLogPath, config.SharedDriver, config.SharedDriver)`

client_plugin/vmdk_plugin/sanity_test.go
luomiao: See my comments to the InitConfig.
This is similar to that one.
This is needed when we have the new shared plugin, which will have a different default log path.

client_plugin/vmdk_plugin/main.go
venilnoronha: Okay. Then, should we use [`go build -tags ...`](https://github.com/golang/go/issues/8772) in this case?

For example, we could create:
```
// config_shared_linux.go

// +build shared,linux

const (
    DefaultConfigPath = ..
    DefaultLogPath = ..
    ...
)
```
```
// config_vsphere_linux.go

// +build vsphere,linux

const (
    DefaultConfigPath = ..
    DefaultLogPath = ..
    ...
)
```
and similar ones for `windows`.

We could then let `Go` pick the correct set of `const`s during build time by invoking `go build -tags shared ...` and `go build -tags vsphere ...`. This way, the `config` API will not have to deal with it's own defaults.

Thoughts?

msterin: Let's keep it simple - we should understand what is going on by simply looking at the code.  If I need to also parse the build system for figuring  out what the code does, it becomes confusing.

venilnoronha: Alright.

client_plugin/utils/config/config.go
```diff
@@ -19,10 +19,27 @@ package config
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/natefinch/lumberjack"
+	"github.com/vmware/docker-volume-vsphere/client_plugin/utils/log_formatter"
 	"io/ioutil"
+	"os"
+	"runtime"
 )
 
 const (
+	// PhotonDriver is a vmdk plugin driver for photon platform
```
msterin: nit:  the existing comments kind of useless here. I'd say 
```go
// driver to handle single attach vmdk-based volumes in Photon Platform
...
// driver to handle single attach vmdk-based in vSphere/vCenter 6.0+ 
...
// driver to  handle shared (multiple client) volumes
```

client_plugin/utils/config/config.go
```diff
@@ -39,6 +56,8 @@ type Config struct {
 	Target        string `json:",omitempty"`
 	Project       string `json:",omitempty"`
 	Host          string `json:",omitempty"`
+	Port          int    `json:",omitempty"`
```
msterin: please add a comment that this is superset and NOT all drivers need all fields

client_plugin/utils/config/config.go
```diff
@@ -70,3 +86,147 @@ func SetDefaults(config *Config) {
 		config.LogLevel = defaultLogLevel
 	}
 }
+
+// LogInit init log with passed logLevel (and get config from configFile if it's present)
+// returns True if using defaults,  False if using config file
+func LogInit(logLevel *string, logFile *string, defaultLogFile string, configFile *string) bool {
```
msterin: generally passing multiple string params is error prone. IMO it's much better to define a struct with named params and pass &{name:value,...}.

client_plugin/utils/config/config.go
```diff
@@ -70,3 +86,147 @@ func SetDefaults(config *Config) {
...
+	level, err := log.ParseLevel(*logLevel)
+	if err != nil {
+		panic(fmt.Sprintf("Failed to parse log level: %v", err))
+	}
+
+	log.SetFormatter(new(log_formatter.VmwareFormatter))
+	log.SetLevel(level)
+
+	if usingConfigDefaults {
+		log.Info("No config file found. Using defaults.")
```
msterin: nit: it would be nice to print (%v) the defaults

client_plugin/utils/config/config.go
```diff
@@ -70,3 +86,147 @@ func SetDefaults(config *Config) {
...
+	// If no driver provided on the command line, use the one in the
+	// config file or the default.
+	if *driverName != "" {
+		c.Driver = *driverName
+	} else if c.Driver == "" {
+		c.Driver = defaultDriver
+	}
+
+	// The windows plugin only supports the vsphere driver.
+	if runtime.GOOS == "windows" && c.Driver != defaultWindowsDriver {
```
msterin: I would made it a method in the driver, so you can call driver.check_config(). It's OK as is , but if/when we need to do another driver-specific decision here, I'd suggest we switch to driver methods

client_plugin/utils/config/config.go
```diff
@@ -70,3 +86,147 @@ func SetDefaults(config *Config) {
...
+		c.Driver = defaultWindowsDriver
+	}
+
+	log.WithFields(log.Fields{
+		"driver":    c.Driver,
+		"log_level": *logLevel,
+		"config":    *configFile,
+	}).Info("Starting plugin ")
+
+	if c.Driver == PhotonDriver && err == nil {
```
msterin: ah, here is another decision. Is there a way to make these checks into a driver-specific method ?

luomiao: In the current code, drivers are initialized until a later point in main function:
```
if cfg.Driver == config.PhotonDriver {
                driver = photon.NewVolumeDriver(cfg.Target, cfg.Project,
                        cfg.Host, config.MountRoot)
        } else if cfg.Driver == config.VSphereDriver {
                driver = vmdk.NewVolumeDriver(cfg.Port, cfg.UseMockEsx, config.MountRoot, cfg.Driver)
```
Is there a way to call into a driver-specific method before the driver is initialized?

client_plugin/utils/config/config.go
luomiao: It seems according to go check, it needs to be in the format of 
// PhotonDriver is ....
PhotonDriver = "photon"

luomiao: It seems according to go check, it needs to be in the format of 
```
// PhotonDriver is ....
PhotonDriver = "photon"
// VMDKDriver is ...
VMDKDriver = "vmdk"
```
I will update the following part accordingly.

msterin: that's OK, my comment was more about content... and it's a nit anyways

client_plugin/utils/config/config.go
msterin: I think we can create driver instance earlier and then use the method on driver interface

client_plugin/utils/config/config.go
luomiao: Here do you mean the part from the check of
`if runtime.GOOS == "windows" && c.Driver != defaultWindowsDriver`?
I don't think this is a driver-specific code. This is the same for all drivers.

client_plugin/utils/config/config.go
luomiao: OK, let me add a switch for that and init the driver instance in main function and see how it looks like :)

client_plugin/utils/config/config.go
msterin: you are right , this is different pivot -  "does this platform support this driver?"

luomiao:
PR is updated to avoid driver-specific checks in common path.
They are moved inside each driver's NewVolumeDriver function.

PR #1568, opened 2017-07-10
[E2E] Validation after admincli vmgroup and config ops

### LINKED ISSUE #1458: [E2E] Improve admincli util functions to verify the business data
Due to issue #1411, the assertion statement in the test logic cannot really guarantee the operation is completed as expected. Before issue #1411 is addressed, we can have more verification - in addition to verify the ssh exit code (which doesn't work right now), but also verify the real business data to make sure the operation is completed as expected.

There are a lot of admincli util functions. Adding extra verification logic to all functions is a huge effort. For now let's improve the following 4 functions:
1. CreateVMgroup()
2. DeleteVMgroup()
3. ConfigInit
4. ConfigRemove

pshahzeb:
This PR adds verification step in following 4 utils

CreateVMgroup()
DeleteVMgroup()
ConfigInit
ConfigRemove

The intention is to catch the regression early with such verification and avoid misleading during debugging as to which test failed and the reason for the same

Testing:
test-e2e passes
```
2017/07/10 19:31:23 END: VsanTestSuite.TestValidPolicy
2017/07/10 19:31:23 Destroying volume [vsanVol_volume_1060437@vsanDatastore]
2017/07/10 19:31:24 Removing policy [validPolicy] on esx [10.192.248.54]
OK: 34 passed
--- PASS: Test (1391.83s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	1391.836s
make: Leaving directory `/go/src/github.com/vmware/docker-volume-vsphere/client_plugin'
```

Fixes #1458

### DIFF
--- a/tests/utils/admincli/vmgroupmgmt.go
+++ b/tests/utils/admincli/vmgroupmgmt.go
@@ -128,12 +128,12 @@ func ConfigInit(ip string) (string, error) {
 	}
 
 	// verify the DB init
-	if GetDBmode(ip) != admincli.DBSingleNode {
-		msg := "Could not init DB to SingleNode on esx " + ip
-		return msg, fmt.Errorf(msg)
+	if GetDBmode(ip) == admincli.DBSingleNode {
+		return "DB init successful on esx " + ip, nil
 	}
 
-	return "DB init successful on esx " + ip, nil
+	msg := "Could not init DB to SingleNode on esx " + ip
+	return msg, fmt.Errorf(msg)
 }
 
 // ConfigRemove - Remove the (local) Single Node Config DB
@@ -146,13 +146,12 @@ func ConfigRemove(ip string) (string, error) {
 	}
 
 	// verify the DB removal
-	if GetDBmode(ip) != admincli.DBNotConfigured {
-		msg := "Could not remove DB on esx " + ip
-		return msg, fmt.Errorf(msg)
+	if GetDBmode(ip) == admincli.DBNotConfigured {
+		return "DB remove successful on esx " + ip, nil
 	}
 
-	return "DB remove successful on esx " + ip, nil
-
+	msg := "Could not remove DB on esx " + ip
+	return msg, fmt.Errorf(msg)
 }
 
 // IsVmgroupPresent - checks for a vmgroup in list of vmgroups
--- a/tests/utils/admincli/vmgroupmgmt.go
+++ b/tests/utils/admincli/vmgroupmgmt.go
@@ -18,9 +18,10 @@
 package admincli
 
 import (
+	"fmt"
 	"log"
 	"strings"
-	"fmt"
+
 	"github.com/vmware/docker-volume-vsphere/tests/constants/admincli"
 	"github.com/vmware/docker-volume-vsphere/tests/utils/ssh"
 )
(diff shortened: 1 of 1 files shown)


tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -18,27 +18,55 @@
...
-		return ssh.InvokeCommand(ip, admincli.RemoveVMgroup+name)
+		out, err = ssh.InvokeCommand(ip, admincli.RemoveVMgroup+name)
+	}
+
+	if err != nil {
+		return out, err
+	}
+
+	// Verify the vmgroup removal
+	if IsVmgroupPresent(ip, name) != true {
```
lipingxue: When "delete_vol" is set to true, should also check volume that associated with that vmgroup have been removed.

pshahzeb: Currently this util doesn't have knowledge about which volumes fall into the given VMgroup. So directly we cannot verify.
What we can do is retrieve the list of volumes from admin cli and in verification make sure those volumes are deleted. But I would prefer doing it out of this PR depending on the need of such verification.

lipingxue: OK, make sense.

tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -93,13 +121,38 @@ func ConfigInit(ip string) (string, error) {
...
 	log.Printf("Initializing the SingleNode Config DB on esx [%s] \n", ip)
-	return ssh.InvokeCommand(ip, admincli.InitLocalConfigDb)
+	out, err := ssh.InvokeCommand(ip, admincli.InitLocalConfigDb)
+
+	if err != nil {
+		return out, err
+	}
+
+	// verify the DB init
+	if GetDBmode(ip) != admincli.DBSingleNode {
```
shaominchen: This logic seems not exactly consistent with what we are doing in above CreateVmGroup and DeleteVmGroup. If we follow the same style, the logic here should be something like this:

```golang
if GetDBmode(ip) == admincli.DBSingleNode {
    return out, nil
}

msg := ...
return msg, fmt.Errorf(msg)
```
Not a big concern, but I'd like to know if there are any specific reason for this difference.

tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -93,13 +121,38 @@ func ConfigInit(ip string) (string, error) {
...
 	log.Printf("Removing the SingleNode Config DB on esx [%s] \n", ip)
-	return ssh.InvokeCommand(ip, admincli.RemoveLocalConfigDb)
+	out, err := ssh.InvokeCommand(ip, admincli.RemoveLocalConfigDb)
+
+	if err != nil {
+		return out, err
+	}
+
+	// verify the DB removal
+	if GetDBmode(ip) != admincli.DBNotConfigured {
```
shaominchen: Same as above.

tests/utils/admincli/vmgroupmgmt.go
pshahzeb: No specific reason. Just happened to write the if condition in that way. Would you want it to be changed as per style in CreateVmGroup and DeleteVmGroup?

shaominchen: The logic in CreateVmGroup and DeleteVmGroup makes more sense to me because we just want to add one extra verification, but don't really want to change the original successful command output.

pshahzeb: Sure. Will make it consistent.

tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -93,13 +121,37 @@ func ConfigInit(ip string) (string, error) {
...
-	return ssh.InvokeCommand(ip, admincli.InitLocalConfigDb)
+	out, err := ssh.InvokeCommand(ip, admincli.InitLocalConfigDb)
+
+	if err != nil {
+		return out, err
+	}
+
+	// verify the DB init
+	if GetDBmode(ip) == admincli.DBSingleNode {
+		return "DB init successful on esx " + ip, nil
```
shaominchen: Shouldn't we return original command output in this case?

```golang
return out, nil
```
Basically the point is: we don't want to change the original command output if the extra verification passes.

tests/utils/admincli/vmgroupmgmt.go
```diff
@@ -93,13 +121,37 @@ func ConfigInit(ip string) (string, error) {
...
-	return ssh.InvokeCommand(ip, admincli.RemoveLocalConfigDb)
+	out, err := ssh.InvokeCommand(ip, admincli.RemoveLocalConfigDb)
+
+	if err != nil {
+		return out, err
+	}
+
+	// verify the DB removal
+	if GetDBmode(ip) == admincli.DBNotConfigured {
+		return "DB remove successful on esx " + ip, nil
```
shaominchen: Same as above.

tests/utils/admincli/vmgroupmgmt.go
pshahzeb: done

PR #1573, opened 2017-07-11
Making CI builds serialized

shuklanirdesh82:
TBD


misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# datastore name refers storage backed by base ESX (physical ESX) where
```
msterin: nit: refers `to` storage backed...

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
+# ongoing builds are going to create file which helps to serialize CI run
```
msterin: nit: `ongoing/xxx contains the directory named after the currently running CI/ build`

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
 DS='-ds=datastore1'
-unit_test_array=($TEST_URL_ARRAY)
-numServers=${#unit_test_array[@]}
+
+# Lock variables to be consumed
```
msterin: nit: just `lock  names, to use in lock/unlock`

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
-while [[ "$prevBuild" != "" ]];
-do
-    prevBuildStatus=`drone build info vmware/docker-volume-vsphere $prevBuild`
-    outArray=($prevBuildStatus)
-    if [[ ${outArray[2]} != *"running"* || ${outArray[2]} == *"running"* && $prevBuild == $DRONE_BUILD_NUMBER ]]
+    # verifies status value OR request is resulted from *Restart* event
+    if [[ ${buildInfoObj[2]} != *"running"* || $1 == $DRONE_BUILD_NUMBER ]]
     then
-        govc datastore.rm $DS docker-volume-vsphere/ongoing/$prevBuild
+        return 0
```
msterin: nit: I'd print (debug print) the buildInfoObj and $1 here , just in case we need to debug

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
-        echo "Waiting 2 minutes for previous build $prevBuild to complete";
-        sleep 120;
+        return 1
+    fi
+}
+
+# Util to remove created directory (consumed as locking mechanism or
+# preserving metadata for ongoing build)
+function release_lock {
+#    echo "Releasing lock file [$1]"
```
msterin: either drop $1 check in acquire_lock, or add it here - acquire_lock and release_lock should handle the args the same way

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
+    then
+        # Grab ongoing build information from drone.
+        runningBuild=`govc datastore.ls $DS docker-volume-vsphere/ongoing`
+        runningBuildArr=($runningBuild)
+        # Let's check cleanup is needed or not
+        # cleans up if return value is 0 otherwise not; checking build is finished or not
+        if is_cleanup_needed ${runningBuildArr[0]}
+        then
+            echo "Cleanning stale data..."
+            release_lock "ongoing/${runningBuildArr[0]}"
```
msterin: what is this lock ? Or are you simply removing the folder using release_lock() ? It is confusing and error prone, just call govc datastore.rm here.

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
+            release_lock $build_lock
+            isRunAllowed=0
```
msterin: `isRunAllowed` is not needed, as well as lines 97-100 below. instead, just do  (line 92 +) 
```
   datastore.rm <ongoing/id>
   # release_lock $build_lock <= NOT NEEDED AS  YOU NEED TO REACQUIRE IT ANYWAYS
   release_lock $check_lock
   break # this will exit with still holding build lock, which is what you need

File: misc/drone-scripts/validate-and-serialize.sh [hunk 8]
```diff
@@ -13,33 +13,101 @@
...
+        fi
+        # Let's release build_check_lock
+        release_lock $check_lock
+        if [[ isRunAllowed -eq 0 ]]
```
msterin: 97-100 is not needed, break should happen above

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,101 @@
...
+        fi
     fi
-    prevBuild=`govc datastore.ls $DS docker-volume-vsphere/ongoing|tail -n 1`
+    # wait for some time to poll again
+    echo "Waiting $interval seconds for build ${runningBuildArr[0]} to complete";
+    sleep $interval;
 done
 
+# build_lock is acquired and adding current build# entry to ongoing folder
+acquire_lock $build_lock
```
msterin: don't need this as the lock is either acquired in while(), or in `if is_cleanup_needed ${runningBuildArr[0]}`

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: In the worst case scenario (for the waiting thread) there will be too much logging. And Drone UI has limitation to show 5MB build execution output .. we might cross the limit.

Let me know what do you think. 

I'm not sure whether we have `DEBUG` log level supported for such scripts.

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: done!

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: done!

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: done!

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: I am adding a check for `release_lock`

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: Okay. In that case let me invoke `govc datastore.rm` directly.

misc/drone-scripts/validate-and-serialize.sh
msterin: yeah, good point. Ignore this comment then

shuklanirdesh82: Sure thanks!

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,95 @@
...
 
-# Check if any other build is in the ongoing folder, if present, check if entry is stale.
-# If entry is stale clean it up.
-# If entry is an on going build wait for it.
+# wait time in seconds
+interval=30
 
-prevBuild=`govc datastore.ls $DS docker-volume-vsphere/ongoing|tail -n 1`
+# Decide whether cleanup is needed or not.
+# grab ongoing build# and parse build info, returns true when build_status
```
msterin: nit: there is no need to repeat the code in the comment. 
Something like "returns true if the ongoing/build info is stale and the build is not running anymore, or if the user wants to restart the build so the ongoing/build and current build have the same ID"

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,95 @@
...
+        runningBuildArr=($runningBuild)
+        # Let's check cleanup is needed or not
+        # cleans up if return value is 0 otherwise not; checking build is finished or not
+        if is_cleanup_needed ${runningBuildArr[0]}
+        then
+            echo "Cleanning stale data..."
+            govc datastore.rm $DS docker-volume-vsphere/ongoing/${runningBuildArr[0]}
+            release_lock $check_lock
+            break
+        fi
```
msterin: You still need to release check_lock. even if the cleanup is not needed
Missing `; else release_lock $check_lock ; fi` for `if is _cleanup_needed`

shuklanirdesh82: yeah I have identified while doing self code review ;)

misc/drone-scripts/validate-and-serialize.sh
shuklanirdesh82: make sense. Improving it.

misc/drone-scripts/validate-and-serialize.sh
```diff
@@ -13,33 +13,99 @@
...
-        govc datastore.rm $DS docker-volume-vsphere/ongoing/$prevBuild
+        return 0
     else
-        echo "Waiting 2 minutes for previous build $prevBuild to complete";
-        sleep 120;
+        return 1
+    fi
+}
+
+# Util to remove created directory (consumed as locking mechanism or
```
msterin: the comment is not fully accurate anymore. This function is *only* for releasing a lock

shuklanirdesh82: well after addressing https://github.com/vmware/docker-volume-vsphere/pull/1573#discussion_r126824716 comment .. function is meant for releasing lock only.

I will fix the inline comment before merging.

Thanks!

PR #1581, opened 2017-07-13
Adding vendor etcd client v3

shivanshu21:
Considered the latest version of the etcd client on Mark's suggestion
Though v3 is a larger code base, it has better support for transactions
It still has code from some private individuals like the previous v2 client which etcd client itself refers to.


shivanshu21:
@msterin The V3 client adds 9 MB to the size of the binary.

luomiao:
LGTM

Issue #1505, opened 2017-06-28
[E2E] Volume data is missing after failover across swarm nodes


shaominchen:
One verification step in existing TestFailoverAcrosssSwarmNods has been commented out because it always fails during auto run. However, the same verification passes during manual run. Need to investigate the issue, figure out the root cause, and re-enable this verification in the test.

shaominchen:
I've figured out the root cause. It has been very confusing that the same test steps always pass manually, but fail in automatic run. The truth is: the govc poweroff command is doing a real abrupt power off on the given VM, but the power off button triggered from NGC UI is actually doing a guest shutdown (by default). So, when we run the test steps manually (by clicking power off from UI), the VM is always shut down gracefully without data corruption. But when we run the test steps automatically, the VM is powered off abruptly, and causing data corruption. Please refer to this document for more information about VM shut down and power off: https://www.vmware.com/support/ws5/doc/ws_running_poweroff_vs_shutdown.html.

When the volume is mounted to a container and some data are written into the volume, the data may not be flushed into disk immediately. During automatic test run, the VM is powered off right after the data being written to the volume. That's why the data corruption happens.

The fix is straightforward: we should shut down (instead of powering off) the VM in this test. Abrupt power-off should not be used in most cases unless there's no data to be protected.

shaominchen:
This is confirmed not a product issue, so moving back to P1. The fix is ready, but being blocked on the failure on one of CI testbeds.

ashahi1:
While trying to debug this issues as to why this test only fails on CI, we managed to found bug #1621.
After Govindan fixed that bug in PR #1626, last two runs on CI are green and one more is going on.
If current also finishes successfully, we can merge this change.

ashahi1:
Last three runs are green on CI. Closing this issue.

PR #1585, opened 2017-07-13
Use vm shutdown instead of power off. Fixes issue #1505.

shaominchen:
Please refer to the issue #1505 for more information about the root cause and proposed solution.

Testing Done:

OK: 46 passed
--- PASS: Test (1830.20s)
PASS
ok  	github.com/vmware/docker-volume-vsphere/tests/e2e	1830.211s

### DIFF
--- a/tests/utils/esx/govc.go
+++ b/tests/utils/esx/govc.go
@@ -25,9 +25,9 @@ If some test is misbehaving, then the developer can enable that log and test.
 package esx
 
 import (
+	"fmt"
 	"log"
 	"strings"
-	"fmt"
 
 	"github.com/vmware/docker-volume-vsphere/tests/constants/esx"
 	"github.com/vmware/docker-volume-vsphere/tests/utils/misc"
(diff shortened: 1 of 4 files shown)


tests/utils/esx/govc.go
```diff
@@ -59,21 +59,32 @@ func GetVMPowerState(vmName string) string {
...
 	cmd := esx.PowerOffVM + vmName
 	out, _ := ssh.InvokeCommandLocally(cmd)
 	return out
 }
 
+// ShutDownVM util shuts down the VM
+// govc vm.power -s=true photon
+func ShutDownVM(vmName string) string {
+	log.Printf("Shuting down VM [%s]\n", vmName)
+	cmd := esx.ShutDownVM + vmName
```
msterin: why it is done with ssh, and not govmomi ?

shaominchen: This is done with govmomi. This go file name (ssh.go) is confusing I think. The call ssh.InvokeCommandLocally() is calling govc command in local shell.

msterin: Yeas, but it's still go code calling shell to call local command line (govc) to use govmomi to talk to VC. Why not cut the middleman and call govmomi directly ? It's a generic question to all test, not to your PR - I understand you are just being consistent.

The rest LGTM ; I just want to understand the above

shaominchen: I see your point. So you are asking why we are not using govmomi API/library directly. @shuklanirdesh82 did the early investigation. My guess is that that govc CLI seems easier to integrate with. Let me file a issue to track this so that the current fix will not be blocked.

shaominchen: Filed issue #1586.

shaominchen:
I'm holding on this PR because of the CI failure: https://ci.vmware.run/vmware/docker-volume-vsphere/1220
In CI there are two different setups. The same test (TestFailoverAcrossSwarmNodes) passed during 1st run, but failed during 2nd run.
Need to discuss with @shuklanirdesh82 to figure out the difference between those 2 CI setups.

ashahi1:
@shaominchen No, that's not correct. This test fails on both the setups - sometimes on 6.0 and other times on 6.5 as well.

Failed run on 6.5 esx build: https://ci.vmware.run/vmware/docker-volume-vsphere/1224

shaominchen:
@ashahi1 So far the test passed 100% in my local run on ESX 6.5, and failed 100% on CI ESX 6.0. To debug this issue, I'd suggest to use ESX 6.0 and enable TRACE logging. That will help us to figure out the root cause more easily.

ashahi1:
@shaominchen I ran locally with 6.0 and test passed. And since test failed on CI even with 6.5 build, so test failure has nothing to do with esx version. Looks like we need to revisit the test steps here.

Failed CI run with 6.5 esx build: https://ci.vmware.run/vmware/docker-volume-vsphere/1224

shaominchen:
@ashahi1 As you have analyzed, it may not be related to esx versions. So please go ahead to debug the issue further. You'll need to enable TRACE logging to make debugging easier.

ashahi1:
While trying to debug this issues as to why this test only fails on CI, we managed to found bug #1621.
After Govindan fixed that bug in PR #1626, last three runs on CI are green.

PR #1611, opened 2017-07-20
Shared plugin: etcd starting up functions.

luomiao:
Design:
Swarm leader node starts an etcd cluster when the shared volume driver is initialized.
Swarm manager nodes join the etcd cluster through the leader node.
Swarm worker nodes do nothing.

The etcd cluster resides inside plugin runc container.
Thus when plugin is stopped, the etcd member inside the container dies.
When the plugin/VM is restarted, if the plugin find there is already a started member
in the etcd cluster with the same peer url, this entry is removed by current plugin and continue to re-join.
`msterin> ` Who should do it ? Is it a part of this PR of future PRs ? 
`miao> ` This PR include the remove and re-join already. modified the statement above to clarify.

Note 1:
Due to etcd limitation, users should make sure the swarm cluster has the number of managers
(including leader) = 1 or >2, in order to avoid the situation of quorum loss.
`msterin> ` what happens if they don't ? Do we print a log message with recommendation? What happens if they loose quorum ? 
`miao> ` I updated the statement above to be more accurate. The quorum loss in our case mostly happens when 1 member adding into 1-node cluster and left. This member cannot rejoin due to quorum loss. But if 2 members joined to 1-node cluster (became 3-node cluster) and 1 member left, the quorum still exist (https://github.com/coreos/etcd/issues/6420). When the quorum loss happens, etcd cluster can still work but in unhealthy mode, which means it cannot add/remove members. We throw out error in log if users tried to start a new master plugin and failed. This is not recoverable for now but etcd has an opening progressing issue on it. I think it's ok for us to add recommendation in Readme and simply fail the new manager plugin when quorum loss happened, before ETCD make progress on this quorum loss situation.

Note 2:
For now the etcd cluster start/join only happens when the plugin is initialized.
Thus if a swarm worker node has the plugin installed first and promoted to manager later,
it won't join the etcd cluster. Before this is resolved, users should reinstall the plugin
on a promoted node.
`msterin> ` - is there an issue for this (or work item on internal tracking) ? Also, can we print a message if this happens ? We can do a periodic (each 5 min ? ) poll to check the node status and the etcd node status. Unless the issue is planned to be fixed soon. 
`miao> ` This needs more investigation and I will create a work item on this issue. We can do periodic poll to trigger etcd member operations if the plugin cannot get notification when a node is promoted. 

Note 3:
We are using the default docker swarm network (IP address from docker info) for now. And as a result, the new plugin configuration requires Network type = host.
`msterin> `  Same comment - we need either an issue or internal work item to track, and in any case please write up specific commands to configure plugin
`miao> ` yes we have a work item on it. There is no command needed to be done by users though. The config.json of the plugin build already has it by default, included with this PR.

Dockerfile change:
Create another dockerfile template so we have two dockerfile templates for the two plugins separately. This is because the new shared plugin requires etcd to be installed. The difference is too big for using the same template.

Test:
e2e test passed.

### DIFF
--- a/client_plugin/drivers/shared/etcdops.go
+++ b/client_plugin/drivers/shared/etcdops.go
@@ -1,0 +1,282 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package shared
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	etcdClient "github.com/coreos/etcd/clientv3"
+	"github.com/docker/engine-api/types"
+)
+
+const (
+	etcdClientPort           = ":2379"
+	etcdPeerPort             = ":2380"
+	etcdClusterToken         = "vsphere-shared-etcd-cluster"
+	etcdListenURL            = "0.0.0.0"
+	etcdScheme               = "http://"
+	etcdClusterStateNew      = "new"
+	etcdClusterStateExisting = "existing"
+)
+
+// initEtcd start or join ETCD cluster depending on the role of the node
+func (d *VolumeDriver) initEtcd() error {
+	ctx := context.Background()
+	cli := d.dockerd
+
+	// get NodeID from docker client
+	info, err := cli.Info(ctx)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"error": err},
+		).Error("Failed to get Info from docker client ")
+		return err
+	}
+
+	// worker just returns
+	nodeID := info.Swarm.NodeID
+	if info.Swarm.ControlAvailable == false {
+		log.WithFields(
+			log.Fields{"nodeID": nodeID},
+		).Info("Swarm node role: worker. Action: return from InitEtcd ")
+		return nil
+	}
+
+	// check my local role
+	node, _, err := cli.NodeInspectWithRaw(ctx, nodeID)
+	if err != nil {
+		log.WithFields(log.Fields{"nodeID": nodeID,
+			"error": err}).Error("Failed to inspect node ")
+		return err
+	}
+
+	// get the IP address of current node
+	addr := info.Swarm.NodeAddr
+
+	// if leader, proceed to start ETCD cluster
+	if node.ManagerStatus.Leader {
+		log.WithFields(
+			log.Fields{"nodeID": nodeID},
+		).Info("Swarm node role: leader, start etcd cluster")
+		err = startEtcdCluster(addr, nodeID)
+		if err != nil {
+			log.WithFields(log.Fields{"nodeID": nodeID,
+				"error": err}).Error("Failed to start ETCD Cluster")
+			return err
+		}
+		return nil
+	}
+
+	// if manager, first find out who's leader, then proceed to join ETCD cluster
+	nodes, err := cli.NodeList(ctx, types.NodeListOptions{})
+	if err != nil {
+		log.WithFields(log.Fields{"nodeID": nodeID,
+			"error": err}).Error("Failed to get NodeList from swarm manager")
+		return err
+	}
+	for _, n := range nodes {
+		if n.ManagerStatus != nil && n.ManagerStatus.Leader == true {
+			log.WithFields(
+				log.Fields{"leader ID": n.ID,
+					"manager ID": nodeID},
+			).Info("Swarm node role: manager. Action: find leader ")
+
+			joinEtcdCluster(addr, n.ManagerStatus.Addr, nodeID)
+			return nil
+		}
+	}
+
+	err = fmt.Errorf("Failed to get leader for swarm manager %s", nodeID)
+	return err
+}
+
+// startEtcdCluster function is called by swarm leader to start a ETCD cluster
+func startEtcdCluster(nodeAddr string, nodeID string) error {
+	lines := []string{
+		"--name", nodeID,
+		"--advertise-client-urls", etcdScheme + nodeAddr + etcdClientPort,
+		"--initial-advertise-peer-urls", etcdScheme + nodeAddr + etcdPeerPort,
+		"--listen-client-urls", etcdScheme + etcdListenURL + etcdClientPort,
+		"--listen-peer-urls", etcdScheme + etcdListenURL + etcdPeerPort,
+		"--initial-cluster-token", etcdClusterToken,
+		"--initial-cluster", nodeID + "=" + etcdScheme + nodeAddr + etcdPeerPort,
+		"--initial-cluster-state", etcdClusterStateNew,
+	}
+	go etcdService(lines)
+
+	return nil
+}
+
+// joinEtcdCluster function is called by a non-leader swarm manager to join a ETCD cluster
+func joinEtcdCluster(nodeAddr string, leaderAddr string, nodeID string) error {
+	cli, err := addrToEtcdClient(leaderAddr)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"nodeAddr": nodeAddr,
+				"leaderAddr": leaderAddr,
+				"nodeID":     nodeID},
+		).Error("Failed to join ETCD cluster on manager ")
+	}
+
+	// list all current ETCD members, check if this node is already added as a member
+	lresp, err := cli.MemberList(context.Background())
+	if err != nil {
+		log.WithFields(
+			log.Fields{"leaderAddr": leaderAddr,
+				"error":       err,
+				"members len": len(lresp.Members)},
+		).Error("Failed to list member for ETCD")
+		return err
+	}
+
+	peerAddr := etcdScheme + nodeAddr + etcdPeerPort
+	existing := false
+	for _, member := range lresp.Members {
+		// loop all current etcd members to find if there is already a member with the same peerAddr
+		if member.PeerURLs[0] == peerAddr {
+			if member.Name == "" {
+				// same peerAddr already existing
+				// empty name indicates this member is not started, continue the join process
+				log.WithFields(
+					log.Fields{"nodeID": nodeID,
+						"peerAddr": peerAddr},
+				).Info("Already joined as etcd member but not started. ")
+
+				existing = true
+			} else {
+				// same peerAddr already existing and started, need to remove before re-join
+				// we need the remove since etcd data directory is not persistent
+				// thus the node cannot re-join as the same member as before
+				log.WithFields(
+					log.Fields{"nodeID": nodeID,
+						"peerAddr": peerAddr},
+				).Info("Already joined as a etcd member and started. Action: remove self before re-join ")
+
+				_, err = cli.MemberRemove(context.Background(), member.ID)
+				if err != nil {
+					log.WithFields(
+						log.Fields{"peerAddr": peerAddr,
+							"member.ID": member.ID},
+					).Error("Failed to remove member for ETCD")
+					return err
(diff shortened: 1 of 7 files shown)


client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const (
+	apiVersion               = "v1.24" // docker engine 1.24 and above support this api version
+	dockerUSocket            = "unix:///var/run/docker.sock"
+	etcdClientPort           = ":2379"
+	etcdPeerPort             = ":2380"
+	etcdClusterToken         = "vsphere-shared-etcd-cluster"
+	etcdListenURL            = "0.0.0.0"
+	etcdScheme               = "http://"
```
shivanshu21: etcdProtocol? Or etcdTransport?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+	etcdClusterStateNew      = "new"
+	etcdClusterStateExisting = "existing"
+)
+
+// InitEtcd start or join ETCD cluster depending on the role of the node
+func InitEtcd() error {
+	// create new docker client
```
shivanshu21: I was adding similar things to shared_plugin.go. Should we move the docker Client to VolumeDriver structure and the docker info part to a function in shared_plugin.go?

We might want to call docker info multiple times at various places. Perhaps a better place will be shared_plugin.go. A function getsDockerInfo() can return a structure similar to this:

53 // Docker swarm information
 54 type SwarmMetadata struct {
 55         isManager bool
 56         isLeader  bool
 57         selfIP    string
 58         managers  []string
 59 }

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+			log.Fields{"nodeID": nodeID},
+		).Info("Swarm node role: worker, return from InitEtcd ")
+		return nil
+	}
+
+	// check my local role
+	node, _, err := cli.NodeInspectWithRaw(ctx, nodeID)
+	if err != nil {
+		log.WithFields(log.Fields{"nodeID": nodeID,
+			"error": err}).Error("Failted to inspect node ")
```
shivanshu21: Failed. Typo.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+		return err
+	}
+
+	// get the IP address of current node
```
shivanshu21: info.Swarm.NodeAddr can give you this node's IP address.

BTW this is docker CLI code. Might be helpful.

https://github.com/docker/cli/blob/2bfac7fcdafeafbd2f450abb6d1bb3106e4f3ccb/cli/command/system/info.go

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+			return err
+		}
+		return nil
+	}
+
+	// if manager, first find out who's leader, then proceed to join ETCD cluster
+	nodes, err := cli.NodeList(ctx, types.NodeListOptions{})
+	if err != nil {
+		log.WithFields(log.Fields{"nodeID": nodeID,
+			"error": err}).Error("Failted get NodeList from swarm manager")
```
shivanshu21: Typo. Failed.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		log.WithFields(
+			log.Fields{"endpoint": endpoint,
+				"error":       err,
+				"members len": len(lresp.Members)},
+		).Error("Failed to list member for ETCD")
+		return err
+	}
+
+	peerAddr := etcdScheme + nodeAddr + etcdPeerPort
```
shivanshu21: By functionality, peerAddr is appearing to be the address of this node. If so, can this be renamed to nodeAddr for clarity?

shivanshu21: can peerAddr be renamed to nodeURL or something? this address is of local node. "Peer" word is confusing.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+					return err
+				}
+			}
+			break
+		}
+	}
+
+	initCluster := ""
+	if !existing {
+		peerAddrs := []string{peerAddr}
```
shivanshu21: I am guessing member add() accepts a list of strings so we converted it to string[]. But "peer" is confusing. If peerAddrs holds addr of this node, can the name nodeURLs be picked?

client_plugin/drivers/shared/etcdops.go
luomiao: "scheme" is the corresponding member name of URL struct: https://golang.org/pkg/net/url/#URL

client_plugin/drivers/shared/etcdops.go
luomiao: It will depend on how and where we will use docker client.
I am fine to put it into the VolumeDriver structure later when it's used in other situations.

I am not sure about getsDockerInfo() function though.
The returned info already has those information, why bother setting them into another struct?
And those information are changing so we cannot reuse them...

client_plugin/drivers/shared/etcdops.go
luomiao: peerAddr is the address matching with peerURL inside etcd member structure.
And it's matching with etcd MemberAdd function:
https://github.com/coreos/etcd/blob/d51d381eca406156fd7e44118d4b0dab03589354/clientv3/cluster.go#L36

client_plugin/drivers/shared/etcdops.go
luomiao: https://github.com/coreos/etcd/blob/d51d381eca406156fd7e44118d4b0dab03589354/clientv3/cluster.go#L36
It's the right name in ETCD world...
It especially means the URL addresses for listening peer traffic.

client_plugin/drivers/shared/etcdops.go
luomiao: Yes for here we can get NodeAddr from Info.Swarm.NodeAddr.
However later when we need to get the address of leader from a non-leader manager,
we cannot do "docker info" for a remote node, but can only do "docker node inspect".
The returned struct of NodeInspectWithRaw is swarm.Node structure, instead of types.Info.
So we have to go through swarm.Node.ManagerStatus.Addr to get the addresses...
And then I decide to reuse that code :)

luomiao:
Since we anyway need to call docker info every time and there is no extra processing needed, IMO there is no need to add an extra function wrapper...

How it looks like with the function wrapper:
```
type SwarmMetadata struct {
isManager bool
isLeader bool
selfIP string
managers []string
}

func getsDockerInfo() SwarmMetadata {
var metadata SwarmMetadata
info, err := cli.Info(ctx)
metadata.isManager = info.ControlAvailable
metadata.selfIP = info.NodeAddr
for _, manager := range info.swarm.RemoteManagers {
append(metadata.managers, manager.Addr)
}
...
return metadata
}

// Below are repeated used code
// when the new info is needed
metadata := getsDockerInfo()

// when need to check if manager
if metadata.isManager != true ...

// when need to get selfIP
something := metadata.selfIP

// when need to loop managers
for _, managerIP := range metadata.managers {
...
do something to managerIP
}
```

How it looks like without function wrapper:
```
// Below are repeated used code
// when the new info is needed 
info, err := cli.Info(ctx)

// when need to check if manager
if info.ControlAvailable != true ...

// when need to get selfIP
something := info.NodeAddr

// when need to loop managers
for _, manager := range info.Swarm.RemoteManagers {
...
do something to manager.Addr
}
```

So if we compare above, the function wrapper is not saving any code reuse. BTW it can add an extra loop in order to get manager IP addresses. So I would prefer not using a wrapper here.

msterin:
@luomiao  - I have a few questions, added with `msterin> ` to the description. Please take a look

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,271 @@
...
+		"--initial-cluster", initCluster + nodeID + "=" + etcdScheme + nodeAddr + etcdPeerPort,
+		"--initial-cluster-state", etcdClusterStateExisting,
+	}
+
+	go etcdRoutine(lines)
+
+	return nil
+}
+
+func etcdRoutine(cmd []string) {
```
msterin: the name feels wrong. etcdService() or etcdLoop() or even etcdDaemon() are closer to what's really going on

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+		).Error("Failed to get Info from docker client ")
+		return err
+	}
+
+	// worker just returns
+	nodeID := info.Swarm.NodeID
+	if info.Swarm.ControlAvailable == false {
+		log.WithFields(
+			log.Fields{"nodeID": nodeID},
+		).Info("Swarm node role: worker, return from InitEtcd ")
```
msterin: this info message is confusing.... maybe `"Swarm node role: worker. Action: return from InitEtcd "` (same comment to other tracking Info messages)

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+			return err
+		}
+		return nil
+	}
+
+	// if manager, first find out who's leader, then proceed to join ETCD cluster
+	nodes, err := cli.NodeList(ctx, types.NodeListOptions{})
+	if err != nil {
+		log.WithFields(log.Fields{"nodeID": nodeID,
+			"error": err}).Error("Failed get NodeList from swarm manager")
```
msterin: 'failed **to** get...'

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+				"error":       err,
+				"members len": len(lresp.Members)},
+		).Error("Failed to list member for ETCD")
+		return err
+	}
+
+	peerAddr := etcdScheme + nodeAddr + etcdPeerPort
+	existing := false
+	for _, member := range lresp.Members {
+		if member.PeerURLs[0] == peerAddr {
```
msterin: there should be a better way to find it . Have you checked existing packages ?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	for _, manager := range info.Swarm.RemoteManagers {
+		cli, err := addrToEtcdClient(manager.Addr)
+		if err == nil {
+			return cli
+		}
+	}
+
+	log.Errorf("Failed to create etcd client according to manager info")
```
msterin: Need to print extra inffo here; if we got here it's likely there was no debug print yet

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+	return nil
+}
+
+func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
+	s := strings.Split(addr, ":")
```
msterin: comment the function please. Expecially the string parsing part...

client_plugin/drivers/shared/etcdops.go
```diff
@@ -0,0 +1,269 @@
...
+				_, err = cli.MemberRemove(context.Background(), member.ID)
+				if err != nil {
+					log.WithFields(
+						log.Fields{"peerAddr": peerAddr,
+							"member.ID": member.ID},
+					).Error("Failed to remove member for ETCD")
+					return err
+				}
+			}
+			break
```
msterin: why break? Generally, this loop nesting looks too deep , so it's either need to be flattened out  or (more likley) more comments added at each node of the decision tree here.

client_plugin/drivers/shared/etcdops.go
luomiao: Etcd clientv3 only have 4 member API functions:
https://github.com/coreos/etcd/blob/d51d381eca406156fd7e44118d4b0dab03589354/clientv3/cluster.go#L31
And there is no API function to locate a certain member...

client_plugin/drivers/shared/etcdops.go
luomiao: Because we don't need to continue the comparison since one IP address can only join at once. No need to continue the check.
I will add some more comments here.

Issue #1612, opened 2017-07-20
Add sshpass to the vibauthor container


govint:
Presently the vibauthor container used to run the builds has all the tools that are needed for the build. As part of #1515 a new make target is being added to allow a developer to deploy a testbed (1master + 2 workers) and then deploy the plugin on those. Thus allowing the developer run all their tests on the testbed without having to set those up manually.

Its convenient to use the vibauthor container to run the script to deploy the VMs as it already has govc installed in it and usable to import the OVAs and create the VMs.

The issue faced in #1515 is to be able to run commands via SSH on the VMs, as those need the password to be entered. Even to do a password-less SSH command we need to first copy the public key to the VM which again needs the password.

sshpass allows passing the password to an ssh command. Allowing the deploy script to get the password via an env. var. and then using sshpass to run ssh commands.

The container already contains ssh utils like - ssh  ssh-add  ssh-agent  ssh-copy-id  ssh-keygen  ssh-keyscan.

So, ..... this issue is to add sshpass to the container so the deploy script can work.

govint:
#1515 to create a test bed for a developer doesn't seem like its a much needed capability and users should be able to set up their workspaces of VMs by themselves. Moreover, developers (outside of the core dev. team) who want to run e2e tests (do they) would typically use a set up for a longer time vs. creating and discarding those setups often. I'm pushing this out to a future release unless this is really worth implementing.

Creating a script to make a test bed has nothing to do with the plugins we have and really shouldn't be in this code base either.

shuklanirdesh82:
we can revisit after kubecon release not mustfix

govint:
This is not a must fix and the original reason was to be able to create a test set up via a make target. Which, IMO, is not needed at all. Given users already have the means to set up a cluster of hosts the plugin doesn't need to provide any such capability. Closing this.

PR #1613, opened 2017-07-20
Create test setup script [SKIP CI]

govint:
This change is for review only. The script is presently being tested as a stand alone script and needs the change in #1612 to be able to run as a make target. The script needs _sshpass_ utiil and if run as a stand alone script then user needs to have installed sshpass on their VM.

1. Needs three variables to be exported thats used by the script.
2. Uses govc to deploy and manage creating the testbed (1master + 2 workers).
3. Supports create/delete of a test bed.
4. Deploys the given OVA (user gives file path or a URL) as a "template" VM and creates the testbed as linked clones off that vs. using up space for three separate VMs.
5. Checks and installs docker on the VMs and creates a swarm on those.

User can provide an OVA that has docker already installed on it in which case the script won't bother installing docker.

Photon OVA has an issue that the first login needs the root password to be changed, This is difficult to handle in a script and hence unable to use Photon presently. I can use expect to work around this but thats not done yet. Will need expect to be installed in vibauthor-go container so it can be used or user should have Tcl/Tk installed on the VM to have expect available for the script. If the OVA has the password for root user already set (which is ideal as the user can set up their VM the way they want and then use that as the VM type for their test deployments) then the script should work ok.

### DIFF
--- a/client_plugin/Makefile
+++ b/client_plugin/Makefile
@@ -63,6 +63,11 @@ SYSTEMD_LIB  := $(PACKAGE)/lib/systemd/system/
 INSTALL_BIN  := $(PACKAGE)/usr/local/bin
 UPSTART_INIT := $(PACKAGE)/etc/init/
 
+#
+# Scripts to set up VMs to create a Docker swarm test bed
+#
+SETUP_VMS := $(SCRIPTS)/testbed.sh
+
 # esx service for docker volume ops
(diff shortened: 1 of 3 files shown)


govint:
Adding command line options and turning off actions that can't be supported (OVF install) between docker host and ESX. Will repost the PR again tomorrow.

PR #1620, opened 2017-07-21
Fix in tenancy.md. [SKIP CI]

lipingxue:
I found that in tenancy.md,  "default_datastore" is missing in the output of "vmgroup ls" command. This PR is to fix that.


shuklanirdesh82:
@lipingxue After merging your changes, steps mentioned at https://github.com/vmware/docker-volume-vsphere/blob/master/CONTRIBUTING.md#documentation-pr needs to be followed.

Issue #1621, opened 2017-07-21
Docker cli lists volume as attached whereas admin cli shows same volume as detached


ashahi1:
Test runs on a swarm setup of one master node and two worker nodes.

Test starts a service on swarm worker-1. Then first worker node is shutdown and test expects docker service to start on second worker. After the service successfully starts on the second worker, but one of the verification fails because docker cli shows volume as attached whereas admin cli shows volume as attached.

https://ci.vmware.run/vmware/docker-volume-vsphere/1266

```
Test steps:
Test vDVS usage during failover across different swarm nodes
Test steps:
1. Create a service with volume mounted
2. Verify the service is up and running on one node
3. Verify one container is spawned
4. Verify the volume is in attached status
5. Write data to the volume
6. Shutdown the node on which the service is running
7. Verify the service is restarted on a different node
8. Verify the volume is in attached status    <<<<<<<<<<<<<<<<< Fails here
9. Verify the data from this node
10. Remove the service
11. Verify the service is gone
12. Verify the volume is in detached status
```

Test logs:
```
2017/07/21 15:28:09 Shutting down VM [Ubuntu.1404.vsan]
2017/07/21 15:28:09 Sleep for 5 seconds
2017/07/21 15:28:14 Sleep for 5 seconds
2017/07/21 15:28:19 Sleep for 5 seconds
2017/07/21 15:28:24 Sleep for 5 seconds
2017/07/21 15:28:29 Checking docker service [swarm_test_service_933036] status on VM [192.168.31.241]
2017/07/21 15:28:29 Listing docker service [swarm_test_service_933036] running on VM [192.168.31.241]
2017/07/21 15:28:29 Checking running containers for docker service [swarm_test_service_933036] on docker hosts: [192.168.31.163]
2017/07/21 15:28:29 Containers running on docker host [192.168.31.163]: swarm_test_service_933036.1.i8dtp2ko28psxqz79xv4x1fg5
2017/07/21 15:28:29 Printing Docker volume inspect for volume swarm_test_volume_1046104
2017/07/21 15:28:30 OUT: [
    {
        "Driver": "vsphere:latest",
        "Labels": null,
        "Mountpoint": "/mnt/vmdk/swarm_test_volume_1046104@datastore1",
        "Name": "swarm_test_volume_1046104@datastore1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "access": "read-write",
            "attach-as": "independent_persistent",
[...]

PR #1626, opened 2017-07-24
Support test and set of KV via vm listener

govint:
Modified the disk_detach_int code path to include optional key, value params. The KV will be written with the given dict only if the current KV data matches the given key, value pair.

Ensuring a test-and-set of the KV in specific scenarios - right now just the VM listener.


govint:
Fixes #1621

esx_service/vmdk_ops.py
```diff
@@ -1426,7 +1426,7 @@ def disk_detach_int(vmdk_path, vm, device):
         logging.warning("%s\n%s", msg, "".join(traceback.format_tb(ex_traceback)))
         return err(msg)
 
-    setStatusDetached(vmdk_path)
+    setStatusDetached(vmdk_path, key=None, value=None)
     logging.info("Disk detached %s", vmdk_path)
```
msterin: There is no need to pass key and value here as None foe them is already default.

govint: Sorry, just cut-paste, fixed.

PR #1631, opened 2017-07-25
Shared volume create

shivanshu21:
**Create and inspect workflow**

Creates initial metadata
Creates internal volume
Updates metadata
No watcher being set. Will do in mount code.
Functions to read / write / delete from etcd
All writes of multiple keys are transactional
Fetches metadata from etcd and implements inspect

** Testing done**
Raised issue #1660 to track writing automated tests

Manually tested create and inspect
Positive cases:
*Run the commands create and inspect
Negative cases:
*Try create from worker node (not supported yet)
*Try inspect of non existent volumes
*Create same volume twice (After successful get, docker does nothing)

 
**Logs**
```
root@ubuntu16A ~: docker volume create --driver kahuna vol1
Error response from daemon: create vol1: VolumeDriver.Create: This node is not a swarm manager. Shared Volume creation is only possible from swarm manager nodes.
```

```
root@ubuntu16B ~: docker volume create --driver kahuna vol88
vol88


root@ubuntu16B ~: docker volume inspect vol88
[
    {
        "Driver": "kahuna:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/vol88",
        "Name": "vol88",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "ubuntu16B"
            ],
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "Not started yet",
            "Volume Status": "Ready"
        }
    }
]
root@ubuntu16B ~: docker volume inspect vol87
[]
Error: No such volume: vol87
root@ubuntu16B ~: 
```

** Logs **
2017-07-26 18:21:23.795093003 -0700 PDT [INFO] VolumeDriver Get
2017-07-26 18:21:23.811413203 -0700 PDT [INFO] Entering VolumeDriver Create
2017-07-26 18:21:23.822709884 -0700 PDT [INFO] Attempting to write initial metadata entry for vol88
2017-07-26 18:21:23.836121481 -0700 PDT [INFO] Attempting to create internal volume for vol88
2017-07-26 18:21:24.575397104 -0700 PDT [INFO] Attempting to set watcher for volume: vol88
2017-07-26 18:21:24.575876087 -0700 PDT [INFO] Attempting to update volume state to ready for volume: vol88
2017-07-26 18:21:24.590800048 -0700 PDT [INFO] Successfully created volume: vol88

================================================================

**After addressing review comments**

root@ubuntu16B ~: docker volume inspect vol1
[]
Error: No such volume: vol1
root@ubuntu16B ~: docker volume create --driver kahuna vol1
vol1
root@ubuntu16B ~: docker volume inspect vol1
[
    {
        "Driver": "kahuna:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/vol1",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Ready"
        }
    }
]
root@ubuntu16B ~: 

========================================================

**After reapplying miao's lost change**
Squashing all comment redressals together
Earlier git rebase / stash pop overwrote one of Miao's changes
Fixed that now

**Testing logs**

root@ubuntu16B ~: docker volume ls
root@ubuntu16B ~: docker volume create --driver kahuna vol1
vol1
root@ubuntu16B ~: docker volume create --driver kahuna vol2
vol2
root@ubuntu16B ~: docker volume ls
vsphere:latest      InternalVolvol1@datastore1
vsphere:latest      InternalVolvol2@datastore1
kahuna:latest       vol1
kahuna:latest       vol2
root@ubuntu16B ~: 
root@ubuntu16B ~: docker volume inspect vol1
[
    {
        "Driver": "kahuna:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/vol1",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Ready"
        }
    }
]
root@ubuntu16B ~:


client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +147,109 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+
+	// Ensure that the node running this command is a manager
+	info, err := d.dockerd.Info(context.Background())
+	if err != nil {
+		msg = fmt.Sprintf("Failed to get Info from docker client. Reason: %v", err)
+		log.Warningf(msg)
+		return volume.Response{Err: msg}
+	}
+	if info.Swarm.ControlAvailable == false {
+		msg = fmt.Sprintf("This node is not a swarm manager. Shared Volume creation is only possible from swarm manager nodes.")
```
govint: Is it possible to list, inspect, volumes from all nodes or only on the swarm manager node?  Is it possible to create a volume via the vmdk plugin and then be able to share that volume across nodes via the shared plugin? If the datastores are the same then can't the shared plugin allow using volumes created by the vmdk plugin?

luomiao: It's possible to list/inspect volumes from all the nodes.
It's possible to support a volume created by vmdk plugin to share it through shared plugin. But this is not a planned feature for this release.

luomiao:
@shivanshu21 
The PR is not rebased correctly.
My PR for etcd is already in the master.
It should not show up here in your PR.
You need to first checkout the latest master and cherry-pick your commit only to it.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -33,6 +36,11 @@ const (
 	etcdScheme               = "http://"
 	etcdClusterStateNew      = "new"
 	etcdClusterStateExisting = "existing"
+	kvStoreRequestTimeout    = 5
```
luomiao: How about change this to 
`kvStoreRequestTimeout = 5*time.Second`
And replace all the following `kvStoreRequestTimeout*time.Second` to `kvStoreRequestTimeout`

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	var value string
+	var err error
+
+	value = fmt.Sprintf("%d:", record.port)
+	value += record.username + ":"
+	value += record.password + ":"
+	for _, element := range record.clientList {
+		value += element
+		value += "?"
+	}
```
luomiao: Maybe we should use json methods to write/read the metadata in strings.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	words := strings.Split(value, ":")
+	record.port, _ = strconv.Atoi(words[0])
+	record.username = words[1]
+	record.password = words[2]
+	hostnames := strings.Split(words[3], "?")
+	for _, element := range hostnames {
+		if element != "" {
+			record.clientList = append(record.clientList, element)
+		}
+	}
```
luomiao: Same as above. We maybe able to use json methods to simply the read processing.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -62,6 +86,18 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 	d.dockerd = cli
 
+	// Pull Samba image
+	log.Infof("Pulling Samba server image")
```
luomiao: Is it possible to download the image when we create the plugin?
And we only need to load it when it's needed.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +147,109 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	volRecord := VolumeMetadata{
+		status:         volStateCreating,
+		globalRefcount: 0,
+		port:           0,
+		username:       "root",
+		password:       "badpass",
+		clientList:     hnames,
+	}
+	log.Infof("Attempting to write initial metadata entry for %s", r.Name)
+	err = d.writeVolMetadata(r.Name, volRecord)
```
luomiao: What if there is already someone else is creating the volume?
It seems this writeVolMetadata doesn't check if a key is already there with the same name.
We need to read to make sure there is no key with the same volume name existing.
(That's the Get() in the workflow!)

client_plugin/utils/config/config.go
```diff
@@ -51,6 +51,7 @@ const (
 // Config stores the configuration for the plugin
 type Config struct {
 	Driver        string `json:",omitempty"`
+	SlaveDriver   string `json:",omitempty"`
```
luomiao: This should not be here.
This is a shared plugin's specific flag.
You should put it inside struct `VolumeDriver` shared_driver.go and read it through flag.
Check how `project` is used inside photon_driver.go.

client_plugin/drivers/shared/etcdops.go
msterin: It's not maybe, it's certainly.

client_plugin/drivers/shared/shared_driver.go
msterin: Yes, I think pull on install is reasonable, also guarantees no out-of-disk-space failures on runtime.

shivanshu21: Was thinking about that. But downloading during plugin creation would mean that the samba server will become a part of the plugin.

We want to run it outside of plugin as a docker service. Correct me if I am wrong.

luomiao: So if we use the same dockerd to load the image, it will be loaded into to the host's dockerd.
The same as what you have done here. But `ImagePull` -> `ImageLoad`.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
+	volStateReady        = "Ready"
+	volStateMounted      = "Mounted"
+	volStateIntermediate = "Intermediate"
+	sambaImageName       = "dperson/samba" // TODO Replace with our own samba later
 )
 
+// VolumeMetadata structure contains all the metadata about a volume
+// that will be put in etcd
+type VolumeMetadata struct {
+	status         string
```
luomiao: Better to use a type for the status:
```
type volState string

const (
	volStateCreating volState = "creating"
	volStateReady volState = "ready"
	volStateMounted volState = "mounted"
	volStateIntermediate volState = "intermediate"
)

type VolumeMedata struct {
...
status volState
}
```

client_plugin/drivers/shared/etcdops.go
```diff
@@ -33,6 +36,11 @@ const (
 	etcdScheme               = "http://"
 	etcdClusterStateNew      = "new"
 	etcdClusterStateExisting = "existing"
+	kvStoreRequestTimeout    = 5
+	keyPrefix                = "SVOL_"
+	dataKey                  = "_data"
```
msterin: please do add comments explaining the data

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 
 	return cli, nil
 }
+
+// Update or Create volume metadata in KV store
+func (d *VolumeDriver) writeVolMetadata(name string,
```
msterin: is it supposed to be transactional?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	record VolumeMetadata) error {
+
+	var msg string
+	var value string
+	var err error
+
+	value = fmt.Sprintf("%d:", record.port)
```
msterin: if you are already using spintf, use it to add username/pasword as well

client_plugin/drivers/shared/etcdops.go
msterin: yes, this is actually better. Please try to avoid sprintfs/etc and instead just keep json

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	value += record.password + ":"
+	for _, element := range record.clientList {
+		value += element
+		value += "?"
+	}
+
+	etcdAPI := d.createEtcdClient()
+	defer etcdAPI.Close()
+
+	// Write grefc
```
msterin: I can assume grefc is global ref counter. However, the struct field needs to be commented, and please, do not save 
a few key strokes in the comments. `// update global ref counter'` is probably what you wanted to say here. Also I suggest separating this in a function (optional) to improve readability , writeVolMetadata() is getting too long

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	ctx, cancel := context.WithTimeout(context.Background(),
+		kvStoreRequestTimeout*time.Second)
+	_, err = etcdAPI.Put(ctx, keyPrefix+name+refKey, fmt.Sprintf("%d", record.globalRefcount))
+	cancel()
```
msterin: what dies cancel() do here ? it is cancelling timeout  ?  a comment would help

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	if err != nil {
+		msg = fmt.Sprintf("Failed to write grefc metadata for volume %s. Reason: %v", name, err)
+		log.Warningf(msg)
+		return errors.New(msg)
+	}
+
+	// Write volume status
+	ctx, cancel = context.WithTimeout(context.Background(),
+		kvStoreRequestTimeout*time.Second)
+	_, err = etcdAPI.Put(ctx, keyPrefix+name+statusKey, record.status)
```
msterin: does the API have "bulk put"  or transactions ? There are too many race opportunities here. 
Maybe you should have a lock (via etcd KV pair) and have others wait on the lock release ?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +288,122 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	resp, err := etcdAPI.Get(ctx, keyPrefix+name, etcdClient.WithPrefix(),
+		etcdClient.WithSort(etcdClient.SortByKey, etcdClient.SortDescend))
+	cancel()
+	if err != nil {
+		msg = fmt.Sprintf("Failed to read metadata entry for volume %s. Reason: %v", name, err)
+		log.Warningf(msg)
+		return record, errors.New(msg)
+	}
+
+	for _, ev := range resp.Kvs {
```
msterin: json is your friend, please avoid doing this kind of string manipulations

client_plugin/drivers/shared/shared_driver.go
shivanshu21: The docker engine calls Get() before it calls Create(). That get() will respond with "volume found" if there is an entry.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
 )
 
 const (
-	version       = "vSphere Shared Volume Driver v0.2"
-	apiVersion    = "v1.24" // docker engine 1.24 and above support this api version
-	dockerUSocket = "unix:///var/run/docker.sock"
+	version              = "vSphere Shared Volume Driver v0.2"
+	apiVersion           = "v1.24" // docker engine 1.24 and above support this api version
+	dockerUSocket        = "unix:///var/run/docker.sock"
+	tradVolumePrefix     = "InternalTrad"
```
msterin: what is trad ? a comment would be good

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
+	volStateCreating     = "Creating"
+	volStateReady        = "Ready"
+	volStateMounted      = "Mounted"
```
msterin: The states are not obvious, a comment would be good

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
+	volStateIntermediate = "Intermediate"
+	sambaImageName       = "dperson/samba" // TODO Replace with our own samba later
```
msterin: I suggest to replace it now. Just push manually to cnastorage/shared-vmdk-samba or something like that. 
Would be easier that to track it

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
 
+// VolumeMetadata structure contains all the metadata about a volume
+// that will be put in etcd
+type VolumeMetadata struct {
+	status         string
+	globalRefcount int
+	port           int
+	// Username and pass are just placeholders for now
+	// Later they can be used for multitenancy.
+	username   string
```
msterin: username/password for what ? Why do we need it here ?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,48 @@ package shared
...
+	password   string
+	clientList []string
```
msterin: comments please. Is it IPs ? something else ?

client_plugin/drivers/shared/shared_driver.go
msterin: Your guessing is  correct 

 > But downloading during plugin creation would mean that the samba server will become a part of the plugin.

No, it will mean that samba container will be pre-loaded into Docker cache, and we will know for sure it will fit

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -62,6 +87,18 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 	d.dockerd = cli
 
+	// Pull Samba image
+	log.Infof("Pulling Samba server image")
+	imgOptions := dockerTypes.ImagePullOptions{
+		All:           false,
```
msterin: please wrap individual functionality pieces in dedicated functions

client_plugin/drivers/shared/shared_driver.go
luomiao: Thanks. That makes sense.

client_plugin/utils/config/config.go
```diff
@@ -155,6 +156,7 @@ func InitConfig(defaultConfigPath string, defaultLogPath string, defaultDriver s
 	}
 	configFile := flag.String("config", defaultConfigPath, "Configuration file path")
 	driverName := flag.String("driver", "", "Volume driver")
+	sharedDriverSlave := flag.String("sharedDriverSlave", "", "Driver which creates internal volumes")
```
msterin: the name is not semantically accurate. Just call it storageVolimeDriver or internalVolumeDriver something like that

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +148,114 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	_, err = d.dockerd.VolumeCreate(context.Background(), dockerVolOptions)
+	if err != nil {
+		omsg := fmt.Sprintf("Failed to create volume %s. Reason: %v",
+			r.Name, err)
+		log.Warningf(omsg)
+
+		// If failed, attempt to delete the metadata for this volume
+		err = d.deleteVolMetadata(r.Name)
+		if err != nil {
+			msg = fmt.Sprintf("%v. Please remove metadata entry manually using etcd CLI.", err)
```
msterin: I do not think this is very helpful. Which metadata ?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +148,114 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	log.Infof(msg)
+	internalVolname := tradVolumePrefix + r.Name
+	dockerVolOptions := dockerTypes.VolumeCreateRequest{
+		Driver:     d.sharedDriverSlave,
+		Name:       internalVolname,
+		DriverOpts: r.Options,
+	}
+	_, err = d.dockerd.VolumeCreate(context.Background(), dockerVolOptions)
+	if err != nil {
+		omsg := fmt.Sprintf("Failed to create volume %s. Reason: %v",
```
msterin: can it be renamed to something meaninful  ?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +148,114 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	log.Infof("Attempting to set watcher for volume: %s", r.Name)
+
+	// Update metadata to indicate successful volume creation
+	log.Infof("Attempting to update volume state to ready for volume: %s", r.Name)
+	volRecord = VolumeMetadata{
+		status:         volStateReady,
+		globalRefcount: 0,
+		port:           0,
+		username:       "root",
+		password:       "badpass",
```
msterin: ????

If you need somethig to test, at least put it in constants with proper comments

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +148,114 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	err = d.writeVolMetadata(r.Name, volRecord)
+	if err != nil {
+		omsg := fmt.Sprintf("Failed to create volume %s. Reason: %v", r.Name, err)
+		log.Warningf(omsg)
+
+		// If failed, attempt to remove the backing trad volume
+		log.Infof("Attempting to delete internal volume")
+		err = d.dockerd.VolumeRemove(context.Background(), internalVolname)
+		if err != nil {
+			msg = fmt.Sprintf(" Failed to remove internal volume. Please remove")
```
msterin: I think I saw this somewhere already. Can you separate message building functions and put them close to each other in the code ?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +148,114 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+			msg += fmt.Sprintf("this volume manually using docker cli. Reason: %v ", err)
+			log.Warningf(msg)
+			omsg = omsg + msg
+		}
+
+		// Attempt to delete the metadata for this volume
+		err = d.deleteVolMetadata(r.Name)
+		if err != nil {
+			// Ques: Do we print etcddtl (cli) command here as corrective action?
+			msg = fmt.Sprintf(". Please remove metadata entry manually using etcd CLI.")
```
msterin: I certainly saw this already 

Also, please do not use "ques" , put TODO instead (easier to search for)

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +223,118 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+
+		// If failed, attempt to delete the metadata for this volume
+		err = d.deleteVolMetadata(r.Name)
+		if err != nil {
+			log.Warningf("Failed to remove metadata entry for volume: %s. Reason: %v", r.Name, err)
+		}
+		return volume.Response{Err: msg}
+	}
+
+	// TODO Set a watcher for this volume
```
luomiao: This can be removed. We won't start watcher here.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +223,118 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	log.Infof(msg)
+	internalVolname := tradVolumePrefix + r.Name
+	dockerVolOptions := dockerTypes.VolumeCreateRequest{
+		Driver:     d.internalVolumeDriver,
+		Name:       internalVolname,
+		DriverOpts: r.Options,
+	}
+	_, err = d.dockerd.VolumeCreate(context.Background(), dockerVolOptions)
+	if err != nil {
+		msg = fmt.Sprintf("Failed to create volume %s. Reason: %v", r.Name, err)
```
luomiao: Need to give the internal volume driver's name in the log message.
And provide suggestion to check the status of the internal volume driver.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
 )
 
+// volStatus: Datatype for keeping status of a shared volume
+type volStatus string
+
+/* Constants
+   version:             Version of the shared plugin driver
+   apiVersion:          docker engine 1.24 and above support this api version
+   dockerUSocket:       Unix socket on which Docker engine is listening
+   tradVolumePrefix:    Prefix for the names of internal volumes. These
```
luomiao: What's "trad" means? :)

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -52,6 +118,16 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	d.MountIDtoName = make(map[string]string)
 	d.MountRoot = mountDir
 
+	// Read flag from CLI. If not provided, use cfg value
+	internalVolumeParam := flag.String("sharedDriverSlave", "",
```
luomiao: sharedDriverSlave -> internalDriver?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +233,118 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	// TODO Set a watcher for this volume
+	log.Infof("Attempting to set watcher for volume: %s", r.Name)
+
+	// Update metadata to indicate successful volume creation
+	log.Infof("Attempting to update volume state to ready for volume: %s", r.Name)
+	entries = nil
+	entries = append(entries, kvPair{key: etcdPrefixState + r.Name, value: string(volStateReady)})
+	err = d.writeVolMetadata(entries)
+	if err != nil {
+		outerMessage := fmt.Sprintf("Failed to create volume %s. Reason: %v", r.Name, err)
```
luomiao: Error message is not accurate.
At this step, we already created the entry and created the internal volume.
But the error is because it failed to update the state of the volume to ready.

msterin:
Looks like lots of comments are addressed but the changes are squashed, so there is no way to see what was changed since the last review. Next time, please do keep "review fixes" and "original change" commits separately, it allow to avoid multiple re-reviews of the same code.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
+	Status         volStatus `json:"-"` // Field won't be marshalled
+	GlobalRefcount int       `json:"-"` // Field won't be marshalled
+	Port           int       `json:"port"`
+	ServiceName    string    `json:"serviceName"`
+	Username       string    `json:"username"`
+	Password       string    `json:"password"`
+	ClientList     []string  `json:"clientList"`
+}
+
+/* VolumeDriver - vsphere shared plugin volume driver struct
```
shuklanirdesh82:
```
drivers/shared/shared_driver.go:78:1: comment on exported type VolumeMetadata should be of the form "VolumeMetadata ..." (with optional leading article)
drivers/shared/shared_driver.go:101:1: comment on exported type VolumeDriver should be of the form "VolumeDriver ..." (with optional leading article)
drivers/shared/etcdops.go
```

need to make sure about turning off above warnings.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+		// come to an end. In this case if Get is still running
+		// in background, it will cancel it
+		cancel()
+		if err != nil {
+			msg = fmt.Sprintf("Failed to read metadata. Reason: %v", err)
+			log.Warningf(msg)
+			return entries, errors.New(msg)
+		}
+
+		// If you didnt find a key, there wont be
```
msterin: who is the "you" here ?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+		// an error returned. It will just return an empty resp
+                // Return whatever you have now. Caller will have to
```
msterin: nit - please alight the comments

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -91,8 +174,47 @@ func (d *VolumeDriver) List(r volume.Request) volume.Response {
...
+		log.Warningf(msg)
+		return statusMap, errors.New(msg)
+	}
+	if len(entries) == 0 {
+		// Volume does not exist!
+		msg := fmt.Sprintf("Volume does not exist")
+		return statusMap, errors.New(msg)
+	} else if len(entries) != len(keys) {
+		// Volume exists but we didn't find all its keys
+		msg := fmt.Sprintf("Failed to get volume: Couldn't find all keys!")
```
msterin: What does it mean ? Different etcd versoin? Data corruption ? Something else ?  If an admin sees this in the log. what do we want him to do ?

luomiao: Logically this should not happen. We create and delete all three type of keys at the same time.
So when this situation happens, I cannot think of a reason other than data corruption for now.
And I am not sure if etcd can have data corruption?

msterin: then assert

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +233,118 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	// Create traditional volume as backend to shared volume
+	msg = fmt.Sprintf("Attempting to create internal volume for %s", r.Name)
+	log.Infof(msg)
+	internalVolname := tradVolumePrefix + r.Name
+	dockerVolOptions := dockerTypes.VolumeCreateRequest{
+		Driver:     d.internalVolumeDriver,
+		Name:       internalVolname,
+		DriverOpts: r.Options,
+	}
+	_, err = d.dockerd.VolumeCreate(context.Background(), dockerVolOptions)
```
govint: Ok, if the host crashes before or after volumeCreate by the backing volume driver then on startup this plugin must go thru and figure which volumes are in creating state and remove those or update to _ready_ state if a backing volume is successfully created.

govint: What happens if this docker host crashes now and leaves the volume in creating state and then doesn't restart or plugin doesn't restart. The volume is in creating state, Can another node access the volume and can the user delete that volume?

What if two nodes are creating a volume of the same name? Can we elaborate how this logic works there.

This function should handle the case that the backing volume already exists/

Function should also handle being able to create volumes of the same name on different datastores, does the user mention the volume as <vol@DS> to the shared plugin?

Should include tests that try these scenarios

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+		ops = append(ops, etcdClient.OpPut(elem.key, elem.value))
+	}
+
+	// Lets write the metadata in a single transaction
+	// Use a transaction if more than one entries are to be written
+	if len(entries) > 1 {
+		_, err = etcdAPI.Txn(context.TODO()).Then(ops...).Commit()
+	} else {
+		_, err = etcdAPI.Do(context.TODO(), ops[0])
+	}
```
luomiao: I don't think txn is needed for a single entry though.
Even using txn won't prevent the situation you mentioned.
ETCD itself will keep strong consistency for a single entry.

"Also, would the caller need an atomic test and set kind of functionality"
Yes, this will come in a later PR when it's needed.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+	etcdAPI := d.createEtcdClient()
+	defer etcdAPI.Close()
+
+	// Read does not require transactions
+	// Just read all the keys required
+	for _, k := range keys {
+		elem := kvPair{key: k, value: ""}
+		ctx, cancel := context.WithTimeout(context.Background(),
+			requestTimeout)
+		resp, err := etcdAPI.Get(ctx, k)
```
luomiao: Agree. This read function should use Txn too, as same as the write function.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
+	dockerUSocket    = "unix:///var/run/docker.sock"
+	tradVolumePrefix = "InternalTrad"
+	// TODO Replace with our own samba later
+	sambaImageName                 = "dperson/samba"
+	sambaUsername                  = "root"
+	sambaPassword                  = "badpass"
+	volStateCreating     volStatus = "Creating"
+	volStateReady        volStatus = "Ready"
+	volStateMounted      volStatus = "Mounted"
+	volStateIntermediate volStatus = "Intermediate"
```
luomiao: There are more states needed, such as "Error". We can add more when they are used :)

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -111,7 +233,118 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+		msg = fmt.Sprintf("Failed to fetch hostname during volume creation")
+		log.Warningf(msg)
+		return volume.Response{Err: msg}
+	}
+	hnames = append(hnames, hostname)
+	volRecord := VolumeMetadata{
+		Status:         volStateCreating,
+		GlobalRefcount: 0,
+		Port:           0,
+		ServiceName:    "Not started yet",
```
luomiao: My suggestion is, do not set all the fields below "Port" during initialization and add omitempty tag in the struct definition:

```
	Port           int       `json:"port,omitempty"`
 	ServiceName    string    `json:"serviceName,omitempty"`
 	Username       string    `json:"username,omitempty"`
 	Password       string    `json:"password,omitempty"`
 	ClientList     []string  `json:"clientList,omitempty"`
```

So they will be empty fields when they are not set.

luomiao:
@govint 
For some of your questions, I directly edited your comments for reply :)

client_plugin/drivers/shared/etcdops.go
```diff
@@ -33,8 +35,18 @@ const (
...
 )
 
+// kvPair : Key Value pair holder
+type kvPair struct {
+	key   string
+	value string
+}
+
 // initEtcd start or join ETCD cluster depending on the role of the node
 func (d *VolumeDriver) initEtcd() error {
```
luomiao: Good suggestion. I will include this change in the next etcd PR.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -33,8 +35,18 @@ const (
 	etcdScheme               = "http://"
 	etcdClusterStateNew      = "new"
 	etcdClusterStateExisting = "existing"
+	etcdPrefixState          = "SVOLS_stat_"
+	etcdPrefixGRef           = "SVOLS_gref_"
+	etcdPrefixInfo           = "SVOLS_info_"
+	requestTimeout           = 10 * time.Second
```
shivanshu21: There is another PR that will have a lot of rebasing to do if I do this now. So I will move them in another checkin which is due soon.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+			log.Warningf(msg)
+			return entries, errors.New(msg)
+		}
+
+		// If you didnt find a key, there wont be
+		// an error returned. It will just return an empty resp
+                // Return whatever you have now. Caller will have to
+                // match the number of keys demanded to keys returned
+		if resp.Count == 0 {
+			return entries, nil
```
shivanshu21: The intent is that user checks the len() of the returned slice. The len() for a just initialized slice is 0, so even if it fails on the first get, it will return a slice of len 0. I have verified this by writing a small piece of code.

client_plugin/drivers/shared/shared_driver.go
shivanshu21: Intermediate was supposed to be a state that signals other jobs not to proceed. The plan was to use this state for any meta data update, including delete.

Changing its text to "MetadataUpdateInProgress". Will introduce Error state in another PR.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
+	GlobalRefcount int       `json:"-"` // Field won't be marshalled
+	Port           int       `json:"port"`
+	ServiceName    string    `json:"serviceName"`
+	Username       string    `json:"username"`
+	Password       string    `json:"password"`
+	ClientList     []string  `json:"clientList"`
+}
+
+/* VolumeDriver - vsphere shared plugin volume driver struct
+   dockerd:                 Client used for talking to Docker
```
shivanshu21: Noted all changes that need to go in the next PR. Other PRs that are yet to be merged currently will be impacted if I make this change in this PR.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -280,3 +292,104 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
...
+		resp, err := etcdAPI.Get(ctx, k)
+		// cancel is context cancelFunc()
+		// Its purpose is to ensure that the previous etcd op has
+		// come to an end. In this case if Get is still running
+		// in background, it will cancel it
+		cancel()
+		if err != nil {
+			msg = fmt.Sprintf("Failed to read metadata. Reason: %v", err)
+			log.Warningf(msg)
+			return entries, errors.New(msg)
```
shivanshu21: Once entries is declared, even though its nil, its len() will be 0. Caller of this function is supposed to match the len() of received entries to len() of requested keys.

So this will work fine. Verified the same in a separate small piece of code.

client_plugin/drivers/shared/etcdops.go
shivanshu21: Transaction create a single revision number for all changes (write/delete) that happen to a set of keys. **They are not locks.**

Read operations don't modify the KV store. Hence no revision number is assigned to them.

I understand that if we read status=Mounted but Gref=0 that is a problem.

One possible solution is that we always write all 3 keys when updating any metadata. This will ensure that all three keys have same revision number at all times.

During read if the revision numbers don't match, we will fetch the keys again.

luomiao: From what I understand, wrapping GET operations inside a transaction will prevent modifying one of the keys in the middle.
According to https://coreos.com/blog/transactional-memory-with-etcd3.html
```
getresp, err := etcd.Txn(ctx.TODO()).Then(OpGet(from), OpGet(to)).Commit()
    if err != nil {
         return false, err
    }
```
indicate:

>  All Get requests are wrapped in a transaction so interleaved writes cannot update from while fetching to.

client_plugin/drivers/shared/etcdops.go
govint: But this is expecting that the caller of this interface works in a specific way, If entries is accessed as is then that would be a problem. This function should be uniform in what it returns as values - either empty slice or nil in case of error, not both.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
+   internalVolumeDriver:    Name of the plugin used by shared volume
```
govint: So, isn't the version needed?

shivanshu21: We have a pre requisite for the user to install both the vDVS (or any other backing plugin) and the shared volume plugin on the system. This is something the user has to do.

But once this setup is correctly done, the shared volume plugin only talks to docker API over unix socket and mentions the "driver" it wants to use. The default is "vsphere" which refers to vDVS plugin.

It the same as when we do volume creation manually via CLI: docker volume create --driver vsphere vol1

Version information is not needed to make this API call, hence it is not required in the shared volume metadata.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -23,24 +23,90 @@ package shared
...
+                            plugin to create internal volumes
+*/
 type VolumeDriver struct {
 	utils.PluginDriver
-	dockerd *dockerClient.Client
+	dockerd              *dockerClient.Client
+	internalVolumeDriver string
```
shivanshu21: Side note: vShared sounds like a great name for the shared volume plugin btw. :-)

client_plugin/drivers/shared/shared_driver.go
govint: Ah, thats correct, the shared plugin talks only to Docker. Sorry missed that.

luomiao:
LGTM

client_plugin/drivers/shared/etcdops.go
govint: Ha, absolutely we don't need a txn for single entries, but it will be good to know if Etcd maintains any order to updating key values.

Issue #1615, opened 2017-07-21
File system for created volume is N/A


ashahi1:
Steps:
1. Verified that there are no volumes.
2. Ran a container by passing -v option - Command throws an error ```docker: Error response from daemon: create Vol22@sharedVmfs-0: Post http://%2Frun%2Fdocker%2Fplugins%2Fa2341bfd26d15ba333e64eb30eb53e8174d87716ed1168313e9fecc779f7b759%2Fvsphere.sock/VolumeDriver.Create: http: ContentLength=42 with Body length 0.```
3. Admin cli and docker cli lists the volume
4. File system for volume is listed as N/A on admin cli

Steps and their output are as follows:

-------------------------------

1. Verified that there are no volumes
```
root@sc-rdops-vm02-dhcp-52-237:~# docker volume ls
DRIVER              VOLUME NAME
root@sc-rdops-vm02-dhcp-52-237:~#
```
2. Ran a container by passing -v option and specifying volume name as - Vol22@sharedVmfs-0
```
root@sc-rdops-vm02-dhcp-52-237:~# docker run -it --volume-driver=vsphere -v Vol22@sharedVmfs-0:/v1 --name=abc busybox
docker: Error response from daemon: create Vol22@sharedVmfs-0: Post http://%2Frun%2Fdocker%2Fplugins%2Fa2341bfd26d15ba333e64eb30eb53e8174d87716ed1168313e9fecc779f7b759%2Fvsphere.sock/VolumeDriver.Create: http: ContentLength=42 with Body length 0.
See 'docker run --help'.
```

3. Admin cli ls and docker volume ls lists the volume - Vol22@sharedVmfs-0
```
[root@sc2-rdops-vm01-dhcp-40-37:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
Volume  Datastore     VMGroup   Capacity  Used  Filesystem  Policy  Disk Format  Attached-to   Access      Attach-as               Created By    Created Date
------  ------------  --------  --------  ----  ----------  ------  -----------  ------------  ----------  ----------------------  ------------  ------------------------
Vol22   sharedVmfs-0  _DEFAULT  100MB     0     N/A         N/A     thin         ubuntu-VM2.2  read-write  independent_persistent  ubuntu-VM2.2  Fri Jul 21 00:49:52 2017

[root@sc2-rdops-vm01-dhcp-40-37:~]


[...]

PR #1636, opened 2017-07-26
Fixed request options init for create op.

govint:
Verified via the command line posted in the issue,

root@photon-machine [ /vol/vpl-dv/docker-volume-vsphere ]# docker run -it --volume-driver=vsphere -v debugvol14@sharedVmfs-0:/v1 --name=abc busybox
/ #


lipingxue:
@govint So this bug only exist in photon_driver code, right?
In #1615, you said that, fix from docker side is needed.  Please file an issue against docker and put the link in issue #1615.

govint:
@lipingxue, no, there is a bug in the vmdk and photon driver as well which is seen from 0.16 onwards. The docker side of  the issue is a separate part and I'll raise that. This change fixes the issue in the plugin code.

PR #1642, opened 2017-07-27
Implement a PowerShell script to handle vDVS windows plugin installation [SKIP CI]

shaominchen:
Implement a PowerShell script to handle vDVS windows plugin download, installation and reinstallation (overwrite existing installation). Later we can improve the same script to handle uninstallation as well.

Testing Done:

**1. Fresh installation:**
```
PS C:\Users\Administrator\go\src\github.com\vmware\docker-volume-vsphere> .\install-vdvs.ps1 http://engweb.vmware.com/~shchen/vdvs/docker-volume-vsphere.zip
Downloading from http://engweb.vmware.com/~shchen/vdvs/docker-volume-vsphere.zip...
Extracting docker-volume-vsphere.zip into C:\Program Files\VMware\vmdkops...
Installing Windows service vdvs from C:\Program Files\VMware\vmdkops\vdvs.exe...

Status   Name               DisplayName
------   ----               -----------
Stopped  vdvs               vSphere Docker Volume Service
Starting Windows service vdvs...
Running  vdvs               vSphere Docker Volume Service
```
**2. Re-installation (overwrite existing installation):**
```
PS C:\Users\Administrator\go\src\github.com\vmware\docker-volume-vsphere> .\install-vdvs.ps1 http://engweb.vmware.com/~shchen/vdvs/docker-volume-vsphere.zip
Windows service vdvs is already installed.
Do you want to reinstall [Y/N]?: y
Stopping Windows service vdvs...
Uninstalling Windows service vdvs...
[SC] DeleteService SUCCESS
Downloading from http://engweb.vmware.com/~shchen/vdvs/docker-volume-vsphere.zip...
Extracting docker-volume-vsphere.zip into C:\Program Files\VMware\vmdkops...
Installing Windows service vdvs from C:\Program Files\VMware\vmdkops\vdvs.exe...

Status   Name               DisplayName
------   ----               -----------
Stopped  vdvs               vSphere Docker Volume Service
Starting Windows service vdvs...
Running  vdvs               vSphere Docker Volume Service
```


install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+# Check if vdvs plugin is already installed
+$svc = Get-Service | Where-Object {$_.Name -eq $svcName}
+if ($svc) {
+    echo "Windows service $svcName is already installed."
+    $reinstall = Read-Host "Do you want to reinstall [Y/N]?"
+    if ($reinstall -eq "Y") {
+        $svc = Get-Service -Name $svcName
+        if ($svc.Status -eq "Running") {
+            echo "Stopping Windows service $svcName..."
+            sc.exe stop $svcName
```
msterin: I suspect Powershell has an cmdlet for this `Stop-Service`

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+            echo "Stopping Windows service $svcName successfully!"
+        }
+        echo "Uninstalling Windows service $svcName..."
+        sc.exe delete $svcName
```
msterin: Same here. No need to use sc.exe, PowerShell has cmdlets for all

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+if ($?) {
+    echo "Starting Windows service $svcName successfully!"
+} else {
+    echo "Failed to start Windows service $svcName."
+    return
+}
+
+# Show the running service
+$svc = Get-Service -Name $svcName
+echo $svc
```
msterin: why do you need echo here ? Get-service will redirect to stdout by default

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+if ($?) {
+    echo "Extracting $zipFileName into $installPath successfully!"
+} else {
+    echo "Failed to extract $zipFileName into $installPath."
+    return
+}
+
+# Install the vdvs plugin as a service
+echo "Installing Windows service $svcName from $exePathName..."
+New-Service -Name $svcName -BinaryPathName $exePathName -DisplayName $svcDisplayName -Description $svcDescription
```
msterin: you don't need the $? here. You can have the same by simply saying something like that
```powershell
if ( New-Service -Name $svcName -BinaryPathName $exePathName ... ) { 
    echo OK
} else { 
   echo bad
   return
}
```

this comment applies to all `if $?` places

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+} else {
+    echo "Failed to download from $uri."
+    return
+}
+
+# Extract to the installation folder
+echo "Extracting $zipFileName into $installPath..."
+Expand-Archive -Path $zipFileName -DestinationPath $installPath -Force
+if ($?) {
+    echo "Extracting $zipFileName into $installPath successfully!"
```
msterin: I don't think this message is needed as you already said "extacting..." so can just move on  here and react to failure only

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+
+# Installation Instructions
+# ============================================================================
+# Run the script with the plugin url as the parameter in a PowerShell session:
+# PS C:\Users\Administrator> .\install-vdvs.ps1 <vdvs-download-url>
+# ============================================================================
+
+# Command line parameters
+param (
+    [Parameter(Mandatory=$true)][string]$uri
```
venilnoronha: Would it be good to have `uri` as an optional parameter? Any naive user should simply be able to do `.\install-vdvs.ps1` and the script could install the latest version of the plugin.

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+$svcName = "vdvs"
+$svcDisplayName = "vSphere Docker Volume Service"
+$svcDescription = "Enables user to run stateful containerized applications on top of VMware vSphere."
+$installPath = "C:\Program Files\VMware\vmdkops"
+$exePathName = $installPath + "\vdvs.exe"
+$zipFileName = "docker-volume-vsphere.zip"
+
+# Check if vdvs plugin is already installed
+$svc = Get-Service | Where-Object {$_.Name -eq $svcName}
+if ($svc) {
```
venilnoronha: Can we simplify it to `if (Get-Service -Name $svcName) {`?

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Installation Instructions
+# ============================================================================
+# Run the script with the plugin url as the parameter in a PowerShell session:
+# PS C:\Users\Administrator> .\install-vdvs.ps1 <vdvs-download-url>
```
venilnoronha: Should we handle a `--help` parameter and print usage?

build.bat
```diff
@@ -108,6 +108,7 @@ echo Moving binaries to the build directory.
 if not exist %winBuildDir% mkdir %winBuildDir%
 move /y %vmdkopsDir%\vmci_client.dll %winBuildDir%
 move /y vdvs.exe %winBuildDir%
+powershell compress-archive -path %winBuildDir%\* -DestinationPath %winBuildDir%\docker-volume-vsphere.zip
```
venilnoronha: nit: `compress-archive -path` -> `Compress-Archive -Path`

install-vdvs.ps1
shaominchen: Yes, there is. Will replace with Stop-Service.

install-vdvs.ps1
shaominchen: "PowerShell has cmdlets for all" - this is not true. I didn't find a cmdlet to remove the service. Of course we can write a statement or a function to do this, but using sc.exe seems simpler. Let me know if you have better suggestions.

install-vdvs.ps1
shaominchen: Will remove.

install-vdvs.ps1
shaominchen: Just personal preference to print a successful message. I'm fine to remove this since you feel it's duplicate.

install-vdvs.ps1
shaominchen: PowerShell does not follow this convention. It uses Get-Help cmdlet to fetch help information for all cmdlets. We need to investigate how to interact with Get-Help. That might be trivial, but I think it can be addressed separately.

install-vdvs.ps1
shaominchen: I was thinking the same. To change it to an optional parameter, we should have a default value here. However, currently the url of the latest version of the plugin is not confirmed yet. Once that is confirmed, we can change this to optional at any time.

install-vdvs.ps1
shaominchen: Sure.

build.bat
shaominchen: Fixed.

install-vdvs.ps1
venilnoronha: Sounds good.

install-vdvs.ps1
venilnoronha: Should we finalize the URL with this PR, or should we track it as a separate issue?

install-vdvs.ps1
venilnoronha: Then, do you want to remove L#40?

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+$svc = Get-Service | Where-Object {$_.Name -eq $svcName}
+if ($svc) {
+    echo "Windows service $svcName is already installed."
+    $reinstall = Read-Host "Do you want to reinstall [Y/N]?"
+    if ($reinstall -eq "Y") {
+        $svc = Get-Service -Name $svcName
+        if ($svc.Status -eq "Running") {
+            echo "Stopping Windows service $svcName..."
+            sc.exe stop $svcName
+            echo "Stopping Windows service $svcName successfully!"
```
venilnoronha: nit: Stopping -> Stopped

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+        }
+        echo "Uninstalling Windows service $svcName..."
+        sc.exe delete $svcName
+        echo "Uninstalling Windows service $svcName successfully!"
```
venilnoronha: nit: Uninstalling -> Uninstalled

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+        return
+    }
+}
+
+# Download the vdvs binary
+echo "Downloading from $uri..."
+Invoke-WebRequest $uri -OutFile $zipFileName
+
+if ($?) {
+    echo "Downloading from $uri successfully!"
```
venilnoronha: nit: Downloading -> Downloaded

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+} else {
+    echo "Failed to extract $zipFileName into $installPath."
+    return
+}
+
+# Install the vdvs plugin as a service
+echo "Installing Windows service $svcName from $exePathName..."
+New-Service -Name $svcName -BinaryPathName $exePathName -DisplayName $svcDisplayName -Description $svcDescription
+if ($?) {
+    echo "Installing Windows service $svcName successfully!"
```
venilnoronha: nit: Installing -> Installed

install-vdvs.ps1
```diff
@@ -0,0 +1,97 @@
...
+} else {
+    echo "Failed to install Windows service $svcName."
+    return
+}
+
+# Start the vdvs service
+echo "Starting Windows service $svcName..."
+Start-Service -Name $svcName
+if ($?) {
+    echo "Starting Windows service $svcName successfully!"
```
venilnoronha: nit: Starting -> Started

install-vdvs.ps1
shaominchen: I don't think we can finalize the URL right now as we are not sure what version number and/or build number will be used in the final binary file name.

install-vdvs.ps1
shaominchen: Yes, L#40 can be removed - good catch! But no, we can't simplify the previous line to this simple statement, because this statement will throw exception if the service with given name does not exist.

install-vdvs.ps1
shaominchen: That's not the case. Above code won't work. We need $? here to check the exit status of the last command. I will change it to if (! $?) { echo bad}. Let me know if you have better suggestions.

install-vdvs.ps1
venilnoronha: Okay, we should create an issue for this.

install-vdvs.ps1
venilnoronha: We could do `$svc = Get-Service -Name $svcName -ErrorAction SilentlyContinue`. Even the current statement is fine.

install-vdvs.ps1
shaominchen: Yes, I will file an issue for the remaining work:
1. Support Get-Help
2. Make uri parameter optional, set default value to the latest version
3. Support uninstallation

venilnoronha: Sounds great!

install-vdvs.ps1
msterin: Hmm... maybe for new-process ? I did not try for it but for get-process it certainly works for me
```powershell
if (Get-Process -Name $svcname -ErrorAction SilentlyContinue)  {
  Write-Host "Found"
} else  {
  Write-Host "Not found"
}
```

shuklanirdesh82:
@shaominchen Actually there is no need to invoke CI run for this PR. I have killed 3 runs invoked from this PR. It is better to use `[SKIP CI]` for such PRs which don't require CI runs.

install-vdvs.ps1
shaominchen: Get-Process is a query which returns something. New-Service does not return anything, so above code will always echo bad.

PR #1661, opened 2017-07-30
Shared plugin: add etcd watcher and event handler

luomiao:
Every docker swarm manager starts a watcher on all the global
refcount keys in etcd. All the managers get events when any global
refcount key is updated. Only one of the managers is able to get
the right to adjust the state of volume and start/stop the file
system server.

This PR also changed some of the etcd functions into kvStore interface:
```
type kvStore interface {
	// Init - initialize the KV store cluster
	Init() error
	// WriteVolMetadata - Update or Create volume metadata in KV store
	WriteVolMetadata(entries []kvPair) error
	// ReadVolMetadata - Read volume metadata in KV store
	ReadVolMetadata(keys []string) ([]kvPair, error)
        // DeleteVolMetadata - Delete volume metadata in KV store
        DeleteVolMetadata(name string) error
	// CompareAndPut - Compare the value of key with oldVal, if equal, replace with newVal
	CompareAndPut(key string, oldVal string, newVal string) bool
	// ListVolumeName - List all the volumes in the KV store
	ListVolumeName() ([]string, error)
}
```

This PR needs to be rebased after PR https://github.com/vmware/docker-volume-vsphere/pull/1631 is merged.

Basic e2e test passed on local testbed.

### DIFF
--- a/client_plugin/drivers/shared/etcdops.go
+++ b/client_plugin/drivers/shared/etcdops.go
@@ -49,6 +49,8 @@ import (
    requestTimeout:             After how long should an etcd request timeout
    etcdClientCreateError:      Error indicating failure to create etcd client
    VolumeDoesNotExistError:    Error indicating that there is no such volume
+   etcdSingleRef:              if global refcount 0 -> 1, start SMB server
+   etcdNoRef:                  if global refcount 1 -> 0, shut down SMB server
 */
 const (
 	etcdClientPort           = ":2379"
@@ -61,9 +63,12 @@ const (
 	etcdPrefixState          = "SVOLS_stat_"
 	etcdPrefixGRef           = "SVOLS_gref_"
 	etcdPrefixInfo           = "SVOLS_info_"
-	requestTimeout           = 10 * time.Second
+	requestTimeout           = 5 * time.Second
+	checkSleepDuration       = time.Second
 	etcdClientCreateError    = "Failed to create etcd client"
 	VolumeDoesNotExistError  = "No such volume"
+	etcdSingleRef            = "1"
+	etcdNoRef                = "0"
 )
 
 // kvPair : Key Value pair holder
@@ -72,10 +77,32 @@ type kvPair struct {
 	value string
 }
 
+type kvStore interface {
+	// Init - initialize the KV store cluster
+	Init() error
+	// WriteVolMetadata - Update or Create volume metadata in KV store
+	WriteVolMetadata(entries []kvPair) error
+	// ReadVolMetadata - Read volume metadata in KV store
+	ReadVolMetadata(keys []string) ([]kvPair, error)
+	// DeleteVolMetadata - Delete volume metadata in KV store
+	DeleteVolMetadata(name string) error
+	// CompareAndPut - Compare the value of key with oldVal, if equal, replace with newVal
+	CompareAndPut(key string, oldVal string, newVal string) bool
+	// ListVolumeName - List all the volumes in the KV store
+	ListVolumeName() ([]string, error)
+}
+
+type etcdInfo struct {
+	driver   *VolumeDriver
+	nodeID   string
+	nodeAddr string
+	client   *etcdClient.Client
+}
+
 // initEtcd start or join ETCD cluster depending on the role of the node
-func (d *VolumeDriver) initEtcd() error {
+func (e *etcdInfo) Init() error {
 	ctx := context.Background()
-	docker := d.dockerd
+	docker := e.driver.dockerd
 
 	// get NodeID from docker client
 	info, err := docker.Info(ctx)
@@ -88,6 +115,7 @@ func (d *VolumeDriver) initEtcd() error {
 
 	// worker just returns
 	nodeID := info.Swarm.NodeID
+	e.nodeID = nodeID
 	if info.Swarm.ControlAvailable == false {
 		log.WithFields(
 			log.Fields{"nodeID": nodeID},
@@ -105,13 +133,14 @@ func (d *VolumeDriver) initEtcd() error {
 
 	// get the IP address of current node
 	addr := info.Swarm.NodeAddr
+	e.nodeAddr = addr
 
 	// if leader, proceed to start ETCD cluster
 	if node.ManagerStatus.Leader {
 		log.WithFields(
 			log.Fields{"nodeID": nodeID},
 		).Info("Swarm node role: leader, start etcd cluster")
-		err = startEtcdCluster(addr, nodeID)
+		err = e.startEtcdCluster()
 		if err != nil {
 			log.WithFields(log.Fields{"nodeID": nodeID,
 				"error": err}).Error("Failed to start ETCD Cluster")
@@ -134,7 +163,7 @@ func (d *VolumeDriver) initEtcd() error {
 					"manager ID": nodeID},
 			).Info("Swarm node role: manager. Action: find leader ")
 
-			joinEtcdCluster(addr, n.ManagerStatus.Addr, nodeID)
+			e.joinEtcdCluster(n.ManagerStatus.Addr)
 			return nil
 		}
 	}
@@ -144,7 +173,9 @@ func (d *VolumeDriver) initEtcd() error {
 }
 
 // startEtcdCluster function is called by swarm leader to start a ETCD cluster
-func startEtcdCluster(nodeAddr string, nodeID string) error {
+func (e *etcdInfo) startEtcdCluster() error {
+	nodeID := e.nodeID
+	nodeAddr := e.nodeAddr
 	lines := []string{
 		"--name", nodeID,
 		"--advertise-client-urls", etcdScheme + nodeAddr + etcdClientPort,
@@ -155,13 +186,19 @@ func startEtcdCluster(nodeAddr string, nodeID string) error {
 		"--initial-cluster", nodeID + "=" + etcdScheme + nodeAddr + etcdPeerPort,
 		"--initial-cluster-state", etcdClusterStateNew,
 	}
+
+	// start the etcd cluster routine
 	go etcdService(lines)
 
-	return nil
+	// check if etcd cluster is successfully started, then start the watcher
+	return e.checkLocalEtcd()
 }
 
 // joinEtcdCluster function is called by a non-leader swarm manager to join a ETCD cluster
-func joinEtcdCluster(nodeAddr string, leaderAddr string, nodeID string) error {
+func (e *etcdInfo) joinEtcdCluster(leaderAddr string) error {
+	nodeAddr := e.nodeAddr
+	nodeID := e.nodeID
+
 	etcd, err := addrToEtcdClient(leaderAddr)
 	if err != nil {
 		log.WithFields(
@@ -255,9 +292,11 @@ func joinEtcdCluster(nodeAddr string, leaderAddr string, nodeID string) error {
 		"--initial-cluster-state", etcdClusterStateExisting,
 	}
 
+	// start the etcd cluster routine
 	go etcdService(lines)
 
-	return nil
+	// check if successfully joined the etcd cluster, then start the watcher
+	return e.checkLocalEtcd()
 }
 
 func etcdService(cmd []string) {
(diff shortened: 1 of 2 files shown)


client_plugin/drivers/shared/etcdops.go
```diff
@@ -134,7 +163,7 @@ func (d *VolumeDriver) initEtcd() error {
 					"manager ID": nodeID},
 			).Info("Swarm node role: manager. Action: find leader ")
 
-			joinEtcdCluster(addr, n.ManagerStatus.Addr, nodeID)
+			e.joinEtcdCluster(n.ManagerStatus.Addr)
 			return nil
```
lipingxue: Why here return error as "nil? I think e.startEtcdCluster returns with err, right?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -255,9 +292,11 @@ func joinEtcdCluster(nodeAddr string, leaderAddr string, nodeID string) error {
 		"--initial-cluster-state", etcdClusterStateExisting,
 	}
 
+	// start the etcd cluster routine
```
lipingxue: Why the comments is start the etcd cluster routine? I think it should be "join the etcd cluster"?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
 	}
 }
 
-func (d *VolumeDriver) createEtcdClient() *etcdClient.Client {
-	dclient := d.dockerd
+// checkLocalEtcd function check if local ETCD endpoint is successfully started or not
+// if yes, start the watcher for volume global refcount
+func (e *etcdInfo) checkLocalEtcd() error {
+	ticker := time.NewTicker(checkSleepDuration)
```
lipingxue: Just curious, how do we decide the right value of "checkSleepDuration" and "requestTimeout"?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+				go e.etcdWatcher()
+				return nil
+			}
+		case <-timer.C:
+			return fmt.Errorf("Timeout reached; etcd cluster is not started")
+		}
+	}
+}
+
+func (e *etcdInfo) etcdWatcher() {
```
lipingxue: Please add comments for this function.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+	watchCh := e.client.Watch(context.Background(), etcdPrefixGRef,
+		etcdClient.WithPrefix(), etcdClient.WithPrevKV())
+	for wresp := range watchCh {
+		for _, ev := range wresp.Events {
+			e.etcdEventHandler(ev)
+		}
+	}
+}
+
+func (e *etcdInfo) etcdEventHandler(ev *etcdClient.Event) {
```
lipingxue: Please add comments for this function.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+	log.WithFields(
+		log.Fields{"type": ev.Type},
+	).Infof("Watcher on global refcount returns event ")
+
+	if ev.Type == etcdClient.EventTypePut {
+		if string(ev.Kv.Value) == etcdSingleRef &&
+			ev.PrevKv != nil &&
+			string(ev.PrevKv.Value) == etcdNoRef {
+			// watcher observes global refcount from 0 to 1
+			// transactional edit state first
```
lipingxue: Where is the code to make sure the state edit is transactional?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+			} else {
+				// failed to start SMB server, set to state Error
+				e.CompareAndPut(etcdPrefixState+volName,
+					string(volStateIntermediate),
+					string(volStateError))
+			}
+			return
+		}
+
+		if string(ev.Kv.Value) == etcdNoRef &&
```
lipingxue: This code snippet is very similar to the above code which handle "etcdSingleRef" to "etcdNonRef". Is it possible to create a  common util function to reuse the code?

client_plugin/drivers/shared/etcdops.go
lipingxue: I see. Inside function "CompareAndPut", you call the transaction interface. But you call function "CompareAndPut" 2 or 3 times here. How do we make sure it is transactional too?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -346,14 +520,14 @@ func (d *VolumeDriver) etcdList() ([]string, error) {
 }
 
 // writeVolMetadata - Update or Create volume metadata in KV store
```
lipingxue: Nit: "writeVolMetadata" -> "WriteVolMetadata" to be consistent with the function name.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -379,13 +553,13 @@ func (d *VolumeDriver) writeVolMetadata(entries []kvPair) error {
 }
 
 // readVolMetadata - Read volume metadata in KV store
```
lipingxue: Same above.

client_plugin/drivers/shared/etcdops.go
shivanshu21: Join etcd cluster is called by manager nodes who are not initializing the cluster. To join, they must start their own etcd, hence the comment.

client_plugin/drivers/shared/etcdops.go
shivanshu21: I am not sure we want the different compare and put steps to be in the same transaction. State of the volume has to be updated first which will then act as a lock for all other operations. Miao can throw some light maybe.

client_plugin/drivers/shared/etcdops.go
luomiao: Thanks for catching this... I changed the logic of return last time and forgot to update return error here!

client_plugin/drivers/shared/etcdops.go
luomiao: These values are mainly set according to other projects. Will need to do more testing and adjust values to be more accurate.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -145,7 +147,8 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	d.dockerd = cli
 
 	// initialize built-in etcd cluster
-	err = d.initEtcd()
+	d.etcd.driver = &d
```
shivanshu21: Why are we doing this? 
Putting etcdInfo obj in VolumeDriver
And also putting VolumeDriver in etcdInfo obj?

shivanshu21: I mean, we do have access to d.<functions> in etcdops.go right?

client_plugin/drivers/shared/etcdops.go
luomiao: We **cannot** have those CompareAndPut as a single transaction :)
As Shivanshu has mentioned above, please consider the _CompareAndPut to change the state from Ready to Intermediate_ as "lock", and consider the _CompareAndPut to change state from Intermediate to Ready_ as "unlock". Only when the "lock" is successful, we go ahead to do corresponding operations (that is, startSMBServer), and "unlock" later.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+				// this handler doesn't get the right to stop server
+				return
+			}
+
+			if e.driver.stopSMBServer(volName) {
+				// SMB server started, set Mounted state
+				if e.CompareAndPut(etcdPrefixState+volName,
+					string(volStateIntermediate),
+					string(volStateReady)) == false {
+					// Failed to set state Intermediate->Mounted,
```
shivanshu21: Comment fix: Intermediate -> ready. SMB server has stopped not started.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+					// set to state Error
+					e.CompareAndPut(etcdPrefixState+volName,
+						string(volStateIntermediate),
+						string(volStateError))
+				}
+			} else {
+				// failed to start SMB server, set to state Error
```
shivanshu21: Same. Failed to stop smb server.

client_plugin/drivers/shared/shared_driver.go
luomiao: No really :)
For example, from `etcdEventHandler` function (which is a method of `etcdInfo` struct), we need to call `startSMBServer` function (which is a method of `VolumeDriver` struct). So from `e *etcdInfo` we need to get the VolumeDriver associate with it, and the best way is to link back VolumeDriver inside etcdInfo.

client_plugin/drivers/shared/etcdops.go
luomiao: So it really means "start" ... "routine". will update the comment to avoid confusion :)

client_plugin/drivers/shared/etcdops.go
```diff
@@ -319,10 +493,10 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 }
 
 // etcdList function lists all the volume names associated with this KV store
```
shaominchen: Nit: update the comment

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,10 +77,32 @@ type kvPair struct {
...
+}
+
+type etcdInfo struct {
+	driver   *VolumeDriver
+	nodeID   string
+	nodeAddr string
+	client   *etcdClient.Client
+}
+
 // initEtcd start or join ETCD cluster depending on the role of the node
```
shaominchen: Nit: comments need to be updated

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,10 +77,32 @@ type kvPair struct {
...
+	ReadVolMetadata(keys []string) ([]kvPair, error)
+	// DeleteVolMetadata - Delete volume metadata in KV store
+	DeleteVolMetadata(name string) error
+	// CompareAndPut - Compare the value of key with oldVal, if equal, replace with newVal
+	CompareAndPut(key string, oldVal string, newVal string) bool
+	// ListVolumeName - List all the volumes in the KV store
+	ListVolumeName() ([]string, error)
+}
+
+type etcdInfo struct {
```
shaominchen: Can we have a better name for this interface? The name "etcdInfo" sounds like a data object.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,10 +77,32 @@ type kvPair struct {
 	value string
 }
 
+type kvStore interface {
```
shaominchen: Is it intended that the interface is not exported, but the functions in it are exported?

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +308,143 @@ func etcdService(cmd []string) {
...
+		}
+	}
+}
+
+func (e *etcdInfo) etcdEventHandler(ev *etcdClient.Event) {
+	log.WithFields(
+		log.Fields{"type": ev.Type},
+	).Infof("Watcher on global refcount returns event ")
+
+	if ev.Type == etcdClient.EventTypePut {
```
shaominchen: Do we plan to add more handlers for other events? If we are just handling PUT event here, we can return fast to avoid too deep nested if conditions.

client_plugin/drivers/shared/etcdops.go
luomiao: Export the interface will be in another refactor PR :)

client_plugin/drivers/shared/etcdops.go
luomiao: Yeah we will need to check other events in the future...

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +305,122 @@ func etcdService(cmd []string) {
...
+			ev.PrevKv != nil &&
+			string(ev.PrevKv.Value) == etcdSingleRef {
+			nested(string(ev.Kv.Key), volStateMounted, volStateReady, e.driver.stopSMBServer)
+		}
+	}
+
+	return
+}
+
+func (e *etcdKVS) CompareAndPut(key string, oldVal string, newVal string) bool {
```
govint: Add a comment.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,60 +77,75 @@ type kvPair struct {
...
 			log.Fields{"error": err},
 		).Error("Failed to get Info from docker client ")
-		return err
+		return nil
 	}
 
-	// worker just returns
+	// get the swarmID and IP address of current node
 	nodeID := info.Swarm.NodeID
+	addr := info.Swarm.NodeAddr
```
luomiao: Swarm is a pre-request for this plugin...
But we need a check and error out here too. will update.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,60 +77,75 @@ type kvPair struct {
...
+		driver:   driver,
+		nodeID:   nodeID,
+		nodeAddr: addr,
+	}
+
+	// worker just returns
 	if info.Swarm.ControlAvailable == false {
 		log.WithFields(
 			log.Fields{"nodeID": nodeID},
 		).Info("Swarm node role: worker. Action: return from InitEtcd ")
```
luomiao: Thanks for catching the function name. This one in fact is not a error message but a log info for debugging use.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,60 +77,75 @@ type kvPair struct {
...
+		return e
 	}
 
 	// if manager, first find out who's leader, then proceed to join ETCD cluster
 	nodes, err := docker.NodeList(ctx, types.NodeListOptions{})
 	if err != nil {
 		log.WithFields(log.Fields{"nodeID": nodeID,
 			"error": err}).Error("Failed to get NodeList from swarm manager")
-		return err
+		return nil
```
luomiao: For the first question:
It's OK because only swarm **leader** will call the startEtcdCluster, and non-leader managers won't call startEtcdCluster.

For the second question:
No. Because swarm leader doesn't need to call NodeList. The workflow is: if this node is a leader, startEtcdCluster and leave; if this node is a non-leader manager, do NodeList to find who's the leader, then joinEtcdCluster and leave.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -72,60 +77,75 @@ type kvPair struct {
...
 	}
 	for _, n := range nodes {
```
luomiao: Because when the code comes to here, it's not a leader :)
The leader already left this function after startEtcdCluster.
Don't forget there is a return after that!
This loop is for non-leader managers to find the leader...

client_plugin/drivers/shared/etcdops.go
```diff
@@ -134,17 +154,24 @@ func (d *VolumeDriver) initEtcd() error {
 					"manager ID": nodeID},
 			).Info("Swarm node role: manager. Action: find leader ")
```
luomiao: Yes it's accurate.

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +305,122 @@ func etcdService(cmd []string) {
...
+	).Infof("Watcher on global refcount returns event ")
+
+	nested := func(key string, fromState volStatus, toState volStatus, fn func(string) bool) {
+		// watcher observes global refcount critical change
+		// transactional edit state first
+		volName := strings.TrimPrefix(key, etcdPrefixGRef)
+		succeeded := e.CompareAndPut(etcdPrefixState+volName,
+			string(fromState), string(volStateIntermediate))
+		if !succeeded {
+			// this handler doesn't get the right to start server
```
luomiao: The preceding line is a return :)

client_plugin/drivers/shared/etcdops.go
```diff
@@ -269,8 +305,122 @@ func etcdService(cmd []string) {
...
+		if fn(volName) {
+			// server start/stop succeed, set desired state
+			if e.CompareAndPut(etcdPrefixState+volName,
+				string(volStateIntermediate),
+				string(toState)) == false {
+				// Failed to set state to desired state
+				// set to state Error
+				e.CompareAndPut(etcdPrefixState+volName,
+					string(volStateIntermediate),
+					string(volStateError))
```
luomiao: Yes, users can still delete the volume with "--force" flag. And we already have the log message inside CompareAndPut function.

PR #1670, opened 2017-08-01
Added check to skip vm_list when creating the default tenant

### LINKED ISSUE #1644: Admin CLI should not allow VM list  when re-creating _DEFAULT vmgroup
The admin cli allows the user to remove and recreate the default vmgroup like other (non-default) vmgroups. When re-creating the default vmgroup the admin CLI allows specifying a VM list and unlike how the default vmgroup is defined adds the VM list as belonging to the _DEFAULT vmgroup - which is a bug.

The issue here is once added to the default vmgroup, those VMs can't be removed thereafter from the default group and nor can the default group be removed.

Fix would be to disallow (error out) the vmgroup create if the user is creating the _DEFAULT group and a VM list is specified.

govint:
Fixes #1644. Change is to ignore a vmlist if given when creating the default tenant.

 /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
------------------------------------  --------  -------------------------  -----------------  -------

vmgroup rm --name _DEFAULT

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
Uuid  Name  Description  Default_datastore  VM_list
----  ----  -----------  -----------------  -------

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name _DEFAULT --default-datastore _VM_DS --vm-list master-VM0.0

 /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
Uuid                                  Name      Description  Default_datastore  VM_list
------------------------------------  --------  -----------  -----------------  -------
11111111-1111-1111-1111-111111111111  _DEFAULT               _VM_DS


esx_service/utils/auth_api.py
```diff
@@ -457,7 +457,7 @@ def _tenant_create(name, default_datastore, description="", vm_list=None, privil
 
     # VM list can be empty during tenant create. Validate only if it exists
     vms = None
-    if vm_list:
+    if vm_list and name != auth_data_const.DEFAULT_TENANT:
```
lipingxue: I think we should error out here when user give non-empty vm_list while creating DEFAULT_TENANT. We should not ignore the vm_list.

I think similar check should also apply with "tenant vm add", "tenant vm rm" and "tenant vm replace".

govint: Adding a VM to the default group is already handled,
/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup vm add --name _DEFAULT --vm-list master-VM0.0
This feature is not supported for vmgroup _DEFAULT.


I wasn't sure we wanted to error out and have the user redo the command, may be issue a warning and let the command continue.

govint: Fixed to return error if a vm list is given when creating the default group.

govint:
vmdkops_admin.py vmgroup rm --name _DEFAULT

vmdkops_admin.py vmgroup create --name _DEFAULT --default-datastore _VM_DS --vm-list manager-VM0.0
Invalid Argument(['manager-VM0.0'])

vmdkops_admin.py vmgroup ls
Uuid  Name  Description  Default_datastore  VM_list
----  ----  -----------  -----------------  -------

vmdkops_admin.py vmgroup create --name _DEFAULT --default-datastore _VM
_DS
vmgroup '_DEFAULT' is created.

PR #1672, opened 2017-08-01
Enables Docker plugin registration via plugin_server module

venilnoronha:
This PR moves the existing plugin registration logic from `build.bat` to the `plugin_server` module.

### DIFF
--- a/client_plugin/utils/plugin_server/plugin_server_windows.go
+++ b/client_plugin/utils/plugin_server/plugin_server_windows.go
@@ -18,6 +18,7 @@ package plugin_server
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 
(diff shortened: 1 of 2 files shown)


client_plugin/utils/plugin_server/plugin_server_windows.go
```diff
@@ -26,7 +27,34 @@ import (
...
+
+// pluginConfig is the plugin config.
+var pluginConfig = fmt.Sprintf(`{ "Name": "vsphere", "Addr": "npipe://%s" }`, npipeAddr)
+
+// writePluginConfig writes the plugin config to the Docker plugins directory.
+func writePluginConfig() {
+	if err := os.MkdirAll(pluginsPath, 0755); err != nil {
+		log.WithFields(log.Fields{"path": pluginsPath,
+			"err": err}).Fatal("Failed to create plugin config directory ")
+		panic("Failed to create plugin config directory")
```
shaominchen: Failing with panic doesn't sound a good idea. I'd suggest to return err and fail in a consistent way as we did at line 82.

client_plugin/utils/plugin_server/plugin_server_windows.go
```diff
@@ -26,7 +27,34 @@ import (
 	"github.com/docker/go-plugins-helpers/volume"
 )
 
-const npipeAddr = `\\.\pipe\vsphere-dvs` // Plugin's npipe address
+const (
+	// npipeAddr is the plugin's npipe address.
+	npipeAddr = "//./pipe/vsphere-dvs"
+
+	// pluginsPath is the Docker plugins directory path.
```
shaominchen: Nit: Docker plugins => Docker's plugin...

client_plugin/utils/plugin_server/plugin_server_windows.go
venilnoronha: Addressed in f6ce550.

client_plugin/utils/plugin_server/plugin_server_windows.go
venilnoronha: Addressed in f6ce550.

venilnoronha:
@pshahzeb
1. As described, it moves the existing plugin registration logic from `build.bat` (which is now meant for only building the plugin on a Windows VM) to the `plugin_server` module (which would ensure that the plugin registration happens correctly every time the daemon starts). @shaominchen and I had a brief discussion on this a while back, and I missed to create an issue for this.

2. I've tested it manually, as we're still in the midst of porting e2e tests.

PR #1686, opened 2017-08-02
Allow "--volume-totalsize" to be set for "_VM_DS" and "_ALL_DS".

lipingxue:
This PR includes:
1. change to make "--volume-totalsize" can be set for "_VM_DS" and "_ALL_DS".
2. go test for this change
3.  corresponding change in AdminCLI user guide

Testing Done:
test-e2e is successful on local setup.

### DIFF
--- a/docs/user-guide/admin-cli.md
+++ b/docs/user-guide/admin-cli.md
@@ -407,8 +407,8 @@ optional arguments:
 #### Add
 Grants datastore access to a vmgroup.
 Valid value for "datastore" includes the name of valid datastores in the ESX host , special value "_VM_DS" or "_ALL_DS".
-When DS is set to _VM_DS, access to vm_datastore where vm lives is allowed for vms in vmgroup.
-When DS is set to _ALL_DS, access to any other datastore which has not been granted access to explicitly is allowed for vms in vmgroup.
+When DS is set to _VM_DS, access to datastore where vm lives is allowed for vms in vmgroup.
+When DS is set to _ALL_DS, access to any datastore which has not been granted access to explicitly is allowed for vms in vmgroup.
 Sample:
 
 ```bash
@@ -454,7 +454,7 @@ Datastore   Allow_create  Max_volume_size  Total_size
 datastore1  True          Unset            Unset
 datastore2  True          500.00MB         1.00GB
 ```
-For _VM_DS and _ALL_DS special DS names, --volume-totalzie can also be set.
+For _VM_DS and _ALL_DS special DS names, "--volume-totalsize" can also be set.
 
 When "--volume-totalsize" is set for "_VM_DS", it means the total volume size on datastore where VM lives cannot exceed the value specified by "--volume-totalsize". For example, if "--volume-totalsize" is set to "1GB" for "_VM_DS" and VM lives in "datastore2", where the total volume size is already 800MB and user tries to create another volume with 500MB will not be allowed. However, if the VM is moved to "datastore3", where the total volume size is 500MB and user tries to create another volume with 500MB will be allowed.
 
@@ -469,7 +469,7 @@ datastore1    True          Unset            Unset
 _VM_DS        True          500.00MB         1.00GB
 ```
 
-When "--volume-totalsize" is set for "_ALL_DS", it means the total volume size on any datastore which has not been given access to explicitly cannot exceed the value specified by "--volume-totalsize". In the following example, "vmgroup1" has been given access to "datastore1" and "_ALL_DS". The "Total_size" for "datastore1" is "Unset", which means no limit. The "Total_size" for "_ALL_DS" is "1GB". So for "datastore1", there is no limit on the total volume size. However, for any other datastore, the total volume size on that datastore cannot exceed 1GB.
+When "--volume-totalsize" is set for "_ALL_DS", it means the total volume size on any datastore which has not been given access to explicitly cannot exceed the value specified by "--volume-totalsize". In the following example, "vmgroup1" has been given access to "datastore1" and "_ALL_DS". The "Total_size" for "datastore1" is "Unset", which means no limit. The "Total_size" for "_ALL_DS" is "1GB". So for "datastore1", there is no limit on the total volume size. However, for any datastore other than "datastore1", the total volume size on that datastore cannot exceed 1GB.
 
 ```
 [root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup access add --name=vmgroup1 --datastore=_ALL_DS  --volume-maxsize=500MB --volume-totalsize=1GB --allow-create
@@ -601,7 +601,7 @@ datastore1  True          1000.00MB        2.00GB
 _ALL_DS     True          500.00MB         Unset
 ```
 
-"-volume-totalsize" can also be set to the value other than unlimit when add privilege for special value "_VM_DS" and "_ALL_DS".
+"--volume-totalsize" can also be set to the value other than unlimit when add privilege for special value "_VM_DS" and "_ALL_DS".
 ```
 [root@localhost:~] /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup access ls --name=vmgroup1
 Datastore   Allow_create  Max_volume_size  Total_size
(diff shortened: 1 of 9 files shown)


docs/user-guide/admin-cli.md
```diff
@@ -590,24 +598,33 @@ vmgroup access set succeeded
 Datastore   Allow_create  Max_volume_size  Total_size
 ----------  ------------  ---------------  ----------
 datastore1  True          1000.00MB        2.00GB
-_ALL_DS    True          500.00MB         Unset
+_ALL_DS     True          500.00MB         Unset
 ```
 
-"-volume-totalsize" cannot be set to the value other than unlimit when add privilege for special value "_VM_DS" and "_ALL_DS".
+"-volume-totalsize" can also be set to the value other than unlimit when add privilege for special value "_VM_DS" and "_ALL_DS".
```
govint: --volume-t....

docs/user-guide/admin-cli.md
```diff
@@ -454,25 +454,33 @@ Datastore   Allow_create  Max_volume_size  Total_size
 datastore1  True          Unset            Unset
 datastore2  True          500.00MB         1.00GB
 ```
-For _VM_DS and _ALL_DS special DS names, --volume-totalzie has to be "Unset".
+For _VM_DS and _ALL_DS special DS names, --volume-totalzie can also be set.
+
+When "--volume-totalsize" is set for "_VM_DS", it means the total volume size on datastore where VM lives cannot exceed the value specified by "--volume-totalsize". For example, if "--volume-totalsize" is set to "1GB" for "_VM_DS" and VM lives in "datastore2", where the total volume size is already 800MB and user tries to create another volume with 500MB will not be allowed. However, if the VM is moved to "datastore3", where the total volume size is 500MB and user tries to create another volume with 500MB will be allowed.
```
govint: If volume totalsize isn't set then what is the behavior for _VM_DS and for _ALL_DS. Is it like unlimited?

docs/user-guide/admin-cli.md
```diff
@@ -454,25 +454,33 @@ Datastore   Allow_create  Max_volume_size  Total_size
...
-----------  ------------  ---------------  ----------
-datastore1  True          Unset            Unset
-_ALL_DS    True          500.00MB         Unset
+Datastore     Allow_create  Max_volume_size  Total_size
+------------  ------------  ---------------  ----------
+datastore1    True          Unset            Unset
+_VM_DS        True          500.00MB         1.00GB
+```
+
+When "--volume-totalsize" is set for "_ALL_DS", it means the total volume size on any datastore which has not been given access to explicitly cannot exceed the value specified by "--volume-totalsize". In the following example, "vmgroup1" has been given access to "datastore1" and "_ALL_DS". The "Total_size" for "datastore1" is "Unset", which means no limit. The "Total_size" for "_ALL_DS" is "1GB". So for "datastore1", there is no limit on the total volume size. However, for any other datastore, the total volume size on that datastore cannot exceed 1GB.
```
govint: Could clarify explicitly whether the volume-totalsize applies across all those datastores that aren't part of the vmgroup or applies per each of those datastores. I'm assuming the former, the latter implies that user can create volume-totalzie worth VMDKs on each datastore thats explicitly not in the vmgroup.

lipingxue: I think it applies per each of those datastores. Remember, the access control is per <vmgroup, datastore> pair. We don't have something like total-storage limit for a vmgroup. Access control is a FALLBACK access control,  which means if any datastore which does not have been given an access control will match this FALLBACK access control.  For example, if I have three datastores, "datastore1", "datastore2" and "datastore3", and I have two access privileges, one is to "datastore1" and the other is "_ALL_DS". If the "volume-totalsize" for "_ALL_DS" is set to 1GB, then for any datastore other than "datastore1", I want the total storage usage on THAT datastore can not exceed 1GB.

docs/user-guide/admin-cli.md
```diff
@@ -408,7 +408,7 @@ optional arguments:
 Grants datastore access to a vmgroup.
 Valid value for "datastore" includes the name of valid datastores in the ESX host , special value "_VM_DS" or "_ALL_DS".
 When DS is set to _VM_DS, access to vm_datastore where vm lives is allowed for vms in vmgroup.
```
shaominchen: Not the current change, but I think "vm_datastore" should be simply "datastore".

docs/user-guide/admin-cli.md
```diff
@@ -408,7 +408,7 @@ optional arguments:
...
-When DS is set to _ALL_DS, access to all DS is allowed for vms in vmgroup.
+When DS is set to _ALL_DS, access to any other datastore which has not been granted access to explicitly is allowed for vms in vmgroup.
```
shaominchen: Why "any other datastore"? Do you mean the datastore where vm lives is NOT allowed? That seems counter-intuitive.

docs/user-guide/admin-cli.md
```diff
@@ -454,25 +454,33 @@ Datastore   Allow_create  Max_volume_size  Total_size
 datastore1  True          Unset            Unset
 datastore2  True          500.00MB         1.00GB
 ```
-For _VM_DS and _ALL_DS special DS names, --volume-totalzie has to be "Unset".
+For _VM_DS and _ALL_DS special DS names, --volume-totalzie can also be set.
```
shaominchen: Typo: totalzie => totalsize

docs/user-guide/admin-cli.md
shaominchen: To add on @govint's question, is the default behavior of "Unset" consistent between _VM_DS, _ALL_DS and a specific datastore? If not, we should clarify this.

esx_service/utils/auth.py
```diff
@@ -424,8 +398,17 @@ def authorize(vm_uuid, datastore_url, cmd, opts):
         - tenant_name: If the VM belongs to a tenant, return tenant_name, otherwise, return
         None
     """
-    logging.debug("Authorize: cmd=%s opts=`%s' vm_uuid=%s, datastore_url=%s", cmd, opts, vm_uuid, datastore_url)
-    # The possible value of "datastore_url" can be url of real datastore or "_VM_DS://"
+    # "datastore_url" is the url of datastore which the volume to be created on
+    # The possible value of "datastore_url" can be url of a real datastore or "_VM_DS"
+    # "privilege_ds_url" is the url of datastore which we need to get the privilege of
```
shaominchen: I'm a bit confused here. Why datastore_url and privilege_ds_url can be different? In other words, why do we need to check the privilege of a different datastore other than the one on which the volume will be created?

esx_service/vmdk_ops.py
```diff
@@ -823,25 +840,57 @@ def get_datastore_name(datastore_url):
 
     return datastore_name
 
-def authorize_check(vm_uuid, datastore_url, cmd, opts, use_default_ds, datastore, vm_datastore):
+def authorize_check(vm_uuid, datastore_url, cmd, opts, use_default_ds, datastore, vm_datastore_url, vm_datastore):
```
shaominchen: To do an authorize check, we need to pass 5 datastore related parameters. This is an indication that something is wrong either in our implementation or in our design. I don't have good suggestions right now as the existing code are just too convoluted. It seems not straightforward to improve the quality without a significant refactoring (or design change).

docs/user-guide/admin-cli.md
lipingxue: If "--volume-totalsize" is not set, then the default value is "Unset", which means no limit. 
The behavior of "Unset" is consistent on _VM_DS, _ALL_DS and specific datastore.

docs/user-guide/admin-cli.md
lipingxue: Done.

docs/user-guide/admin-cli.md
lipingxue: Done.

docs/user-guide/admin-cli.md
lipingxue: Fixed.

docs/user-guide/admin-cli.md
lipingxue: Done.

esx_service/utils/auth.py
lipingxue: We have discussed offline.

lipingxue:
@govint @shaominchen I have addressed your comments, and please review it.

shuklanirdesh82:
@lipingxue There is a CI failure.

```
Error response from daemon: create volume_20170804152930: VolumeDriver.Create: Server returned an error: UnboundLocalError("local variable 'vmdk_path' referenced before assignment",)
```

As per our code contributing guideline, it is always good to have "Testing Done" summary in the PR description.

/CC @tusharnt

lipingxue:
@govint I have addressed comments from You and Sam. Please take a look. Thannks!

PR #1716, opened 2017-08-08
Shared plugin: fix to make sure error out when swarm is not active.

luomiao:
When the swarm status of a node is not active, NewKvStore returns an empty result.
This should be checked before assigned to d.kvStore, which won't be empty.
This minor fix makes sure when the returned result is empty, NewVolumeDriver will error out and retry.

Local test passed.

### DIFF
--- a/client_plugin/drivers/shared/shared_driver.go
+++ b/client_plugin/drivers/shared/shared_driver.go
@@ -124,11 +124,12 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 
 	// initialize built-in etcd cluster
-	d.kvStore = etcdops.NewKvStore(d.dockerOps)
-	if d.kvStore == nil {
+	etcdKVS := etcdops.NewKvStore(d.dockerOps)
+	if etcdKVS == nil {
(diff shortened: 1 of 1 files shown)


client_plugin/drivers/shared/shared_driver.go
```diff
@@ -124,11 +124,12 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 
 	// initialize built-in etcd cluster
-	d.kvStore = etcdops.NewKvStore(d.dockerOps)
-	if d.kvStore == nil {
+	etcdKVS := etcdops.NewKvStore(d.dockerOps)
```
lipingxue: Why you need a new variable here, I think in the original code, when d.kvStore is nil, you also error out. Do I miss anything here?

luomiao: I also had the same understanding...
But with the original code, d.kvStore won't be nil even if etcdops.NewKvStore returns nil. I think the reason is because d.kvStore is an interface instead of a struct.
The fix is exactly to handle this issue.

Issue #1625, opened 2017-07-22
E2E test util is failing when space is embedded in datastore name


shuklanirdesh82:
Steps to repro: create a testbed setup which is having space in the name e.g. `datastore (2)` > run E2E tests

CI Run: https://ci.vmware.run/vmware/docker-volume-vsphere/1281

```
2017/07/22 03:20:45 Attached VM name from Docker CLI is [bash: -c: line 0: syntax error near unexpected token `('
bash: -c: line 0: `docker volume inspect  --format '{{index .Status "attached to VM"}}' BasicTestSuite.TestVolumeLifecycle_volume_884362@datastore1 (2)']
```

/CC @govint @tusharnt

govint:
Suggest please revert the CI datastorr name to have no spaces so that everyone doesn't need to be blocked on this and instead let one person try out e2e tests locally with a datastorr with spaces and fix all the places where the code is broken. There are multiple places where fixes need to be made and no point bringing CI down to have this fixed

shaominchen:
Agree with @govint.

ashahi1:
Fixed as part of PR #1717

PR #1717, opened 2017-08-08
Fix e2e tests to run if datastore name has space

ashahi1:
Fixes issue #1625

Have modified E2E tests such that these tests can work even if datastore name has space.

Ran all E2E tests on datastore have spaces in their names.


tests/e2e/restart_test.go
```diff
@@ -330,7 +330,8 @@ func (s *RestartTestData) TestDuplicateVolumeName(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Cleanup volume on the second datastore
-	out, err = dockercli.DeleteVolume(s.config.DockerHosts[1], s.volumeName+"@"+ds2)
+	volName := "'" + s.volumeName + "@" + ds2 + "'"
```
govint: Suggest making a util function CreateFQVN() that handles adding the _'_ to create a full volume name with its datastore. So this code isn't repeated. The function returns a string that can be used in any command string.

tests/e2e/vmgroups_test.go
```diff
@@ -66,7 +66,7 @@ func (vg *VmGroupTest) SetUpSuite(c *C) {
 		log.Printf(out)
 	}
 	// Create the test VM group1
-	cmd = adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore " + vg.config.Datastores[0]
+	cmd = adminconst.CreateVMgroup + vgTestVMgroup1 + " --default-datastore '" + vg.config.Datastores[0] + "'"
```
govint: Will it work if GetTestConfig returns the datastore names with the _'_ added?

govint: This works and I'm assuming that any command line (need to check) done this way for the datastore name _should_ work.

_docker volume inspect snapvol2@'sharedVmfs-0'_
[
    {
        "Driver": "vsphere",
        "Labels": null,
        "Mountpoint": "/mnt/vmdk/snapvol2@sharedVmfs-0",
        "Name": "snapvol2@sharedVmfs-0",
        "Options": {},
        "Scope": "global",
        "Status": {
            "access": "read-write",
            "capacity": {
                "allocated": "15MB",
                "size": "100MB"
            },
            "clone-from": "None",
            "created": "Sun Aug  6 13:01:34 2017",
            "created by VM": "worker1-VM1.0",
            "datastore": "sharedVmfs-0",
            "diskformat": "thin",
            "fstype": "ext4",
            "status": "detached"
        }
    }
]

govint: So, could try out just changing GetTestConfig() to return the datastore names to prefix-suffix with _'_

tests/e2e/restart_test.go
```diff
@@ -73,7 +74,8 @@ func (s *RestartTestData) SetUpSuite(c *C) {
 	c.Assert(err, IsNil, Commentf(out))
 
 	// Create a volume on the second datastore
-	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], s.volumeName+"@"+ds2)
+	ds2 = strings.Replace(ds2, "'", "", -1)
+	out, err = dockercli.CreateVolume(s.config.DockerHosts[1], inputparams.CreateFQVN(s.volumeName, ds2))
```
govint: Is it failing if the command line is formatted with the DS name in "'"

govint: API isn't being used anywhere so the command lines should work with the quoted DS names. Isn't that working?

tests/e2e/restart_test.go
```diff
@@ -89,7 +91,7 @@ func (s *RestartTestData) TearDownSuite(c *C) {
 		c.Assert(err, IsNil, Commentf(out))
 
 		// Remove access for second datastore from the default vmgroup
-		out, err = adminutil.RemoveDatastoreFromVmgroup(s.config.EsxHost, adminconst.DefaultVMgroup, s.config.Datastores[1])
+		out, err = adminutil.RemoveDatastoreFromVmgroup(s.config.EsxHost, adminconst.DefaultVMgroup, "'"+ds2+"'")
```
govint: Is this adding back the quotes? Do we need to remove it earlier?

ashahi1: Yes, we remove it in SetUpSuite as we do strings.Compare() there and since config.Datastores has names with quotes whereas status["datastore"] has names without quotes, comparison fails.
```
if strings.Compare(status["datastore"], strings.Replace(s.config.Datastores[0], "'", "", -1)) == 0 {
		ds2 = s.config.Datastores[1]
	}
```

tests/e2e/restart_test.go
ashahi1: config.Datastores already has quotes attached to ds names so adding two back-to-back single quotes fails the command
```
root@sc-rdops-vm02-dhcp-52-237:~# docker volume create -d vsphere --name ''RefcountTest_volume_634502@ shared   Vmfs   - (10)''
-bash: syntax error near unexpected token `('
```
So either we can call CreateFQVN() to add quotes or we can go back to the way it was - s.volumeName+"@"+ds2

ashahi1:
@govint I have pushed a fresh commit and in that commit, I have removed the changes that were made to the tests. I have modified the test utils accordingly.

govint:
Looks good, lets retain the FQVN() api, its better than creating names with quotes in every test.

tests/e2e/vmgroups_test.go
```diff
@@ -113,7 +113,7 @@ func (vg *VmGroupTest) TearDownSuite(c *C) {
 
 	cmd := adminconst.ListVMgroups
 	out, err = ssh.InvokeCommand(vg.config.EsxHost, cmd)
-	log.Printf(out)
+	// log.Printf(out)
```
pshahzeb: Why to silence these prints?

tests/utils/dockercli/containers.go
```diff
@@ -122,7 +124,10 @@ func GetVolumeStatus(hostName, volumeName string) (map[string]string, error) {
 	}
 
 	status := make(map[string]string)
-	val := strings.Fields(out)
+	val := strings.SplitAfter(out, volmPropertyDelimiter)
```
pshahzeb: will this not work https://golang.org/pkg/strings/#Split
with sep as volmPropertyDelimiter

tests/utils/esx/govc.go
```diff
@@ -85,12 +85,17 @@ func ShutDownVM(vmName string) string {
...
 }
 
-// GetDatastoreList returns a list of datastore names available
+// GetDatastoreList returns a list of datastore names available.
 func GetDatastoreList() []string {
 	// log.Printf("Finding Datastores available on ESX")
 	cmd := esx.DatastoreInfo + esxcliJSON + esx.DatastoreList
 	out, _ := ssh.InvokeCommandLocally(cmd)
-	return strings.Fields(out)
+	val := strings.SplitAfter(out, "\n")
```
pshahzeb: Similar comment to use Split instead of SplitAfter. With using split, you can also get rid of line 95

tests/utils/verification/volumeproperties.go
```diff
@@ -145,7 +145,6 @@ func getVolumeStatusHost(name, hostName string) string {
 // vsphere shared driver
 func GetSharedVolumeStatusHost(name, hostName string) string {
 	cmd := dockercli.InspectVolume + " --format \"{{index .Status \\\"Volume Status\\\"}}\" " + name
-	log.Printf("GetSharedVolumeStatusHost: cmd[]", cmd)
```
pshahzeb: why remove this print?

tests/e2e/vmgroups_test.go
ashahi1: This is how it looks like on CI console. I just feel it unnecessarily uses up the log space.
```
2017/08/11 02:12:54 stty: standard input: Inappropriate ioctl for device
Uuid            Name           Description      Default_datas..  VM_list          
--------------  -------------  ---------------  ---------------  ---------------  
11111111-111..  _DEFAULT       This is a def..  _VM_DS                            
e3290eef-591..  vmgroup_test1                   datastore1       photon.vmfs,U..  
ca6e7108-704..  vmgroup_test2                   vsandatastore
```
And we can always enable it for debug purposes.

govint: Ok, I'd added this earlier but with all the tests out there this may be is not needed any more except for debug.

PR #1727, opened 2017-08-09
vShared plugin: Add Mount() support for worker

luomiao:
* Add new methods for KvStore interface:
  AtomicIncr, AtomicDecr, and BlockingWaitAndGet
* Implement Mount() function

With this PR, user can execute "docker run -v vol:/path container" with a volume created with vShared plugin.
This is the implementation on worker side.
Corresponding server side implementation is required for a complete docker run functionality.

### DIFF
--- a/client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
+++ b/client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -386,7 +387,7 @@ func (e *EtcdKVS) CompareAndPut(key string, oldVal string, newVal string) bool {
 	// Create a client to talk to etcd
 	etcdAPI := e.createEtcdClient()
 	if etcdAPI == nil {
-                log.Warningf(etcdClientCreateError)
+		log.Warningf(etcdClientCreateError)
 		return false
 	}
 	defer etcdAPI.Close()
@@ -534,13 +535,14 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 func (e *EtcdKVS) List(prefix string) ([]string, error) {
 	var keys []string
 
-	etcd := e.createEtcdClient()
-	if etcd == nil {
-		return nil, fmt.Errorf(etcdClientCreateError)
+	etcdAPI := e.createEtcdClient()
+	if etcdAPI == nil {
+		return keys, fmt.Errorf(etcdClientCreateError)
 	}
+	defer etcdAPI.Close()
 
 	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
-	resp, err := etcd.Get(ctx, prefix, etcdClient.WithPrefix(),
+	resp, err := etcdAPI.Get(ctx, prefix, etcdClient.WithPrefix(),
 		etcdClient.WithSort(etcdClient.SortByKey, etcdClient.SortDescend))
 	cancel()
 	if err != nil {
@@ -678,4 +680,133 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
 		return errors.New(msg)
 	}
 	return nil
+}
+
+// AtomicIncr - Increase a key value by 1
+func (e *EtcdKVS) AtomicIncr(key string) error {
+	// Create a client to talk to etcd
+	etcdAPI := e.createEtcdClient()
+	if etcdAPI == nil {
+		return fmt.Errorf(etcdClientCreateError)
+	}
+	defer etcdAPI.Close()
+
+	ticker := time.NewTicker(checkSleepDuration)
+	defer ticker.Stop()
+	timer := time.NewTimer(requestTimeout)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+			resp, err := etcdAPI.Get(ctx, key)
+			cancel()
+			if err != nil {
+				log.WithFields(
+					log.Fields{"key": key,
+						"error": err},
+				).Error("Failed to Get key-value from ETCD Client ")
+				return err
+			}
+
+			oldVal := string(resp.Kvs[0].Value)
+			num, _ := strconv.Atoi(oldVal)
+			num++
+			newVal := strconv.Itoa(num)
+			if e.CompareAndPut(key, oldVal, newVal) {
+				return nil
+			}
+		case <-timer.C:
+			return fmt.Errorf("Timeout reached; AtomicIncr is not complete")
+		}
+	}
+}
+
+// AtomicDecr - Decrease a key value by 1
+func (e *EtcdKVS) AtomicDecr(key string) error {
+	// Create a client to talk to etcd
+	etcdAPI := e.createEtcdClient()
+	if etcdAPI == nil {
+		return fmt.Errorf(etcdClientCreateError)
+	}
+	defer etcdAPI.Close()
+
+	ticker := time.NewTicker(checkSleepDuration)
+	defer ticker.Stop()
+	timer := time.NewTimer(requestTimeout)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+			resp, err := etcdAPI.Get(ctx, key)
+			cancel()
+			if err != nil {
+				log.WithFields(
+					log.Fields{"key": key,
+						"error": err},
+				).Error("Failed to Get key-value from ETCD Client ")
+				return err
+			}
+
+			oldVal := string(resp.Kvs[0].Value)
+			num, _ := strconv.Atoi(oldVal)
+			if num == 0 {
+				return fmt.Errorf("Cannot decrease a value equal to 0")
+			}
+			num--
+			newVal := strconv.Itoa(num)
+			if e.CompareAndPut(key, oldVal, newVal) {
+				return nil
+			}
+		case <-timer.C:
+			return fmt.Errorf("Timeout reached; AtomicDecr is not complete")
+		}
+	}
+}
+
+// BlockingWaitAndGet - Blocking wait until a key value becomes equal to a specific value
+// then read the value of another key
+func (e *EtcdKVS) BlockingWaitAndGet(key string, value string, newKey string) (string, error) {
+	// Create a client to talk to etcd
+	etcdAPI := e.createEtcdClient()
+	if etcdAPI == nil {
+		return "", fmt.Errorf(etcdClientCreateError)
+	}
+	defer etcdAPI.Close()
+
+	ticker := time.NewTicker(checkSleepDuration)
+	defer ticker.Stop()
+	timer := time.NewTimer(2 * requestTimeout)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			txresp, err := e.client.Txn(context.TODO()).If(
+				etcdClient.Compare(etcdClient.Value(key), "=", value),
+			).Then(
+				etcdClient.OpGet(newKey),
+			).Commit()
+
+			if err != nil {
+				log.WithFields(
+					log.Fields{"key": key,
+						"value":   value,
+						"new key": newKey,
+						"error":   err},
+				).Error("Failed to compare and get from ETCD client")
+				return "", err
+			}
+
+			if txresp.Succeeded {
+				resp := txresp.Responses[0].GetResponseRange()
+				return string(resp.Kvs[0].Value), nil
+			}
+		case <-timer.C:
+			return "", fmt.Errorf("Timeout reached; BlockingWait is not complete")
+		}
+	}
 }
--- a/client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
+++ b/client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
@@ -535,14 +535,14 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 func (e *EtcdKVS) List(prefix string) ([]string, error) {
 	var keys []string
 
-	etcdAPI := e.createEtcdClient()
-	if etcdAPI == nil {
+	client := e.createEtcdClient()
+	if client == nil {
 		return keys, fmt.Errorf(etcdClientCreateError)
 	}
-	defer etcdAPI.Close()
+	defer client.Close()
 
 	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
-	resp, err := etcdAPI.Get(ctx, prefix, etcdClient.WithPrefix(),
+	resp, err := client.Get(ctx, prefix, etcdClient.WithPrefix(),
 		etcdClient.WithSort(etcdClient.SortByKey, etcdClient.SortDescend))
 	cancel()
 	if err != nil {
@@ -568,11 +568,11 @@ func (e *EtcdKVS) WriteMetaData(entries []kvstore.KvPair) error {
 	var err error
 
 	// Create a client to talk to etcd
-	etcdAPI := e.createEtcdClient()
-	if etcdAPI == nil {
+	client := e.createEtcdClient()
+	if client == nil {
 		return errors.New(etcdClientCreateError)
 	}
-	defer etcdAPI.Close()
+	defer client.Close()
 
 	// ops contain multiple operations that will be done to etcd
 	// in a single revision
@@ -583,9 +583,9 @@ func (e *EtcdKVS) WriteMetaData(entries []kvstore.KvPair) error {
 	// Lets write the metadata in a single transaction
 	// Use a transaction if more than one entries are to be written
 	if len(entries) > 1 {
-		_, err = etcdAPI.Txn(context.TODO()).Then(ops...).Commit()
+		_, err = client.Txn(context.TODO()).Then(ops...).Commit()
 	} else {
-		_, err = etcdAPI.Do(context.TODO(), ops[0])
+		_, err = client.Do(context.TODO(), ops[0])
 	}
 
 	if err != nil {
@@ -603,11 +603,11 @@ func (e *EtcdKVS) ReadMetaData(keys []string) ([]kvstore.KvPair, error) {
 	var missedCount int
 
 	// Create a client to talk to etcd
-	etcdAPI := e.createEtcdClient()
-	if etcdAPI == nil {
+	client := e.createEtcdClient()
+	if client == nil {
 		return entries, errors.New(etcdClientCreateError)
 	}
-	defer etcdAPI.Close()
+	defer client.Close()
 
 	// Lets build the request which will be executed
 	// in a single transaction
@@ -617,7 +617,7 @@ func (e *EtcdKVS) ReadMetaData(keys []string) ([]kvstore.KvPair, error) {
 	}
 
 	// Read all requested keys in one transaction
-	getresp, err := etcdAPI.Txn(context.TODO()).Then(ops...).Commit()
+	getresp, err := client.Txn(context.TODO()).Then(ops...).Commit()
 	if err != nil {
 		log.Warningf("Transactional metadata read failed: %v", err)
 		return entries, err
@@ -658,11 +658,11 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
 	var err error
 
 	// Create a client to talk to etcd
-	etcdAPI := e.createEtcdClient()
-	if etcdAPI == nil {
+	client := e.createEtcdClient()
+	if client == nil {
 		return errors.New(etcdClientCreateError)
 	}
-	defer etcdAPI.Close()
+	defer client.Close()
 
 	// ops hold multiple operations that will be done to etcd
 	// in a single revision. Add all keys for this volname.
@@ -673,7 +673,7 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
 	}
 
 	// Delete the metadata in a single transaction
-	_, err = etcdAPI.Txn(context.TODO()).Then(ops...).Commit()
+	_, err = client.Txn(context.TODO()).Then(ops...).Commit()
 	if err != nil {
 		msg = fmt.Sprintf("Failed to delete metadata for volume %s. Reason: %v", name, err)
 		log.Warningf(msg)
@@ -685,11 +685,11 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
 // AtomicIncr - Increase a key value by 1
 func (e *EtcdKVS) AtomicIncr(key string) error {
 	// Create a client to talk to etcd
-	etcdAPI := e.createEtcdClient()
-	if etcdAPI == nil {
+	client := e.createEtcdClient()
+	if client == nil {
 		return fmt.Errorf(etcdClientCreateError)
 	}
-	defer etcdAPI.Close()
+	defer client.Close()
 
 	ticker := time.NewTicker(checkSleepDuration)
 	defer ticker.Stop()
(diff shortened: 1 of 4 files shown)


client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -657,3 +668,132 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
...
+					log.Fields{"key": key,
+						"value":   value,
+						"new key": newKey,
+						"error":   err},
+				).Error("Failed to compare and get from ETCD client")
+				return "", err
+			}
+
+			if txresp.Succeeded {
+				resp := txresp.Responses[0].GetResponseRange()
```
lipingxue: Does the etcd API guarantee txresp.Responses is not a nil here?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+
+	if plugin_utils.AlreadyMounted(r.Name, d.MountRoot) {
+		log.WithFields(log.Fields{"name": r.Name}).Info("Already mounted, skipping mount. ")
+		return volume.Response{Mountpoint: d.GetMountPoint(r.Name)}
+	}
+
+	mountpoint, err := d.MountVolume(r.Name, "", "", false, true)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": r.Name, "error": err},
```
lipingxue: Nit: better to split "name" and "error" into two lines.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+			log.Infof("Detaching %s - it is not used anymore", r.Name)
+			// TODO: umount here
+		}
+		return volume.Response{Err: err.Error()}
+	}
+
+	return volume.Response{Mountpoint: mountpoint}
+}
+
+// MountVolume - Request attach and them mounts the volume.
```
lipingxue: Typo: them->then

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isReadOnly bool, skipAttach bool) (string, error) {
+	mountpoint := d.GetMountPoint(name)
+	// First, make sure  that mountpoint exists.
+	err := fs.Mkdir(mountpoint)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "dir": mountpoint},
```
lipingxue: Nit: better split into two lines as above.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+
+// processMount -  process a mount request
+func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
+	d.MountIDtoName[r.ID] = r.Name
+
+	// If the volume is already mounted , just increase the refcount.
+	// Note: for new keys, GO maps return zero value, so no need for if_exists.
+	refcnt := d.IncrRefCount(r.Name) // save map traversal
+	log.Debugf("volume name=%s refcnt=%d", r.Name, refcnt)
+	if refcnt > 1 {
```
lipingxue: Here the ref count is local refcnt? What does refcnt>1 mean here? Does it mean more than one container has mounted the volume?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		log.WithFields(
+			log.Fields{"name": r.Name, "refcount": refcnt},
+		).Info("Already mounted, skipping mount. ")
+		return volume.Response{Mountpoint: d.GetMountPoint(r.Name)}
+	}
+
+	if plugin_utils.AlreadyMounted(r.Name, d.MountRoot) {
```
lipingxue: In which condition, refcnt is not >1 and the volume is already mounted?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		).Error("Failed to make directory for volume mount ")
+		return mountpoint, err
+	}
+
+	// Increase GRef
+	log.Infof("Before AtomicIncr")
+	err = d.kvStore.AtomicIncr(kvstore.VolPrefixGRef + name)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name, "error": err},
```
lipingxue: Nit: break into two lines.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+	// Blocking wait till Mounted
+	info, err := d.kvStore.BlockingWaitAndGet(kvstore.VolPrefixState+name,
+		string(kvstore.VolStateMounted), kvstore.VolPrefixInfo+name)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to blocking wait for Mounted state. Error: %v.", err)
+		err = d.kvStore.AtomicDecr(kvstore.VolPrefixGRef + name)
+		if err != nil {
+			msg += fmt.Sprintf(" Also failed to decrease global refcount. Error: %v.", err)
+		}
+		log.WithFields(log.Fields{"name": name, "error": msg}).Error("")
```
lipingxue: Nit, same above.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		return "", errors.New(msg)
+	}
+
+	// Start mounting
+	log.Infof("Volume state mounted, prepare to mounting locally")
+	var volRecord VolumeMetadata
+	// Unmarshal Info key
+	err = json.Unmarshal([]byte(info), &volRecord)
+	if err != nil {
+		log.WithFields(log.Fields{"name": name, "error": err}).Error("Failed to unmarshal info data ")
```
lipingxue: Nit: same above.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		"port=" + strconv.Itoa(volRecord.Port),
+		"vers=3.0",
+	}
+	mountArgs = append(mountArgs, "-o", strings.Join(options, ","))
+
+	_, addr, _, err := d.dockerOps.GetSwarmInfo()
+	if err != nil {
+		log.WithFields(
+			log.Fields{"volume name": volName,
+				"error": err,
```
lipingxue: Nit: Alignment.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+			}).Error("Failed to get IP address from docker swarm ")
+		return err
+	}
+	source := "//" + addr + "/" + fileShareName
+	mountArgs = append(mountArgs, source)
+	mountArgs = append(mountArgs, mountpoint)
+
+	log.WithFields(
+		log.Fields{"volume name": volName,
+			"arguments": mountArgs,
```
lipingxue: Nit: alignment.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,161 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		}).Info("Mounting volume with options ")
+	command := exec.Command("mount", mountArgs...)
+	output, err := command.CombinedOutput()
+	if err != nil {
+		log.WithFields(
+			log.Fields{"volume name": volName,
```
lipingxue: Alignment.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -521,13 +522,14 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 func (e *EtcdKVS) List(prefix string) ([]string, error) {
 	var keys []string
 
-	etcd := e.createEtcdClient()
-	if etcd == nil {
-		return nil, fmt.Errorf(etcdClientCreateError)
+	etcdAPI := e.createEtcdClient()
```
luomiao: It's because we already have the import package named as etcdClient:
`etcdClient "github.com/coreos/etcd/clientv3"`
I will rename the package to etcd and change etcdAPI to etcdClient.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -657,3 +668,132 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
...
+	for {
+		select {
+		case <-ticker.C:
+			ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+			resp, err := etcdAPI.Get(ctx, key)
+			cancel()
+			if err != nil {
+				log.WithFields(
+					log.Fields{"key": key,
+						"error": err},
```
luomiao: This is already automatically aligned by golang.
I cannot change it :)

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -657,3 +668,132 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
...
+	// Create a client to talk to etcd
+	etcdAPI := e.createEtcdClient()
+	if etcdAPI == nil {
+		return "", fmt.Errorf(etcdClientCreateError)
+	}
+	defer etcdAPI.Close()
+
+	ticker := time.NewTicker(checkSleepDuration)
+	defer ticker.Stop()
+	timer := time.NewTimer(2 * requestTimeout)
```
luomiao: This operation may need to wait remote SMB server starts. So it can take up to 5 to 10 seconds. So we doubled the requestTimeout, which is 5 seconds.
All the timeout currently are just estimation. We need more testing to get a more accurate timeout value.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
luomiao: It's not guaranteed. I will add a check.

client_plugin/drivers/shared/shared_driver.go
luomiao: Yes, this is the local refcnt.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
luomiao: The above change doesn't look correct.
clientv3 doesn't mean too much.
So I kept where etcdClient still as etcdClient, but changed etcdAPI to client since it's limited inside each function.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		return volume.Response{Mountpoint: d.GetMountPoint(r.Name)}
+	}
+
+	if plugin_utils.AlreadyMounted(r.Name, d.MountRoot) {
+		log.WithFields(log.Fields{"name": r.Name}).Info("Already mounted, skipping mount. ")
+		return volume.Response{Mountpoint: d.GetMountPoint(r.Name)}
+	}
+
+	mountpoint, err := d.MountVolume(r.Name, "", "", false, true)
+	if err != nil {
```
govint: If this if the first ever mount of this volume does an error also stop the samba server and detach the backing volume?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		log.WithFields(
+			log.Fields{"name": name,
+				"error": err},
+		).Error("Failed to increase global refcount when processMount ")
+		return "", err
+	}
+
+	// Blocking wait till Mounted
+	info, err := d.kvStore.BlockingWaitAndGet(kvstore.VolPrefixState+name,
```
govint: Could specify more clearly exactly what the blocking mount wait is waiting for and what error conditions are possible. What triggers the change to _Mounted_ state for the volume?

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -649,11 +660,152 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
...
+			}
+
+			if len(resp.Kvs) == 0 {
+				return fmt.Errorf("AtomicIncr: no key found for %s", key)
+			}
+
+			oldVal := string(resp.Kvs[0].Value)
+			num, _ := strconv.Atoi(oldVal)
+			num++
+			newVal := strconv.Itoa(num)
```
govint: Can this be like newVal := strconv.Itoa(++num)

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -649,11 +660,152 @@ func (e *EtcdKVS) DeleteMetaData(name string) error {
...
+			}
+
+			if len(resp.Kvs) == 0 {
+				return fmt.Errorf("AtomicIncr: no key found for %s", key)
+			}
+
+			oldVal := string(resp.Kvs[0].Value)
+			num, _ := strconv.Atoi(oldVal)
+			if num == 0 {
+				return fmt.Errorf("Cannot decrease a value equal to 0")
```
govint: Is this an expected scenario or should this just return OK?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+
+	// Increase GRef
+	log.Infof("Before AtomicIncr")
+	err = d.kvStore.AtomicIncr(kvstore.VolPrefixGRef + name)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name,
+				"error": err},
+		).Error("Failed to increase global refcount when processMount ")
+		return "", err
```
govint: Remove the mount dir also?

client_plugin/drivers/shared/shared_driver.go
govint: Its possible the etcd service is down, is there a way for these volume ops to identify that case and report that?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+	}
+
+	// Blocking wait till Mounted
+	info, err := d.kvStore.BlockingWaitAndGet(kvstore.VolPrefixState+name,
+		string(kvstore.VolStateMounted), kvstore.VolPrefixInfo+name)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to blocking wait for Mounted state. Error: %v.", err)
```
govint: Failed to _do_ or _complete_ transit to _Mounted_ state or simply Failed to transit to _Mounted_ state for volume _volname_?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+	log.Infof("Volume state mounted, prepare to mounting locally")
+	var volRecord VolumeMetadata
+	// Unmarshal Info key
+	err = json.Unmarshal([]byte(info), &volRecord)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"name": name,
+				"error": err},
+		).Error("Failed to unmarshal info data ")
+		return "", err
```
govint: Shouldn't the volume be unmounted in the KV store and stop the samba service and detach the backing volume?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+		log.Fields{"name": name,
+			"Port":        volRecord.Port,
+			"ServiceName": volRecord.ServiceName,
+		}).Info("Get info for mounting ")
+	err = d.mountSharedVolume(name, mountpoint, &volRecord)
+	if err != nil {
+		msg := fmt.Sprintf("Failed to mount shared volume. Error: %v.", err)
+		err = d.kvStore.AtomicDecr(kvstore.VolPrefixGRef + name)
+		if err != nil {
+			msg += fmt.Sprintf(" Also failed to decrease global refcount. Error: %v.", err)
```
govint: Same comment as above for resetting volume state and detaching backing volume.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +401,171 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+			log.Fields{"name": name,
+				"error": msg}).Error("")
+		return "", errors.New(msg)
+	}
+
+	return mountpoint, nil
+}
+
+// mountSharedVolume - mount the shared volume.
+func (d *VolumeDriver) mountSharedVolume(volName string, mountpoint string, volRecord *VolumeMetadata) error {
```
govint: Could add more comments mentioning exactly from where the shared volume is going to be mounted, prerequisite is that the volume must be in _mounted_ state.

govint: Plus there can be spaces in the volume name and should be handled (for any args that could have spaces in them.

client_plugin/drivers/shared/shared_driver.go
luomiao: "Is there a check needed to verify ... the samba service for this volume and the backing VMDK volume are present. "
The caller of Mount() can only proceed when the state of volume becomes "Mounted". And it's the responsibility of the event handler on one of the manager nodes to start a SMB server, the event handler will fail if they are not able to start the SMB server, and thus the volume will never be "Mounted". And the caller of Mount() will timeout eventually.

"What happens if a mount and a remove are in a race?"
There won't be a mount and remove in a race. This is because all the change of states of the volumes are atomic transactions. A mount can only proceed, if the event handler on one of the manager plugins can change the state from "Ready" to "Mounting". Similarly, a remove can only proceed, if the Remove() function can change the state from "Ready" to "Deleting". If the plugins failed to change the state, they will either wait, or return error, depending on the current state.
So the caller of Mount() will get an error in the case that this volume is being removed (or already removed) by other nodes.

client_plugin/drivers/shared/shared_driver.go
luomiao: If the first mount failed, this plugin will revert back the global refcount from 1 to 0. And the watcher on one of the managers will catch this event and stop smb server.

client_plugin/drivers/shared/shared_driver.go
luomiao: It seems both vmdk driver and photon driver don't remove the dir when it's failed. Is there a reason why or why not to delete the dir?

client_plugin/drivers/shared/shared_driver.go
luomiao: Will update the comment.
The change to Mounted state is done by manager plugin's event handler.

"Its possible the etcd service is down, is there a way for these volume ops to identify that case and report that?"
Yes if etcd service is done, this BlockingWaitAndGet will return error accordingly.

client_plugin/drivers/shared/shared_driver.go
luomiao: No it failed to wait for the change of state.
This caller doesn't do the transit. It's "waiting" for the event handler on the manager nodes to do the transit.
It could fail because timeout (event handler fails to update the state somehow) or because of error (etcd is not available).

client_plugin/drivers/shared/shared_driver.go
luomiao: Again this is done by event handler, it cannot unmount volume and stop smb service here :)

client_plugin/drivers/shared/shared_driver.go
luomiao: same reply as above :)

client_plugin/drivers/shared/shared_driver.go
luomiao: I will create an issue to handle that.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
luomiao: yep. will update.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
luomiao: No this is not expected. It should error out.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -55,6 +58,8 @@ const (
 	sambaImageName = "dperson/samba"
 	sambaUsername  = "root"
 	sambaPassword  = "badpass"
+	fileShareName  = "public"
```
shivanshu21: This is "share1" as we discussed. I will move this to dockerops and use the name share1 once I rebase.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -55,6 +58,8 @@ const (
 	sambaImageName = "dperson/samba"
 	sambaUsername  = "root"
```
shivanshu21: This will be moved to dockerops after I rebase.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -234,6 +234,7 @@ func (d *VolumeDriver) Create(r volume.Request) volume.Response {
 	volRecord := VolumeMetadata{
 		Status:         kvstore.VolStateCreating,
 		GlobalRefcount: 0,
+		Port:           1445,
```
shivanshu21: Why 1445? Earlier it was 0.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -402,8 +403,175 @@ func (d *VolumeDriver) Path(r volume.Request) volume.Response {
...
+			log.Fields{"name": name,
+				"error": msg}).Error("")
+		return "", errors.New(msg)
+	}
+
+	// Start mounting
+	log.Infof("Volume state mounted, prepare to mounting locally")
+	var volRecord VolumeMetadata
+	// Unmarshal Info key
+	err = json.Unmarshal([]byte(info), &volRecord)
```
shivanshu21: I merged my changes with yours in a local branch. This unmarshalling is returning volRecord that doesn't have username and password.

I directly supplied the const username and password to cifs mounting logic for demo purposes. Can you check why username and password are not showing up? Print them in the log line below.

Here is what I saw:
```
2017-08-10 22:23:45.502876484 -0700 PDT [INFO] Mounting volume with options volume name=vol222 arguments=[-t cifs -o username=,password=,port=30000,vers=3.0 //192.168.213.140/public /mnt/vshared/vol222]
```

client_plugin/drivers/shared/shared_driver.go
govint: Thanks for detailing, am able to get the code flow much better now.

client_plugin/drivers/shared/shared_driver.go
luomiao: This is just for making this partial part working.
This should be removed after your PR is in.

client_plugin/drivers/shared/shared_driver.go
luomiao: With only this PR, there is no problem to achieve username and password.
I think what you observed here is because you re-write the info value with the marshal string from sharedVolConnectivityData, which only has port and service name. As a result username and password is overwritten. I added a comment in your PR regarding to this problem: https://github.com/vmware/docker-volume-vsphere/pull/1742/files#diff-d20300e752878bf03d64ac84f8f56dc5R368

Issue #1709, opened 2017-08-06
[Debuggability/Serviceability] Add logging for important workflows


pdhamdhere:
We do not log error in Plugin for "Get" request;

```
	status, err := d.GetVolume(r.Name)
	if err != nil {                              <<<<<No logging
		return volume.Response{Err: err.Error()}
	}
```

```
func (d *VolumeDriver) GetVolume(name string) (map[string]interface{}, error) {
	return d.ops.Get(name)
}
```

```
func (v VmdkOps) Get(name string) (map[string]interface{}, error) {
	log.Debugf("vmdkOps.Get name=%s", name)
	str, err := v.Cmd.Run("get", name, make(map[string]string))
	if err != nil {            <<<<< No logging
		return nil, err
	}
```

PR #1737, opened 2017-08-10
Added log for errors on get volume meta-data

govint:
Added a log for get op. Fixes the issue in #1709. Checked and added logs for Get() and List() ops.


govint:
vshared is under development now and I'd defer to @luomiao and @shivanshu21  to handle logging there.

luomiao:
I will add this logging in https://github.com/vmware/docker-volume-vsphere/pull/1727

PR #1742, opened 2017-08-11
Server side Mounting

shivanshu21:
**About the PR**
Automatically loads Samba container into Docker on install
Starts Samba server when volume global refcount goes from 0 -> 1
Waits till Samba container comes up
Updates volume metadata accordingly

**Testing done**
Create volumes from manager and worker nodes
Use Etcdctl to up the global refcount(client mounting code being written)
Watch the file server container come up as part of Samba service
Repeat for other volumes
Delete Samba service manually
Bring volume Refcount, Status to 0, Ready using etcdctl
Delete the volumes

**Logs**

```
root@ubuntu16B ~: docker volume ls
local               428d9e505ee6154dbf2ea85ce2f8236b683b59edd1b63d55f6c66e55472bf05a
vsphere:latest      InternalVolcat1@datastore1
vsphere:latest      InternalVolvol1@datastore1
kahuna:latest       cat1
kahuna:latest       vol1
root@ubuntu16B ~: 
root@ubuntu16B ~: docker service ls
ID                  NAME                MODE                REPLICAS            IMAGE                  PORTS
o860evvlzpxy        vol1fileServer      replicated          1/1                 dperson/samba:latest   *:0->445/tcp
suwue94lgeeo        cat1fileServer      replicated          1/1                 dperson/samba:latest   *:0->445/tcp
root@ubuntu16B ~: 
root@ubuntu16B ~: docker volume inspect cat1
[
    {
        "Driver": "kahuna:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/cat1",
        "Name": "cat1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 30001,
            "Global Refcount": 1,
            "Service name": "cat1fileServer",
            "Volume Status": "Mounted"
        }
    }
]
root@ubuntu16B ~: 
```

```
2017-08-10 17:57:31.634402577 -0700 PDT [INFO] Attempting to create internal volume for cat1
2017-08-10 17:57:32.364146854 -0700 PDT [INFO] Attempting to update volume state to ready for volume: cat1
2017-08-10 17:57:32.379628157 -0700 PDT [INFO] Successfully created volume: cat1
2017-08-10 17:57:41.168559747 -0700 PDT [INFO] Watcher on global refcount returns event type=PUT
2017-08-10 17:59:20.328577314 -0700 PDT [INFO] VolumeDriver Get: cat1
2017-08-10 17:59:36.50982949 -0700 PDT [INFO] Watcher on global refcount returns event type=PUT
2017-08-10 17:59:37.510754501 -0700 PDT [INFO] Attempting to change volume state to Mounting
2017-08-10 17:59:44.050209024 -0700 PDT [INFO] Checking status of file server container...
2017-08-10 17:59:44.05406241 -0700 PDT [INFO] Updating port and file service name for cat1
```

### DIFF
--- a/client_plugin/drivers/shared/dockerops/dockerops.go
+++ b/client_plugin/drivers/shared/dockerops/dockerops.go
@@ -28,7 +28,11 @@ import (
 	log "github.com/Sirupsen/logrus"
 	dockerClient "github.com/docker/engine-api/client"
 	dockerTypes "github.com/docker/engine-api/types"
+	"github.com/docker/engine-api/types/filters"
 	"github.com/docker/engine-api/types/swarm"
+	"io"
+	"os"
+	"time"
 )
 
 const (
@@ -36,6 +40,31 @@ const (
 	dockerAPIVersion = "v1.24"
 	// dockerUSocket: Unix socket on which Docker engine is listening
 	dockerUSocket = "unix:///var/run/docker.sock"
+	// Postfix added to names of Samba services for volumes
+	serviceNamePostfix = "fileServer"
+	// Path where the file server image resides in plugin
+	fileServerPath = "/usr/lib/vmware/samba.tar"
+	// Driver for the network which Samba services will use
+	// for communicating to clients
+	networkDriver = "overlay"
+	// Name of the Samba server docker image
+	sambaImageName = "dperson/samba"
+	// Name of the Samba share used to expose a volume
+	FileShareName = "share1"
+	// Default username for all accessing Samba server mounts
+	SambaUsername = "root"
+	// Default password for all accessing Samba server mounts
+	SambaPassword = "badpass"
+	// Port number inside Samba container on which
+	// Samba service listens
+	defaultSambaPort = 445
+	// Time between successive checks for Samba service
+	// status to see if service container was launched
+	checkDuration = 5 * time.Second
+	// Timeout to mark Samba service launch as unsuccessful
+	sambaRequestTimeout = 30 * time.Second
+	// Prefix for internal volume names
+	internalVolumePrefix = "InternalVol"
 )
 
 // DockerOps is the interface for docker host related operations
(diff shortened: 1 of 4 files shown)


client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
 // StartSMBServer - Start SMB server
-func (d *DockerOps) StartSMBServer(volName string) bool {
-	log.Errorf("startSMBServer to be implemented")
-	return true
+func (d *DockerOps) StartSMBServer(volName string) (int, string, bool) {
+	var service swarm.ServiceSpec
+	var options dockerTypes.ServiceCreateOptions
+
+	// Name of the service
+	service.Name = volName + serviceNamePostfix
```
shivanshu21: Add a random number to service name. This will be great precaution against orphan Samba services running.

If an orphan is running for the same volume name, and someone mounts from global refcount 0 -> 1, new samba service creation will fail because of service name conflict.

Adding a random number will avoid a name conflict.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
 }
 
 // StartSMBServer - Start SMB server
-func (d *DockerOps) StartSMBServer(volName string) bool {
-	log.Errorf("startSMBServer to be implemented")
-	return true
+func (d *DockerOps) StartSMBServer(volName string) (int, string, bool) {
```
luomiao: Can you comment what's the returned values meaning?
int: port number
string: service name
bool: success or not

Also I think it might be better to return port number as string instead of int?

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -338,12 +346,44 @@ func (e *EtcdKVS) etcdEventHandler(ev *etcdClient.Event) {
...
+				log.Warningf("Failed to update port and file service name: %v",
+					err)
+				e.CompareAndPut(kvstore.VolPrefixState+volName,
+					string(interimState),
+					string(kvstore.VolStateError))
+				return
+			}
+			entries = append(entries, kvstore.KvPair{
+				Key:   kvstore.VolPrefixInfo + volName,
+				Value: string(byteRecord)})
```
luomiao: Question: the sharedVolConnectivityData only has Port and ServiceName field. So the byteRecord will only contain these two fields. If we writeMetaData with this byteRecord string, won't this operation overwrite other existing fields for the value of Info key? such as the username/password, clients...

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -28,19 +28,52 @@ import (
...
 
 const (
 	// dockerAPIVersion: docker engine 1.24 and above support this api version
 	dockerAPIVersion = "v1.24"
 	// dockerUSocket: Unix socket on which Docker engine is listening
 	dockerUSocket = "unix:///var/run/docker.sock"
+	// Postfix added to names of Samba services for volumes
+	serviceNamePostfix = "fileServer"
+	// Path where the file server image resides in plugin
+	fileServerPath = "/usr/lib/samba.tar"
```
govint: All VMW artifacts are typically located in /usr/lib/vmware/..., for vdvs its under /usr/lib/vmware/vmdkops  - may be something similar for this plugin too to locate this server?

govint: OTOH if this is a path inside the container does this path get propagated to Docker for Docker to be able to access the file? Ideally better to keep plugin artifacts outside of access to any user outside of the plugin.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	// service.EndpointSpec is an input for service create.
+	// It carries the previous data structure as well as Mode.
+	// ResolutionModeVIP indicates that networking uses virtual IPs.
+	service.EndpointSpec = &swarm.EndpointSpec{
+		Mode:  swarm.ResolutionModeVIP,
+		Ports: exposedPorts,
+	}
+
+	//Start the service
+	resp, err := d.Dockerd.ServiceCreate(context.Background(),
```
govint: could add a note or doc somewhere as to how this service is scaled/replicated across nodes or within a node. How volumes are accessed across nodes. How is a service instance mapped to a volume as each volume gets mounted (gref 0 -> 1)

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+func (d *DockerOps) isFileServiceRunning(servID string, volName string) (uint32, bool) {
+	var port uint32
+	// Grep the samba service running for this volume using service ID
+	serviceFilters := filters.NewArgs()
+	serviceFilters.Add("id", servID)
+	services, err := d.Dockerd.ServiceList(context.Background(),
+		dockerTypes.ServiceListOptions{Filter: serviceFilters})
+	if err != nil {
+		log.Warningf("Failed to check if file server for volume %s was started. %v", volName, err)
+		return port, false
```
lipingxue: Why return "port" during failure, and it seems to me that "port" has not been assigned a value yet.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	}
+	if len(services) < 1 {
+		log.Warningf("No service returned for volume %s. Service not started properly.", volName)
+		return port, false
```
lipingxue: same as above.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	}
+
+	port = services[0].Endpoint.Ports[0].PublishedPort
```
lipingxue: I think we need to check services[0].Engpoint.Ports is not nil here.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+
+	// Grep all tasks for the service returned and verify that their state is running
```
lipingxue: Nit: "their state is running" -> "their states are running"

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	taskFilter := filters.NewArgs()
+	for _, service := range services {
+		taskFilter.Add("service", service.ID)
+	}
+	tasks, err := d.Dockerd.TaskList(context.Background(),
+		dockerTypes.TaskListOptions{Filter: taskFilter})
+	if err != nil {
+		log.Warningf("Failed to get task list for file service for volume %s. %v", volName, err)
+		return port, false
```
lipingxue: same as above.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	}
+	for _, task := range tasks {
+		if task.Status.State != swarm.TaskStateRunning {
+			log.Infof("File server not running for volume %s", volName)
+			return port, false
```
lipingxue: same question as above.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
-func (d *DockerOps) StopSMBServer(volName string) bool {
+func (d *DockerOps) StopSMBServer(volName string) (int, string, bool) {
 	log.Errorf("stopSMBServer to be implemented")
+	return 0, "", true
+}
+
+// loadFileServerImage - Load the file server image present
+// in the plugin to Docker images
+func (d *DockerOps) LoadFileServerImage() bool {
+	file, err := os.Open(fileServerPath)
```
lipingxue: File "fileServerPath" is opened but not closed.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -338,12 +346,44 @@ func (e *EtcdKVS) etcdEventHandler(ev *etcdClient.Event) {
...
+			// port number and file service name
+			var entries []kvstore.KvPair
+			var volRecord sharedVolConnectivityData
+			volRecord.Port = port
+			volRecord.ServiceName = servName
+			byteRecord, err := json.Marshal(volRecord)
+			if err != nil {
+				log.Warningf("Failed to update port and file service name: %v",
+					err)
+				e.CompareAndPut(kvstore.VolPrefixState+volName,
```
lipingxue: Please add a comment for this line to explain what is this line doing. It will help others to understand the code.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
lipingxue: +1 on this.

client_plugin/drivers/shared/dockerops/dockerops.go
luomiao: In golang, if the int is not assigned a value, it will be 0 by default.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: A Samba service that is running without a volume is called an Orphan samba service.
This can be a result of an improper cleanup during unmount. I was hitting this now because as of now we don't have unmount.

Once we have an orphan service stuck for a volume, we cannot remount that volume as mount() will try to launch service with the same name. The name will conflict.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: From what I could gather from the Go Doc, ImageLoad() takes an io.Reader. This is what I got from Open().

ImageLoad() returns an io.ReadCloser() which I called at line 333: err = resp.Body.Close(). 

Is this not sufficient to close the file? Conventionally after an open() a close() should be called on the same object. But then what is the io.ReadCloser() for?

https://godoc.org/github.com/moby/moby/client#Client.ImageLoad

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Will move it to /usr/lib/vmware/. The path is in the plugin runC container and is not exposed to outside world directly. The ImageLoad() method takes an io.reader() to read this file and asks Docker to load it as an image.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: As per discussion with @luomiao, We don't want this.

client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -142,13 +175,165 @@ func (d *DockerOps) VolumeRemove(volName string) error {
...
+	var exposedPorts []swarm.PortConfig
+	exposedPorts = append(exposedPorts, swarm.PortConfig{
+		Protocol:   swarm.PortConfigProtocolTCP,
+		TargetPort: defaultSambaPort,
+	})
+	// service.EndpointSpec is an input for service create.
+	// It carries the previous data structure as well as Mode.
+	// ResolutionModeVIP indicates that networking uses virtual IPs.
+	service.EndpointSpec = &swarm.EndpointSpec{
+		Mode:  swarm.ResolutionModeVIP,
```
shivanshu21: Updated comment.

```
        // service.EndpointSpec is an input for service create.
        // It carries the previous data structure as well as Mode.

        // Mode here is the mode we want to use for service discovery.
        // Outside clients do not know on which node is the service
        // running or how many containers are running inside or their
        // IP addresses. Service discovery mechanisms like Virtual IPs
        // or DNS round robin are used to route packets from
        // 127.0.0.1:port to the service container.
        // swarm.ResolutionModeVIP implies that we want to use
        // virtual IPs for service resolution.
```
https://docs.docker.com/engine/swarm/networking/#configure-service-discovery

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Port will be zero by default.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Port will be zero by default.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Port will be zero by default.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Explored a little. This is sufficient.

client_plugin/drivers/shared/dockerops/dockerops.go
shivanshu21: Done

luomiao:
LGTM

Issue #1722, opened 2017-08-08
Allow deletion of volumes stuck in deleting state


shivanshu21:
Volumes that experience failure in deletion (eg. If the internal volume deletion failed) are stuck in Deleting state.

The next time, the plugin doesn't allow them to proceed to deletion as it is expecting the state "Ready". If the state is already deleting, the deletion should just proceed.

root@ubuntu16B ~: docker volume rm vol1
Error response from daemon: unable to remove volume: remove vol1: VolumeDriver.Remove: Failed to remove internal volume vol1. Reason: Error response from daemon: {"message":"unable to remove volume: remove InternalVolvol1: volume is in use - [572265485e36d56edcd3a234972c5a1d9e3251e53f05943ca0fe6418d563aa3b, 8719b601cd469d3f24d92833c7ff5e61aaf0b0fe2616700d819cc58a2467606e]"} Check the status of the volumes belonging to driver vsphere
root@ubuntu16B ~: docker volume ls
DRIVER              VOLUME NAME
vsphere:latest      InternalVolvol1@datastore1
kahuna:latest       vol1
root@ubuntu16B ~: docker volume inspect vol1
[
    {
        "Driver": "kahuna:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vmdk/vol1",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Deleting"
        }
    }
]
root@ubuntu16B ~: 

root@ubuntu16B ~: docker volume rm vol1
Error response from daemon: unable to remove volume: remove vol1: VolumeDriver.Remove: Remove failed: Failed to set volume state to deleting. Containers on other host VM are still using volume vol1. Host VMs using this volume:
root@ubuntu16B ~:

luomiao:
When the deleting of internal volume is failed, Remove() should set the state of volume back from Deleting to Ready before return error.

PR #1743, opened 2017-08-11
vShared plugin: improve volume state handling during Remove()

luomiao:
Fixes: https://github.com/vmware/docker-volume-vsphere/issues/1722

When Remove() function failed to update state of shared volume from
Ready to Deleting:
1. Volume is in Mounted state -> get clients and return error
2. Volume is already in Deleting state and timeout -> continue deleting
3. Volume is in other states -> return error

When the deletion of internal volume failed, depending on the state of
the internal volume, Remove() function should return error, or continue
deleting the metadata.

### DIFF
--- a/client_plugin/drivers/shared/shared_driver.go
+++ b/client_plugin/drivers/shared/shared_driver.go
@@ -373,7 +373,7 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 			log.Errorf(msg)
 			return volume.Response{Err: msg}
 		} else {
-			msg = fmt.Sprintf("Remove failed: cannot delete from current state.")
+			msg = fmt.Sprintf("Remove failed: cannot delete from current state %s.", state)
 			log.Errorf(msg)
 			return volume.Response{Err: msg}
 		}
--- a/client_plugin/drivers/shared/shared_driver.go
+++ b/client_plugin/drivers/shared/shared_driver.go
@@ -336,32 +336,47 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
 	if !d.kvStore.CompareAndPutStateOrBusywait(kvstore.VolPrefixState+r.Name,
 		string(kvstore.VolStateReady),
 		string(kvstore.VolStateDeleting)) {
-		clientFetchSucceeded := true
+		// Failed to change state from Ready to Deleting
+		// 1. Volume is in Mounted state -> get clients and return error
+		// 2. Volume is already in Deleting state and timeout -> continue delete
+		// 3. Volume is in other states -> return error
+
 		// Get a list of host VMs using this volume, if any
 		keys := []string{
+			kvstore.VolPrefixState + r.Name,
 			kvstore.VolPrefixInfo + r.Name,
 		}
 		entries, err := d.kvStore.ReadMetaData(keys)
 		if err != nil {
-			clientFetchSucceeded = false
-			log.Warningf("Failed to check which host VMs are using volume %s", r.Name)
-		}
-		// Unmarshal Info key
-		err = json.Unmarshal([]byte(entries[0].Value), &volRecord)
-		if err != nil {
-			clientFetchSucceeded = false
-			log.Warningf("Failed to unmarshal data. %v", err)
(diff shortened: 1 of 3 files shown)


client_plugin/drivers/shared/shared_driver.go
```diff
@@ -336,43 +336,66 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
+			msg = fmt.Sprintf("Remove failed: cannot read metadata of volume %s", r.Name)
+			log.Errorf(msg)
+			return volume.Response{Err: msg}
 		}
 
-		msg = fmt.Sprintf("Remove failed: Failed to set volume state to deleting.")
-		if clientFetchSucceeded {
-			msg += fmt.Sprintf(" Containers on other host VM are still using volume %s.",
-				r.Name)
+		state := entries[0].Value
```
lipingxue: Is is possible that entries is nil here?

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -336,43 +336,66 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
+				return volume.Response{Err: msg}
+			}
+
+			msg = fmt.Sprintf("Remove failed: volume state is Mounted.")
 			msg += fmt.Sprintf(" Host VMs using this volume: %s",
 				strings.Join(volRecord.ClientList, ","))
+			log.Errorf(msg)
+			return volume.Response{Err: msg}
+		} else {
+			msg = fmt.Sprintf("Remove failed: cannot delete from current state.")
```
lipingxue: Please also log the current state value here.

client_plugin/drivers/shared/shared_driver.go
```diff
@@ -336,43 +336,66 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
 	log.Infof("Attempting to delete internal volume for %s", r.Name)
 	internalVolname := internalVolumePrefix + r.Name
 	err := d.dockerOps.VolumeRemove(internalVolname)
 	if err != nil {
-		msg = fmt.Sprintf("Failed to remove internal volume %s. Reason: %v", r.Name, err)
-		msg += fmt.Sprintf(" Check the status of the volumes belonging to driver %s", d.internalVolumeDriver)
+		msg = fmt.Sprintf("Failed to remove internal volume from driver %s for volume %s. Reason: %v.",
+			d.internalVolumeDriver, r.Name, err)
+
+		// check the internal volume state:
```
lipingxue: Why we need the following check here?

client_plugin/drivers/shared/shared_driver.go
shivanshu21: No, if entries is nil, read metadata returns an error.

client_plugin/drivers/shared/shared_driver.go
luomiao: When VolumeRemove failed, it could because the volume is in use, or the volume is not accessible.
VolumeInspect can tell us which status the volume is.
If it's still existing (in use), we should stop the deleting process.
Or else, we should proceed to delete the metadata since the internal volume is already inaccessible.

luomiao: When VolumeRemove failed, it could because the volume is in use, or the volume is not accessible.
VolumeInspect can tell us which status the volume is.
If it's still existing (in use), we should stop the deleting process.
Or else, we should proceed to delete the metadata since the internal volume is already inaccessible.

luomiao:
@shivanshu21 
I decided to leave it in Deleting state for the following two reasons:

1.
If we allow Deleting->Ready, there is one more state to wait for Ready->Mounting change then.
It will change our state machine of volumes to be more complex.

2.
When a "docker volume rm" command is executed, even when the internal volume is still in use (SMB server not stopping yet), the SMB server may not in a working state anymore. It maybe in the progress of stopping, or it could hang. So internal volume still in use doesn't mean the shared volume can be set back to Ready state. So a better way is, if "docker volume rm" is already executed for a volume, this volume should be marked with Deleting state, until it's successfully deleted.

PR #1754, opened 2017-08-15
vShared plugin: Address timeout error due to docker load image

luomiao:
The loading of SMB server image can take more time than expected.
And a delay in the initialization of the plugin can cause error:
"shared.sock: connect: no such file or directory"

This fix changes the load image function into a separate thread.
Thus initialization won't be blocked on image loading.

### DIFF
--- a/client_plugin/drivers/shared/shared_driver.go
+++ b/client_plugin/drivers/shared/shared_driver.go
@@ -124,10 +124,8 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 
 	// Load the file server image
-	if d.dockerOps.LoadFileServerImage() == false {
-		return nil
-	}
-	log.Infof("Loaded file server image")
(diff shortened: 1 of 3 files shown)


client_plugin/drivers/shared/shared_driver.go
```diff
@@ -124,10 +124,8 @@ func NewVolumeDriver(cfg config.Config, mountDir string) *VolumeDriver {
 	}
 
 	// Load the file server image
-	if d.dockerOps.LoadFileServerImage() == false {
-		return nil
-	}
-	log.Infof("Loaded file server image")
+	go d.dockerOps.LoadFileServerImage()
```
lipingxue: I assume "initialize built-in etcd cluster" is not dependent on whether loading file image is done or not, right?

shivanshu21: Correct

PR #1758, opened 2017-08-15
[SKIP CI] Update Readme.md with instruction to upgrade driver or plugin

### LINKED ISSUE #1700: Document update for vDVS upgrade instruction and expected behavior
One of our customer reached out through gitter channel, it seems our document should provide better detail about expected behavior of upgrade workflow with recommended steps.

for your reference some of the questions to address in our document:

```
1. how to make upgrade of vdvs a safe operation?
2. what actually happens to data which container tries to write to volume during plugin upgrade? If I run docker plugin rm command the /var/run/docker/plugins/*/vsphere.sock is deleted so how container accesses vmdk file in datastore?
```

For more information, please refer https://gitter.im/vmware/docker-volume-vsphere.

/CC @msterin @pdhamdhere

ashahi1:
Fixes #1700

Updated Readme.md with instructions for upgrade of vDVS driver or plugin.


README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
 $ /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
 ```
 
+## Upgrade Instructions
```
shuklanirdesh82: 1. This section should be after installation section (https://github.com/ashahi1/docker-volume-vsphere/blob/198f25c61af6948ccc4d454c3db09eec909b7970/README.md#installation-instructions)

2. content should be part of Installation.md (https://github.com/ashahi1/docker-volume-vsphere/blob/master/docs/user-guide/install.md) and have a reference on this page.

README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
...
+
+User's data is not impacted during driver or plugin upgrade.
+vDVS managed plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like attach/detach/create/delete volume using vDVS driver may run into an error for the time being until the upgrade process is finished.
```
shuklanirdesh82: nit: `attach/detach/create/delete` => `create/attach/detach/delete`  ??

README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
...
+1. Uninstall VIB (vDVS driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the vDVS release page
+```esxcli software vib install -v /tmp/<vib_name>.vib```
+3. Uninstall the managed plugin from the docker host
+```docker plugin rm -f <plugin name>```
+4. Install the latest release of vDVS managed plugin
+```docker plugin install --grant-all-permissions --alias vsphere vmware/docker-volume-vsphere:latest```
+
+We do not require user to stop containers to do plugin upgrade.
```
shuklanirdesh82: I would suggest to move this block before steps (L118-L121)

govint: Sure? if container is running and then plugin is removed and then container stops who detaches the plugin. Same for service.

README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
...
+If plugin upgrade was done while volume was still mounted to a container, user may not see the attached volume after upgrade. Docker restart will fix the issue and all volumes will start getting listed.[Issue #1753](https://github.com/vmware/docker-volume-vsphere/issues/1753)
```
govint: This isn't really true. At least till issue is confirmed.

One more thing to take care is user must ensure that plugin install command _must_ use the same alias that they used for the earlier version of the plugin. Otherwise existing volumes will not be usable till at least docker is restarted (haven't verified).

README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
...
+
+Similarly, no change in behavior of vmgroup operations is expected due to vDVS driver upgrade.
```
govint: What about DB?

govint:
Pls. also check if docker plugin upgrade is feasible to be used vs. remove and re-install.

README.md
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
 $ /usr/lib/vmware/vmdkops/bin/vmdkops_admin.py volume ls
 ```
 
+## Upgrade Instructions
+
+User's data is not impacted during driver or plugin upgrade.
+vDVS managed plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like attach/detach/create/delete volume using vDVS driver may run into an error for the time being until the upgrade process is finished.
+
+Steps for safe upgrade:
+1. Uninstall VIB (vDVS driver)
```
ashahi1: I backed-up and restored the auth-db and everything was working fine.

I performed following steps:
```
1. Created a new vmgroup - vmgroup1 and added vm-1 to it.
2. Created a volume from vm-1.
3. Copied the current auth-db file to different location and then deleted the current auth-db file
4. Upgraded the vib
5. After upgrade, copied back the auth-db file from saved location
6. Did vmgroup ls and I could still the vmgroup1 and vm-1 still belong to the vmgroup1
7. Also vm-1 could only see the volumes belonging to vmgroup1
``

File: README.md [hunk 7]
```diff
@@ -100,6 +100,26 @@ $ docker volume rm MyVolume
...
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the vDVS release page
+```esxcli software vib install -v /tmp/<vib_name>.vib```
```
ashahi1: Added this:
```
To verify Admin CLI service for vDVS is running after upgrade, please run the command `/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status` and verify status is running.
```

README.md
ashahi1: @govint  I verified that if we change the alias name during plugin install and use a newer alias name, I didn't have to restart docker to use the volumes.

Please see the steps below:
```
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin install --grant-all-permissions --alias vspherexxxxx vmware/docker-volume-vsphere:0.15
0.15: Pulling from vmware/docker-volume-vsphere
967f49dfadaf: Download complete
Digest: sha256:64a50830c2fa9befbede6efca3f28d2e2d8b4e595703f1b5984470579656e935
Status: Downloaded newer image for vmware/docker-volume-vsphere:0.15
Installed plugin vmware/docker-volume-vsphere:0.15
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin ls
ID                  NAME                  DESCRIPTION                           ENABLED
5b4d7b7f3c38        vspherexxxxx:latest   VMWare vSphere Docker Volume plugin   true
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~# docker volume create -d vspherexxxxx --name TestVolumexxxxx
TestVolumexxxxx
root@sc-rdops-vm02-dhcp-52-237:~# docker volume ls
DRIVER                VOLUME NAME
vspherexxxxx:latest   TestAAA@shared    Vmfs   -   0
vspherexxxxx:latest   TestVolume123@shared    Vmfs   -   0
vspherexxxxx:latest   TestVolumexxxxx@shared    Vmfs   -   0
vspherexxxxx:latest   vsphere15VOL1@shared    Vmfs   -   0
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin ls
ID                  NAME                  DESCRIPTION                           ENABLED
5b4d7b7f3c38        vspherexxxxx:latest   VMWare vSphere Docker Volume plugin   true
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin rm -f vspherexxxxx
vspherexxxxx
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin install --grant-all-permissions --alias vsphereyyyyy vmware/docker-volume-vsphere:0.16
0.16: Pulling from vmware/docker-volume-vsphere
1d54a760dc2a: Download complete
Digest: sha256:5281d47506740ee927001ea5fc16d8584ed7fc04de0fb25969f0b11eb3243345
Status: Downloaded newer image for vmware/docker-volume-vsphere:0.16
Installed plugin vmware/docker-volume-vsphere:0.16
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~# docker plugin ls
ID                  NAME                  DESCRIPTION                           ENABLED
22380aa655b0        vsphereyyyyy:latest   VMWare vSphere Docker Volume plugin   true
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~# docker volume ls
DRIVER                VOLUME NAME
vsphereyyyyy:latest   TestAAA@shared    Vmfs   -   0
vsphereyyyyy:latest   TestVolume123@shared    Vmfs   -   0
vsphereyyyyy:latest   TestVolumexxxxx@shared    Vmfs   -   0
vsphereyyyyy:latest   vsphere15VOL1@shared    Vmfs   -   0
root@sc-rdops-vm02-dhcp-52-237:~#
root@sc-rdops-vm02-dhcp-52-237:~# docker run -it --volume-driver=vsphereyyyyy:latest -v 'TestVolumexxxxx@shared    Vmfs   -   0':/vol1 --name ub_xxxxx ubuntu
root@ff3b0ec54308:/#
```

README.md
ashahi1: If plugin is removed while volume is still attached, then volume will remain attached. I can document this as part of this PR.

ashahi1:
@govint I have addressed your comments. Please take a look.
Thanks!

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
 ID                  NAME                DESCRIPTION         ENABLED
 ```
 
+## Upgrade Instructions
+
+The volume driver can be upgraded or reinstalled any time with no impact to the user data. User data in volumes existing before an upgrade of the volume driver is retained as it is and is accessible post an upgrade.
+
+Our volume managed plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like create/attach/detach/delete volume using driver may run into an error for the time being until the upgrade process is finished.
```
shuklanirdesh82: nit: `volume managed plugin` => `volume plugin` (managed plugin is the docker term)

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
 ID                  NAME                DESCRIPTION         ENABLED
 ```
 
+## Upgrade Instructions
+
+The volume driver can be upgraded or reinstalled any time with no impact to the user data. User data in volumes existing before an upgrade of the volume driver is retained as it is and is accessible post an upgrade.
```
shuklanirdesh82: `post an upgrade` ??

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
...
+
+Our volume managed plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like create/attach/detach/delete volume using driver may run into an error for the time being until the upgrade process is finished.
+If the plugin is removed while volume is still attached to a vm, then volume will remain attached.
+
+We do not require user to stop containers to do plugin upgrade.
```
shuklanirdesh82: 1. `We do not require` => `We do not recommend` how about this?
2. Please move this line at TOP (*Note:* We do not ....)

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
...
+
+Similarly, no change in behavior of vmgroup operations is expected due to volume driver upgrade.
+
+An upgrade shouldn't have to always require un-install and install of both driver or plugin. At times only one may change.
```
shuklanirdesh82: L105 is confusing to me .. Please rephrase it. BTW, it should be our recommendation to user staying on the same release version. Can you please explain why do we need such information here?

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
...
+But it is recommended to stay on the latest same versions.
+
+ESX Driver upgrade:
+1. Uninstall VIB (driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the vsphere-storage-for-docker release page
+```esxcli software vib install -v /<vib_name>.vib```
```
shuklanirdesh82: `/<vib` ..it leads to confusion about `/` .. you may want to mention about the vib path or vib name that's it.

docs/user-guide/install.md
ashahi1: I think this seems better: ```Plugin upgrade can be done even while container is running.```

docs/user-guide/install.md
ashahi1: I will rephrase it but it was added to have more clarity like driver and plugin upgrade can be independent or each other.

ashahi1: I have added this: ```We recommended to stay on the latest same versions of driver and plugin.```

README.md
govint: Ok, should the back up and restore steps also be documented for the upgrade?

docs/user-guide/install.md
```diff
@@ -91,3 +91,30 @@ vsphere
...
+
+To verify Admin CLI service for vDVS is running after upgrade, please run the command `/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py status` and verify status is running.
+
+Plugin upgrade:
+1. Uninstall the managed plugin from the docker host
+```docker plugin rm -f <plugin name>```
+2. Install the latest release of vsphere-storage-for-docker managed plugin
```
pdhamdhere: Uninstall/Install is broken. Following sequence worked for me;

`docker plugin install --grant-all-permissions --alias vsphere vmware/docker-volume-vsphere:0.15`
...
`docker plugin upgrade --grant-all-permissions vsphere:latest vmware/docker-volume-vsphere:0.16`
...

Please test before recommending

pshahzeb:
For Plugin Upgrade section:

Doing plugin rm -f and then plugin install can result in intermittent issue from docker saying”unable to dial to vsphere.sock”, where it tries to connect to the sock in the previous
plugin id and not the new installed one.
After filing issue with moby, they advices to use upgrade.

While testing plugin upgrade command, all works fine if container’s aren’t running during upgrade.
If they are, and we do plugin upgrade, there appears to be a data inaccessibility to the data persisted before upgrade.
This was a know bug on docker side which they fixed it in 17.06. See this https://github.com/moby/moby/issues/32515

With 17.06,  after the upgrade, we are hitting a different issue while enabling the plugin. Docker is unable to talk to vsphere.sock; if plugin was disabled using -f flag.

So I suggest documenting upgrade instruction as follows:

1. docker plugin disable vsphere:latest (please refrain from using disable -f do avoid issue https://github.com/moby/moby/issues/32515)

2. docker plugin upgrade --grant-all-permissions vsphere:latest vmware/docker-volume-vsphere:0.16

We might also want to let customer know as to why not to do plugin rm and then plugin install for upgrading the plugin and point to https://github.com/moby/moby/issues/34545

pdhamdhere:
Do not recommend what not to do but recommend what to do to upgrade plugin. Makes sense?

ashahi1:
>docker plugin disable vsphere:latest (please refrain from using disable -f do avoid issue moby/moby#32515)
>
So are we suggesting that user will have to stop the container prior to doing plugin upgrade? Because if they are running, user will have to use -f flag.

/cc @pdhamdhere @govint @pshahzeb

govint:
@ashahi1, no, when upgrade is done, (after doing a disable -f of the plugin) the daemon is unmounting all bind mounts of the plugin volumes from the container while its still in use. Immediately after the upgrade none of the written data is visible even if the volume is remounted later. Thats data loss. So. _disable -f" is a strict no-no when upgrading. I'm checking if this is some bind mount related issue or docker is misusing the way these mounts can be used in the first place ( unmount when in use).

govint:
Another issue - once the plugin is upgraded the old bind mounts are first removed but then don't seem to be getting restored once the new plugin is in place. This may be by design and hence even if allowed by docker, forced disable must not be  used at all when plugin is upgraded.

msterin:
> Another issue - once the plugin is upgraded the old bind mounts are first removed but then don't seem to be getting restored once the new plugin is in place. 

i suppose Docker folks assume that plugin is on the data path, Then, removing bind mounts is somewhat understandable. Still, the IO will fail anyways so I guess it's just housekeeping. They still could have removed the bind mounts on container stop or rm only . 

@govint - is there an issue against Docker for plugin updates ?

pshahzeb:
@msterin @govint  

https://github.com/moby/moby/issues/32515
This is the issue against docker for plugin upgrade which explains the issue we are facing.

msterin:
@pshahzeb thanks for the reference. However, this is explaining why do they need to remove the mounts on removing the plugin. Can you elaborate why do we need -f and what are the plugin live upgrade steps suggested by docker (or us) ? It seems the use case is blocked, and if it is, we need to file a  issue with Docker with a suggested approach.

pshahzeb:
> Can you elaborate why do we need -f and what are the plugin live upgrade steps suggested by docker (or us) ?

We need to allow user to do plugin upgrade even when containers (who have vDVS volumes attached) are running. Without using -f, docker won't allow disable saying "plugin is in use".

From plugin upgrade instructions and example from [here](https://docs.docker.com/engine/reference/commandline/plugin_upgrade/#examples) ,
no specific note or instruction for live upgrade of plugin but they do indicate the usage of disable -f flag.

From moby/moby#32515, the issue with respect to dataloss is fixed with 17.06 and we tested it. Previously written data is available after volume upgrade. I verified through a live running container and reading data after plugin upgrade.
The only thing is we are hitting altogether different issue with 17.06 to enable the plugin
```
"root@promc-2n-dhcp105-23:~# docker plugin enable vsphere
Error response from daemon: dial unix /run/docker/plugins/3bc1676f62f894ed4776ca5609d807a55d64bc95f8d12129e664fcdfabee3ea9/vsphere.sock: connect: no such file or directory
root@promc-2n-dhcp105-23:~#"
```
Strange thing is it is only occuring when plugin is disabled through -f.

msterin:
So there is an issue , which seems to be with Docker (it hits stale socket, isn't it)? Is there a tracking issue in moby repo ?

pshahzeb:
The plugin id is preserved during upgrade. So the socket isn't stale as such ie. the path where docker is trying to dial is a valid path. Need more debugging on this. As of now the only sure thing is it only happens when plugin was disabled through -f flag.

docs/user-guide/install.md
pshahzeb: These three instructions look sufficient to me for upgrade.
```
1. docker plugin disable -f vsphere:latest 

2. docker plugin upgrade --grant-all-permissions vsphere:latest vmware/docker-volume-vsphere:latest

3. docker plugin enable vsphere
```

pshahzeb:
@govint @shuklanirdesh82 @pdhamdhere 
Please take a look at the updated diff.

docs/user-guide/install.md
```diff
@@ -158,3 +158,40 @@ Windows service vdvs uninstalled successfully!
 ```
 PS C:\> Set-ExecutionPolicy Unrestricted -Scope Process -Force
 ```
+
+## Upgrade Instructions
+
+The volume driver can be upgraded or reinstalled any time with no impact to the user data. User data in volumes existing before an upgrade of the volume driver is retained as it is and is accessible post an upgrade.
+
+Our volume managed plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like create/attach/detach/delete volume using driver may run into an error for the time being until the upgrade process is finished.
```
govint: "The volume plugin ....

govint: "Management operations like create/attach/detach/delete volume using driver may run into an error for the time being ..." Could be like

"Management operations like create/attach/detach/delete volume may fail until ......"

docs/user-guide/install.md
```diff
@@ -158,3 +158,40 @@ Windows service vdvs uninstalled successfully!
...
+If the plugin is removed while volume is still attached to a vm, then volume will remain attached.
```
govint: " If the plugin is removed while _a_ volume ..."

docs/user-guide/install.md
```diff
@@ -158,3 +158,40 @@ Windows service vdvs uninstalled successfully!
...
+
+We do not require user to stop containers to do plugin upgrade.
+
+Similarly, no change in behavior of vmgroup operations is expected due to volume driver upgrade.
```
govint: Should ESX side upgrade be mentioned in a separate section - basically upgrade can be in two parts - the plugin and the ESX service.  They can be mentioned in separate sections.

docs/user-guide/install.md
govint: Please mention if user must always upgrade both service and plugin or only one. If both then is there an order in which that should happen? For example can containers be running when ESX service is upgraded? Because those may exit and leave the volumes attached to the VM. Forcing the user to bounce the plugin so it can detect and detach unused volumes.

docs/user-guide/install.md
```diff
@@ -158,3 +158,40 @@ Windows service vdvs uninstalled successfully!
...
+An upgrade shouldn't have to always require un-install and install of both driver or plugin. At times only one may change.
+But it is recommended to stay on the latest same versions.
+
+**ESX Driver upgrade:**
+1. Uninstall VIB (driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the vsphere-storage-for-docker release page
+```esxcli software vib install -v /<vib_name>.vib```
+
+To verify Admin CLI service for vDVS is running after upgrade, please run the command `esxcli storage guestvol status` and verify status is running.
```
govint: How does user fall back if the new version is causing an issue.

pshahzeb: Mentioning the downgrade steps specifically.

README.md
```diff
@@ -93,6 +93,9 @@ sudo rpm -ivh <name>.rpm # Photon or rpm based distros
 
 Please refer the [Installation Guide](https://github.com/vmware/docker-volume-vsphere/blob/master/docs/user-guide/install.md#installation-on-windows-docker-hosts) for instructions.
 
+## Upgrade Instructions
```
shuklanirdesh82: let's not add this section here. This is already part of customer facing document and README.md is just redirecting from my PR (https://github.com/vmware/docker-volume-vsphere/pull/1808)

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
 ```
 PS C:\> Set-ExecutionPolicy Unrestricted -Scope Process -Force
 ```
+
+## Upgrade Instructions
+
+The volume driver can be upgraded or reinstalled any time with no impact to the user data. User data in volumes existing before an upgrade of the volume driver is retained as it is and is accessible post an upgrade.
```
shuklanirdesh82: Please reword this:`User data in volumes existing before an upgrade of the volume driver is retained as it is and is accessible post an upgrade.`

something like: User data in volumes stays intact and remains accessible post upgrade.

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
...
+But it is recommended to stay on the latest same versions.
+The order of upgrade of driver or plugin doesn't matter.
+
+We do not require user to stop containers to do the driver or plugin upgrade.
+No change in behavior of vmgroup operations is expected due to driver or plugin upgrade.
+
+**ESX Driver upgrade:**
+1. Uninstall VIB (driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the vsphere-storage-for-docker release page
```
shuklanirdesh82: vsphere-storage-for-docker => docker-volume-vsphere :) (we are back to old name)

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
...
+But it is recommended to stay on the latest same versions.
+The order of upgrade of driver or plugin doesn't matter.
+
+We do not require user to stop containers to do the driver or plugin upgrade.
+No change in behavior of vmgroup operations is expected due to driver or plugin upgrade.
+
+**ESX Driver upgrade:**
+1. Uninstall VIB (driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the docker-volume-vspher release page
```
shuklanirdesh82: need to fix typo

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
...
+Note: If the container using volumes exits during the upgrade of ESX driver (i.e. after vib remove but before vib install), the volumes may remain attached to VM. In such a case, please disable and then enable (restart) the volume plugin to ensure volumes are properly detached.
+
+**Plugin upgrade:**
+The volume plugin is taking care of the control path (management layer) and doesn't deal with data path. So during the upgrade process, only management related operations will be affected and data will be intact. Management operations like create/attach/detach/delete volume may fail until the upgrade process is finished. If the plugin is removed while a volume is still attached to a vm, then volume will remain attached.
+
+1. Disable the managed plugin from the docker host
+```
+docker plugin disable -f vsphere
+```
+2. Upgrade the latest release of vsphere-storage-for-docker managed plugin
```
shuklanirdesh82: need to use correct name instead of "vsphere-storage-for-docker"

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
...
+```
+docker plugin upgrade --grant-all-permissions vsphere:latest vmware/docker-volume-vsphere:latest
+```
+3. Enable the managed plugin from the docker host
+```
+docker plugin enable vsphere
+```
+Note: If you have used a different alias during existing installation of plugin, use that instead of ```vsphere``` in the steps above
```
shuklanirdesh82: 1. Note: => `**Note:**`
2. minor:`existing installation of plugin` => `existing plugin installation` (rewording is more readable)
3. It is better move this "Note" above step#1, user will be aware well in advance.

docs/user-guide/install.md
```diff
@@ -158,3 +158,53 @@ Windows service vdvs uninstalled successfully!
...
+
+## Downgrade to a previous version
+The steps to downgrade to a previous version are exactly the same as the upgrade with following note:
+1. In case of ESX driver the vib_name is the complete path to vib you want to downgrade to. For example, to downgrade the driver to 0.16 release.
```
shuklanirdesh82: isn't it a good ideal to call out explicitly that you have un-install VIB first before performing following step.

pshahzeb: As discussed offline, the earlier pointer to indicate the steps are exactly same as upgrade mentioned above, should suffice. The only thing that's different is the name of vib and plugin which is explained for the sake of example.

docs/user-guide/install.md
```diff
@@ -158,3 +158,54 @@ Windows service vdvs uninstalled successfully!
 ```
 PS C:\> Set-ExecutionPolicy Unrestricted -Scope Process -Force
 ```
+
+## Upgrade Instructions
+
+The volume driver can be upgraded or reinstalled any time with no impact to the user data. User data in the volumes stays intact and remains accessible post upgrade.
```
pdhamdhere: Upgrade of Guest Plugin or ESX Server has no impact on data path and running containers. It is recommend to have Guest Plugin and ESX Server from same release. However, it is not mandatory, Plugin and ESX Server version are backward compatible.

docs/user-guide/install.md
```diff
@@ -158,3 +158,54 @@ Windows service vdvs uninstalled successfully!
...
+We do not require user to stop containers to do the driver or plugin upgrade.
+No change in behavior of vmgroup operations is expected due to driver or plugin upgrade.
+
+**ESX Driver upgrade:**
+1. Uninstall VIB (driver)
+```esxcli software vib remove -n esx-vmdkops-service```
+2. Download and install the latest VIB from the docker-volume-vsphere release page
+```esxcli software vib install -v /<vib_name>.vib```
+
+To verify Admin CLI service for vDVS is running after upgrade, please run the command `esxcli storage guestvol status` and verify status is running.
```
pdhamdhere: Please run  `esxcli storage guestvol status` and verify status is running.

Issue #1553, opened 2017-07-07
Skylight performance improvement - Investigate accelerating fs_windows implementation


venilnoronha:
## Description
The `fs_windows.go` and `device_watcher_windows.go` implementations heavily rely on `PowerShell` scripts to perform `WMI` operations, due to lack of a functional `Go` library to do the same. These `PowerShell` scripts are also embedded within the `.go` files as of today.

The following changes are suggested, in priority order.
1. Replace `PowerShell` scripts with native/third-party `Go` library based implementations.
2. Move `PowerShell` scripts to separate `.ps1` script files (if point 1 isn't immediate).

See #1540 for more details.

## Investigation
The [x/sys/windows](https://godoc.org/golang.org/x/sys/windows) and [syscall_windows](https://golang.org/src/syscall/syscall_windows.go) modules from `Go` don't support device operations. 

Following are a few alternatives to `PowerShell` scripts to implement `fs` functionalities in `Go`.
1. [wmi](https://github.com/a-palchikov/wmi) - a `WMI` wrapper based on Google's legacy [com-and-go](https://code.google.com/archive/p/com-and-go/) package.
2. [go-ole](https://github.com/go-ole/go-ole) based `WMI` queries. StackOverflow discussion [here](https://stackoverflow.com/questions/20365286/query-wmi-from-go).
3. Write a custom wrapper around `kernel32.dll`. See example [here](https://github.com/LukeEuler/bodhi/blob/master/filesystem/filesystem_windows.go).
4. [StackExchange/wmi](https://github.com/StackExchange/wmi) - `IDispatch` based third-party wrapper around `WMI`.
5. [go-wmi](https://github.com/gabriel-samfira/go-wmi) - yet another `IDispatch` based `WMI` wrapper.
6. [walk](https://github.com/lxn/walk) and [win](https://github.com/lxn/win) - multiple `DLL` wrappers.

These libraries/implementations don't fully support device operations as-is, so there is a fair amount of work in getting them to work for our `fs` module.

msterin:
Thanks for the list of API choices. It seems that as long as our PS implementation is not a perf bottleneck (maybe pre-load PS to avoid 1st time perf hit?) we should be fine with PS, but let's keep this Issue around.

venilnoronha:
Since we already have a fully functional plugin built in `Go` (for Linux of course), I would just implement the device handling logic in `C` or `C#` and use it within the `Go` plugin, to maximize reuse.

[...]

PR #1771, opened 2017-08-17
Vendors go-powershell package

venilnoronha:
This PR vendors `go-powershell` and related packages which will be later used for developing the single-session `powershell` utility.

 `go-powershell` creates a regular PowerShell instance and talks to it over `std` IO. The reason to use that package is that it abstracts the PowerShell command interface, and manages `std` IO streams for us. Additionally, `go-powershell` depends on the `juju/errors` package.

Also, have a look at #1553 and [this](https://github.com/vmware/vsphere-storage-for-docker/issues/1553#issuecomment-316847782) comment.


venilnoronha:
Sure. `go-powershell` creates a regular PowerShell instance and talks to it over `std` IO. The reason to use that package is that it abstracts the PowerShell command interface, and manages `std` IO streams for us. Additionally, `go-powershell` depends on the `juju/errors` package.

PR #1774, opened 2017-08-18
vShared: remove orphan services and internal volumes

luomiao:
Due to unexpected errors, the docker services and internal volumes may not be removed completely when the shared volume is deleted. The existing of those orphan services and internal volumes will result in a waste of resources and can hurt the docker performance. It can also result in error when creating volumes with the previous existed names.

This change adds handling for those orphan services and orphan internal volumes. We first query all the shared volumes and their states from kv store. Then we compare if there is any service or internal volume doesn't have a matching shared volume or state. No matching shared volume means the service or internal volume is orphan, which need to be removed.

### DIFF
--- a/client_plugin/drivers/shared/dockerops/dockerops.go
+++ b/client_plugin/drivers/shared/dockerops/dockerops.go
@@ -358,7 +358,7 @@ func (d *DockerOps) isFileServiceRunning(servID string, volName string) (uint32,
 //      error:   error returned when it can not can service ID and port number
 func (d *DockerOps) getServiceIDAndPort(volName string) (string, uint32, error) {
 	// Grep the samba service running using service name
-	serviceName := volName + serviceNamePostfix
+	serviceName := serviceNamePrefix + volName
 	serviceFilters := filters.NewArgs()
 	serviceFilters.Add("name", serviceName)
 	services, err := d.Dockerd.ServiceList(context.Background(),
--- a/client_plugin/drivers/shared/dockerops/dockerops.go
+++ b/client_plugin/drivers/shared/dockerops/dockerops.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -42,7 +43,7 @@ const (
 	// dockerUSocket: Unix socket on which Docker engine is listening
 	dockerUSocket = "unix:///var/run/docker.sock"
 	// Postfix added to names of Samba services for volumes
-	serviceNamePostfix = "fileServer"
+	serviceNamePrefix = "vSharedServer_"
 	// Path where the file server image resides in plugin
 	fileServerPath = "/usr/lib/vmware/samba.tar"
 	// Driver for the network which Samba services will use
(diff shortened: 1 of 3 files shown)


client_plugin/drivers/shared/dockerops/dockerops.go
```diff
@@ -41,7 +43,7 @@ const (
 	// dockerUSocket: Unix socket on which Docker engine is listening
 	dockerUSocket = "unix:///var/run/docker.sock"
 	// Postfix added to names of Samba services for volumes
-	serviceNamePostfix = "fileServer"
```
shivanshu21: is this working with underscore? I was receiving an error for service name not being DNS compliant while running automation.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -59,6 +60,7 @@ const (
 	etcdClusterStateExisting = "existing"
 	requestTimeout           = 5 * time.Second
 	checkSleepDuration       = time.Second
+	gcTicker                 = 5 * time.Second
```
shivanshu21: Isn't this too expensive to run every 5 seconds? Maybe 30 seconds gap should be fine.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
```diff
@@ -332,6 +335,63 @@ func (e *EtcdKVS) etcdWatcher(cli *etcdClient.Client) {
...
+// trimVolName: trim the volume name if there is special split characters existing in the name
+func trimVolName(volName string) string {
+	// Currently we only take @ as the split character
+	// TODO: need to take care of volumes with same name but on different datastores
+	s := strings.Split(volName, "@")
+	return s[0]
+}
+
+// cleanOrphanServiceAndVolume: stop orphan services and delete orphan internal volumes
+func (e *EtcdKVS) cleanOrphanServiceAndVolume(volumesToVerify []string, stopService bool) {
```
shivanshu21: While we are removing internal volume and samba service in this function, will the metadata for a shared volume in "Error" state still remain in etcd?

client_plugin/drivers/shared/dockerops/dockerops.go
luomiao: This seems works for me. I can remove the underscore. But do you know which component the DNS complaint is coming from?

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
luomiao: Volume with "Error" state will stay until the user executes "docker volume rm" on this shared volume. This function will only remove service and internal volumes if the corresponding shared volume metadata is not in etcd, or the state of shared volume is "Deleting" or "Error". It won't delete shared volume metadata here.

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
lipingxue: +1 on this

client_plugin/drivers/shared/kvstore/etcdops/etcdops.go
lipingxue: Why we want to delete the internal volume if the shared volume when the state is  "Error"?  I think deleting the internal volume when shared volume is "Deleting" make sense since user wants to delete that shared volume. But for other cases which cause the state to "Error", why do we want to delete the internal volume?

luomiao: Makes sense. I will update to stop the SMB service but not remove the internal volume when it's Error state.

PR #1783, opened 2017-08-18
Updating e2e tests to make use of esxcli instead of vmdkops_admin.py

pshahzeb:
## Description

1. This PR updates our e2e tests to make use of esxcli storage vdvs
2. Since -c flag (toggle what columns we need for volume ls) isn't supported through esxcli, I had to tweak some list utils to make use of appropriate columns from volumn ls output
3. Error of a command is prefixed with string "ERROR". This string is consumed by esxcli. So it only shows the actual error message. Good thing is it does set exit status to 1 i.e. command failed.

### DIFF
--- a/client_plugin/Makefile
+++ b/client_plugin/Makefile
@@ -424,19 +424,19 @@ test-e2e-shared:
 .PHONY: test-e2e-runalways test-e2e-runonce test-e2e-runonce-windows
 test-e2e-runalways:
 	$(log_target)
-	$(GO) test -v -timeout 40m -tags "runalways" $(E2E_Tests)
+	$(GO) test -v -timeout 50m -tags "runalways" $(E2E_Tests)
 
 test-e2e-runonce:
 	$(log_target)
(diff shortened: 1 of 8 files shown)


client_plugin/Makefile
```diff
@@ -424,19 +424,19 @@ test-e2e-shared:
 .PHONY: test-e2e-runalways test-e2e-runonce test-e2e-runonce-windows
 test-e2e-runalways:
 	$(log_target)
-	$(GO) test -v -timeout 40m -tags "runalways" $(E2E_Tests)
+	$(GO) test -v -timeout 50m -tags "runalways" $(E2E_Tests)
```
lipingxue: Why we need to increase the timeout value?

pshahzeb: Since all of the commands are triggered through esxcli, its adding a bit more time to overall time of running tests (since we fire so much admin commands over ssh)

lipingxue: ok, make sense.

PR #1795, opened 2017-08-22
Use driver name to detect vmdk volumes

govint:
Use the driver name for a volume vs. the vol path to figure if its a vsphere managed volume.

Will update test status once its done on local setup.

### DIFF
--- a/client_plugin/utils/refcount/refcnt.go
+++ b/client_plugin/utils/refcount/refcnt.go
@@ -316,20 +316,13 @@ func (r *RefCountsMap) Decr(vol string) (uint, error) {
 }
 
 // check if volume with source as mount_source belongs to vmdk plugin
-func isVMDKMount(mount_source string) bool {
-	managedPluginMountStart := "/var/lib/docker/plugins/"
+func isVMDKMount(driver string) bool {
+	pluginDriver := "vsphere"
 
 	// if plugin is used as a service
-	if strings.HasPrefix(mount_source, mountRoot) {
+	if strings.HasPrefix(driver, pluginDriver) {
 		return true
 	}
-	// if plugin is used as managed plugin
-	// managed plugin has mount source in format:
-	// '/var/lib/docker/plugins/{plugin uuid}/rootfs/mnt/vmdk/{volume name}'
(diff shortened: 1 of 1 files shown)


client_plugin/utils/refcount/refcnt.go
```diff
@@ -387,7 +380,7 @@ func (r *RefCountsMap) discoverAndSync(c *client.Client, d drivers.VolumeDriver)
 		log.Debugf("  Mounts for %v", ct.Names)
 		for _, mount := range containerJSONInfo.Mounts {
 			// check if the mount location belongs to vmdk plugin
-			if isVMDKMount(mount.Source) != true {
+			if isVMDKMount(mount.Driver) != true {
```
pshahzeb: Will this work if the user sets a different alias while installing?

govint: Good point and no. But it could be documented to use a _recommended_ alias - something that has  _vsphere_ in it. I mean if they install the vsphere plugin and call it by another org's name that won't be right. Plus playing with alias names can mean that volumes won't be accessible across aliases though they are managed by the same plugin. What do you suggest.

pshahzeb: > I mean if they install the vsphere plugin and call it by another org's name that won't be right.

They can have a good enough reason to name the alias something else. May be as part of automated deployment itself. The way I look at alias is basically to provide the convenience to user to give a name of choice. Limiting it to vsphere is too tight.
Also we have different name for shared plugin. @luomiao 

Second issue I see is with the existing customers who have  aliased the plugin something else. As part of plugin upgrade, they ought to preserve the alias and they can't switch to vsphere. They'll have to uninstall and install it.

Given the limited alternatives we have, it boils to the question as:
Probability of a customer not using "/var/lib/docker " for runtime (this is by default) VS probability of customer not naming the alias to vsphere.
We can afford the one with lesser probability.

I have seen neither of such cases so I don't have a choice.

govint: In which case lets leave this as is and I'm going to close this as a won't fix. If user wishes to alias the plugin as _pivotal_ thats fine. As mentioned in the issue its not going to be common practice till a user does do it given Docker allows the daemon to run with a _-g_ option to locate the daemon's files.

Issue #1712, opened 2017-08-07
Support case-insensitive match for datastore names


venilnoronha:
Docker, by default, turns volume names to lower-case before providing it to the vDVS plugin on Windows. For example, the volume name in `docker volume create --driver=vsphere vol@vsanDatastore` is transformed to `vol@vsandatastore` in Docker API requests, causing a failure. Therefore, we should employ a first-match (or any other) mechanism on the VMDKOps Service and still enable volume creation on `vsanDatastore` for a better customer experience.

/cc @msterin

msterin:
To clarify - this issue is to support datastore names like vsanDatastore. if Docker passes us a volume name all lowcase  e.g. `myvolume@vsandatastore` our ESX-side code needs to:

* [ ] (low priority) know that this request is coming from Windows VM (we need to add something in the protocol). Generally, we can do the item below for all platforms since 2 datastores different only in letter case would requires customer manually messing it up, and it's super unlikeley
* [ ] (top priority) match `vsandatastore` with `vsanDatastore`. 
* [ ] (low priority) we can check there are no case-sensitive names matching the same, and issue a warning if there is (very low priority as it's a real bad practice anyways).

pshahzeb:
If we plan to go with top priority fix, i.e. match vsandatastore with vsanDatastore:

The fix would be to do a case insensitive match on ESX-side code.

The obvious follow-up question is how do we handle two datastores with same name but different cases mounted on ESX:
eg: vsandatastore and vsanDatastore are two different datastores mounted. This is allowed by vcenter.
> we can check there are no case-sensitive names matching the same, and issue a warning if there is (very low priority as it's a real bad practice anyways).

In such a case should we return an error ? We cannot issue just a warning because we need to make a choice out of the available datastores.
What do you guys think?
@msterin @lipingxue @pdhamdhere

shaominchen:
Here's my suggestion:
1. For GA release, we need a minimal fix that at lease makes sure the default "vsanDatastore" works for Windows. Instead of changing existing case-sensitive logic on ESX side, I'd suggest just do a hard-coded mapping from "vsandatastore" to "vsanDatastore". This quick fix (workaround actually) will be straightforward and most importantly, it's absolutely safe - it won't cause any regression at this point.
2. After GA, we should consider a right fix to take care of all comprehensive scenarios.
3. As for the scenario where two datastores with same name but different cases, as Mark said, this is really bad practice. I think we can just return an error in such case.

[...]

PR #1800, opened 2017-08-23
Support case-insensitive match for datastore names

shaominchen:
Fixes issue #1712 

Testing Done:
1. Create a volume with datastore name 'vsanDatastore':

C:\Users\root\go\src\github.com\vmware>docker volume create -d vsphere vol7@vsanDatastore
vol7@vsandatastore

2. Create a volume with datastore name 'sharedvmfs-0' when there's a same name but different case datastore 'sharedVmfs-0':

C:\Users\root\go\src\github.com\vmware>docker volume create -d vsphere vol9@sharedvmfs-0
Error response from daemon: create vol9@sharedvmfs-0: VolumeDriver.Create: "Found multiple datastores with same name (ignoring case difference): ['sharedvmfs-0', 'sharedVmfs-0']"

3. test-e2e-windows passed with camel-case datastore names

OK: 5 passed
--- PASS: Test (347.47s)
PASS


esx_service/vmdk_ops.py
```diff
@@ -837,12 +837,33 @@ def parse_vol_name(full_vol_name):
...
+
+        matching_datastores = find_matching_datastores(ds_name)
+
+        # Return error if more than one datastores found
+        if len(matching_datastores) > 1:
+            raise ValidationError("Found multiple datastores with same name (ignoring case difference): {0}".format(matching_datastores))
+
+        # Found exactly one match
+        if len(matching_datastores) == 1:
+            ds_name = matching_datastores[0]
```
msterin: why do you need this line ? I'd expect ds_name in this case to be exactly matching_datastore[0] .

esx_service/vmdk_ops.py
```diff
@@ -837,12 +837,33 @@ def parse_vol_name(full_vol_name):
...
 
     # Return qualified volume name and datastore name
     return vol_name, ds_name
 
+def find_matching_datastores(ds_name):
+    """
+    Returns a case-insensitive match for the given datastore name
+    """
+    matching_datastores = []
+    for name in get_datastore_names_list():
```
msterin: lines 861-864 in more python-ish way: 
```python
matching_datastores = [d for d in get_datastore_names_list() if d.lower() == ds_name.lower()]
```

esx_service/vmdk_ops.py
shuklanirdesh82: I guess to avoid falling into null pointer or empty value in case of `len(matching_datastores) == 0` when datastore name is not valid/notExist. e.g. `myvolume@IamNotExitDatastore`

esx_service/vmdk_ops.py
```diff
@@ -837,12 +837,33 @@ def parse_vol_name(full_vol_name):
...
+            raise ValidationError("Datastore name is too long (max len is {0})".format(MAX_DS_NAME_LEN))
+
+        matching_datastores = find_matching_datastores(ds_name)
+
+        # Return error if more than one datastores found
+        if len(matching_datastores) > 1:
+            raise ValidationError("Found multiple datastores with same name (ignoring case difference): {0}".format(matching_datastores))
+
+        # Found exactly one match
+        if len(matching_datastores) == 1:
```
pshahzeb: Shouldn't check for len(matching_datastores) == 0 be explicit?

esx_service/vmdk_ops.py
shaominchen: As @shuklanirdesh82 mentioned, matching_datastores might be empty if the given datastore name does not exist.

I agree the logic here looks a little bit weird, because we should have handled all cases, including len(matching_datastores) == 0. I'm not doing that because that case has already been taken care of in the follow-up validate_datastore() logic.

Ideally the logic here should be incorporated with validate_datastore, but that will need more refactoring, which I'd like to avoid at this point.

Hope this explains your question/confusion @msterin

esx_service/vmdk_ops.py
shaominchen: This has already been taken care of by the follow-up validate_datastore() logic. Please see my answers above.

esx_service/vmdk_ops.py
shaominchen: Sure, will follow your suggestion.

esx_service/vmdk_ops.py
msterin: my point was that if there is 1matching DS., then matching_datastores[0] is already guaranteed to contain ds_name.
So why would you need 'ds_name =  matching_datastores[0]` if they are identical at the moment ?

shaominchen: No, they may not be identical. They are identical on linux, but on windows, ds_name is always lower case (i.e. "vsandatastore"), but matching_datastores[0] can be "vsanDatastore".

That's the whole point of this fix actually.

msterin: Ah, so on Linux this is a redundant op, but on Windows it corrects the case. Makes sense, I missed it. Can you please add the comment ("on Linux this is a redundant op, but on Windows it corrects the case")? Lgtm then

shaominchen: Sure, will add some comments.

shuklanirdesh82: my 2 cents ... 

datastore name is case insensitive from vc/esx point of view. You are able to keep 2 datastore  (datastore 1 and datastore2) name as `localDatastore` & `LocalDatastore` any point of time no restriction to have such names for datastore. 

/CC @shaominchen @msterin

shaominchen: @shuklanirdesh82 Mark, Shahzeb and I have discussed this. We think reporting an error message is good enough to let the customers know that we don't support such a case. There's no need to add a known issue, since we don't plan to support such a bad practice.

esx_service/vmdk_ops.py
```diff
@@ -837,12 +837,32 @@ def parse_vol_name(full_vol_name):
             raise ValidationError("Volume name contains illegal characters: {0}".format(c))
 
     # Check if the datastore name is too long
-    if ds_name and len(ds_name) > MAX_DS_NAME_LEN:
-        raise ValidationError("Datastore name is too long (max len is {0})".format(MAX_DS_NAME_LEN))
+    if ds_name:
+        if len(ds_name) > MAX_DS_NAME_LEN:
+            raise ValidationError("Datastore name is too long (max len is {0})".format(MAX_DS_NAME_LEN))
+
+        matching_datastores = find_matching_datastores(ds_name)
```
msterin: nit (feel free to ignore): it's probably easier to say 
```python
    matching_datastores = [d for d in get_datastore_names_list() if d.lower() == ds_name.lower()]
``` 
and drop find_matching_datastores(ds_name) function all together.

shaominchen: Sure, will update.

shuklanirdesh82: +1 to bring it inline

tests/utils/verification/volumeproperties.go
```diff
@@ -81,7 +81,7 @@ func CheckVolumeListAvailability(hostName string, reqVolList []string) bool {
 	for _, name := range reqVolList {
 		name = strings.Replace(name, "'", "", -1)
 		name = strings.Replace(name, "\"", "", -1)
-		if strings.Contains(volumes, name) != true {
+		if strings.Contains(strings.ToLower(volumes), strings.ToLower(name)) != true {
```
pshahzeb: So even if we create a volume with a datastore eg. vol@vsanDatastore,
it is listed in docker cli volume list as vol@vsandatastore

We need to document this from the skylight's feature's point of view.

shaominchen: Agreed. Will document this difference.

PR #1803, opened 2017-08-23
Create dockvols only when required on specific datastores.

govint:
Added a new optional arg to get_vol_path to allow caller to decide whether the dockvols dir is created or not.  In the init_datastoreCache() path the dockvols dir isn't created anymore and instead created only as needed. Didn't remove the call to get_vol_path() in init_datastoreCache as we need the path to the dockvols  dir to be stored in the cache and the call to get_vol_path() is the only place that decides how to create that path.

### DIFF
--- a/esx_service/utils/vmdk_utils.py
+++ b/esx_service/utils/vmdk_utils.py
@@ -90,7 +90,7 @@ def init_datastoreCache(force=False):
         tmp_ds = []
 
         for datastore in ds_objects:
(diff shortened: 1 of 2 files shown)


govint:
Swarm test failed and looks like CI VMs need to be restarted.

shuklanirdesh82:
> Swarm test failed and looks like CI VMs need to be restarted.

It is fixed and this PR's run is restarted.

govint:
CI error due to swarm configuration.

PR #1824, opened 2017-08-26
Cutting release steps for windows [SKIP CI]

pshahzeb:
Steps to build the windows plugin
Files that need to be uploaded to bintray. Since windows plugin is an executable, we need to upload one powershell script and one zip file (which contains .exe and a .dll) to bintray


CONTRIBUTING.md
```diff
@@ -338,6 +338,19 @@ DOCKER_HUB_REPO=vmware EXTRA_TAG= VERSION_TAG=<version_tag> make all
 
 Update documentation following steps listed below.
```
shuklanirdesh82: Can you move this line below? Haven't realized while merging @luomiao change for vFile release step

Issue #1844, opened 2017-08-28
vFile plugin: volume create/rm stuck forever when ETCD is broken


luomiao:
**What happened:**
When the ETCD cluster went to unhealthy mode due to reasons like quorum loss, `docker volume create` and `docker volume rm` with vFile driver will stuck forever.

**What you expected to happen:**
vFile plugin should return error with proper message indicating the status of ETCD or swarm cluster is unhealthy.

luomiao:
Fixed by https://github.com/vmware/docker-volume-vsphere/pull/1850

PR #1850, opened 2017-08-29
vFile: include timeout for ETCD request

### LINKED ISSUE #1792: vShared plugin: ETCD error when some of the managers don't have plugin installed
When an ETCD operation is needed, the plugin will go over the list of managers and try to create an ETCD client.
Currently even for the addresses of managers who even don't have the plugin installed (not part of the ETCD cluster) can also return an ETCD client without an error.
This will result in the failures for the following ETCD operations:
```
2017-08-21 22:30:39.454301325 -0700 PDT [ERROR] Failed to call ETCD Get for listing all keys with prefix error="context deadline exceeded" prefix="SVOLS_stat_"
2017-08-21 22:30:39.45445512 -0700 PDT [ERROR] Failed to get volume list error="context deadline exceeded"
```

luomiao:
Fixes #1792 and #1844 

1. Add DialTimeout when creating new ETCD client. With DialTimeout, ETCD won't create a client unless the remote ETCD endpoint is accessible.
This can avoid problems when some of the swarm managers don't have the vFile plugin installed.

2. Add etcdRequestTimeout for all ETCD requests. Thus an ETCD request will return error when the ETCD endpoint is not accessible, instead of blocking wait, which can cause volume operations stuck forever.

Local tests passed.

### DIFF
--- a/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
+++ b/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
@@ -42,11 +42,13 @@ import (
                                cluster
    etcdClusterStateExisting:   Used to indicate that this node is joining
                                an existing etcd cluster
-   requestTimeout:             After how long should an etcd request timeout
+   etcdRequestTimeout:         After how long should an etcd request timeout
+   etcdUpdateTimeout:          Timeout for waiting etcd server change status
    checkSleepDuration:         How long to wait in any busy waiting situation
                                before checking again
    gcTicker:                   ticker for garbage collector to run a collection
    etcdClientCreateError:      Error indicating failure to create etcd client
+   swarmErrorMsg:              Message indicating swarm cluster is unhealthy
    etcdSingleRef:              if global refcount 0 -> 1, start SMB server
    etcdNoRef:                  if global refcount 1 -> 0, shut down SMB server
 */
@@ -58,10 +60,12 @@ const (
 	etcdScheme               = "http://"
 	etcdClusterStateNew      = "new"
 	etcdClusterStateExisting = "existing"
-	requestTimeout           = 5 * time.Second
+	etcdRequestTimeout       = 2 * time.Second
+	etcdUpdateTimeout        = 10 * time.Second
 	checkSleepDuration       = time.Second
 	gcTicker                 = 5 * time.Second
 	etcdClientCreateError    = "Failed to create etcd client"
+	swarmErrorMsg            = "Swarm cluster maybe unhealthy"
 	etcdSingleRef            = "1"
 	etcdNoRef                = "0"
 )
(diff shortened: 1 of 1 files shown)


client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -42,11 +42,13 @@ import (
...
    etcdClusterStateExisting:   Used to indicate that this node is joining
                                an existing etcd cluster
-   requestTimeout:             After how long should an etcd request timeout
+   etcdRequestTimeout:         After how long should an etcd request timeout
+   etcdUpdateTimeout:          Timeout for waiting etcd server change status
    checkSleepDuration:         How long to wait in any busy waiting situation
                                before checking again
    gcTicker:                   ticker for garbage collector to run a collection
    etcdClientCreateError:      Error indicating failure to create etcd client
+   swarmErrorMsg:              Message indicating swarm cluster is unhealthy
```
lipingxue: nit: better use name "swarmUnhealthyErrorMsg".

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -58,10 +60,12 @@ const (
 	etcdScheme               = "http://"
 	etcdClusterStateNew      = "new"
 	etcdClusterStateExisting = "existing"
-	requestTimeout           = 5 * time.Second
+	etcdRequestTimeout       = 2 * time.Second
```
lipingxue: How do you know the timeout value used here is the right one?

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -670,15 +687,12 @@ func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 	s := strings.Split(addr, ":")
 	endpoint := s[0] + etcdClientPort
 	cfg := etcdClient.Config{
-		Endpoints: []string{endpoint},
+		Endpoints:   []string{endpoint},
+		DialTimeout: etcdRequestTimeout,
 	}
 
 	etcd, err := etcdClient.New(cfg)
 	if err != nil {
```
lipingxue: Why the error log is removed?

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
luomiao: So the etcdUpdateTimeout is in fact the old requestTimeout now.
In the previous code, requestTimeout or 2 * requestTimeout is used for waiting status change of ETCD server, or waiting for a key-value inside ETCD getting updated by other managers. So it's basically keeping the old value.
And the new etcdRequestTimeout is timeout for waiting an ETCD operation to be done. I checked the example code in ETCD codebase and 2 seconds are used for requests timeout.

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
luomiao: Because this function is called by the garbage collector and List function too.
When there is some of the swarm managers don't have plugin installed (no ETCD on it), which is still a valid environment for vFile, this message will pollute the logs. However, it's not really necessary. This is because a plugin will loop all the managers to find a valid one with plugin installed. We only need the error log when none of the managers are valid, which is currently printed out at: https://github.com/vmware/docker-volume-vsphere/pull/1850/files/a7f1b6a45f7bf37d664018534cfed073fc3d2e24#diff-ac388eb52d1ec58c6967185c9f547fe2R674

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
luomiao: will update.

luomiao:
@lipingxue 
I reproduced the ETCD errors in #1792 and #1844 and make sure the PR resolved those errors when 1) managers without vFile plugin installed existing in the swarm; 2) too many managers left the cluster which results in quorum loss situation.

Issue #1858, opened 2017-08-31
Optimize vmdkops_admin.get_policy() function


shaominchen:
The command "vmdkops_admin volume ls" took around 28 seconds against a 500 volume setup. One of the major bottlenecks is the function vmdkops_admin.get_policy(), which took about 20 seconds to complete. We need to optimize this function to get rid of the biggest bottleneck.

Please refer to https://github.com/vmware/docker-volume-vsphere/issues/1083 for detailed performance analysis.

Issue #1083, opened 2017-03-24
[Improvement] Admin cli volume ls is slow to list all the volumes if there are large number of volumes


ashahi1:
```Admin cli volume ls``` is very slow to list all the volumes if the setup has lots of volumes. 

I ran the test multiple times and saw slow behavior across all the runs. I tested with 500 volumes. 

Test Steps:
```
1. Create 500 volumes of 100 mb each sequentially.
2. Do docker volume ls and admin cli volume ls.
3. Remove all the 500 volumes.
```

Setup:
```
Setup:

4 ESXs - 364 VMs - 4 shared Vmfs datastores
Each shared vmfs has ~90 VMs.
All the VMs are powered-on.

```

Details of each run is as follows:

Run-1:
```
Individual volume creation of 100 mb : 2-3 seconds (mostly 3 and sometimes even 4)
500 Volumes creation : ~27 mins (1642 seconds)
admin cli volume ls : 32 secs
500 Volumes removal : 23 mins
Individual volume removal of 100 mb : 2-3 seconds (mostly 2)
```

Run-2:
```
Individual volume creation of 100 mb : 2-3 seconds (mostly 3)
500 Volumes creation : ~26 mins (1564 seconds)
admin cli volume ls : ~27 secs
500 Volumes removal : ~23 mins
Individual volume removal of 100 mb : ~2 seconds
```

Run-3:
```
Individual volume creation of 100 mb : 2-3 seconds (mostly 3)
500 Volumes creation : ~26 mins (1602 seconds )
admin cli volume ls : 27 secs
500 Volumes removal : ~23 mins (1409 seconds)
Individual volume removal of 100 mb : ~2 seconds
```

Run-4:
```
Individual volume creation of 100 mb : mostly 3 seconds
500 Volumes creation : ~26 mins (1560 seconds)
admin cli volume ls : 27 secs
500 Volumes removal : ~23 mins (1421 seconds)
Individual volume removal of 100 mb : ~2 seconds
```

Run-5:
```
Individual volume creation of 100 mb : mostly 3 seconds
500 Volumes creation : ~25 mins (1558 seconds)
docker cli volume ls : ~3 seconds
admin cli volume ls : ~25 seconds
500 Volumes removal : ~23 mins (1412 seconds)
Individual volume removal of 100 mb : mostly 2 seconds
```

Run-6:
```
Individual volume creation of 100 mb : mostly 3 seconds
[...]

Issue #1864, opened 2017-09-05
[CI] Failover test for swarm failed as vDVS was not running on the worker node after vm power-on


ashahi1:
CI: https://ci.vmware.run/vmware/docker-volume-vsphere/1569

Plugin and driver logs have been attached at the end.

Test log:
```
2017/09/05 18:56:35 vDVS is not running on VM: 192.168.31.85
2017/09/05 18:56:35 vDVS is not running on VM [Ubuntu.16.10]. Please check if the previous VM IP [192.168.31.85] has changed.
2017/09/05 18:56:35 Destroying volume [swarm_test_volume_194316]

----------------------------------------------------------------------
FAIL: swarm_test.go:73: SwarmTestSuite.TearDownTest

swarm_test.go:76:
    c.Assert(err, IsNil, Commentf(out))
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc4201ad8a0), Stderr:[]uint8(nil)} ("exit status 1")
... Error response from daemon: unable to remove volume: remove swarm_test_volume_194316: VolumeDriver.Remove: Failed to remove volume swarm_test_volume_194316, in use by VM = Ubuntu.14.04.


----------------------------------------------------------------------
PANIC: swarm_test.go:96: SwarmTestSuite.TestFailoverAcrossSwarmNodes

swarm_test.go:164:
    c.Assert(isVDVSRunning, Equals, true, Commentf("vDVS is not running after VM [%s] being restarted", hostName))
... obtained bool = false
... expected bool = true
... vDVS is not running after VM [Ubuntu.16.10] being restarted
```

Logs:
[docker-volume-vsphere.txt](https://github.com/vmware/docker-volume-vsphere/files/1278717/docker-volume-vsphere.txt)
[vmdk_ops.txt](https://github.com/vmware/docker-volume-vsphere/files/1278718/vmdk_ops.txt)

ashahi1:
@shaominchen Can you please take a look?

shaominchen:
Anup and I debugged this issue. After VM was powered off and then powered on, it took about 2.5 minutes for the vDVS plugin to get up and running. Here are the plugin logs:

```
2017-09-05 21:50:16.821252145 +0000 UTC [WARNING] Received signal signal=terminated 
2017-09-05 21:52:34.103082066 +0000 UTC [INFO] No config file found. Using defaults.
[...]

Issue #1859, opened 2017-08-31
Optimize vmdkops_admin.get_metadata() function


shaominchen:
The command "vmdkops_admin volume ls" took around 28 seconds against a 500 volume setup. One of the major bottlenecks is the function vmdkops_admin.get_metadata(), which took about 6 seconds to complete. We need to optimize this function to get rid of the bottleneck.

Please refer to #1083 for detailed performance analysis.

shuklanirdesh82:
Performance enhancement should be a `mustfix`

PR #1867, opened 2017-09-05
Optimize vmdkops_admin ls operation

shaominchen:
Fixes issue #1858: Optimize vmdkops_admin.get_policy() function. The fix is to cache vSAN datastore managed object. The vSAN datastore should be preconfigured, so it doesn't make sense to fetch it every time, which is extremely time consuming. Please refer to the last 2 comments in issue #1083 for more detailed analysis.

Here's a comparison before and after the fix being applied (time in seconds):

| Volume Number |   Before Fix    |    After Fix     |
| ---------------- | ------------- | ------------- |
|        100              |      9.854        |       2.996       |
|        200              |     14.053       |       4.743       |
|        400              |     23.991       |       8.554       |
|        500              |     28.787      |       10.493      |
|        600              |     36.201       |       12.267      |
|        800              |     47.249      |       16.886      |
|        1000            |      57.168       |      21.483      |

As we can see, the performance was improved significantly. Even though for very large number of volumes (>1000), it's still a bit slow, but at least it's reasonably acceptable now if volume number is less than 500.


shaominchen:
@msterin Thanks for the review! To answer your questions:
1. Yes, it works. I verified with these steps:
    1) Create a vol at vsanDatastore
    2) Run "docker volume ls" and "vmdkops_admin volume ls"
    3) Rename vsanDatastore to vsanDS
    4) Run "docker volume ls" and "vmdkops_admin volume ls"
    5) Try to create a vol at vsanDatastore - fail
    All above steps got expected result.
2. I don't have the environment right now, will check if I can set up one to verify that.
3. Yes - issue #1859. Please see my last 2 comments in issue #1083 for more detailed analysis.
4. Two reasons:
    1) Logically I feel it's more natural (and clean) to patch in is_on_vsan()
    2) The optimization effect is better to patch in is_on_vsan(). If we patch in get_vsan_datastore - let's say cache "si" and/or "stores" instance, the remaining logic is still a unnecessary waste, even if it's trivial.

shaominchen:
BTW, CI failure is due to issue #1864.

esx_service/utils/vsan_info.py
```diff
@@ -56,11 +58,13 @@ def get_vsan_dockvols_path():
 
 def is_on_vsan(vmdk_path):
     """Returns True if path is on VSAN datastore, False otherwise"""
-    ds = get_vsan_datastore()
-    if not ds:
-        return False
+    global _vsan_ds
+    if not _vsan_ds:
+        _vsan_ds = get_vsan_datastore()
```
govint: When the service starts it calls init_datastoreCache(), can't the datastore type be be cached in that map as well in addition to the name, URL and dockvols path? This code path of is_vsan_datastore() seems to be extra when we have a DS cache already.

shaominchen:
@govint Thanks for the review! Yes, this issue get reproduced on VMFS as well. Actually it doesn't matter if the volumes are created on VMFS or VSAN datastore. The logic get_policy/is_on_vsan/get_vsan_datastore is always invoked as part of generate_ls_rows workflow.

esx_service/utils/vsan_info.py
shaominchen: I agree it's possible to improve&reuse existing datastore cache. Some minor concerns:
1. Adding a new entry to existing datastore cache might break existing code paths, so we need thorough regression test
2. The existing datastore cache is not very well encapsulated. The init_datastoreCache() is being called in multiple places. Ideally this should be avoided.

To reuse this cache, I will need to refactor existing implementation to improve the encapsulation and make it more robust. Otherwise I will have to call init_datastoreCache() in one more code path, which is really something I want to avoid.

Both are not critical concerns though. But it just means much more work to do. Considering that the current vsan_info.py is a module specifically for VSAN, I feel it's acceptable to maintain this minor cache in this module. So personally I don't have strong motivation to switch to the global datastore cache right now. Let me know what you think.

shaominchen:
@msterin Regarding your 2nd question, what I can do it to add a new VSAN cluster to get the 2nd VSAN datastore. However, I can't add the same host to two different VSAN clusters. So for the docker hosts (VMs) in one host, I can only create volumes on the attached VSAN datastore only. The other VSAN datastore which belongs to a different cluster is not visible/accessible to the docker hosts (VMs) in the current VSAN cluster.

Issue #1865, opened 2017-09-05
Rearrange markdown files


shuklanirdesh82:
Reference: https://github.com/vmware/docker-volume-vsphere/issues/1370#issue-234384162

one action item is left to streamline our documentation effort.

Our existing markdown files should be organized into `docs/external` & `docs/internal` folder.

- [x] 1.  `docs/external`: docs/external/_all_stuff_we_need_for_doc_site_including_actual_content. 
- [x] 2. `docs/internal`: We can keep docs/internal for specs and the likes
- [x] 3. Update references in markdown files after rearranging files
- [x] 4. Update automated script to read files from `docs/external`
- [x] 5. Need to update `docs/README.md` file to reflect latest information.

PR #1871, opened 2017-09-07
[SKIP CI] Organizing markdown files into docs/external & docs/internal folder

ashahi1:
This PR addresses first two issues mentioned in #1865

```
Organizing markdown files into docs/external & docs/internal folder.
 1. docs/external: docs/external/_all_stuff_we_need_for_doc_site_including_actual_content.
 2. docs/internal: We can keep docs/internal for specs and the likes
```


docs/internal/tenancy.md
```diff
@@ -0,0 +1,139 @@
+---
```
shuklanirdesh82: can it be deleted?

Issue #1816, opened 2017-08-25
Failed to mount vFile volume. Error: exit status 255 on Swarm master


pdhamdhere:
Setup: 3 node swarm cluster
Steps:
1. Try to start container using shared volume on *master*
2. Mount fails with `Failed to mount vFile volume. Error: exit status 255`

```
root@u-1:~# docker run --rm -it -v doesitwork2:/x busybox
docker: Error response from daemon: error while mounting volume '/mnt/vfile/doesitwork2/': VolumeDriver.Mount: Failed to mount vFile volume. Error: exit status 255.
root@u-1:~# vi /var/log/vfile.log 
root@u-1:~# docker volume inspect doesitwork2
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/doesitwork2/",
        "Name": "doesitwork2",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Ready"
        }
    }
]
root@u-1:~# ip addr
...
    inet 172.16.111.131/24 brd 172.16.111.255 scope global ens160
```

From vfile.log
```
2017-08-25 21:11:25.936768801 +0000 UTC [INFO] Checking status of file server container...
2017-08-25 21:11:25.942214739 +0000 UTC [WARNING] No service returned with name doesitwork2.
2017-08-25 21:11:25.972762754 +0000 UTC [INFO] Updating port and file service name for doesitwork2
2017-08-25 21:12:21.717058057 +0000 UTC [INFO] VolumeDriver Get: doesitwork2
2017-08-25 21:12:21.95763916 +0000 UTC [INFO] Mounting volume name=doesitwork2
2017-08-25 21:12:22.012096338 +0000 UTC [INFO] Before AtomicIncr
2017-08-25 21:12:23.060467684 +0000 UTC [INFO] Watcher on global refcount returns event type=PUT
2017-08-25 21:12:24.076707491 +0000 UTC [INFO] Attempting to change volume state to Mounting
2017-08-25 21:12:29.126624469 +0000 UTC [INFO] Checking status of file server container...
2017-08-25 21:12:29.155411432 +0000 UTC [INFO] Updating port and file service name for doesitwork2
2017-08-25 21:12:30.094439552 +0000 UTC [INFO] Volume state mounted, prepare to mounting locally
[...]

PR #1886, opened 2017-09-14
More code and script change to enable vFile CI

luomiao:
@ashahi1 
Please generate the key. Thanks!

Update:
Due to the slowness of the testbed, CI testing exposed a few problems in the code.
Besides adding scripts change, this PR also includes several fixes:
1. Change the waiting for file server path mounting from one time waiting to a timeout waiting. -> Fixes: #1941 #1816 
2. When some of the nodes in the swarm network failed/rejected for an assigned docker service task, the status of this task will stay in task lists if polling information according to service ID. Current code will return false as long as one task is failed/rejected. However the correct handling should return true as long as one task is running, since we only need one running container for the file server service.  Fixes: #1911


ashahi1:
CI run has started with the new key.

.drone.yml
```diff
@@ -49,6 +49,7 @@ build:
       - go get github.com/golang/lint/golint
       - go get -u gopkg.in/check.v1
       - make -s build
+      - make -s build-vfile
```
shuklanirdesh82: it is better to include the target into the `build` target it self

luomiao: If we include vFile also into build target, all the following development for vDVS will suffer waiting for building vfile too. The build of vfile and vdvs are totally separate from each other. So I think we should have them as different build target.

shuklanirdesh82: building vFile binary wont take much time it is still OK to build binary as part of build having it said building and pushing plugin should be a separate target and I think it is there. Am I missing anything here?

luomiao:
@shuklanirdesh82 
The updated PR has addressed Nirdesh's comment to include vfile as part of build target.
Since Nirdesh is PTO now, I will just dismiss this review since it's already cleared.

luomiao:
@lipingxue 
since there are some changes to vfile code too, could you also please review those changes?
Thanks.

lipingxue:
@luomiao OK. I will review it.

PR #1887, opened 2017-09-15
Add force option to removing a vmgroup.

govint:
Added a --force option for the _vmgroup rm_ command. So, if a VM is say deleted or moved off the host then its still going to be possible to remove the vmgroup on the _old_host. The VMs in the vmgroup get _attached_ to the default vmgroup if the --force option is used to remove the vms. All other options remain as is.

Testing:

vim-cmd solo/register $PWD/dum/dum.vmx dum
25

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup create --name dummyvg --default-datastore=sharedVmfs-1 --vm-list=dum

vmgroup 'dummyvg' is created. Do not forget to run 'vmgroup vm add' to add vm to vmgroup.

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup ls
------------------------------------  --------  -------------------------  -----------------  -------------
5e32424c-9e1d-42c7-881d-e8cc2e99c18c  ds-ren                               sharedVmfs-0       worker1-VM1.0
9ff32c5e-b76b-4331-b712-91c33522526d  dummyvg                              sharedVmfs-1       dum

vim-cmd vmsvc/unregister 25

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup rm --name=dummyvg
ERROR:Cannot delete non-empty vmgroup dummyvg. Remove VMs from the vmgroup before deleting it.

/usr/lib/vmware/vmdkops/bin/vmdkops_admin.py vmgroup rm --name=dummyvg --force
vmgroup rm succeeded


esx_service/utils/auth_api.py
```diff
@@ -569,15 +569,21 @@ def _tenant_rm(name, remove_volumes=False):
...
     if error_info:
         return error_info
 
+    if tenant.vms:
+        if not force:
+            error_info = generate_error_info(ErrorCode.TENANT_NOT_EMPTY, name)
+            logging.error(error_info.msg)
+            return error_info
+        else:
+            error_msg = tenant.remove_vms(auth_mgr.conn, tenant.vms)
```
shaominchen: Isn't this call going to return some error due to "Cannot find vm"?

govint: This call is the final step to removing the VM from the vmgroup, the vms list is returned from _get_tenant()_ which returns all the VMs associated with the tenant as (vm_id, vm_name).  tenant.remove_vms() only removes the vms from the DB, any error at that point is treated as _internal error_. No checks are performed on the vm IDs,

shaominchen: I still don't follow. My question is: if the VMs were already deleted, doesn't the call tenant.remove_vms() return some error which will again fail the "vmgroup rm" operation?

shaominchen: As per offline discussion, in this case the VMs are removed from VC, but still exist in vDVS database. So above tenant.remove_vms() call will not fail.

PR #1888, opened 2017-09-15
Fix issue 1073: 'vmdk-opsd stop' needs to drain operations in flight

shaominchen:
The basic idea of the solution is:

1. On 'stop' requests, add a barrier so executeRequest rejects all new incoming requests
2. Add count of ops in flight, and when barrier is enabled make sure the count drops to zero before proceeding
3. Rename today's "stop" to "shutdown" - keeping it as a way to force-kill the service in case something is stuck  (e.g. a 10TB format is running)
4. Add log messages for "stop" that the "service is being stopped, waiting for completion of X operations in flight "

Testing Done (by adding sleep time in the ops):
1. Showing different command options:
```
[root@sc2-rdops-vm06-dhcp-197-205:/etc/init.d] ./vmdk-opsd
./vmdk-opsd start|stop|restart|status|kill
```
2. Stop the service when there's no in-flight ops:

- Console output:
```
[root@sc2-rdops-vm06-dhcp-197-205:/etc/init.d] ./vmdk-opsd stop
Stopping vmdkops-opsd with PID=8396316
8396340
Waiting for operations in flight to complete, please do not interrupt...
vmdkops-opsd is not running
```

- Log output:
```
09/27/17 18:50:05 8396316 [MainThread] [WARNING] Received signal num: 15 - exiting
09/27/17 18:50:05 8396316 [MainThread] [INFO   ] No in-flight operations. Exiting...
09/27/17 18:50:05 8396316 [Dummy-2] [WARNING] Received signal num: 15 - exiting
09/27/17 18:50:05 8396316 [VMChangeListener] [INFO   ] VMChangeListener thread exiting
09/27/17 18:50:05 8396316 [Dummy-2] [INFO   ] No in-flight operations. Exiting...
```
3. Stop the service when there's in-flight ops:

- Console output:
```
[root@sc2-rdops-vm06-dhcp-197-205:/etc/init.d] ./vmdk-opsd stop
Stopping vmdkops-opsd with PID=8425834
8425858
8426066
Waiting for operations in flight to complete, please do not interrupt...
Waiting for operations in flight to complete, please do not interrupt...
Waiting for operations in flight to complete, please do not interrupt...
Waiting for operations in flight to complete, please do not interrupt...
Waiting for operations in flight to complete, please do not interrupt...
vmdkops-opsd is not running
```

- Log output:
```
09/27/17 21:12:17 8425834 [MainThread] [WARNING] Received signal num: 15 - exiting
09/27/17 21:12:17 8425834 [MainThread] [DEBUG  ] Setting stop barrier to true
09/27/17 21:12:17 8425834 [MainThread] [INFO   ] Started new thread : 842593679104 with target <function wait_ops_in_flight at 0xc42cba4e18> and args None
09/27/17 21:12:17 8425834 [MainThread] [DEBUG  ] Currently active threads: [<_MainThread(MainThread, started 842477255456)>, <Thread(VMChangeListener, started daemon 842564798208)>, <Thread(Thread-3, started 842593679104)>, <Thread(ubuntu-VM0.2-nolock-list, started 842565326592)>]
09/27/17 21:12:17 8425834 [MainThread] [DEBUG  ] lib.vmci_get_one_op returns -1, buffer 'b'{"cmd":"list","details":{"Name":""},"version":"2"}''
09/27/17 21:12:17 8425834 [MainThread] [WARNING] vmci_get_one_op failed ret=-1: Invalid argument (errno=22) Retrying...
09/27/17 21:12:17 8425834 [MainThread] [DEBUG  ] lib.vmci_get_one_op: waiting for new request...
09/27/17 21:12:27 8425834 [ubuntu-VM0.2-nolock-list] [DEBUG  ] After decr: counter.value=0
09/27/17 21:12:27 8425834 [ubuntu-VM0.2-nolock-list] [DEBUG  ] Setting counter event
09/27/17 21:12:27 8425834 [Thread-3] [INFO   ] All in-flight operations are completed. Exiting...
```
4. Kill the service
```
[root@sc2-rdops-vm06-dhcp-197-205:/etc/init.d] ./vmdk-opsd kill
Killing vmdkops-opsd with PID=7942958
7942981
vmdkops-opsd is not running
```
5. Verified that on 'stop' request, new incoming requests will be rejected:
- Console output:
```
root@sc-rdops-vm02-dhcp-52-237:~# docker volume create -d vsphere v3
Error response from daemon: create v3: VolumeDriver.Create: Service is being stopped: operation declined!
```
- Log output:
```
09/27/17 21:15:10 8393750 [MainThread] [DEBUG  ] lib.vmci_get_one_op: waiting for new request...
09/27/17 21:15:10 8393750 [MainThread] [DEBUG  ] lib.vmci_get_one_op returns 9, buffer 'b'{"cmd":"create","details":{"Name":"v3","Opts":{"fstype":"ext4"}},"version":"2"}''
09/27/17 21:15:10 8393750 [MainThread] [WARNING] Service is being stopped: operation declined!
09/27/17 21:15:10 8393750 [MainThread] [DEBUG  ] lib.vmci_reply: VMCI replied with errcode 0
09/27/17 21:15:10 8393750 [MainThread] [DEBUG  ] lib.vmci_get_one_op: waiting for new request...
```
6. Verified that if service fails to stop, new incoming requests will be rejected


msterin:
I am not quite sure.what the current logic is expected to do - it seems to be starting a new (unrelated)  python interpreter and won't impact vmdk_opsd.oy
Simply send some.signa (i.e. USER or anything else ) to running daemon, and have signal handler there doing the drain and stop.

shaominchen:
@msterin Thanks for taking a look! Yes, as you said, the current logic is just starting a new (unrelated) python interpreter and won't impact vmdk_opsd.oy, so it not working as expected.

I will do some investigation about the signal mechanism. Meanwhile if you can provide some more information such as samples, that will definitely help.

msterin:
Pllllllll

msterin:
Take a look at signal.signal in vmdk_ops.py and related kill -SIGTERM in vmdk_opsd. Pick up another signal (e.g. USER) and do the same

shaominchen:
@msterin We can use signal to send a stop message to vmdk_ops.py, but how can we get the count of ops in flight from vmdk_ops.py? Python signal API supports one-way communication only, correct?

msterin:
You are correct. My suggestion is to print the info about ops in flight count in the vmdkops log only, and in the shell script just say "waiting for ops in flight to complete"

shaominchen:
OK. That way the admincli stop command is not very user friendly, because it will get stuck for some time (until all ops in flight complete or timeout) without any continuous feedback to the user. Anyway considering this is a rare corner case, I think this is acceptable.

msterin:
> OK. That way the admincli stop command is not very user friendly, because it will get stuck for some time (until all ops in flight complete or timeout) without any continuous feedback to the user. Anyway considering this is a rare corner case, I think this is acceptable.

Agreed. As long as the message is clear (`Waiting for operations in flight to complete, please do not interrupt...` or something like that, IMO

govint:
As @msterin indicated all that needs to be done is to have a signal trapped in the ESX service which will a) decline all new requests and b) cause the service to exit once all current ops are finished and c) service tracks in flight commands as a counter and uses that to track when to exit (if thats requested).

esx_service/vmdk-opsd
```diff
@@ -150,7 +184,7 @@ check_running_status () {
    fi
 
    echo "${OPSD_TAG} is not running"
-   return 6
+   return 5
```
lipingxue: Why change the error code from "6" to "5"?

esx_service/vmdk-opsd
```diff
@@ -112,32 +113,65 @@ stop() {
...
    fi
 
+   cleanup_resource
+   if [ $? -ne 0 ]; then
+      exit $?
+   fi
+
+   check_running_status
+   if [ $? -eq 0 ]; then
+      echo "Failed to stop $OPSD_TAG. Please try shutdown to force kill the process"
```
msterin: nit: `shutdown` feels like a normal shutdown (i.e. with draining in flight ops.
The verb should be ""shutdown" (or "stop") for normal op. and "kill" or "forcestop" for forced op

esx_service/vmdk-opsd
msterin: why 5 and not 6 ? I think these numbers should be constants, but if you use literals - why change value ?

esx_service/vmdk-opsd
```diff
@@ -165,6 +199,9 @@ case $1 in
    "stop")
       stop
       ;;
+   "shutdown")
```
msterin: "kill" ?

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,39 @@ def set_vol_opts(name, tenant_name, options):
...
+    # Set the STOP flag to true
+    stopEvent.set()
+
+    # Wait until the in-flight ops counter goes down to zero
+    while ((opsCounter.value > 0) and (STOP_RETRY_COUNT > 0)):
+        logging.info("Service is being stopped. Waiting for completion of %d operations in flight...", opsCounter.value)
+        time.sleep(STOP_RETRY_SLEEP)
+        STOP_RETRY_COUNT -= 1
+
+    if (opsCounter.value == 0):
```
msterin: how you can guarantee that one it goes to 0 and returns 0 here, nobody starts a new op between now and the line 1699?

esx_service/vmdk_ops.py
```diff
@@ -1757,6 +1791,10 @@ def execRequestThread(client_socket, cartel, request):
         reply_string = err("Server returned an error: {0}".format(repr(ex_thr)))
         send_vmci_reply(client_socket, reply_string)
 
+    # Decrease the in-flight ops counter
+    opsCounter.decr()
+    logging.debug("execRequestThread: opsCounter=%d", opsCounter.value)
```
msterin: the value can change between lines 1795 and 796 if there is any other thread touching it

esx_service/vmdk_ops.py
```diff
@@ -1698,6 +1728,10 @@ def send_vmci_reply(client_socket, reply_string):
                         os.strerror(errno), errno)
 
 def execRequestThread(client_socket, cartel, request):
+    # Increase the in-flight ops counter
+    opsCounter.incr()
+    logging.debug("execRequestThread: opsCounter=%d", opsCounter.value)
```
msterin: the value can change between incr() and logging. You may want to print it inside of counter() code, before releasing the lock

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,39 @@ def set_vol_opts(name, tenant_name, options):
...
+def wait_ops_in_flight():
+    global STOP_RETRY_COUNT
+
+    # Set the STOP flag to true
+    stopEvent.set()
+
+    # Wait until the in-flight ops counter goes down to zero
+    while ((opsCounter.value > 0) and (STOP_RETRY_COUNT > 0)):
+        logging.info("Service is being stopped. Waiting for completion of %d operations in flight...", opsCounter.value)
+        time.sleep(STOP_RETRY_SLEEP)
```
msterin: wait_ops_in_flight() is a signal handler and will be executed on the main thread. 
Which means main thread will be blocked until you exit from `wait_ops_in_flight`, so I expect ALL volume-related docker commands to hang once you started the drain.

In the other words, I do not believe this could work. You should set the flag and spawn the thread for waiting, and exit on thread exit. Or you can do it fully async and get notified (semaphor or conditional) when the counter changes. 
But you should not block in main thread.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1824,16 @@ def handleVmciRequests(port):
...
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the flag and the counter - exit the loop if flag is set and counter equals 0
+        if (stopEvent.isSet() and (opsCounter.value == 0)):
+            logging.info("Stopping service: no operations in flight.")
+            break;
+
+        # Listening on VMCI socket
```
msterin: so if there is an operation in flight, you fall through to like 1835 and start waiting on vmci. Until vmci does not return,  the code is stuck and will not exit. Seems like a bug to me

esx_service/vmdk-opsd
shaominchen: Here's the previous error codes:
```
1 - failed to start (already running)
2 - failed to create resource pool
3 - failed to start (error happens)
4 - failed to delete resource pool
5 - failed to stop vmdkops
6 - vmdkops not running
```
After introducing 'shutdown' option, we need a new error code for "failed to shut down vmdkops". I can use 7 for that, but I feel we should use two consecutive code for these 2 error conditions, so I changed it like this:
```
1 - 4 (same as above)
5 - vmdkops not running
6 - failed to stop vmdkops
7 - failed to shut down vmdkops
```
I don't think these error code are really being used anywhere. If there are concerns about backward compatibility, I can change it back.

msterin: FWIW I have no concerns, I just did not see it matters so I was surprised by the change :-)

esx_service/vmdk-opsd
shaominchen: "kill" does not work - the logic will get stuck in an infinite nesting because the linux system call kill will be invoked. I can use 'forcestop' if you wish.

esx_service/vmdk-opsd
shaominchen: @lipingxue asked the same question. Please see my comments above, and let me know your opinion.

esx_service/vmdk-opsd
shaominchen: Same as above.

esx_service/vmdk-opsd
msterin: why ? I am confused. This is just operation name (string) and it does not even have to be function name

shaominchen: I think it's due to the following statement:

kill -${SIGKILL} ${PID}

esx_service/vmdk-opsd
```diff
@@ -112,32 +113,65 @@ stop() {
    if [ -n "${PID}" ]; then
       echo "Stopping $OPSD_TAG with PID=${PID}"
       kill -$SIGTERM ${PID}
+
       local COUNT=0
       while kill -0 ${PID} > /dev/null 2>&1; do
```
govint: use ${SIGTERM} instead of 0

esx_service/vmdk_ops.py
shaominchen: Do you mean all the ongoing logic in the main thread will be blocked when a signal handler is executed (for us, it mean the "while True" logic in handleVmciRequests() will be blocked by this)? I didn't see this description in Python signal documentation though.

esx_service/vmdk-opsd
```diff
@@ -112,32 +113,65 @@ stop() {
...
          sleep ${WAIT_KILL_SEC};
+         echo "Waiting for operations in flight to complete, please do not interrupt..."
          if [ $((COUNT++)) -gt ${NUM_RETRY} ]; then
-            echo "Sending kill -${SIGKILL} to ${PID} after kill -${SIGTERM} did not stop ${PID}"
-            kill -${SIGKILL} ${PID}
+            echo "Kill -${SIGTERM} did not stop ${PID}"
```
govint: Could use a modified version of the original message - Something like "failed to terminate the service after N retries, killing service..."

esx_service/vmdk_ops.py
shaominchen: As @pshahzeb mentioned, all new incoming requests should be rejected by line 1857. Basically, once the flag is set, the counter should just go down, and never go up.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,39 @@ def set_vol_opts(name, tenant_name, options):
...
+    # Wait until the in-flight ops counter goes down to zero
+    while ((opsCounter.value > 0) and (STOP_RETRY_COUNT > 0)):
+        logging.info("Service is being stopped. Waiting for completion of %d operations in flight...", opsCounter.value)
+        time.sleep(STOP_RETRY_SLEEP)
+        STOP_RETRY_COUNT -= 1
+
+    if (opsCounter.value == 0):
+        logging.info("All in-flight operations are completed. Exiting...")
+    else:
+        logging.info("In-flight operations are taking too long to complete")
```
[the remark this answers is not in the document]
shaominchen: This is just log message. Line 135 in vmdk-opsd will take care of what you said.

esx_service/vmdk-opsd
```diff
@@ -112,32 +113,65 @@ stop() {
...
             break
```
govint: Why remove the kill, rather than have the user issue multiple commands just to kill a process, better do it here. When asked to stop the service is allowed to drain commands and when that doesn't complete the process is killed - which should be rare.

Basically, the user wants to stop the process - the user  isn't going to debug the process anyway. So stop should do the job. User will most likely do a ps -c and grep the PID and kill it if one command doesn't handle.

esx_service/vmdk_ops.py
shaominchen: Agreed. Will fix.

esx_service/vmdk_ops.py
shaominchen: Will fix.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,39 @@ def set_vol_opts(name, tenant_name, options):
...
+
 def msg_about_signal(signalnum, details="exiting"):
     logging.warn("Received signal num: %d, %s.", signalnum, details)
```
govint: Log should also mention that new requests are stopped and waiting for in-flight commands to drain.

esx_service/vmdk_ops.py
shaominchen: Agree. Looks like we should check stopEvent.isSet() only and remove the check (opsCounter.value == 0). But I guess we should wait until all in-flight ops done before invoking vmci_release_listening_socket() at line 1874? @msterin

esx_service/vmdk-opsd
shaominchen: This is just to check if the process with the given PID is running, no signal is really being sent.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1824,16 @@ def handleVmciRequests(port):
     cartel = c_int32()
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the flag and the counter - exit the loop if flag is set and counter equals 0
+        if (stopEvent.isSet() and (opsCounter.value == 0)):
```
govint: Why check for value equal to zero, if stop event is set then this loop should exit and wait for the counter to get to zero and exit after that.

esx_service/vmdk_ops.py
```diff
@@ -1810,6 +1853,13 @@ def handleVmciRequests(port):
 
         client_socket = c # Bind to avoid race conditions.
 
+        # Check the flag - if set, fail new incoming requests
+        if stopEvent.isSet():
```
govint: Can't the clients simply timeout and fail themselves? Thats the error the client will anyway get if the service was successfully stopped (when this loop exited) and the client issued a command after that. This handling of commands during the drain is extra.

esx_service/vmdk_ops.py
govint: wait_ops_in_flight() should just have to set the stop event and then return.  The loop in handleVmciRequests() should check for the stop event and exit that loop and wait for the command count to drain to zero. the signal handler shouldn't have to do anything else besides setting the stop event.

esx_service/vmdk-opsd
msterin: still don't get it. My suggestion is to change "vmdk-opsd shutdown"" to "vmdk-opsd kill" (i.e. replace the command name passed to vmdk_opsd. How does it imact anything else ? No need to have a function kill() {} , the function can be named force_kill() :-)

esx_service/vmdk_ops.py
msterin: Correct, it is not possible in the current implementation - but it depends on other parts which are outside of the function and are not clear here. At the very least a comment is required.  
At any rate , this is not very important as this code needs to change (see my other comment about blocking main thread)

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
msterin: still, would be nice to say `In-flight operations are taking too long to complete, abandoning wait`

esx_service/vmdk-opsd
shaominchen: I see. Will try as you suggested.

esx_service/vmdk_ops.py
msterin: yes, release_listening_socket needs to wait untill all is done. 
Generally.  the way I expected the code was (in pseudo-code) something like this: 

```python
lock= threading.Rlock() # used for counters
condition = threading.Condition(lock) # used to signal on the counter drop
barrier = False

    signal_handler: 
        barrier = True
        spawn a waiting_thread
        return 

   main_thread:
        while get_vmci_op:
           if (barrier) vmci_reply("shutting down"); continue
           increase count
           execute_thread_for_command

thread_for_command:
    count.decr() on exit

Count:
   decr:
        with Rlock()  
          count -= 1
          if (count == 0) :
                 condition.notifyAll()


   incr:
       if (barrier): 
         assert_or_yell
      with Rlock:
       count += 1

waiting_thread":
      condition.acquire()
     try:  
        condition.wait(timeout)
    except RuntimeError:
         log.warning("dropped the wait due to timeout")
     sys.exit()
 ```

or you can use threading.event() to signal/wait instead of condition (events are simpler) 
In the current code, it's harder to navigate the logic for issues ....

File: esx_service/vmdk_ops.py (hunk 7 above)
Comment (msterin) [reply esx_service/vmdk_ops.py]: @shaominchen  0 https://docs.python.org/2/library/signal.html says

Some care must be taken if both signals and threads are used in the same program. The fundamental thing to remember in using signals and threads simultaneously is: always perform signal() operations in the main thread of execution. Any thread can perform an alarm(), getsignal(), pause(), setitimer() or getitimer(); only the main thread can set a new signal handler, and the **main thread will be the only one to receive signals (this is enforced by the Python signal module, even if the underlying thread implementation supports sending signals to individual threads)***.

I read it that if the main thread executes the signal handler and the code stays in the handler forever, nothing else will be executed on the main thread

File: tests/e2e/vmlistener_test.go [hunk 16]
```diff
@@ -101,6 +101,9 @@ func (s *VMListenerTestParams) TestServiceRestart(c *C) {
 	out, err = admincli.RestartVmdkopsService(s.esx)
 	c.Assert(err, IsNil, Commentf(out))
 
+	// Give some time for vmdkops service initialization
+	misc.SleepForSec(5)
```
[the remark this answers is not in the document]
shaominchen: This PR is already very complicated, so I'm not going to add anything unrelated to this PR. There are actually a few misc.SleepForSec(15) in vmlistener_test. I will file a separate issue for this.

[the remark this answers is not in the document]
shaominchen: Filed issue #1894.

esx_service/vmdk-opsd
shaominchen: That's not the intent - we are not trying to kill the service after N retries.

esx_service/vmdk-opsd
shaominchen: I don't have strong opinion about this - either way seems acceptable to me. @msterin do you have any opinion about this?

esx_service/vmdk_ops.py
shaominchen: We have those logs inside wait_ops_in_flight().

esx_service/vmdk_ops.py
shaominchen: Yes agree, will fix.

esx_service/vmdk_ops.py
shaominchen: We can, but I feel if possible, fail the request immediately is still better than timeout, from user experience point of view.

esx_service/vmdk-opsd
msterin: If stopping fails because drain failq, the whole command fails will clear message. User can always do force kill.if needed, but we should not force kill after a request for soft stop, IMO

shuklanirdesh82:
@shaominchen 

Currently vDVS driver coverage data is missing (#1895). As you are introducing new code it is likely to be a drop in coverage percentage. It is better to add some unit tests or E2E test as part of this PR.

If you are planning to add unit test/E2E test into separate PR, please raise a tracking issue.

Thanks!

esx_service/vmdk-opsd
shaominchen: Yes. For example, in a long running task (e.g. formatting a huge disk), we may not want to force kill the operation in a stop request. It's better to let users do it explicitly by themselves.

esx_service/vmdk_ops.py
shaominchen: I will do some more investigation/testing about this.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: Sure.

esx_service/vmdk_ops.py
shaominchen: Thanks for sharing your thoughts. Will try to address in follow-up changes.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,44 @@ def set_vol_opts(name, tenant_name, options):
...
+def wait_ops_in_flight():
+    global WAIT_STOP_COUNT
+
+    # Set the STOP flag to true
+    stopEvent.set()
+
+    # Wait until the in-flight ops counter goes down to zero
+    while ((opsCounter.value > 0) and (WAIT_STOP_COUNT > 0)):
+        logging.info("Service is being stopped. Waiting for completion of %d operations in flight...", opsCounter.value)
+        time.sleep(WAIT_STOP_INTERVAL)
```
msterin: I think this is the  major issue here... it should no wait in signal handler.

shaominchen: Actually we should wait in signal handler.

As you also mentioned, once the signal handler takes over the main thread, the logic in the handleVmciRequests main loop will be hung. So if we move this wait into handleVmciRequests() - no matter it's in the loop or after the loop - it will never get a chance to run, because the logic in handleVmciRequests() will (almost) always gets stuck on the listening socket call - lib.vmci_get_one_op(). I have already verified this.

However, blocking on lib.vmci_get_one_op() has no impact to the system exit(0) call, which is triggered from the signal handler.

The logic is quite complicated here - I hope I have explained it clearly :)

esx_service/vmdk_ops.py
shaominchen: @govint Please see my latest answer to @msterin 's comments below.

esx_service/vmdk_ops.py
msterin: wait, who suggested to move waiting logic into handleVmciRequests()  ? 

I think the right approach is as follows: 
- signal handler either exit() right away (if count is 0) or sets a barrier and spawns a waiting thread
- waiting thread waits on an avent and when someone signals on the event, exits
- count.decr() signals on the event when barrier is set and count drops to 0

It is fairly standard drain-end-exit steps.

shaominchen: Ah, too many comments got messed up:) Your suggestion above makes sense. I will try it out. Thanks!

shaominchen:
@msterin @govint Please review the latest update. Thanks!

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1826,16 @@ def handleVmciRequests(port):
...
 
-
     while True:
+        # Check the STOP flag - exit the loop if the flag is set
+        if stopEvent.isSet():
+            logging.info("Received stop request - stop processing new request")
+            break;
+
+        # Listening on VMCI socket
+        logging.info("lib.vmci_get_one_op: waiting for new request...")
```
msterin: There is no need for info here. We used to have it and it polluted the logs. If you need it for debug , feel free to use log.debug.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1826,16 @@ def handleVmciRequests(port):
...
         c = lib.vmci_get_one_op(g_vmci_listening_socket, byref(cartel), txt, c_int(bsize))
-        logging.debug("lib.vmci_get_one_op returns %d, buffer '%s'",
-                      c, txt.value)
+        logging.info("lib.vmci_get_one_op returns %d, buffer '%s'", c, txt.value)
```
msterin: same as above

esx_service/vmdk_ops.py
msterin: @govint - why make them suffer and timeout, when we can return a clear error message right away ? 
It really adds 3 lines to our code, with is about 3 times less than the number of lines we already spent discussing :-)

esx_service/vmdk_ops.py
```diff
@@ -1810,18 +1855,30 @@ def handleVmciRequests(port):
...
+            send_vmci_reply(client_socket, err(svc_stop_err))
+            continue
+
         if not get_si():
             svc_connect_err = 'Service is presently unavailable, ensure the ESXi Host Agent is running on this host'
             logging.warning(svc_connect_err)
             send_vmci_reply(client_socket, err(svc_connect_err))
             continue
 
+        # Increase the in-flight ops counter
```
msterin: comment is useless here, just takes up real estate

esx_service/vmdk_ops.py
```diff
@@ -1810,18 +1855,30 @@ def handleVmciRequests(port):
...
+
         # Fire a thread to execute the request
         threadutils.start_new_thread(
             target=execRequestThread,
             args=(client_socket, cartel.value, txt.value))
 
-    vmci_release_listening_socket() # close listening socket when the loop is over
+    # Close listening socket when the loop is over
+    logging.info("Closing VMCI listening socket...")
+    vmci_release_listening_socket()
```
msterin: wait , why do you want to close the listening socket ? Will it be reopened on the next cycle ? where ?

esx_service/utils/counter.py
```diff
@@ -0,0 +1,50 @@
...
+# A counter implementation that can be shared by multiple threads.
+
+import logging
+import threading
+
+class Counter:
+    '''
+    A counter object that can be shared by multiple threads.
+    '''
+    _lock = threading.RLock()
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/utils/counter.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, initial_value = 0):
+        self._value = initial_value
+
+    def incr(self, delta = 1):
+        '''
+        Increment the counter with locking
+        '''
+        with Counter._lock:
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/utils/counter.py
```diff
@@ -0,0 +1,50 @@
...
+            logging.debug("Before incr: counter.value=%d", self._value)
+            self._value += delta
+            logging.debug("After incr: counter.value=%d", self._value)
```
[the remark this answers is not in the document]
shaominchen: OK.

esx_service/vmdk-opsd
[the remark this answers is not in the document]
shaominchen: Updated.

esx_service/vmdk_ops.py
```diff
@@ -93,6 +94,8 @@
 from error_code import ErrorCode
 from error_code import error_code_to_message
 import vm_listener
+import counter
+from counter import Counter
```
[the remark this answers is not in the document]
shaominchen: Agree, will update.

esx_service/vmdk-opsd
```diff
@@ -112,32 +113,65 @@ stop() {
...
+
+cleanup_resource() {
    ${LOCAL_CLI_SCHED} list --group-path=${OPSD_GROUP} &> /dev/null
    if [ $? -eq 0 ]; then
       ${LOCAL_CLI_SCHED} delete --group-path=${OPSD_GROUP}
       if [ $? -ne 0 ]; then
          echo "Failed to delete resource pool=${OPSD_SCHED_PARAM}"
          check_running_status
-         exit 4
+         return 4
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/vmdk_ops.py
```diff
@@ -138,6 +141,16 @@
 # For managing resource locks.
 lockManager = threadutils.LockManager()
 
+# Flag indicating whether stop has been requested
+stopEvent = threading.Event()
```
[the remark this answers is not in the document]
shaominchen: OK. Not expert on Python - a simple boolean flag is not thread safe in Java. I will change it to a simple boolean then. Thanks for clarifying.

esx_service/vmdk_ops.py
```diff
@@ -138,6 +141,16 @@
...
+
+# Counter of operations in flight
+opsCounter = Counter()
+
+# Number of times and sleep interval to wait on stop process
+WAIT_STOP_COUNT = 20
```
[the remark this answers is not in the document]
shaominchen: Using Event or Condition is fine, but one good thing with polling is that it shows "progress" (i.e. time elapsed) in the log. I'm fine to change it to Event though.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,46 @@ def set_vol_opts(name, tenant_name, options):
 
     return False
 
-def msg_about_signal(signalnum, details="exiting"):
+def wait_ops_in_flight():
+    stop_count = WAIT_STOP_COUNT
+
+    # Wait until the in-flight ops counter goes down to zero
+    while ((opsCounter.value > 0) and (stop_count > 0)):
```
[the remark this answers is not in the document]
shaominchen: I will use event here. However, coupling this event with Counter implementation may not be a good idea. I introduced this Counter class for generic purposes, so I'd rather keep the implementation generic so that it can be used for other purposes.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,46 @@ def set_vol_opts(name, tenant_name, options):
...
+        logging.warn("In-flight operations are taking too long to complete - abandoning wait")
+
+def handle_signal(signalnum, details="exiting"):
     logging.warn("Received signal num: %d, %s.", signalnum, details)
-    logging.warn("Operations in flight will be terminated")
+
+    if (opsCounter.value == 0):
+        logging.info("No in-flight operations. Exiting...")
+        sys.exit(0)
+    else:
```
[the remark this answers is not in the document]
shaominchen: Agree, fixed.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,46 @@ def set_vol_opts(name, tenant_name, options):
...
+        # Set the STOP flag to true
+        logging.debug("Setting STOP event flag to True")
+        stopEvent.set()
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1683,46 @@ def set_vol_opts(name, tenant_name, options):
...
+
+        # Fire a thread to wait for ops in flight to drain
+        threadutils.start_new_thread(target=wait_ops_in_flight)
 
 def signal_handler_stop(signalnum, frame):
-    msg_about_signal(signalnum, "exiting")
-    sys.exit(0)
+    handle_signal(signalnum, "exiting")
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/vmdk_ops.py
```diff
@@ -1757,6 +1794,9 @@ def execRequestThread(client_socket, cartel, request):
         reply_string = err("Server returned an error: {0}".format(repr(ex_thr)))
         send_vmci_reply(client_socket, reply_string)
 
+    # Decrease the in-flight ops counter
+    opsCounter.decr()
```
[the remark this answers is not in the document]
shaominchen: Good catch! Fixed.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
msterin: Python has GIL (global interpreter lock) so single ops (like assign or test) are atomic.
I suppose in the use case like our Java bool would have been fine too, as you just set the barrier once and read many times, so as long as bool set or get is atomic (which I suspect it always is) , we'd be fine

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
msterin: progress can be (and I think already is?)  shown from the shell script.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
msterin: I don't think about is as a counter. It's more of a reverse counting semaphore - when drops to 0, releases the waiters

esx_service/vmdk_ops.py
shaominchen: This is existing code, I just added a log:) I suppose it's just for releasing the socket resource.

esx_service/vmdk_ops.py
shaominchen: Sometimes I feel it can improve the readability a little bit by adding one line comment for each separate logic segment in a big chunk of code. This is just my personal preference. I hope it's not a big concern to you:)

esx_service/vmdk_ops.py
shaominchen: Will change it to debug - it's very useful for debugging.

esx_service/vmdk_ops.py
shaominchen: Done.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1826,16 @@ def handleVmciRequests(port):
     cartel = c_int32()
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the STOP flag - exit the loop if the flag is set
+        if stopEvent.isSet():
+            logging.info("Received stop request - stop processing new request")
```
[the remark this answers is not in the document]
shaominchen: Done.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1826,16 @@ def handleVmciRequests(port):
     cartel = c_int32()
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the STOP flag - exit the loop if the flag is set
+        if stopEvent.isSet():
```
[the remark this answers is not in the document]
shaominchen: Fixed.

esx_service/vmdk_ops.py
msterin: I totally agree. But the comment is useless and is a distraction in the exam below
```
#increment ops counter
OpsCounter.incr()
```
IMO the comment should add something g- intent, or brief summary , etc...

Not a big deal and was a neat anyways

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1826,16 @@ def handleVmciRequests(port):
     cartel = c_int32()
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the STOP flag - exit the loop if the flag is set
```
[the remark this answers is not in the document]
shaominchen: I agree it's not really required, since we already have a check after the op is received. But this won't prevent client from seeing proper error messages, since we don't even try to get the ops at all after breaking the loop. Anyway, I will remove this check.

esx_service/vmdk_ops.py
msterin: I see. Ok, I'll assume.ots.corect then :-)

msterin: Sorry about the typos, using phone....

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: Shell does not show the progress of decreasing in-flight ops counter. Anyway, I've changed it to event.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: No, that's not safe in Java due to its memory model. Java language does guarantee that reading or writing a variable is atomic (except long or double). Without synchronization, however, one thread’s changes might not even be visible to other threads. Synchronization is required for reliable communication between threads as well as for mutual exclusion in Java language.

I did some research - it looks like this is safe in Python.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: After debugging this further - it proves that we do need this check. Here's what I observed:

1. After sending the SIGTERM signal, the blocking vmci_get_one_op() call will return with an error:
```
09/26/17 21:03:39 8172365 [MainThread] [DEBUG  ] lib.vmci_get_one_op returns -1, buffer 'b'{"cmd":"list","details":{"Name":""},"version":"2"}''
09/26/17 21:03:39 8172365 [MainThread] [WARNING] vmci_get_one_op failed ret=-1: Invalid argument (errno=22) Retrying...
```
2. Above error will cause the workflow continue and again being blocked on the vmci_get_one_op() call.

So, if we don't have above check, the loop will never end - it's being blocked on vmci_get_one_op() for ever. Please note that we don't call exit(0) from main thread anymore, as the wait_ops_in_flight() is now being invoked from a child thread.

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: @msterin I've addressed all your comments except this one in the latest commit. I'm working on this one right now - will send a new commit if this change looks better.

esx_service/utils/counter.py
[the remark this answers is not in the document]
msterin: I do not see the fix :-(

esx_service/vmdk-opsd
```diff
@@ -48,6 +49,15 @@ NUM_RETRY=10
...
 WAIT_START_SEC=1
 
+ERR_INVALID_PARAM=1
+ERR_START_RUNNING=2
+ERR_START_FAILURE=3
+ERR_CREATE_RESOURCE=4
+ERR_DELETE_RESOURCE=5
+ERR_NOT_RUNNING=6
+ERR_STOP_SERVICE=7
+ERR_KILL_SERVICE=8
```
msterin: thank you

esx_service/vmdk-opsd
```diff
@@ -112,32 +122,65 @@ stop() {
...
          fi
       done
    fi
 
+   cleanup_resource
+   if [ $? -ne 0 ]; then
+      exit $?
+   fi
+
+   check_running_status
```
msterin: nit: it's also possible to use
```shell
if check_running_status # this will be true if $?==0 for check_running_status 
then 
  do this and that
fi
```
Not worth delaying the PR for the fix so just FYI

esx_service/vmdk-opsd
```diff
@@ -112,32 +122,65 @@ stop() {
...
       kill -$SIGTERM ${PID}
+
       local COUNT=0
       while kill -0 ${PID} > /dev/null 2>&1; do
          sleep ${WAIT_KILL_SEC};
+         echo "Waiting for operations in flight to complete, please do not interrupt..."
          if [ $((COUNT++)) -gt ${NUM_RETRY} ]; then
-            echo "Sending kill -${SIGKILL} to ${PID} after kill -${SIGTERM} did not stop ${PID}"
-            kill -${SIGKILL} ${PID}
+            echo "Kill -${SIGTERM} failed to stop ${PID}"
```
msterin: so if it fails and we back off, don't we need to restore the watchdog to oversee vmdk_ops.py daemon ?

esx_service/vmdk-opsd
shaominchen: Above change makes the code shorter, but it might cause ambiguity for new readers. Also it's consistent with existing style in this file, so I will keep it as is for now.

esx_service/vmdk_ops.py
```diff
@@ -1810,18 +1851,31 @@ def handleVmciRequests(port):
...
+
         if not get_si():
             svc_connect_err = 'Service is presently unavailable, ensure the ESXi Host Agent is running on this host'
             logging.warning(svc_connect_err)
             send_vmci_reply(client_socket, err(svc_connect_err))
             continue
 
+        opsCounter.incr()
+        logging.debug("Clearing opsDrained event")
+        opsDrained.clear()
```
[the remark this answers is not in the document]
shaominchen: Please see the latest update.

esx_service/vmdk_ops.py
```diff
@@ -1810,18 +1851,31 @@ def handleVmciRequests(port):
 
         client_socket = c # Bind to avoid race conditions.
 
+        # Check the stop barrier - if set, fail new incoming requests and exit the loop
+        if stopBarrier:
+            svc_stop_err = 'Service is being stopped: operation declined!'
+            logging.warning(svc_stop_err)
+            send_vmci_reply(client_socket, err(svc_stop_err))
```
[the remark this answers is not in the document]
shaominchen: We do have a debug log inside send_vmci_reply(). Anyway, I will not do anything for this for now.

esx_service/utils/counter.py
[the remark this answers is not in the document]
shaominchen: Fixed other places, but this variable also needs to be fixed. Will address in new update.

esx_service/vmdk_ops.py
```diff
@@ -1786,11 +1822,16 @@ def handleVmciRequests(port):
     cartel = c_int32()
     vmci_grab_listening_socket(port)
 
-
     while True:
+        # Check the stop barrier - if set, exit the loop
+        if stopBarrier:
+            logging.info("The service is being stopped and does not accept client requests")
```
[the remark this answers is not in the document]
shaominchen: I have already replied in previous comments - copy it here:
==========================================
After debugging this further - it proves that we do need this check. Here's what I observed:

1. After sending the SIGTERM signal, the blocking vmci_get_one_op() call will return with an error:
```
09/26/17 21:03:39 8172365 [MainThread] [DEBUG  ] lib.vmci_get_one_op returns -1, buffer 'b'{"cmd":"list","details":{"Name":""},"version":"2"}''
09/26/17 21:03:39 8172365 [MainThread] [WARNING] vmci_get_one_op failed ret=-1: Invalid argument (errno=22) Retrying...
```
2. Above error will cause the workflow continue and again being blocked on the vmci_get_one_op() call.

So, if we don't have above check, the loop will never end - it's being blocked on vmci_get_one_op() for ever. Please note that we don't call exit(0) from main thread anymore, as the wait_ops_in_flight() is now being invoked from a child thread.

esx_service/vmdk_ops.py
```diff
@@ -1756,6 +1787,11 @@ def execRequestThread(client_socket, cartel, request):
         logging.exception("Unhandled Exception:")
         reply_string = err("Server returned an error: {0}".format(repr(ex_thr)))
         send_vmci_reply(client_socket, reply_string)
+    finally:
+        opsCounter.decr()
+        if (opsCounter.value == 0):
```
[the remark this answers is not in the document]
shaominchen: Please see the latest update - https://github.com/vmware/docker-volume-vsphere/pull/1888/commits/18561e402ca7a592caf08380b52ae276a7b6ec3e

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1684,39 @@ def set_vol_opts(name, tenant_name, options):
...
 
-def msg_about_signal(signalnum, details="exiting"):
-    logging.warn("Received signal num: %d, %s.", signalnum, details)
-    logging.warn("Operations in flight will be terminated")
+def wait_ops_in_flight():
+    # Wait for the event indicating all in-flight ops are drained
+    eventReceived = opsDrained.wait(WAIT_OPS_TIMEOUT)
+    if (eventReceived):
+        logging.info("All in-flight operations are completed. Exiting...")
+        sys.exit(0)
```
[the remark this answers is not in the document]
shaominchen: You are right about the behavior of sys.exit(), but the current solution does not really rely on this behavior. We spawned  a thread to wait for ops in flight, and break/exit the main loop when barrier is set. So it still works.

esx_service/vmdk-opsd
shaominchen: I did think about this. My opinion is that since user wants to stop the service, and we also prompt user "Please try to force kill the process", do we really care about the watchdog in this case? We do expect user to force kill the daemon, and if they don't, the daemon is still running in un-watched status, which I feel is acceptable.

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1684,39 @@ def set_vol_opts(name, tenant_name, options):
...
+    else:
+        logging.warn("In-flight operations are taking too long to complete - abandoning wait")
```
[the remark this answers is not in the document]
shaominchen: Good question! This is definitely an open question:) I think restoring working service might be better than half dead. A running daemon which does nothing doesn't sound a good idea to me. I will restore it for now unless you or others have different opinions.

esx_service/vmdk-opsd
msterin: Makes sense, but I would recommend to add to a message "Failed to stop... keeps running with limited functionality" or doc

shaominchen: I agree, but how about "Failed to stop... keeps running without watchdog"? The statement "with limited functionality" is kind of unclear and might cause ambiguity/scare to the user:)

esx_service/vmdk_ops.py
```diff
@@ -1670,22 +1681,43 @@ def set_vol_opts(name, tenant_name, options):
...
 
-def msg_about_signal(signalnum, details="exiting"):
-    logging.warn("Received signal num: %d, %s.", signalnum, details)
-    logging.warn("Operations in flight will be terminated")
+def wait_ops_in_flight():
+    # Wait for the event indicating all in-flight ops are drained
+    eventReceived = opsCounter.wait(WAIT_OPS_TIMEOUT)
+    if (eventReceived):
+        logging.info("All in-flight operations are completed. Exiting...")
+        sys.exit(0)
```
msterin: sys.exit ony works on the main thread. I believe I mentioned it in some other comments. You may need to use os.signal(os.getpid(), signal.SIGKILL)

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
msterin: Unfortunately it did not  explain the need the first time, and the copy-n-paste of it stilll does not explain the need. Here is my understanding of the correct flow: 
 
signal from outside interrupts the wait in vmci_get-one_op, we log the error and enter the get_one_op again. Which is perfectly fine - If a client requests an op, the get_one_op will return and the ops will be declined in line 1856. When all the in-flight ops land, the waiting thread is awaken and forces exit (os.kill), no matter what main thread is doing..  Lines 1826-1829 are not needed. If you believe they are needed, lets take it offline.

esx_service/vmdk_ops.py
shaominchen: I've replied in previous comments - copy it here:

"You are right about the behavior of sys.exit(), but the current solution does not really rely on this behavior. We spawned a thread to wait for ops in flight, and also break/exit the main loop when barrier is set. So it still work as expected."

Do you still think we need to use os.signal(os.getpid(), signal.SIGKILL)?

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: Actually that's not the case. Here's what I observed:

1. "Signal from outside interrupts the wait in vmci_get-one_op, we log the error and enter the get_one_op again" - as you said, this is perfectly fine
2. "If a client requests an op, the get_one_op will return and the ops will be declined in line 1856" - this is not true, line 1856 was never exercised actually. After receiving signal and returning an error, the get_one_op does not return any new incoming ops any more - it's simply being blocked there for ever. If I run any op (such as "docker volume ls"), the request does not reach to the driver - the op simply returns result just like vdvs plugin does not exist.

I'm not sure if this is a bug in our vmci code, or it's the expected behavior.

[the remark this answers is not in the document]
msterin: And that's exactly my point. The current code tries to compensate for unknown bug, and it just does not read nor feels right. Our previous theory was that the wait_for_ops blocking in main thread was preventing the get_one_op from returning. Now that the wait is moved to a different thread, there is no explanation at all. I feel super uncomfortable  with the code which does confusing-to-read compensation (lines 1826-1829) for unknown bug. To reiterate - do you still see this blockage (and get_one_op never returns) after the wait is moved to a separate thread ?

esx_service/vmdk_ops.py
msterin: Ah, sorry I must have missed your reply (high traffic !) .

Ah, I see what's going on. The waiting thread does not really forces the service to exit, thus the main thread will wait on get_one_op forever, or until there is a request from the client, and thus forces this strange code in the main loop to avoid the block. (there is another issue of NOT returning from get_one_op on client request, but I think it's unlikely since it's basic functionality). 

So, this is how it should work: 
- waiting thread forces the service to quit when the wait is done. This will interrupt whatever main thread is doing
- main thread does not have to have any special handling of the break on wait complete. It is not it's responsibility. Instead, it just rejects operations when the barrier is UP
- if you see the ops not getting through when the barrier is up, let's investigate offline. 

this should make the overall flow much simpler and clearer

esx_service/vmdk_ops.py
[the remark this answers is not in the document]
shaominchen: I don't see that the blockage (get_one_op never returns) any more. I agree with you, I don't feel comfortable with this part either. Let me debug it further.

esx_service/vmdk_ops.py
shaominchen: I will push one more commit, which I feel more comfortable with. Let's discuss based on the new commit.

shaominchen:
@msterin I've addressed last 2 concerns. Please take a look.

Moreover, I've decided to reject new requests if the service fails to stop in a timely fashion. Keeping it half dead might cause even more confusion. User should try to kill or restart the service and continue.

shaominchen:
@msterin Yes, I've updated the PR description about all the tests that have been done with the latest code. Again, thanks for your time for taking a thorough review for this complicated PR!

PR #1901, opened 2017-09-25
[SKIP CI] (Trivial) Fixed formatting

tusharnt:


shuklanirdesh82:
We need to add [SKIP CI] while merging to master otherwise CI will be kicked off unnecessary.

PR #1908, opened 2017-09-27
Populate client list on mount and depopulate on unmount

### LINKED ISSUE #1761: Populate client list on mount and depopulate on unmount
A shared volume could be running anywhere on the swarm cluster. If a user tries to delete this volume and repeatedly gets the message "The volume is in use", it will be frustrating to keep looking for containers on every host.

Volume metadata already has a place to store the name of host VMs where the volume is mounted. This list should be populated during mount and unpopulated during unmount.

lipingxue:
Fixed #1761 .
I have done the following test manually.
Configuration: 3 VMs in a swarm cluster, 1 master and 2 workers
VM1: 10.160.112.117 (master)
VM2: 10.160.120.245 (worker)
VM3: 10.160.122.147(worker)
1. create a volume "vol1" using vFile plugin
```
vmware@ubuntu16:~$ docker volume ls
vmware@ubuntu16:~$ docker volume create --driver=vfile --name=vol1
vol1
vmware@ubuntu16:~$ 
vmware@ubuntu16:~$ docker volume ls
vsphere:latest      _vF_vol1@vsanDatastore
vfile:latest        vol1
vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Ready"
        }
    }
]
```
2. mount volume from VM1, then check the "docker volume inspect" result, we can see the IP of VM1 has been added to the "Clients" section.
```
vmware@ubuntu16:~$ docker run -it -v vol1:/vol busybox /bin/sh
/ # vmware@ubuntu16:~$ 
vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "10.160.112.117"
            ],
            "File server Port": 30000,
            "Global Refcount": 1,
            "Service name": "vFileServervol1",
            "Volume Status": "Mounted"
        }
    }
]
```

3. Try to mount "vol1" from VM2, then check the "docker volume inspect" result, we can see the IP of VM2 has been added to the "Clients" section.
```
vmware@ubuntu16:~$ docker run -it -v vol1:/vol busybox /bin/sh
/ # vmware@ubuntu16:~$ 

vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "10.160.112.117",
                "10.160.120.245"
            ],
            "File server Port": 30000,
            "Global Refcount": 2,
            "Service name": "vFileServervol1",
            "Volume Status": "Mounted"
        }
    }
]
```

4. Try to mount "vol1" from VM3, then check the "docker volume inspect" result, we can see the IP of VM3 has been added to the "Clients" section.
```
vmware@ubuntu16:~$ docker run -it -v vol1:/vol busybox /bin/sh

vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "10.160.112.117",
                "10.160.120.245",
                "10.160.122.147"
            ],
            "File server Port": 30000,
            "Global Refcount": 3,
            "Service name": "vFileServervol1",
            "Volume Status": "Mounted"
        }
    }
]
```

5. unmount "vol1" from VM3,  then check the "docker volume inspect" result, we can see the IP of VM3 has been removed from the "Clients" section.
```
vmware@ubuntu16:~$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
ae31c3b8768b        busybox             "/bin/sh"                22 seconds ago      Up 18 seconds                           keen_bartik
55e8d79a426f        nginx:latest        "nginx -g 'daemon ..."   8 minutes ago       Up 8 minutes        80/tcp              nginx_serivce.1.q2t8o15bwcsi6lmplx2ud7s9q

vmware@ubuntu16:~$ docker rm ae31c3b8768b -f
ae31c3b8768b

vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "10.160.112.117",
                "10.160.120.245"
            ],
            "File server Port": 30000,
            "Global Refcount": 2,
            "Service name": "vFileServervol1",
            "Volume Status": "Mounted"
        }
    }
]
```

6. unmount "vol1" from VM2,  then check the "docker volume inspect" result, we can see the IP of VM2 has been removed from the "Clients" section.
```
vmware@ubuntu16:~$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED              STATUS              PORTS               NAMES
1d841ae5b9ba        busybox             "/bin/sh"           About a minute ago   Up About a minute                       musing_bardeen

vmware@ubuntu16:~$ docker rm 1d841ae5b9ba   -f
1d841ae5b9ba

vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": [
                "10.160.112.117"
            ],
            "File server Port": 30000,
            "Global Refcount": 1,
            "Service name": "vFileServervol1",
            "Volume Status": "Mounted"
        }
    }
]
```

7. unmount "vol1" from VM1,  then check the "docker volume inspect" result, we can see the IP of VM1 has been removed from the "Clients" section.
```
vmware@ubuntu16:~$ docker ps
CONTAINER ID        IMAGE                  COMMAND                  CREATED             STATUS              PORTS                           NAMES
8e12fa40c92a        dperson/samba:latest   "samba.sh -s share..."   3 minutes ago       Up 3 minutes        139/tcp, 137-138/udp, 445/tcp   vFileServervol1.1.e78e71txfldl3caoaabyadjhh
144d7f74f58b        busybox                "/bin/sh"                3 minutes ago       Up 3 minutes                                        hardcore_benz

vmware@ubuntu16:~$ docker rm 144d7f74f58b  -f
144d7f74f58b
vmware@ubuntu16:~$ 
vmware@ubuntu16:~$ 
vmware@ubuntu16:~$ docker volume inspect vol1
[
    {
        "Driver": "vfile:latest",
        "Labels": {},
        "Mountpoint": "/mnt/vfile/vol1/",
        "Name": "vol1",
        "Options": {},
        "Scope": "global",
        "Status": {
            "Clients": null,
            "File server Port": 0,
            "Global Refcount": 0,
            "Service name": "",
            "Volume Status": "Ready"
        }
    }
]
```
```

### DIFF
--- a/client_plugin/drivers/vfile/vfile_driver.go
+++ b/client_plugin/drivers/vfile/vfile_driver.go
@@ -772,7 +772,6 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
 			}).Error("Failed to remove VM IP from ClientList")
 		return err
 	}
-
 	return nil
 }
 
--- a/client_plugin/drivers/vfile/vfile_driver.go
+++ b/client_plugin/drivers/vfile/vfile_driver.go
@@ -674,25 +674,6 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 		return "", errors.New(msg)
 	}
 
-	// mount is successful, update the clientList
-	_, addr, _, err := d.dockerOps.GetSwarmInfo()
-	if err != nil {
-		log.WithFields(
-			log.Fields{"volume name": name,
-				"error": err,
-			}).Error("Failed to get IP address from docker swarm ")
-		return "", err
-	}
-
-	err = d.addVMToClientList(name, addr)
-	if err != nil {
-		log.WithFields(
-			log.Fields{"volume name": name,
-				"error": err,
-			}).Error("Failed to add VM IP to ClientList")
-		return "", err
-	}
-
 	return mountpoint, nil
 }
 
@@ -735,6 +716,15 @@ func (d *VolumeDriver) mountVFileVolume(volName string, mountpoint string, volRe
 				"output": string(output),
 				"error":  err,
 			}).Error("Mount failed: ")
+		return err
+	}
+
+	err = d.addVMToClientList(volName, addr)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"volume name": volName,
+				"error": err,
+			}).Error("Failed to add VM IP to ClientList")
 		return err
 	}
 
--- a/client_plugin/drivers/vfile/vfile_driver.go
+++ b/client_plugin/drivers/vfile/vfile_driver.go
@@ -195,6 +195,21 @@ func (d *VolumeDriver) List(r volume.Request) volume.Response {
 	return volume.Response{Volumes: responseVolumes}
 }
 
+// GetClientList - return client list which is mounted to volume
+func (d *VolumeDriver) GetClientList(name string) []string {
+	clientListMap, err := d.kvStore.KvMapFromPrefix(kvstore.VolPrefixClient + name)
+	var clientList []string
+	if err != nil {
+		log.Warnf("Failed to get client list for %s", name)
+		return clientList
+	}
+	for key, vmIP := range clientListMap {
+		clientList = append(clientList, vmIP)
+		log.Infof("GetClientList key=%s IP=%s", key, vmIP)
+	}
+	return clientList
+}
+
 // GetVolume - return volume meta-data.
 func (d *VolumeDriver) GetVolume(name string) (map[string]interface{}, error) {
 	var statusMap map[string]interface{}
@@ -232,7 +247,7 @@ func (d *VolumeDriver) GetVolume(name string) (map[string]interface{}, error) {
 	}
 	statusMap["File server Port"] = volRecord.Port
 	statusMap["Service name"] = volRecord.ServiceName
-	statusMap["Clients"] = volRecord.ClientList
+	statusMap["Clients"] = d.GetClientList(name)
 
 	return statusMap, nil
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: client_plugin/drivers/vfile/vfile_driver.go [hunk 1]
```diff
@@ -568,6 +674,25 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 		return "", errors.New(msg)
 	}
 
+	// mount is successful, update the clientList
```
luomiao: Inside mountVFileVolume, there is already code to call GetSwarmInfo to get the IP address. Can we move the following part of the new code to the end of mountVFileVolume function and merge with existing code there? So we can avoid calling GetSwarmInfo two times.

lipingxue:
@luomiao I have addressed your comments.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -613,6 +719,15 @@ func (d *VolumeDriver) mountVFileVolume(volName string, mountpoint string, volRe
 		return err
 	}
 
+	err = d.addVMToClientList(volName, addr)
+	if err != nil {
```
SandeepPissay: Leys say the volume was successfully mounted in the VM but updating the metadata in etcd failed. Basically the metadata does not match the reality. How are you doing to handle this scenario?

lipingxue: In that case,  function "mountVFileVolume" will return with error, then in function "MountVolume", it will handle this error and mark the "volume mount failed".  Do I miss anything?

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -494,6 +494,112 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
...
+		log.Warningf("Failed to marshal JSON for writing metadata: %v",
+			err)
+		return err
+	}
+	writeEntries = append(writeEntries, kvstore.KvPair{
+		Key:   kvstore.VolPrefixInfo + volName,
+		Value: string(byteRecord)})
+
+	log.Infof("Adding VM[%s] to CLientList for %s", vmIP, volName)
+	err = d.kvStore.WriteMetaData(writeEntries)
```
SandeepPissay: How are you going to address concurrent invocations of addVMToClientList method? Say invocation-1 is adding IP1 and invocation-2 is adding IP2, and these could be running on different VFile programs. There is a possibility that invocaiton-2 could be overriding what invocation-1 updated in etcd. Basically, updating the kvstore has to be an atomic operation.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -494,6 +494,112 @@ func (d *VolumeDriver) processMount(r volume.MountRequest) volume.Response {
...
```
SandeepPissay: Same comment here related to concurrenty. One execution of this method could be overwritten by another execution of this method or addVMToClientList method. For example, invocation-1 is removing an entry and invocation-2 is adding another entry. Does the kvstore support optimistic concurrency control?

luomiao:
Regarding to the concurrency issue.

Previously for updating the state or refcount, we indeed have functions like AtomicIncr and AtomicDecr to make sure the updating is atomic. For clients info, it might be difficult to use AtomicIncr and AtomicDecr.
I am thinking of another way. Since etcd v3 doesn't support directory anymore, how about just use key range for clients information?
We can have another new prefix `VolPrefixClient = "SVOLS_client_"`, and when a node mounted a vFile volume, it will do etcd set with `key=SVOLS_client_[volume name]_[swarm ID], value=[IP address]`. And for `docker volume inspect` operation, we call the current `kvMapFromPrefix` function with input as `SVOLS_client_[volume name]_`. And accordingly when a node unmounted a vFile volume, it should just do etcd del with `key=SVOLS_client_[volume name]_[swarm ID]`. In this way we can avoid using transactions or comparisons, which is not easy to do with long values such as multiple strings in this case.

client_plugin/drivers/vfile/vfile_driver.go
SandeepPissay: But the volume is *actually* mounted in the case I mentioned. Dont you have to unmount it in case of failure? I dont see code that does unmount in case of failure. 

I looked at MountVolume function, I see that we decrement global refcount in case of error. What if this decrement operation fails due to network problem or etcd down etc?

SandeepPissay:
@luomiao I did not follow your comment. Lets have a chat on this.

client_plugin/drivers/vfile/vfile_driver.go
lipingxue: Fixed.

lipingxue:
@luomiao @SandeepPissay I have addressed your comments. I fixed the concurrency issue based on Miao's suggestion, please take a look. Thanks!

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -562,6 +613,11 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
 		if err != nil {
```
SandeepPissay: As we discussed, it is still possible that the mount volume could fail and the atomic decrement and removeVMFromClientList fails. In such a case, the data in etcd will say that the volume is successfully attached, whereas in reality that has not happened.

I though more on this, could we do this instead?
- Mark the volume to be mounted as 'mounting' to node
- Mount the volume
- If mount is success, mark the volume as 'mounted' and increment the global ref count.
- If failure, try to undo the above operations in kv. If this fails, we just log the error.

We can have a background task that continuously keeps monitoring volumes that are in 'mounting' state, and mark as failure after some timeout.

Would this approach work?

luomiao: No this won't work...
The reason is, "increase global ref count" has to happen before "Mount the volume".
This is because global ref count change triggered the swarm managers to start the mounting process.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -698,6 +745,43 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
+	err = d.removeVMFromClientList(name, nodeID, addr)
+	if err != nil {
+		log.WithFields(
+			log.Fields{"volume name": name,
+				"error": err,
+			}).Error("Failed to remove VM IP from ClientList")
+		return err
+	}
+	mountpoint := d.GetMountPoint(name)
+	err = fs.Unmount(mountpoint)
```
SandeepPissay: Can you explain why we cannot unmount first and then decrement the gref? 

Otherwise the code looks good. The error handling is not done correctly and it seems like we need to think a bit more on it. This can be done as a different code change.

lipingxue: We can do that.  Current code is just to make sure it consistent with the mount workflow (increase gref and then do mount). Is there any reason you think we should do umount first and then decrement the gref?

SandeepPissay: As we discussed, the auto decrease of gref(1->0) will trigger deletion of the SMB container and that could fail since it is mounted/used by another container. If manager node gets to delete the SMB service successfully and if the worker node fails unmounting, we are increasing the gref and will lead to inconsistent state.

As I mentioned, the error handling needs to be thought through properly since there are different parties involved (manager nodes, worker nodes) and different failures (etcd failure, SMB container deploy failure, mount/unmount failure etc).

lipingxue: @SandeepPissay I addressed your comments to unmount first and then decrease gref. Please reivew it.

Issue #1947, opened 2017-10-24
vFile CI/CD tests need to be executed against ESX 6.0 


tusharnt:
Currently they are run only against ESX 6.5

ashahi1:
vFile test are now enabled on 6.0 ESX. Changes were checked in two PRs - #1987 and #1949 

Closing this issue now.

PR #1949, opened 2017-10-26
Part 1: Enabling vFile tests on 6.0 ESX

ashahi1:
This PR fixes issue #1947


shuklanirdesh82:
@ashahi1 your proposed changes look good, you may want to check upon CI failure.

```
FAIL: advanced_vfile_test.go:69: AdvancedVFileTestSuite.TestVFileVolumeLifecycle

advanced_vfile_test.go:76:
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc420011ee0), Stderr:[]uint8(nil)} ("exit status 1")
... Error response from daemon: create vfilevolume623934: VolumeDriver.Create: Failed to create volume vfilevolume623934. Reason: Failed to create etcd client
```

luomiao:
This is the same problem we have met with 6.5.

This means the port for ETCD is not open on those 6.0 photon os VMs.
@ashahi1 you need to open them the same way as what we have done at 6.5 side.

misc/drone-scripts/setup.sh
```diff
@@ -13,9 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo "Resetting testbed"
+echo "Resetting testbed!!"
 govc snapshot.revert -vm $ESX_6_5 init
-govc snapshot.revert -vm $ESX_6_0 init
+# will revert the snapshot name to init before merging the changes to master
```
shuklanirdesh82: remove this file from the PR and make sure CI esx is having `init` snapshot correctly set

ashahi1: Done.

PR #1953, opened 2017-10-26
[SKIP CI]Add more log in vFile code 

### LINKED ISSUE #1940: [vFile] Increase the logs in vFile plugin
Filing this issue as it was observed that we do not have enough logs to debug vFile plugin quickly.

@luomiao Assigning it to you for the time being.

lipingxue:
Fixed  #1940 
Add more code in vFile code.


luomiao:
I guess in order to skip CI, it should be [SKIP CI] instead of [SKIP CI]?

luomiao:
The logs generally look good to me.
Just wondering if this the first part of logging enhancement? more will come later rt?

lipingxue:
@luomiao This is NOT the first part and it is the log I think we need to add right now. More will come if we later find some other logs are needed to triage the bug. Right now, I am not sure more log will come or not.

Issue #1959, opened 2017-10-30
VFile: Reduce vfile log file size and/or reduce number of log files


govint:
the vFile logs can grow quiet large and can fill the root partition on a Linux desktop if let unchecked, as below
-rw-r--r-- 1 root      root 100M Sep 12 06:29 vfile-2017-09-12T13-29-48.158.log
-rw-r--r-- 1 root      root 100M Sep 12 06:32 vfile-2017-09-12T13-32-46.970.log
-rw-r--r-- 1 root      root 100M Sep 12 06:34 vfile-2017-09-12T13-34-26.209.log
-rw-r--r-- 1 root      root 100M Sep 12 06:35 vfile-2017-09-12T13-35-55.402.log
-rw-r--r-- 1 root      root 100M Sep 12 06:37 vfile-2017-09-12T13-37-20.429.log
-rw-r--r-- 1 root      root 100M Sep 12 06:38 vfile-2017-09-12T13-38-37.306.log
-rw-r--r-- 1 root      root 100M Sep 12 06:39 vfile-2017-09-12T13-39-48.654.log
-rw-r--r-- 1 root      root 100M Sep 12 06:41 vfile-2017-09-12T13-41-27.220.log
-rw-r--r-- 1 root      root 100M Sep 12 06:42 vfile-2017-09-12T13-42-54.285.log
-rw-r--r-- 1 root      root 100M Sep 12 06:44 vfile-2017-09-12T13-44-44.633.log

Suggest reducing the file size and/or the number of the log files.

lipingxue:
@govint If you still have the log, could you attached the log in this issue? I want to see why the log is flooded.

lipingxue:
Record the offline discussion with Govindan.

For this issue I'm not sure if a repro is needed. IN my case the vfile plugin was experiencing a continuous failure (most likely because I was installing and removing the vdvs plugin, stopping and starting nodes) and was generating a lot of logs. I had to remove the files as the VM hung up. For this issue we could just limit the size of each file and the number of files so that the disk isn't overrun.

lipingxue:
Fix of issue #1964 will solve this problem.

luomiao:
Closed by #1973

PR #1973, opened 2017-11-02
Limit the number of log files generated by vDVS/vFile

### LINKED ISSUE #1964: Limit the number of log files generated by vDVS/vFile
Limit the number of log files generated by vDVS/vFile

shuklanirdesh82:
Fixes #1964 

tested locally and working as expected.

e.g.
```
root@sc-rdops-vm18-dhcp-57-89:/var/log# ls -alh
total 22M
drwxrwxr-x 3 root   syslog  4.0K Nov  2 23:21 .
drwxr-xr-x 7 root   root    4.0K Sep 21 18:47 ..
-rw-r--r-- 1 root   root    185K Nov  2 23:18 abc.log
-rw-r----- 1 syslog adm     316K Nov  2 23:17 auth.log
-rw------- 1 root   utmp       0 Sep 21 18:47 btmp
-rw-r--r-- 1 root   root    2.0M Nov  2 23:21 docker-volume-vsphere-2017-11-02T23-21-32.418.log
-rw-r--r-- 1 root   root     764 Nov  2 23:21 docker-volume-vsphere.log
```


README.md
```diff
@@ -58,6 +58,7 @@ journalctl -fu docker.service # Journalctl/Systemd
  and log location. Example:
 ```
  {"MaxLogAgeDays": 28,
+ "MaxLogFiles": 3,
```
tusharnt: Why not make this 10?

README.md
```diff
@@ -58,6 +58,7 @@ journalctl -fu docker.service # Journalctl/Systemd
...
  "MaxLogSizeMb": 100,
```
tusharnt: Not sure why this is 100? Isn't it cumbersome to load 100mb files while debugging? I suggest changing this to 10.

README.md
shuklanirdesh82: This is just an example value for the reference and customer/user has to supply appropriate value as per need.

README.md
shuklanirdesh82: as explained above.

README.md
tusharnt: Even though these values are configurable we need to set a sensible default value.
I suggested the settings that most of the services in vSphere use as it seems to work well for vSphere customers.

README.md
govint: I had 10 x 100MB files and those filled up the root FS, most distros for docker are pretty small and 1GB or even 500MB of logs is going to be impossible to host. We need small log files and few of those.

client_plugin/utils/config/config.go
```diff
@@ -58,6 +60,7 @@ type Config struct {
 	LogPath        string `json:",omitempty"`
 	MaxLogSizeMb   int    `json:",omitempty"`
 	MaxLogAgeDays  int    `json:",omitempty"`
+	MaxLogFiles    int    `json:",omitempty"`
```
govint: 1 x 10MB or 10  x 1MB doesn't matter, the log just rotates when the max log size has been achieved. Configuring the max log files seems like something the user doesn't have to do. Why would the user care how the MaxLogSizeMb is sliced into how many files?

The fix for this issue is to cap the MaxLogSizeMb to something reasonable (vFile had 10 x 100MB, which is too large). Say 100MB cap on the max log size is sufficient vs. adding a new config.

README.md
shuklanirdesh82: Okay sure, I will set the sample values as shown below.

```
"MaxLogFiles": 3,
  "MaxLogSizeMb": 10,
```

At any given time, there will be ~30MB size will be used mostly. Please share your comment to close on this one. I will update the diff then after.

README.md
shuklanirdesh82: sure!

client_plugin/utils/config/config.go
shuklanirdesh82: > I'd suggest not adding a new config and instead cap the max log size instead to a reasonable maximum.

```
	// MaxBackups is the maximum number of old log files to retain.  The default
	// is to retain all old log files (though MaxAge may still cause them to get
	// deleted.)
	MaxBackups int `json:"maxbackups" yaml:"maxbackups"`
```

With the above reference, I agree it can be controlled through `MaxAge` but it doesn't favor always. By taking example of https://github.com/vmware/docker-volume-vsphere/issues/1959#issue-269565208, `MaxAge` doesn't help always, let's say vFile/vDVS plugin spits out log heavily, disk will be filled up quickly whereas it can be avoided through `MaxBackups` config option.

govint: Not MaxAge, MaxLogSize is what I'm referring to, fix the max log size and the number of files doesn't matter.

shuklanirdesh82: @govint As discussed offline, `MaxLogSize ` doesn't help alone when `MaxAge ` is in place. 

some more reference:

MaxSize=10MB
MaxAge=28 days
MaxBackup=3

if plugin behaves crazy .. there will be max 3 files 30MB at any given time .. otherwise MaxAge expires one as its value.

Please let me know if there is any more comment.

shuklanirdesh82:
@tusharnt @govint I have responded to your comments and ready for your review.

PR #1988, opened 2017-11-14
allow user to choose ETCD ports for vFile usage

### LINKED ISSUE #1970: [vFile] allow user to choose ETCD ports for vFile usage
In order to avoid conflicting with other ETCD usage by the user.

lipingxue:
Fixed #1970

With this change, user can specify etcd cluster port number other than default value (default etcdClientPort is 2379 and etcdClientPort is 2380) used by vFile. 
User need to specify those port numbers through environment variable "VFILE_ETCD_CLIENT_PORT" and "VFILE_ETCD_PEER_PORT", see the following example:

```
docker plugin install --grant-all-permissions --alias vfile lipingxue/vfile:vfile-etcd_port VFILE_TIMEOUT_IN_SECOND=300 VFILE_ETCD_CLIENT_PORT=4001 VFILE_ETCD_PEER_PORT=4002
```

After install the vFile plugin, test basic volume create/mount/unmount, works as expected.

### DIFF
--- a/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
+++ b/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
@@ -57,8 +57,8 @@ import (
 */
 const (
 	etcdDataDir              = "/etcd-data"
-	defaultEtcdClient        = ":2379"
-	defaultEtPeerPort        = ":2380"
+	defaultEtcdClientPort    = ":2379"
+	defaultEtcdPeerPort      = ":2380"
 	etcdClusterToken         = "vfile-etcd-cluster"
 	etcdListenURL            = "0.0.0.0"
 	etcdScheme               = "http://"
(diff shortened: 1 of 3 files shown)


client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -893,6 +916,7 @@ func (e *EtcdKVS) createEtcdClient() *etcdClient.Client {
 func addrToEtcdClient(addr string) (*etcdClient.Client, error) {
 	// input address are RemoteManagers from docker info or ManagerStatus.Addr from docker inspect
 	// in the format of [host]:[docker manager port]
+	etcdClientPort, _ := getEtcdPorts()
```
luomiao: This addrToEtcdClient function is called inside createEtcdClient function and will be called for almost every ETCD operations. So calling getEtcdPorts inside this function is not efficient (means everytime we read ENVs again although they are installation flags).
I suggest that we should call getEtcdPorts at the very beginning and then store the information inside the local EtcdKVS struct. And later in addrToEtcdClient we just need to grab this information locally, instead of reading the ENV flags again and again.

lipingxue: @luomiao I have addressed your comments, please take a look. Thanks!

Issue #1943, opened 2017-10-18
Revisit the error handling for vFile design


luomiao:
Current design has a flaw when handling multiple errors in ETCD updating at the same time.
Consider the following scenario:

1. Worker creates a new volume and start to use it, internally global refcount is atomic increased by one.
2. All the managers are triggered to try to update the state from "Ready" to "Mounting". Assume ETCD cluster is not accessible suddenly, and thus the update of the state is failed. State of the volume will stay as "Ready"
3. Worker will timeout when waiting for state to be "Mounted". As an error handling, worker will try to reduce the global refcount. However the ETCD cluster is not accessible again, and the worker may fail to reduce the global refcount and returned error back.
4. Now the volume's state is "Ready" while the global refcount of the volume is 1. From now on, even ETCD cluster is back to normal, no one is able to use this volume correctly, since the global refcount won't be changed from 0 to 1, which means no filer server is able to be started for it.

To handle the above race condition, the design needs to be adjusted to introduce locks/helper threads to check the unmatched global refcount and users.

luomiao:
Closed by #2001

PR #2001, opened 2017-11-28
Implement the new locking and notification system for vFile

luomiao:
The PR is to solve issue https://github.com/vmware/docker-volume-vsphere/issues/1943

The new design has the following changes:

1. The start and stop of file server is not triggered by global refcount change now. Instead, every global mount/umount request will increase the value of StartTrigger/StopTrigger in the KV store. Two separate watchers (on each master node) will generate events according to the PUT operations to these two triggers.
2. The states of volumes are reduced to Ready and Mounted only. No intermediate states are needed. As a result, locks are required when the states need to be updated.
3. To avoid overlapping operations, locks are also required for updating global refcounts. The locks for global refcount are different from the locks for volume states. Usually, the workers are the ones who grab global refcount locks, and the managers (watchers) are the ones who need state locks.
4. Two fields StartMarker and StopMarker are used to guarantee only one manager's watcher is able to proceed to do the start/stop server operations. And thus other watchers will be able to return and handle events for different volumes in parallel.
5. To make sure when an error happens, the volumes won't be left in an error state, the order for updating metadata in KV store has been changed too. First of all, global refcount will only be increased after the state of the volume is Mounted; Second, the state of volume can be changed to Mounted only after the file server is up and running. On the other side, during unmount, global refcount and client list will be updated first in the same transaction; then stop trigger is increased to trigger the event to stop the file server.
6. During the volume deletion, both global refcount and state locks are required. The plugin is responsible to reset the global refcount to 0 and reset state to Ready, and then should increase the StopTrigger to shut down the file server service, if for some reason it's still running.

### DIFF
--- a/client_plugin/drivers/vfile/dockerops/dockerops.go
+++ b/client_plugin/drivers/vfile/dockerops/dockerops.go
@@ -51,7 +51,7 @@ const (
 	// for communicating to clients
 	networkDriver = "overlay"
 	// Name of the Samba server docker image
-	sambaImageName = "dperson/samba"
+	sambaImageName = "luomiao/samba-debian"
 	// Name of the Samba share used to expose a volume
 	FileShareName = "share1"
 	// Default username for all accessing Samba server mounts
@@ -200,6 +200,16 @@ func (d *DockerOps) VolumeInspect(volName string) error {
 //      bool:    Indicated success/failure of the function. If
 //               false, ignore other output values.
 func (d *DockerOps) StartSMBServer(volName string) (int, string, bool) {
+	_, port, err := d.getServiceIDAndPort(volName)
+	if err == nil {
+		// if the service is already existing, just return corresponding port number and name
+		return int(port), serviceNamePrefix + volName, true
+	}
+
+	if err != nil && err.Error() != noSambaServiceError {
+		log.Warningf("Failed to check the if a server is running for volume %s", volName)
+	}
+
 	var service swarm.ServiceSpec
 	var options dockerTypes.ServiceCreateOptions
 
@@ -463,6 +473,9 @@ func (d *DockerOps) DeleteInternalVolume(volName string) {
 				// volume exists, continue waiting and retry removing
 				msg += fmt.Sprintf(" Internal volume still in use. Wait and retry before timeout.")
 				log.Warningf(msg)
+			} else {
+				log.Info("Successfully deleted internal volume.")
+				return
 			}
 		case <-timer.C:
 			// The deletion of internal volume will be handled by garbage collector
@@ -489,7 +502,12 @@ func (d *DockerOps) StopSMBServer(volName string) (int, string, bool) {
 	log.Infof("StopSMBServer for vol %s", volName)
 	serviceID, _, err := d.getServiceIDAndPort(volName)
 	if err != nil {
-		return 0, "", false
+		if err.Error() != noSambaServiceError {
+			return 0, "", false
+		} else {
+			// if the service is already stopped
+			return 0, "", true
+		}
 	}
 
 	//Stop the service
--- a/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
+++ b/client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
@@ -30,6 +30,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	etcdClient "github.com/coreos/etcd/clientv3"
+	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/vmware/docker-volume-vsphere/client_plugin/drivers/vfile/dockerops"
 	"github.com/vmware/docker-volume-vsphere/client_plugin/drivers/vfile/kvstore"
 )
@@ -52,8 +53,10 @@ import (
    gcTicker:                   ticker for garbage collector to run a collection
    etcdClientCreateError:      Error indicating failure to create etcd client
    swarmUnhealthyErrorMsg:     Message indicating swarm cluster is unhealthy
-   etcdSingleRef:              if global refcount 0 -> 1, start SMB server
-   etcdNoRef:                  if global refcount 1 -> 0, shut down SMB server
+   etcdNoRef:                  global refcount = 0, no dependency on this volume
+   etcdLockTicker:             ticker for blocking wait a ETCD lock
+   etcdLockTimer:              timeout for blocking wait a ETCD lock
+   etcdLockTimeoutErrMsg:      Message indicating the failure to block wait a ETCD lock
 */
 const (
 	etcdDataDir              = "/etcd-data"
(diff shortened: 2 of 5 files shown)


client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -1043,6 +1271,69 @@ func (e *EtcdKVS) WriteMetaData(entries []kvstore.KvPair) error {
...
+	}
+
+	// Post all requested operations in one transaction
+	ctx, cancel := context.WithTimeout(context.Background(), etcdRequestTimeout)
+	resp, err := client.Txn(ctx).Then(ops...).Commit()
+	cancel()
+	if err != nil {
+		msg := fmt.Sprintf("Transactional metadata update failed: %v.", err)
+		if err == context.DeadlineExceeded {
+			msg += fmt.Sprintf(swarmUnhealthyErrorMsg)
```
lipingxue: Why here we add error for "swarmUnhealthy"?

client_plugin/drivers/vfile/kvstore/kvstore.go
```diff
@@ -57,13 +57,36 @@ const (
 	VolPrefixGRef                     = "SVOLS_gref_"
 	VolPrefixInfo                     = "SVOLS_info_"
 	VolPrefixClient                   = "SVOLS_client_"
+	VolPrefixStartTrigger             = "SVOLS_start_trigger_"
```
lipingxue: I am still not quite clear with VolPrefixStartTrigger and VolPrefixStartMarker. Could you explain when those two values are used?

client_plugin/drivers/vfile/dockerops/dockerops.go
```diff
@@ -489,7 +502,12 @@ func (d *DockerOps) StopSMBServer(volName string) (int, string, bool) {
 	log.Infof("StopSMBServer for vol %s", volName)
 	serviceID, _, err := d.getServiceIDAndPort(volName)
 	if err != nil {
-		return 0, "", false
+		if err.Error() != noSambaServiceError {
+			return 0, "", false
+		} else {
+			// if the service is already stopped
+			return 0, "", true
```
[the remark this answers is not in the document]
luomiao: Because the purpose of this function is to stop the service.
When the service is already stopped, we just return true so the event handlers can proceed to next steps.

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
```diff
@@ -650,140 +678,311 @@ func (e *EtcdKVS) cleanOrphanService(volumesToVerify []string) {
...
+	// monitor PUT requests on start triggers, excluding the PUT for creation
+	if ev.Type == etcdClient.EventTypePut && ev.PrevKv != nil {
+		log.Debugf("Watcher got a increase event on watcher, new value is %s", string(ev.Kv.Value))
+		volName := strings.TrimPrefix(string(ev.Kv.Key), kvstore.VolPrefixStartTrigger)
+
+		// Compare the value of start marker, only one watcher will be able to successfully update the value to
+		// the new value of start trigger
+		success, err := e.CompareAndPutIfNotEqual(kvstore.VolPrefixStartMarker+volName, string(ev.Kv.Value))
+		if err != nil || success == false {
 			return
```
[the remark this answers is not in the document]
luomiao: This is not a real error. Only one watcher will be able to successfully update the marker.
Other watchers will fail, but they should just return. I will add more comments here.

client_plugin/drivers/vfile/kvstore/etcdops/etcdops.go
luomiao: Should be etcdUnhealthy. Will update.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -371,61 +386,69 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
-				log.Errorf(msg)
-				return volume.Response{Err: msg}
+	// Try to lock for changing state. At this point, there should be no lock on the state
+	stateLock, err := d.kvStore.CreateLock(kvstore.VolPrefixState + r.Name)
+	if err != nil {
+		msg = fmt.Sprintf("Failed to create lock for removing volume %s. Error: %v",
+			kvstore.VolPrefixState+r.Name, err)
+		log.Error(msg)
+		grefLock.ReleaseLock()
+		return volume.Response{Err: msg}
```
lipingxue: I think we should call stateLock.ClearLock() here.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -371,61 +386,69 @@ func (d *VolumeDriver) Remove(r volume.Request) volume.Response {
...
-				msg = fmt.Sprintf("Remove failed: cannot unmarshal info data. %v", err)
-				log.Errorf(msg)
-				return volume.Response{Err: msg}
-			}
+	err = stateLock.TryLock()
+	if err != nil {
+		msg = fmt.Sprintf("Failed to try lock for removing volume %s. Error: %v",
+			kvstore.VolPrefixState+r.Name, err)
+		log.Error(msg)
+		stateLock.ClearLock()
```
lipingxue: I think here should be "stateLock.ReleaseLock()"

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -551,14 +550,29 @@ func (d *VolumeDriver) MountVolume(name string, fstype string, id string, isRead
...
 
-	// Increase GRef
-	log.Infof("Before AtomicIncr")
-	err = d.kvStore.AtomicIncr(kvstore.VolPrefixGRef + name)
+	log.Debugf("MountVolume: before get lock for global refcount of %s", name)
+	// Get the lock for changing the global refcount
+	lock, err := d.kvStore.CreateLock(kvstore.VolPrefixGRef + name)
+	if err != nil {
+		log.Errorf("Failed to create lock for mounting volume %s", kvstore.VolPrefixGRef+name)
+		return "", err
```
lipingxue: I think we should call "lock.ClearLock()" before return error.

client_plugin/drivers/vfile/vfile_driver.go
```diff
@@ -758,35 +784,79 @@ func (d *VolumeDriver) UnmountVolume(name string) error {
...
-	// remove VM from ClientList
-	nodeID, addr, _, err := d.dockerOps.GetSwarmInfo()
+	// Get the lock for changing the global refcount
+	lock, err := d.kvStore.CreateLock(kvstore.VolPrefixGRef + name)
 	if err != nil {
-		log.WithFields(
-			log.Fields{"volume name": name,
-				"error": err,
-			}).Error("Failed to get IP address from docker swarm ")
+		log.Errorf("Failed to create lock for mounting volume %s", kvstore.VolPrefixGRef+name)
```
lipingxue: Should call lock.ClearLock() before return error.

client_plugin/drivers/vfile/vfile_driver.go
luomiao: No this error is when it's failed to create state lock. Since it's not created, we don't need to clear it.

client_plugin/drivers/vfile/vfile_driver.go
luomiao: When we are not successfully get the lock, we do ClearLock instead of ReleaseLock.

client_plugin/drivers/vfile/vfile_driver.go
luomiao: Same comment as above.

Issue #2025, opened 2017-12-06
Document plugin and ESX service version compat


govint:
From 0.19, the plugin has to have a ESX service thats at least 0.19 or later. This needs to be documented to avoid new volume creation errors or errors in attaching existing volumes. This applies for plugins on Linux platform.

PR #2026, opened 2017-12-06
[SKIP CI] Document plugin and ESX service version compatibility

ashahi1:
This PR addresses the issue #2025

### DIFF
--- a/docs/external/faq.md
+++ b/docs/external/faq.md
@@ -36,6 +36,8 @@ Edge release: A monthly release with new features and bug fixes.
 
 Stable release: A quarterly release with reliable/stable updates.
 
+***Note: From 0.19 release of vDVS, the plugin has to have a ESX service thats at least 0.19 or later. This applies for plugins on Linux platform.***
+
 Please refer [release convention](https://github.com/vmware/docker-volume-vsphere/blob/master/CONTRIBUTING.md#release-naming-convention) for more details.
 
 ### Can I use the full volume like "vol@datastore" in compose file with vDVS?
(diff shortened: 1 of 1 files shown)


docs/external/faq.md
```diff
@@ -36,6 +36,8 @@ Edge release: A monthly release with new features and bug fixes.
 
 Stable release: A quarterly release with reliable/stable updates.
 
+***Note: From 0.19 release of vDVS, the plugin has to have a ESX service thats at least 0.19 or later. This applies for plugins on Linux platform.***
```
shuklanirdesh82: Let's not add here and move out as a separate FAQ entry for any user to search error message to get quick answer.

ashahi1: Done

docs/external/faq.md
```diff
@@ -88,6 +88,9 @@ e.g.
 systemctl restart docker
 ```
 
+#### I'm not able to create volume and I see "VolumeDriver.Create: Device not found" error
+
+From 0.19 release of vDVS, the plugin has to have a ESX service thats at least 0.19 or later. This applies for plugins on Linux platform. [#2023](https://github.com/vmware/docker-volume-vsphere/issues/2023)
```
shuklanirdesh82: `ESX service thats at least 0.19` => `ESX service at least 0.19` or `thats` -> `that's` What do you mean by `thats`?

shuklanirdesh82: Let's not mention `Linux platform` explicitly, it should be OS independent.

/CC @govint  to confirm

govint: Could mention VIB (use VIB name with the version) perhaps vs. ESX service. 

For Windows the device identification in the plugin is the same as it was, while for Linux the device identification has been upgraded to detect devices on secondary PCI buses.

govint: This issue will happen only with the plugin on Linux, but agree there is no need for the windows plugin to use an older VIB version. Its best that the users always upgrade both the pluigin and the VIB to matching levels.

PR #2051, opened 2018-01-08
Fix checks for pci slot number to use lower case compares.

govint:
Fix has been tested locally and by reporter. Switch to use lower case compares when getting the slot/bus number for the pvscsi controller. 6.5 U1 seems to be use lower case key names in the VM config and hence causing the issue.


shuklanirdesh82:
> CI fails 
> [info] decrypting encrypted secrets: square/go-jose: error in cryptographic primitive
> https://ci.vmware.run/vmware/vsphere-storage-for-docker/42

@govint Please rebase your branch to kick off the CI run.

PR #2053, opened 2018-01-11
Update the smb image to our own simpler version

luomiao:
Update the samba image to our own simpler version.

### DIFF
--- a/client_plugin/drivers/vfile/dockerops/dockerops.go
+++ b/client_plugin/drivers/vfile/dockerops/dockerops.go
@@ -47,18 +47,20 @@ const (
 	// Postfix added to names of Samba services for volumes
 	serviceNamePrefix = "vFileServer"
 	// Path where the file server image resides in plugin
-	fileServerPath = "/usr/lib/vmware/samba.tar"
+	fileServerPath = "/usr/lib/vmware/vfilesmb.tar"
 	// Driver for the network which Samba services will use
 	// for communicating to clients
 	networkDriver = "overlay"
 	// Name of the Samba server docker image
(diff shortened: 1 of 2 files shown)


client_plugin/drivers/vfile/dockerops/dockerops.go
```diff
@@ -220,31 +222,12 @@ func (d *DockerOps) StartSMBServer(volName string) (int, string, bool) {
 	// The Docker image to run in this service
```
lipingxue: Why with this change, there is no need to pass "containerArgs"?

luomiao: No we don't need to create the shares or users since we have pre-defined shares inside the image already.

lipingxue: Then when user need to create a share owned by specific user and with specific access permission, how can we set that? I don't think we can set in image in this case.

luomiao: That's right :)
This is a feature we don't support for now (and didn't support it before the change too).
We can add it later.

lipingxue: My only concern here is in the original image, they are configurable in the code. But with this change, it seems hard coded in the image. Is there any way that make it configurable other than hard coded?

luomiao: Yes we will need to update the vfilesmb image to include customize functions for creating new users. We can add this when we support the user defined username functionality.

Issue #2061, opened 2018-02-01
unable to start containers with existing vsphere volumes after upgrading


bteichner:
After upgrading from version 0.17 of the plugin to version 0.20, we are unable to start services that had existing vsphere Docker volumes. 

The following error is displayed when running `docker service ps service-name --no-trunc`
**VolumeDriver.Mount: EOF**

I'm also seeing this warning in the Docker daemon logs:
```
"Error restoring volume\" driver=\"vsphere:latest\" error=\"error while checking if volume \\\"service-name\\\" exists in driver \\\"vsphere:latest\\\": Error looking up volume plugin vsphere:latest: plugin \\\"vsphere:latest\\\" not found\" volume=volume-name"
```

**Docker version:** 17.06.2-ee-6

**Upgrade Steps**
_Upgrade plugin on esxi hosts_
1. esxcli software vib remove -n esx-vmdkops-service
2. esxcli software vib install -v /<vib_name>.vib

_Upgrade plugin on Docker hosts_
1. docker plugin disable -f vsphere
2. docker plugin upgrade --grant-all-permissions vsphere:latest vmware/vsphere-storage-for-docker:0.20
3. sysetmctl restart docker (required on worker nodes before re-enabling the plugin)
4. docker plugin enable vsphere

govint:
@bteichner could you also attach the logs from the ESX host (/var/log/vmware/vmdk_ops.log) and from the docker host (/var/log/docker-volume-vsphere.log)

govint:
@bteichner - could you respond with the logs. Thanks.

bteichner:
@govint Can I email you the logs, rather than posting them here?

govint:
@bteichner Thanks for mailing the logs. Looks like the KV file for a container volume is missing. Let me check why this would have happened. But if you have the ESX side logs in its entirety that would help understand why/how the KV file was missing. Can you compress the ESX logs and mail them.

govint:
From the logs received, one KV file is reported as not present,
IOError: [Errno 2] No such file or directory: '/vmfs/volumes/NTNX-NFS-DOCKER-PRD/dockvols/UCPEU_PRD/compasseu-solr-prd-73f3739dc6d0707a.vmfd'
[...]

PR #2067, opened 2018-02-13
Restored DVOL_KEY to "docker-volume-vsphere"

govint:
The change to rename the repo made at least one change in kvESX.py to set DVOL_KEY to the new repo name. This shouldn't have been done as the DVOL_KEY is hashed to create a unique name for the KV file for a container volume. Since it was changed, older KV files are not openable as the code is creating a new hash for the new value and hence a new file name.

Pls. merge this change asap.

All 0.21 users would be impacted as their older container volumes and hence data won't be accessible.

@shuklanirdesh82 can we release a patch or make a one-off release 0.22 to have this change out asap.


govint:
Fixes #2061

shuklanirdesh82:
@govint I am planning to refresh the 0.21 VIB itself to avoid such upgrade problem.

